PR #8, opened 2020-01-24
Openshift Dev stream config - App Updates

devinleighsmith:
PIMS-209, PIMS-322, PIMS-80, PIMS-318

Pipeline bcs and dcs for dev branch.
App - remove SWU specific features
App - Add landing page
App - Update keycloak use within the app - provide example of how to check user role.


devinleighsmith:
Note - the above will require some manual intervention in openshift - the bcs and dcs in openshift are still pointing to my fork. I will need to update them after this branch has been merged.

devinleighsmith:
Another note for @Fosol. I will be making another PR with all of my bcs and dcs files supporting a BRANCH_SUFFIX var that will allow us to generate a dev and production pipeline. Planning on basing this off of @asanchezr PR for consistency.

frontend/openshift/frontend-deploy.test.param
```diff
@@ -0,0 +1,15 @@
+#=========================================================
+# OpenShift template parameters for:
+# Component: .
+# Template File: ../frontend/openshift/templates/frontend/frontend-deploy.yaml
+#=========================================================
+# NAME=frontend
+# APP_NAME=pims
+# ROLE=frontend
+# IMAGE_NAMESPACE=jcxjin-tools
```
Fosol: The recommendation here is to push images to each Project rather than all of them reference the images from TOOLS. The result of only using TOOLS is that if the image is deleted it will result in failures in TEST theoretically.

asanchezr: @Fosol we will have to chat when you come back from training. The above doesn't match what I've heard before but maybe there has been changes to best practices

PR #9, opened 2020-01-24
Openshift setup - API

asanchezr:
PIMS-59

### DIFF
--- a/openshift/templates/tools/pipeline.yaml
+++ b/openshift/templates/tools/pipeline.yaml
@@ -1,0 +1,69 @@
+kind: "Template"
+apiVersion: "v1"
+metadata:
+  name: "${NAME}-pipeline-template"
+  creationTimestamp: null
+objects:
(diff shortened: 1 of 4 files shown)


openshift/templates/tools/pipeline.yaml
```diff
@@ -0,0 +1,69 @@
...
+        type: "JenkinsPipeline"
+        jenkinsPipelineStrategy:
+          jenkinsfilePath: "${JENKINSFILE_PATH}"
+      output: {}
+      resources: {}
+      postCommit: {}
+    status:
+      lastVersion: 0
+parameters:
+  - name: "NAME"
```
devinleighsmith: How are these params without names being filled in at the moment? I'm using .param files committed to source, is there another method that I should be aware of?

Fosol: I suspect the CLI.

PR #12, opened 2020-01-28
PIMS-280 Keycloak Openshift configuration

asanchezr:
Inject Keycloak (auth) settings via ENV vars

### DIFF
--- a/openshift/templates/app/pims-api/pims-api-deploy.yaml
+++ b/openshift/templates/app/pims-api/pims-api-deploy.yaml
@@ -78,6 +78,12 @@ objects:
               env:
                 - name: ASPNETCORE_ENVIRONMENT
                   value: "${ASPNETCORE_ENVIRONMENT}"
+                - name: Keycloak__Secret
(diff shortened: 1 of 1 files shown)


asanchezr:
@Fosol I have added env vars for Keycloak to the backend API containers in Openshift. I still have a couple questions on the actual values that need to be set. Let's chat tomorrow.

openshift/templates/app/pims-api/pims-api-deploy.yaml
```diff
@@ -146,6 +152,19 @@ parameters:
       Staging, Production
     required: true
     value: Production
+  - name: KEYCLOAK_SECRET
+    displayName: Keycloak Secret
+    description: The KeyCloak secret to use for authentication
+    required: false
+    value: ""
+  - name: KEYCLOAK_AUTHORITY
```
Fosol: I've sent you the config for Authority and Audience in RocketChat.

asanchezr: Thanks, I have provided those in the template

PR #13, opened 2020-01-30
IE fix, reorganize package.json

devinleighsmith:
missed committing this on Monday, but would rather commit this separately to keep the review small.

### DIFF
--- a/frontend/public/keycloak.json
+++ b/frontend/public/keycloak.json
@@ -1,8 +1,8 @@
 {
-  "realm": "xz0xtue5",
-  "auth-server-url": "https://sso-dev.pathfinder.gov.bc.ca/auth",
+  "realm": "quartech",
+  "auth-server-url": "http://localhost:8080/auth",
   "ssl-required": "external",
-  "resource": "pims-app",
+  "resource": "team-blue-app",
   "public-client": true,
   "confidential-port": 0
 }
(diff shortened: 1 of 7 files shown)


frontend/public/keycloak.json
```diff
@@ -1,8 +1,8 @@
 {
-  "realm": "xz0xtue5",
-  "auth-server-url": "https://sso-dev.pathfinder.gov.bc.ca/auth",
+  "realm": "quartech",
```
Fosol: I don't think this realm is correct for OpenShift

frontend/public/keycloak.json
```diff
@@ -1,8 +1,8 @@
...
+  "auth-server-url": "http://localhost:8080/auth",
```
Fosol: I don't think this URL is correct for OpenShift

frontend/public/keycloak.json
```diff
@@ -1,8 +1,8 @@
...
   "ssl-required": "external",
-  "resource": "pims-app",
+  "resource": "team-blue-app",
```
Fosol: I don't think this resource is correct for OpenShift

frontend/public/keycloak.json
devinleighsmith: This file is now only used locally, so I switched the config back to the local keycloak config.

PR #14, opened 2020-01-31
Openshift setup - v2

asanchezr:
* Add database templates
* Improve documentation
* General clean up of existing configuration


openshift/templates/base-images/dotnet-31.yaml
```diff
@@ -0,0 +1,61 @@
+kind: Template
```
Fosol: I'm curious why this one is needed?  Is it a template to pull in the image with your redhat account?

openshift/templates/base-images/mssql-2019.yaml
```diff
@@ -0,0 +1,25 @@
+kind: Template
```
Fosol: I'm curious why this is needed too?  Doesn't this image already exist in our image streams?

openshift/templates/app/pims-api/build-dev.env
```diff
@@ -0,0 +1,7 @@
+APP_NAME=pims
```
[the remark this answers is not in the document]
asanchezr: I have added the .env files for convenience and because most teams include it (see our own frontend openshift templates, they gave them as .param files). 

Also, none of the information there is sensitive, please note that how the SQL server admin password is auto-generated, not set in the .env file.

However, I'm fine taking them out but we will need to add to the documentation instructions on how to create them as per your comment above with `oc process`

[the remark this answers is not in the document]
Fosol: I lean towards not including them, just documentation on what to do.  Otherwise we risk including sensitive information (by mistake) and it creates duplication of effort when ensuring the default values are correct in the YAML.

openshift/templates/base-images/dotnet-31.yaml
asanchezr: Okay, so the reason I included these is because you "manually" imported them but we have no configuration files reflecting that. The whole idea of having these files in source control is to create everything needed from scratch (if needed). As it is, someone would not be able to use our configs unless the had those images imported into their environments.

Also, we will need to document how to create the pull secret needed to authenticate against RedHat's container registry. That will come in a subsequent PR

openshift/templates/base-images/mssql-2019.yaml
asanchezr: See my earlier comment

openshift/templates/base-images/dotnet-31.yaml
Fosol: Ah yeah perfect!  I forgot about that.  So you're just missing the information for the pull secret.

openshift/templates/app/pims-api/build-dev.env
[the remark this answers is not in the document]
asanchezr: Agreed. I will update the PR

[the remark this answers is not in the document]
devinleighsmith: Just a reminder to update the readme as well after you remove the env files.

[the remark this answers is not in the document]
asanchezr: yeah, thanks. working on it

openshift/templates/app/pims-db/mssql-deploy.yaml
```diff
@@ -0,0 +1,133 @@
...
+    metadata:
+      name: ${COMP_NAME}-${SUFFIX}-data
+    spec:
+      accessModes:
+        - ReadWriteOnce
+      resources:
+        requests:
+          storage: ${VOLUME_CAPACITY}
+parameters:
+  - name: APP_NAME
```
devinleighsmith: You're using COMP_NAME instead of this correct?

openshift/templates/base-images/mssql-2019.yaml
devinleighsmith: thanks for doing this Alejandro!

openshift/templates/app/pims-db/mssql-deploy.yaml
asanchezr: Well, we use APP_NAME as the group (ie pims) and then we append the COMP_NAME (ie database) to form the full name -> pims-database

We can always tweak how we generate the names

PR #18, opened 2020-02-04
PIMS-449 PIMS-444 PIMS-450

devinleighsmith:
Create mock apis for property list and property detail.
Create react-redux store, framework, and two actions (to call above apis) with examples.


frontend/src/customAxios.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+      const status = error.response ? error.response.status : null;
+      const errorMessage =
+        errorToastMessage || (error.response && error.response.data.message) || String.ERROR;
+
+      if (status === UNAUTHORIZED || status === MAINTENANCE) {
+        window.location.reload(false);
+      } else if (status === TEAPOT || process.env.NODE_ENV === "development") {
+        // Do not retry in dev env, or for axios calls made in frontend tests, which are configured to return status code 418
+        //TODO: come up with a way of displaying request failures.
```
devinleighsmith: admittedly this is a bit of a mess, but I don't have UI or requirements for:
1) some kind of progress bar or notification icon for async requests.
2) how to display async request failures.

backend/Models/Property.cs
```diff
@@ -0,0 +1,43 @@
+using System;
```
devinleighsmith: so clearly these models are incredibly not finalized, but I'd like to start working on the frontend/backend connectivity on Openshift sooner rather then later. Once that communication is working I'll worry about the exact structure of the request/response (hopefully the database will exist by then).

backend/Controllers/PropertiesController.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <summary>
+    /// Get the details for a specified property
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet]
+    [Route ("/api/properties/{pid}")]
+    public IActionResult GetPropertyDetails(double pid)
+    {
+      //return a mock response for now.
+      PropertyDetail propertyDetail = new PropertyDetail
```
devinleighsmith: Just return dumb mock data for now.

frontend/src/pages/MapView.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useEffect } from 'react';
```
devinleighsmith: @asanchezr I've updated the mapview with some example components that display the new redux-based API data. These are just for reference, and are not intended to survive this sprint.

frontend/src/customAxios.tsx
Fosol: Ideally all requests would go through common library so that failures would be handled the same way.  Display a little floating div with a relevant, but generic error message when not in a Developer environment.

backend/Models/Property.cs
Fosol: See my other comments related to models.

backend/Controllers/PropertiesController.cs
```diff
@@ -0,0 +1,102 @@
...
+using Property = BackendApi.Models.Property;
+using PropertyDetail = BackendApi.Models.PropertyDetail;
+using Entity = BackendApi.Data.Entities;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace BackendApi.Controllers
```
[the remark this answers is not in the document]
asanchezr: Agree with the rename

backend/Controllers/PropertiesController.cs
```diff
@@ -0,0 +1,102 @@
...
+    #endregion
+
+    #region Endpoints
+
+    /// <summary>
+    /// Get a list of properties that are within the provided bounds
+    /// </summary>
+    /// <returns></returns>
+    [HttpGet]
+    public IActionResult GetProperties(double? neLat, double? neLong, double? swLat, double? swLong)
```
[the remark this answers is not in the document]
asanchezr: I like the "rectangular" filter (4 corners) better than the "circular" filter (center + diameter) because the most intuitive way of selecting on a map is to click, drag, click to create a rectangle with your area of interest.

Having users click a center-point and then having them drag to choose how big the circle should be is not the best UX I've seen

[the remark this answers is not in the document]
devinleighsmith: The idea behind this is based on the Google maps API and Leaflet API. Both use the concept of bounds when fetching map data. Primary use case is fetching points based on the current viewport, which I'm assuming we would be doing as well (rather then getting the user to scroll the map and then create an area to request pins for).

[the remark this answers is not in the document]
Fosol: Sounds good to me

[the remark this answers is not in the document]
asanchezr: 👍

PR #19, opened 2020-02-04
Pims 63 - Create initial DB migration

Fosol:


Fosol:
I've renamed the API project to `pims.api`.
It now uses the MSSQL DB instead of PostgreSQL.
I've updated the README documents to reflect the changes and how to use it.
Please note that we will most likely blow away the initial migration with a new one once we finalize.

database/README.md
```diff
@@ -1,33 +1,96 @@
...
+Any script that you want to run must be additionally included as build content in the project file.
+
+Edit the `Pims.Api.csproj` file and add the `<Content>` with a `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` (see below example).
+
+```
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk.Web">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <UserSecretsId>17033af2-41c5-4100-a42e-4d11424b0cbb</UserSecretsId>
```
devinleighsmith: /nit could replace with [secret] to be consistent with the keycloak secret documentation.

database/mssql/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM mcr.microsoft.com/mssql/rhel/server:2019-CU1-rhel-8
+
```
asanchezr: @Fosol I think we should use the `2019-latest` tag to match our Openshift environment

database/postgres/scripts/init.sql
```diff
@@ -0,0 +1,23 @@
+DROP TABLE IF EXISTS Places;
```
devinleighsmith: Didn't this get replaced by Parcels?

database/mssql/Dockerfile
Fosol: Fixed

database/README.md
Fosol: Actually this value is related to user secrets and has no meaning outside of someone's developer workstation.  I've updated the README so that it doesn't include the line.

database/postgres/scripts/init.sql
Fosol: I've removed the script as it is irrelevant now.

database/README.md
asanchezr: agree. the thing is that the net core "secret" works a bit different than openshift. I think we need to keep that Id in place as then dotnet goes to C:\<your_user_folder>\<that_Id>\secrets.json

PR #23, opened 2020-02-06
Enable nginx routing of /api to backend pod

devinleighsmith:
the other change was to update the params of the pims-app dc to have an API_SERVICE_NAME of pims-api instead of api

### DIFF
--- a/frontend/openshift/templates/nginx-runtime/s2i/bin/run
+++ b/frontend/openshift/templates/nginx-runtime/s2i/bin/run
@@ -44,7 +44,6 @@ getApiUrl (){
   #
   #   Results in API_URL=https://django-devex-von-dev.pathfinder.gov.bc.ca/api/v1/   
   # ================================================================================
-  echo ${API_SERVICE_NAME}
   if [ ! -z "${API_SERVICE_NAME}" ]; then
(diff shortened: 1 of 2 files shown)


frontend/openshift/templates/nginx-runtime/s2i/bin/run
```diff
@@ -74,5 +73,6 @@ echo "RealIpFrom = ${RealIpFrom:-172.51.0.0/16}"
 echo "API_URL = ${API_URL}"
 
 sed "s~%RealIpFrom%~${RealIpFrom:-172.51.0.0/16}~g; s~%API_URL%~${API_URL}~g;" /tmp/nginx.conf.template > /etc/nginx/nginx.conf
+sed "s~%RealIpFrom%~${RealIpFrom:-172.51.0.0/16}~g; s~%API_URL%~${API_URL}~g;" /tmp/nginx.conf.template
```
asanchezr: I'm curious, why the extra `sed` command?

devinleighsmith: I found the logging in the openshift terminal invaluable, since the logs stick around after a failure and terminal access doesn't. Beats having to debug the dc.

PR #26, opened 2020-02-07
PIMS-45 Create awesome feature

asanchezr:
* Add new class LogRotator
* Enhance docs


devinleighsmith:
@asanchezr Can you kill this whenever you have a minute ;-)

PR #29, opened 2020-02-11
PIMS#448, PIMS#454

devinleighsmith:
Add unit tests for parcels and parcel detail APIs.
Fix the following defects discovered by the tests:
1) NoContent should be returned instead of NotFound when no detail matches id.
2) fix typo in parcels API lat/long filtering.


test/BackendApi.Test/Controllers/ParcelControllerTest.cs
```diff
@@ -0,0 +1,252 @@
...
+            // Act
+            int expectedParcelId = 1;
+            var actualParcelDetail = _parcelController.GetMyParcels(expectedParcelId);
+
+            // Assert
+            Assert.IsType<NoContentResult>(actualParcelDetail);
+        }
+
+        [Fact]
+        public void GetMyParcelDetail_Matching()
```
devinleighsmith: One test that I went back and forth on - a test to check that all of the fields supported by the database are returned by the API. I decided against it for now because I wasn't sure the test would have value due to the automapper, and this also may be more appropriate for an integration test, as I see a test like this being fragile to change.

Interested to hear other's thoughts.

GeoSpatial.sln
```diff
@@ -5,6 +5,8 @@ VisualStudioVersion = 15.0.26124.0
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "pims.api", "backend\pims.api.csproj", "{16BC0468-78F6-4C91-87DA-7403C919E646}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "pims.api.test", "test\BackendApi.Test\pims.api.test.csproj", "{19A6829C-20D8-4F66-9889-55C553ACE2B2}"
```
Fosol: Perhaps rename folder to reflect the change to the API (i.e. /test/backend/api/pims.api.test.csproject)

backend/Models/ParcelDetail.cs
```diff
@@ -56,6 +58,43 @@ public ParcelDetail(Pims.Api.Data.Entities.Parcel parcelDetail)
       this.PropertyStatus = parcelDetail.Status?.Name;
       this.Agency = parcelDetail.Agency?.Name;
     }
-    #endregion
-  }
+
+        public override bool Equals(object obj)
```
Fosol: Need to start adding comments to describe the purpose/function of the methods.  They're useful when using intellisense.

test/BackendApi.Test/Controllers/ParcelControllerTest.cs
```diff
@@ -0,0 +1,252 @@
...
+using System.Linq;
+
+namespace PimsApi.Test.Controllers
+{
+    public class ParcelControllerTest
+    {
+        #region Variables
+        private readonly ParcelController _parcelController;
+        private readonly PIMSContext _dbContext;
+        private readonly Entity.Parcel _expectedParcel = new Entity.Parcel()
```
Fosol: Probably should look for a way to include test objects/data through some form of configuration/feed.  Doesn't need to happen in this PR though.

test/BackendApi.Test/Controllers/ParcelControllerTest.cs
Fosol: I'm generally against testing that all fields exist, Normally all fields are not required to be included in the API results, as the API can change the shape of the literal data based on how we want to present it.  Commonly this occurs when a role based access can't request certain data elements.

That being said, there are times I like tests like this if they are easily modified through templating/configuration.  In other words, there is some kind of logic that is showing/hiding properties.

test/BackendApi.Test/Controllers/ParcelControllerTest.cs
Fosol: I also like seeing helper libraries create to generate objects quickly.  Then the repetitive elements of creating objects is hidden and only the relevant properties are set in the unit test.

GeoSpatial.sln
asanchezr: I was going to suggest we move the whole folder under backend. That would match the frontend. Also the tests should be as close as possible to the code they check

openshift/settings.local.sh
```diff
@@ -0,0 +1,2 @@
+export GIT_URI="https://github.com/devinleighsmith/PIMS.git"
```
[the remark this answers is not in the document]
devinleighsmith: we don't, I'll remove it.

backend/Models/ParcelDetail.cs
devinleighsmith: For the equals method? I'm not sure that I agree. This is generated code for starters, and the equals method is one of the basic building blocks of OOP in my opinion. Perhaps you just mean in general?

test/BackendApi.Test/Controllers/ParcelControllerTest.cs
devinleighsmith: I agree in principle, but in my mind helper methods provide the same functionality. I would switch to helper libraries when we start having a lot of tests creating the same sorts of objects. Either way the tests (the actual facts) don't include the actual object generation other than the function calls.

backend/Models/ParcelDetail.cs
Fosol: In general we should provide a comment for functions so that intelisense provides some useful information.  I know this is a basic override, but the default behaviour is actually very different than what you have provided here.

devinleighsmith: Well, I'll at least check and see if the code generator can provide a comment as well the next time I do this.

PR #31, opened 2020-02-11
PIMS-441 Add RocketChat webhooks

Fosol:


tools/rocketchat/incoming.js
```diff
@@ -0,0 +1,345 @@
...
+    text += "# " + request.content.pull_request.number;
+    text +=
+      " - **[" +
+      request.content.pull_request.title +
+      "](" +
+      request.content.pull_request.html_url +
+      ")**\n\n" +
+      body;
+
+    // switch (request.content.action) {
```
Fosol: Leaving this in for now as it has references to variables that would be useful when I have to to modify this script.

PR #32, opened 2020-02-11
PIMS-417 Show parcels on Map View

asanchezr:
* Request parcels from API on every map update (pan & zoom)
* Request parcel details to populate map popup upon map click
* Also, fix a few frontend linting errors

### DIFF
--- a/frontend/src/pages/MapView.tsx
+++ b/frontend/src/pages/MapView.tsx
@@ -1,14 +1,12 @@
 import React, { useEffect } from 'react';
+import { LatLngBounds } from 'leaflet';
 import Map from '../components/maps/leaflet/Map';
 import './MapView.scss';
 import { fetchParcels, fetchParcelDetail } from 'actionCreators/parcelsActionCreator';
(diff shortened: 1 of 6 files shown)


frontend/src/pages/MapView.tsx
```diff
@@ -18,33 +16,45 @@ const parcelBounds: IParcelListParams = {
...
+  return (
+    <Map
+      lat={48.43}
+      lng={-123.37}
+      zoom={14}
+      parcels={parcels}
+      activeParcel={parcelDetail}
+      onParcelClick={p => dispatch(fetchParcelDetail({ id: p.id }))}
+      onPopupClose={() => dispatch(storeParcelDetail(null))}
+      onViewportChanged={e => {
```
devinleighsmith: I'm curious, how many requests do we get from this when you move the map around?. I'm wondering if we need to debounce this event to prevent the redux action from being fired a ton of times.

asanchezr: Great question. I thought about it but this is a first stab at the requested functionality. Further improvements can come in future PRs

PR #34, opened 2020-02-15
PIMS-127 reorganize .proj and .sln structure.

devinleighsmith:
Pre-PR review. Make sure I got the desired structure right.

I've done basic testing by validating:
dotnet ef commands work
dotnet run works (with appropriate config changes).
docker-compose works (with appropriate config changes).

If this looks good I'll get to work on that PimsService you mentioned in RC.


backend/README.md
```diff
@@ -8,12 +8,12 @@ The current environment is initialized through the environment variable `ASPNETC
 
 When running the solution locally it applies the configuration setting in the following order;
 
-1. appsettings.json
-2. appsettings.`[environment]`.json
-3. UserSecrets `(if environment=Development)`
-4. Environment Variables
+1. appsettings.json (within /api)
+2. appsettings.`[environment]`.json (within /api)
+3. UserSecrets `(if environment=Development)` (within /api and /dal)
```
Fosol: For some reason User Secrets aren't working reliably within VSCode.  They are only supposed to work when the environment = Development, but I can't get them working at all.

backend/README.md
```diff
@@ -8,12 +8,12 @@ The current environment is initialized through the environment variable `ASPNETC
...
+4. Environment Variables (within /api and /dal)
 
-To run the solution with docker-compose create a .env file and populate it with the following;
+To run the solution with docker-compose create a .env file within the /dal and /api directories and populate each with the following;
 
 ```conf
 ASPNETCORE_ENVIRONMENT=Development
```
Fosol: This shouldn't be needed within the `/dal` project.  The `/dal` project can use the `connectionstrings.json` or a `.env` to configure the connection string used by `dotnet ef database update`.

backend/README.md
```diff
@@ -45,9 +45,9 @@ Refer to the CLI documentation [here](https://docs.microsoft.com/en-us/ef/core/m
 
 # Database Migration Management
 
-The DB is setup and configured through Entity Framework Code-First processes.
+The DB is setup and configured through Entity Framework Code-First processes. All dotnet ef commands must be run from the /dal directory.
 
-To use the Entity Framework CLI you will need to install the **.NET SDK version**, **dotnet-ef tool** and add a `connectionstrings.json` configuration file.
+To use the Entity Framework CLI you will need to install the **.NET SDK version**, **dotnet-ef tool** and add a `connectionstrings.json` configuration file to /dal.
```
Fosol: See the above comment related to the choice of using the `connectionstrings.json` or `.env` file.

backend/README.md
```diff
@@ -61,12 +61,12 @@ Install the `dotnet-ef` CLI
 
 > `dotnet tool install --global dotnet-ef`
 
-Create a `connectionstrings.json` configuration file. You can also create one for each environment by creating a file with the naming conventsion `connectionstrings.[environment].json`. Enter the following information into the file;
+Create a `connectionstrings.json` configuration file within the /dal project. You can also create one for each environment by creating a file with the naming conventsion `connectionstrings.[environment].json`. Enter the following information into the file;
```
Fosol: This is optional depending on whether you want to use the `.env` file.  And is only required if you use the `dotnet ef database update`.

backend/README.md
```diff
@@ -142,7 +142,7 @@ namespace Pims.Api.Migrations
 
 Any script that you want to run must be additionally included as build content in the project file.
 
-Edit the `Pims.Api.csproj` file and add the `<Content>` with a `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` (see below example).
+Edit the `pims.dal.csproj` file and add the `<Content>` with a `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` (see below example).
```
Fosol: Probably should keep the project names Pascal Case.

backend/api/appsettings.json
```diff
@@ -10,19 +9,19 @@
   "AllowedHosts": "*",
   "Keycloak": {
     "Authority": "http://host.docker.internal:8080/auth/realms/quartech",
-    "Audience": "pims-api",
+    "Audience": "team-blue-api",
```
Fosol: Keycloak settings are incorrect in this file.

backend/api/libman.json
```diff
@@ -0,0 +1,5 @@
+{
```
Fosol: What's this file for?

backend/dal/Data/Configuration/AddressConfiguration.cs
```diff
@@ -0,0 +1,38 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Pims.Dal.Data.Entities;
+
+namespace Pims.Dal.Data.Configuration
```
Fosol: I wouldn't keep the `Data` folder.  Everything in `Data` should be moved to the root of the `Dal` project.  You'll need to update all your references to this namespace.

backend/dal/Data/Entities/Address.cs
```diff
@@ -1,6 +1,6 @@
 using System;
 
-namespace Pims.Api.Data.Entities
+namespace Pims.Dal.Data.Entities
```
Fosol: I think I'd also extract all the `Entities` into their own project `Pims.Dal.Entities`.  This way we can at some point if required fully control access to the database through the `Dal`, but also share the `Entities` to other projects without the need to reference the `Dal`.

backend/README.md
asanchezr: Running in containers won't carry over the secret file json from your user account folder. For developing with containers, ENV vars are preferred.

backend/README.md
asanchezr: The project name was originally renamed to all lower case. Agree we should go Pascal Case. Please note the the docker file will need to be updated (Linux is case sensitive)

devinleighsmith:
FYI, moving files doesn't prevent git from tracking file history, you just need to use --follow on your git log commands. ie git log --follow GeoSpatial.sln

devinleighsmith:
Closing this out as this was a prelim PR. The actual PR will come with the above requested changes as well as some additional work.

PR #38, opened 2020-02-19
PIMS-873 Add tests for redux actions.

devinleighsmith:


frontend/src/actionCreators/parcelsActionCreator.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const mockResponse = { data: { success: true } };
+    mockAxios.onGet(url).reply(200, mockResponse);
+    return fetchParcels(params)(dispatch).then(() => {
+      expect(requestSpy).toHaveBeenCalledTimes(1);
+      expect(successSpy).toHaveBeenCalledTimes(1);
+      expect(dispatch).toHaveBeenCalledTimes(6);
+    });
+  });
+
+  it("Request failure, dispatches `error` with correct response", () => {
```
Fosol: I'm wondering how much value a failure here is worth testing.  As I don't think the client is doing anything worth testing here.  Unless there is some functionality related to failures within the API.PARCELS(...) function.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const mockResponse = { data: { success: true } };
+    mockAxios.onGet(url).reply(200, mockResponse);
+    return fetchParcelDetail(params)(dispatch).then(() => {
+      expect(requestSpy).toHaveBeenCalledTimes(1);
+      expect(successSpy).toHaveBeenCalledTimes(1);
+      expect(dispatch).toHaveBeenCalledTimes(6);
+    });
+  });
+
+  it("Request failure, dispatches `error` with correct response", () => {
```
Fosol: Same comment as above in regard to testing a failure.

frontend/src/actionCreators/parcelsActionCreator.tsx
```diff
@@ -8,15 +8,15 @@ import CustomAxios from "customAxios";
 import { AxiosResponse } from "axios";
 import { createRequestHeader } from "utils/RequestHeaders";
 
-export const fetchParcels = (parcelBounds: API.IParcelListParams | null) => (dispatch:Function) => {
+export const fetchParcels = (parcelBounds: API.IParcelListParams | null) => (dispatch: Function) => {
```
Fosol: Is there a reason we're returning a function (callback) in this function?  Seems like a bad design and non-intuitive.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
devinleighsmith: The point of this test is just to validate that an error response from the API triggers a dispatch to the redux error action.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
devinleighsmith: Same reply.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
Fosol: Ah so we're testing the Axios and Redux implementation of a error response.  Hmmm, still leaning towards not needing us write tests for their implementation unless we do something ourselves with error responses.

frontend/src/actionCreators/parcelsActionCreator.tsx
devinleighsmith: This is some redux boilerplate which may look confusing. The title of the file helps to explain a bit, this is an just an action creator, which creates an action for future dispatch using the redux dispatch method. This is essentially just a shorthand way of creating an action.
https://redux.js.org/basics/actions/#action-creators
What you are talking about is a bound action creator, which we could easily wrap our action creators with if we wanted (see the same article as above).

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
devinleighsmith: Later on we'll wire up some UI to display errors from the store, but that would be a separate test.

frontend/src/actionCreators/parcelsActionCreator.tsx
asanchezr: For reference, this is how Redux "works" by designing "action creators" ie functions that return a function or also called HOFs (higher-order functions). I agree with you that it is increasingly complex to follow but that's the pattern with Redux unfortunately.

https://redux.js.org/basics/actions/#action-creators

Fosol: Meh, I'm fine with it.  I just am not a fan of **callback hell**.  Any time I see it I wonder why it would be preferred over promises or async - function(...).then(...)

devinleighsmith: No, its not a callback in the way you are thinking. Calling dispatch on the created action returns a promise.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
devinleighsmith: its not their implementation, we have a redux action that I wrote to handle writing errors to the store. I am testing that action has been written to the store after an error occurs.

Fosol: Makes sense

PR #39, opened 2020-02-21
Update documentation with pipeline setup instructions

asanchezr:
With this PR I have cleaned up the generic `pipeline.yaml` template and provided instructions on how to create the various pipelines that support our processes.

- **cicd-pipeline** - this is our main "build" pipeline with auto-deployment to DEV
- **promote-to-test-pipeline** - this is for manual approval to TEST

A future PR will handle building from the `master` branch and promoting to PROD

### DIFF
--- a/openshift/README.md
+++ b/openshift/README.md
@@ -7,49 +7,124 @@ This project uses OpenShift container platform (v3.11) for hosting the applicati
 ## Templates
 
 To configure OpenShift environments first create Templates. Then create the Objects from the Templates.
-1 Login `oc login`
-2 Select the appropriate Project `oc project [name]`
+
+1. Login `oc login`
+2. Select the appropriate Project `oc project [name]`
 
 ### PIMS API Build
 
-1 `oc create -f openshift/templates/app/pims-api/build.yaml`
-2 View the parameters `oc process --parameters pims-api-build`
-3 Create a build.env file that contains the values for the parameters within the template
-4 Create Objects `oc process pims-api-build --param-file=[.env] | oc create -f -`
+1. `oc create -f openshift/templates/app/pims-api/build.yaml`
+2. View the parameters `oc process --parameters pims-api-build`
+3. Create a build.env file that contains the values for the parameters within the template
+4. Create Objects `oc process pims-api-build --param-file=[.env] | oc create -f -`
 
 ### PIMS API Deploy
 
-1 `oc create -f openshift/templates/app/pims-api/deploy.yaml`
-2 View the parameters `oc process --parameters pims-api-deploy`
-3 Create a deploy.env file that contains the values for the parameters within the template
-4 Create Objects `oc process pims-api-deploy --param-file=[.env] | oc create -f -`
+1. `oc create -f openshift/templates/app/pims-api/deploy.yaml`
+2. View the parameters `oc process --parameters pims-api-deploy`
+3. Create a deploy.env file that contains the values for the parameters within the template
+4. Create Objects `oc process pims-api-deploy --param-file=[.env] | oc create -f -`
 
 # Environments
 
 There are 4 environments setup in the OpenShift Platform.
 
-- Tools : Contains all the tools/external services required by the application.
-- Dev : Contains a running application instance of an open pull request branch.
-- Test : Contains a running application instance of the current release branch.
-- Prod : Contains a running application instance of the current state of master.
+- **Tools** - Contains all the tools/external services required by the application.
+- **Dev** - Contains a running application instance of an open pull request branch.
+- **Test** - Contains a running application instance of the current release branch.
+- **Prod** - Contains a running application instance of the current state of master.
 
 # Filename Convention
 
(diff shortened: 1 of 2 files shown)


openshift/README.md
```diff
@@ -7,57 +7,132 @@ This project uses OpenShift container platform (v3.11) for hosting the applicati
...
 # CI/CD Tools
 
 ## Jenkins
 
-Jenkins is being used to orchestrate jobs between Github and OpenShift. Anytime a PR is open against the release/master branch it triggers a job on Jenkins which runs the pipeline stages and reports the status back to Github.
-
-[Link to Documentation](https://docs.openshift.com/container-platform/3.10/using_images/other_images/jenkins.html)
-
-[Link to the jenkins s2i image](https://github.com/BCDevOps/openshift-components/tree/master/cicd/jenkins)
+Jenkins is being used to orchestrate jobs between GitHub and OpenShift. Anytime a Pull Request is opened against the release/master branch, it triggers a job on Jenkins which runs the pipeline stages and reports the status back to GitHub.
```
Fosol: `release/master` probably should be `dev/master`

openshift/README.md
```diff
@@ -7,57 +7,132 @@ This project uses OpenShift container platform (v3.11) for hosting the applicati
...
+
+This project makes use of several pipelines. For each one, use the generic pipeline template located at `/openshift/templates/tools/pipeline.yaml` and customize with appropriate values.
+
+### cicd-pipeline
+
+1. View the parameters `oc process --parameters -f openshift/templates/tools/pipeline.yaml`
+2. Create a **`cicd.env`** file that contains the values for the parameters within the template
+    ```
+    NAME=cicd
```
Fosol: Why does the Jenkinsfile need the extension `cicd`?

openshift/templates/tools/pipeline.yaml
```diff
@@ -40,14 +39,11 @@ objects:
...
-- name: "COMP_NAME"
-  displayName: "Component Name"
-  description: "Name of the application component this pipeline will build and deploy."
+- name: "NAME"
+  displayName: "Pipeline Name (without -pipeline suffix)"
+  description: "Name of the pipeline, e.g cicd, promote-to-test, sonarqube. A -pipeline suffix will be appended to the name provided."
   required: true
+  value: cicd
 - name: "GITHUB_WEBHOOK_SECRET"
   displayName: "GitHub Webhook Secret"
```
Fosol: Probably should generate the secret for the pipeline by default.

openshift/templates/tools/pipeline.yaml
```diff
@@ -40,14 +39,11 @@ objects:
...
-  description: "Name of the application this pipeline will build and deploy."
-  required: true
-- name: "COMP_NAME"
-  displayName: "Component Name"
-  description: "Name of the application component this pipeline will build and deploy."
+- name: "NAME"
+  displayName: "Pipeline Name (without -pipeline suffix)"
+  description: "Name of the pipeline, e.g cicd, promote-to-test, sonarqube. A -pipeline suffix will be appended to the name provided."
   required: true
+  value: cicd
```
Fosol: I’d be tempted to no default the name so that whomever uses the template provides one.

openshift/README.md
asanchezr: fixed

openshift/templates/tools/pipeline.yaml
asanchezr: It DOES generate  secret by default. See the lines below. It also allows a user to supply their own secret (if they want to and are aware of risks)

```
  generate: "expression"
  from: "[a-zA-Z0-9]{40}"
```

openshift/templates/tools/pipeline.yaml
asanchezr: Fixed

openshift/README.md
asanchezr: We can name the Jenkinsfile however we want. In order to have multiple files for various pipelines we need to agree on a pattern for naming them:

I have seen 3 patterns to deal with this on other `bcgov` repos:

1. Use multiple (nested) folders, each one containing a single **Jenkinsfile**
    See https://github.com/bcgov/jag-cullencommission/tree/master/jenkins

2. Having them in a single folder but appending a suffix to indicate the purpose of the Jenkinsfile
    See https://github.com/bcgov/moh-prime
    They have: Jenkinsfile, Jenkinsfile.promotion, Jenkinsfile.tools

3. Having them in a single folder and rename them to *.groovy files, given that groovy is their syntax
     See https://github.com/bcgov/BCSC-SS/tree/dev/jenkins and https://github.com/bcgov/bcrs-business-create-ui/tree/master/Jenkinsfiles

I personally like to keep the reference to a Jenkinsfile-[whatever] vs having a "build.groovy" file. But if the team feels different I'm open to hearing ideas

Fosol: Ah I understand.  I'm fine with your approach, it makes sense.

PR #48, opened 2020-03-02
PIMS-1015 Request Access DB Migration.

devinleighsmith:


backend/api/Models/AccessRequestModel.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Pims.Api.Models
+{
+    public class AccessRequestModel : BaseModel
+    {
+        #region Properties
+        public Guid Id { get; }
+        public UserModel User { get; set; }
+        public IEnumerable<AgencyModel> Agencies { get; set; }
```
Fosol: Only allow for a single agency.  While we support multiple, we will only allow them to select one.

backend/api/Models/AccessRequestModel.cs
```diff
@@ -0,0 +1,21 @@
...
+        public bool? IsGranted { get; set; }
+        public IEnumerable<RoleModel> Roles { get; set; }
```
Fosol: Only allow for a single role.  While we support multiple, we will only allow them to select one.

backend/api/Models/AccessRequestModel.cs
Fosol: However depending on how  you use this in the code.  Still need to review below...  It is possible to be assigned a parent Agency and therefore you would have access to child agencies.

backend/api/Models/AgencyModel.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Pims.Api.Models
+{
+    public class AgencyModel : CodeModel, IEquatable<AgencyModel>
+    {
+        #region Properties
+        public string Description { get; set; }
+        public AgencyModel Parent { get; set; }
+        public ICollection<AgencyModel> Children { get; } = new List<AgencyModel>();
```
Fosol: Just be careful of the infinity situation here with serialization.

backend/api/Models/CodeModel.cs
```diff
@@ -14,6 +14,8 @@ public class CodeModel : IEquatable<CodeModel>
 
         public bool IsDisabled { get; set; }
 
+        public int SortOrder { get; set; }
```
Fosol: Thanks for adding this.

backend/api/Profiles/BaseProfile.cs
```diff
@@ -17,6 +17,8 @@ public BaseProfile()
             CreateMap<BaseModel, Entity.BaseEntity>()
                 .IncludeAllDerived()
                 .ForMember(dest => dest.RowVersion, opt => opt.Ignore())
+                .ForMember(dest => dest.CreatedById, opt => opt.Ignore())
```
Fosol: Do we actually want to pass this information all the time?  I didn't include them originally because I didn't think they had a lot value in the majority of cases.

backend/dal/Configuration/AccessRequestConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// AccessRequestConfiguration class, provides a way to configure users in the database.
+    ///</summary>
+    public class AccessRequestConfiguration : BaseEntityConfiguration<AccessRequest>
+    {
+        #region Methods
+        public override void Configure(EntityTypeBuilder<AccessRequest> builder)
+        {
+            builder.ToTable("AccessRequest");
```
Fosol: We use plural for the other table names.

backend/entities/AccessRequest.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        /// <typeparam name="User"></typeparam>
+        /// <returns></returns>
+        public User User { get; set; }
+
+        /// <summary>
+        /// get - the list of agencies that the user is requesting to be added to.
+        /// </summary>
+        /// <returns></returns>
+        public ICollection<Agency> Agencies { get; set; } = new List<Agency>();
```
Fosol: Don't think we should use plural agencies.  See first comment.

backend/entities/AccessRequest.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// get - the list of roles this user is requesting.
+        /// </summary>
+        /// <typeparam name="UserRole"></typeparam>
+        /// <returns></returns>
+        public ICollection<Role> Roles { get; set; } = new List<Role>();
```
Fosol: Don't think we should use plural roles.  See first comment.

backend/dal/Migrations/20200229012606_AccessRequest.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class AccessRequest : Migration
```
Fosol: Don't need to change this, but at some point we'll need to decide whether to recreate the initial migration or create a new versioned migration.

backend/api/Profiles/BaseProfile.cs
devinleighsmith: The profile is ignoring this information, meaning it will NOT be passed.
I've added this and some of the other Automapper in an effort to get our mapping configurations passing validation: 
 mapperConfig.AssertConfigurationIsValid();

Fosol: Cheer for my reading code quickly abilities :(

asanchezr: If I'm reading correctly, we are ignoring "Created/Updated By" fields when we go from JSON models to DB entities, right? If that's the case I think this code is useful as I wouldn't want to let external actors be able to "override" audit data via a JSON PUT/PUT

Can someone tell me if I'm wrong here?

backend/api/Models/AccessRequestModel.cs
devinleighsmith: I'm wondering if we want to limit this in the database or in the controller/business logic itself? I actually had this as one Agency/Role per request but changed it to conform with the structure of the user entity.

devinleighsmith:
@Fosol This review only includes changes directly related to PIMS-1015. As a result the controller with the POST (PIMS-1014) and the GET (PIMS-979) will be included in PRs for those changes.

backend/api/Models/AccessRequestModel.cs
Fosol: We want the DB to support multiple.  The client hasn't been able to confirm, but they currently said it is most likely only one to one.

devinleighsmith: @Fosol So are you ok with the AccessRequest table to be left as-is (except for the table name, I'' fix that), and then the actual controllers will be responsible for enforcing the one Agency/Role policy?

Fosol: Yes I think so.

PR #49, opened 2020-03-03
Allow remote debugging in API docker container

asanchezr:
* adds Docker debug configuration to VS Code
* updates Dockerfile to match debugging best practices
* allows debugging container to connect to local database (via docker-compose)

### DIFF
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,1 +1,18 @@
 .vs/
+.env
+
+# Build results
+[Dd]ebug/
+[Dd]ebugPublic/
+[Rr]elease/
+[Rr]eleases/
+x64/
+x86/
+build/
+bld/
+[Bb]in/
+[Oo]bj/
+[Oo]ut/
+msbuild.log
+msbuild.err
+msbuild.wrn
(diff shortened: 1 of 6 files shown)


asanchezr:
I left the old docker file intentionally as `Dockerfile.bak` in case we are not happy with the results. I was able to spin up the API container via VS Code and have it connect to the database that was launched via ` docker-compose up database`

We can discuss further details tomorrow.

asanchezr:
@Fosol the `.vscode` configuration settings are a matter of preference. I personally like to be able to open two VS Code instances - one for frontend, one for backend - and be able to debug issues across the full stack. Having individual `.vscode` configurations within the appropriate folders allow for that. 

If instead we have a single `.vscode` folder at the root level we could not have two debuggers running as VS Code only allows one debugger instance running at a time. Besides the debug configs for the frontend (Node) and backend (NET Core) vary greatly and can lead to confusion if they were to be shared under a single `.vscode` folder.

As I said is a matter of preference, let's discuss what the team prefers and I can adapt to that. The only benefit I see of having a .vscode folder at the root level is that you can open the whole thing "as a solution", but it doesn't let you debug frontend and backend together like VS proper would.

PR #50, opened 2020-03-03
PIMS-1014, PIMS-979 POST/GET Access Requests

devinleighsmith:
Notes:
The extra changes to the automapper make our mappings pass automapper validation. Included in this push as it relates to test changes.
I made the decision to group the two new access requests under the user controller, but we could put these in an admin access request controller and non-admin request controller instead.
The restrictions on 1 agency/role are in the add access request controller.

### DIFF
--- a/backend/api/Areas/Admin/Controllers/UserController.cs
+++ b/backend/api/Areas/Admin/Controllers/UserController.cs
@@ -132,9 +132,12 @@ namespace Pims.Api.Areas.Admin.Controllers
         /// Gets all of the access requests that have been submitted to the system.
         /// </summary>
         /// <returns></returns>
-        [HttpGet("/api/accessRequests")]
+        [HttpGet("/api/access/requests")]
         public IActionResult GetAccessRequests(int page = 1, int quantity = 10, string sort = null, bool? isGranted = null)
         {
+            if (page < 1) page = 1;
+            if (quantity < 1) quantity = 1;
+            if (quantity > 20) quantity = 20;
             Entity.Models.Paged<Entity.AccessRequest> result = _pimsAdminService.User.GetAccessRequestsNoTracking(page, quantity, sort, isGranted);
             var entities = _mapper.Map<Api.Models.AccessRequestModel[]>(result.Items);
             var paged = new Pims.Dal.Entities.Models.Paged<Api.Models.AccessRequestModel>(entities, page, quantity, result.Total);
--- a/backend/api/Areas/Admin/Controllers/UserController.cs
+++ b/backend/api/Areas/Admin/Controllers/UserController.cs
@@ -5,7 +5,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Model = Pims.Api.Areas.Admin.Models;
-using Entities = Pims.Dal.Entities;
+using Entity = Pims.Dal.Entities;
 using Pims.Dal.Services.Admin;
 using Microsoft.AspNetCore.Authorization;
 
@@ -92,7 +92,7 @@ namespace Pims.Api.Areas.Admin.Controllers
         [HttpPost]
         public IActionResult AddUser([FromBody] Model.UserModel model)
         {
-            var entity = _mapper.Map<Entities.User>(model); // TODO: Return bad request.
+            var entity = _mapper.Map<Entity.User>(model); // TODO: Return bad request.
             var addedEntity = _pimsAdminService.User.Add(entity);
             var user = _mapper.Map<Model.UserModel>(addedEntity);
             return new JsonResult(user);
@@ -106,7 +106,7 @@ namespace Pims.Api.Areas.Admin.Controllers
         [HttpPut]
         public IActionResult UpdateUser([FromBody] Model.UserModel model)
         {
-            var entity = _mapper.Map<Entities.User>(model);
+            var entity = _mapper.Map<Entity.User>(model);
             var updatedEntity = _pimsAdminService.User.Update(entity);
 
             if (updatedEntity == null) return BadRequest("Item does not exist");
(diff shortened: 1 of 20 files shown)


backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -122,11 +122,24 @@ public IActionResult UpdateUser([FromBody] Model.UserModel model)
...
         }
+
+        /// <summary>
+        /// Gets all of the access requests that have been submitted to the system.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("/api/accessRequests")]
+        public IActionResult GetAccessRequests(int page = 1, int quantity = 10, string sort = null, bool? isGranted = null)
+        {
+            Entity.Models.Paged<Entity.AccessRequest> result = _pimsAdminService.User.GetAccessRequestsNoTracking(page, quantity, sort, isGranted);
```
Fosol: Would love to see a one-liner that gives us the paged results in the correct types (model).  Not this PR.

backend/api/Controllers/UserController.cs
```diff
@@ -32,11 +39,13 @@ public class UserController : ControllerBase
         /// <param name="logger"></param>
         /// <param name="configuration"></param>
         /// <param name="requestClient"></param>
```
Fosol: Missing comment for parameters.  For all the value they bring us currently :P  One day...

backend/api/Controllers/UserController.cs
```diff
@@ -66,6 +75,26 @@ public async Task<IActionResult> UserInfo()
 
             return await response.HandleResponseAsync<MembershipModel.User>();
         }
+
+        /// <summary>
+        /// Allows a user to submit an access request to the system, associating a role and agency to their user.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost("/api/accessRequest")]
```
Fosol: I'm a fan of this format for routes `/api/access/request`

backend/api/Helpers/IRequestClient.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Pims.Api.Helpers
+{
+    public interface IRequestClient
```
Fosol: Thanks for making this

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -87,6 +87,24 @@ public override void Remove(User entity)
...
+        /// <summary>
+        /// Get all the access requests that users have submitted to the system
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        /// <param name="sort"></param>
+        /// <param name="isGranted"></param>
+        public Paged<AccessRequest> GetAccessRequestsNoTracking(int page = 1, int quantity = 10, string sort = null, bool? isGranted = null)
+        {
+            var query = this.Context.AccessRequests.AsNoTracking();
```
Fosol: Might want to handle bad argument values like `page= <1`, etc.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -87,6 +87,24 @@ public override void Remove(User entity)
...
+            if (isGranted.HasValue)
+            {
+                query = query.Where(request => request.IsGranted == isGranted);
+            }
+            var accessRequests = query.Skip((page - 1) * quantity).Take(quantity);
```
Fosol: Might add the basic sort here in the query.  Not critical at this point.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -64,6 +64,19 @@ public User Activate()
...
+
+        /// <summary>
+        /// Activate the new authenticated user with the PIMS datasource.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public AccessRequest AddAccessRequest(AccessRequest request)
+        {
+            if (request == null) throw new ArgumentNullException();
+            this.Context.AccessRequests.Add(request);
```
Fosol: Should add the `CreatedById` to the entity.

backend/api/Controllers/UserController.cs
devinleighsmith: I'll fix this.

backend/api/Controllers/UserController.cs
devinleighsmith: sure, I'll change that.

backend/dal/Services/Admin/Concrete/UserService.cs
devinleighsmith: I think I'll leave that out for the moment.

backend/dal/Services/Admin/Concrete/UserService.cs
devinleighsmith: Agreed.

backend/dal/Services/Concrete/UserService.cs
devinleighsmith: I thought we had some of those base entity fields being filled in automatically, is created by not on that list? I feel like it should be.

devinleighsmith: I see, that only applies to the service type overriding the base class.

How would you feel about extending the commit transaction method to save this information for us? perhaps not with this PR, but something to consider.

backend/api/Controllers/UserController.cs
asanchezr: In the interest of keeping a RESTful approach it should be:

GET /api/access/requests  -- get many
POST /api/access/requests -- add one
PUT /api/access/requests/[:id] -- update one
DELETE /api/access/request/[:id] -- delete one

I've seen APIs where you have both the plural and singular and it gets confusing (in my opinion). Eg, /api/users and /api/user/10

Fosol: Ha I generally lean towards using the plural and singular.  But it's arbitrary, as long as we're consistent.

backend/dal/Services/Concrete/UserService.cs
Fosol: I would prefer to move it to the transaction method.  It requires a bit of extra leg work, as we need every object that is being submitted for an add/update.  I'll look into this later for Tech-debt

backend/api/Controllers/UserController.cs
asanchezr: Again, something to consider and it's based on preference so I'll leave to you two to decide. But I have always thought about REST endpoints as "resources" or "arrays of things" so the naming should follow that mental model (at least for me)

```
GET  /orders          <---> orders 
POST /orders          <---> orders.push(data)
GET  /orders/1        <---> orders[1]
PUT  /orders/1        <---> orders[1] = data
GET  /orders/1/lines  <---> orders[1].lines
POST /orders/1/lines  <---> orders[1].lines.push(data) 
```

By having two different endpoints /orders and /order/1 it becomes confusing when to use one or the other and I like simplicity in the sense that I don't have to remember which one to use - I always use the plural noun and let the HTTP verb (GET, POST, etc) dictate what kind of action is being taken on the "array/REST resource"

At least, that's what I've seen consistently being predicated by larger organizations & teams but again, I'll leave up to you.

https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#define-operations-in-terms-of-http-methods
https://cloud.google.com/blog/products/api-management/restful-api-design-nouns-are-good-verbs-are-bad
https://blog.octo.com/wp-content/uploads/2014/12/OCTO-Refcard_API_Design_EN_3.0.pdf

Fosol: I'm fine with either way.  I've always leaned towards using plural and singular routes to express what they do in human readable format.

The benefit I like about your approach is you can set the default route at the top of the controller and just add the param/path parts to the endpoints.

PR #51, opened 2020-03-03
PIMS-945 Adding way to import Todd's data.

Fosol:
New endpoint `/api/admin/tools/import/properties`
I've created Postman collections to import the data.  Still requires a bit of manual effort.  Would love to find a clean way to iterate through a JSON dataset and submit batches.

### DIFF
--- a/backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
+++ b/backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
@@ -51,7 +51,7 @@ namespace Pims.Api.Areas.Tools.Helpers
         /// Adds or updates the property in the datasource.
         /// Determines if the property is a parcel or a building.
         /// Massages some of the data to align with expected values.
-        /// /// </summary>
+        /// </summary>
         /// <param name="properties"></param>
         /// <returns></returns>
         public IEnumerable<Entity.Parcel> AddUpdateProperties(IEnumerable<PropertyModel> properties)
@@ -61,8 +61,8 @@ namespace Pims.Api.Areas.Tools.Helpers
             {
                 _logger.LogDebug($"Add/Update property pid:{property.ParcelId}, type:{property.PropertyType}, fiscal:{property.FiscalYear}");
 
-                var valid_pid = int.TryParse(property.ParcelId?.Replace("-", ""), out int pid);
-                if (!valid_pid) continue;
+                var validPid = int.TryParse(property.ParcelId?.Replace("-", ""), out int pid);
+                if (!validPid) continue;
 
                 // Change the agency code.
                 switch (property.Agency)
@@ -104,7 +104,6 @@ namespace Pims.Api.Areas.Tools.Helpers
         /// Get or create a new agency for the specified property.
         /// </summary>
         /// <param name="property"></param>
-        /// <param name="agencies"></param>
         /// <returns></returns>
         private Entity.Agency GetOrCreateAgency(PropertyModel property)
         {
@@ -122,20 +121,20 @@ namespace Pims.Api.Areas.Tools.Helpers
             if (!String.IsNullOrWhiteSpace(property.SubAgency))
             {
                 var code = new string(property.SubAgency.GetFirstLetterOfEachWord(true).Take(6).ToArray());
-                var sub_agency = _agencies.FirstOrDefault(a => a.Code == code.Trim() && a.ParentId == agency.Id);
+                var subAgency = _agencies.FirstOrDefault(a => a.Code == code.Trim() && a.ParentId == agency.Id);
 
-                if (sub_agency == null)
+                if (subAgency == null)
                 {
-                    sub_agency = new Entity.Agency(code.Trim(), property.SubAgency)
+                    subAgency = new Entity.Agency(code.Trim(), property.SubAgency)
                     {
                     ParentId = agency.Id,
                     Parent = agency
                     };
-                    _agencies.Add(sub_agency);
-                    _logger.LogDebug($"Adding sub-agency '{sub_agency.Code}' - '{agency.Name}', parent: '{sub_agency.Parent.Code}'.");
+                    _agencies.Add(subAgency);
+                    _logger.LogDebug($"Adding sub-agency '{subAgency.Code}' - '{agency.Name}', parent: '{subAgency.Parent.Code}'.");
                 }
 
-                return sub_agency;
+                return subAgency;
             }
 
             return agency;
@@ -148,7 +147,6 @@ namespace Pims.Api.Areas.Tools.Helpers
         /// </summary>
         /// <param name="code"></param>
         /// <param name="name"></param>
-        /// <param name="cities"></param>
         /// <param name="attempts"></param>
         /// <returns></returns>
         private Entity.City GetOrCreateCity(string code, string name, int attempts = 1)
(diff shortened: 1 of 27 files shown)


backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// Updates the specified agency in the datasource.
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <returns></returns>
+        public override Agency Update(Agency entity)
+        {
+            entity.ThrowIfNull(nameof(entity));
+
+            var building_construction_type = this.Context.Agencies.Find(entity.Id);
```
devinleighsmith: I think building_construction_type may be copy pasta.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -0,0 +1,410 @@
...
+        /// </summary>
+        /// <param name="pimsAdminService"></param>
+        /// <param name="logger"></param>
+        public ImportPropertiesHelper(IPimsAdminService pimsAdminService, ILogger logger)
+        {
+            _pimsAdminService = pimsAdminService;
+            _logger = logger;
+
+            // Preload lookup lists so that they can be references quickly.
+            _buildingConstructionTypes = _pimsAdminService.BuildingConstructionType.GetAll().ToList();
```
devinleighsmith: Do we have any performance considerations here? it seems like we should try and initialize this class as little as possible.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -0,0 +1,410 @@
...
+            var entities = new List<Entity.Parcel>();
+            foreach (var property in properties)
+            {
+                _logger.LogDebug($"Add/Update property pid:{property.ParcelId}, type:{property.PropertyType}, fiscal:{property.FiscalYear}");
+
+                var valid_pid = int.TryParse(property.ParcelId?.Replace("-", ""), out int pid);
+                if (!valid_pid) continue;
+
+                // Change the agency code.
+                switch (property.Agency)
```
devinleighsmith: At some point we may need to extend this, but for now I think its fine.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -0,0 +1,410 @@
...
+                return sub_agency;
+            }
+
+            return agency;
+        }
+
+        /// <summary>
+        /// Get or create a new city for the specified code.
+        /// This is a recursive function so that it can generate a new city with a unique code.
+        /// It has a limitation of which the code must be 4 characters and therefore it can only handle 9 variations (i.e. BRe1, BRe2...)
```
devinleighsmith: if only 9 variations are allowed, shouldn't we put a cap on the max size of attempts?

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -0,0 +1,410 @@
...
+
+            return city;
+        }
+
+        /// <summary>
+        /// Get or create a new city for the specified name.
+        /// It will generate a unique city code based on the first initials of the city name, plus the first three characters of the last word.
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="cities"></param>
```
devinleighsmith: cities isn't a parameter.

devinleighsmith: actually, just double check the param comments on this file, there are a few mismatches.

backend/dal/Helpers/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -22,20 +22,26 @@ public static IServiceCollection AddPimsService(this IServiceCollection services
         }
 
         /// <summary>
-        /// Add PimsAdminService objects to the dependency injection service collection.
+        /// /// Add PimsAdminService objects to the dependency injection service collection.
```
devinleighsmith: /nit

backend/dal/Helpers/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// Extracts the first letter of each word from the text and returns the value.
+        /// </summary>
+        /// <param name="text"></param>
+        /// <param name="uppercase"></param>
+        /// <returns></returns>
+        public static string GetFirstLetterOfEachWord(this string text, bool uppercase = true)
+        {
+            var result = new StringBuilder();
+            var split = text.Split();
```
devinleighsmith: You may need: StringSplitOptions.RemoveEmptyEntries here to handle double spaces or trailing spaces.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -0,0 +1,410 @@
...
+                    b_e.Address.Postal = property.Postal;
+                }
+            }
+
+            // Add a new evaluation for each year.
+            if (!b_e.Evaluations.Any(e => e.FiscalYear == fiscal))
+            {
+                var evaluation = new Entity.BuildingEvaluation(fiscal, b_e)
+                {
+                AssessedValue = float.Parse(property.AssessedValue)
```
devinleighsmith: I'm guessing this is the only field that is in the data?

backend/dal/Services/Admin/Concrete/AgencyService.cs
asanchezr: /nit Also this doesn't conform to usual C# naming conventions (camelCase)

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
Fosol: The reason they're pulled in this way is for that reason.  However multiple separate calls to this (i.e. one parcel at a time) result in a loss of performance.  Since this is more for importing, I lean towards the first.

Yes you are correct, you don't want to initialize this class over and over.  And it really shouldn't be anyways.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
Fosol: Hardcode lists are horrible, but for our use-case I wasn't willing to spend additional time.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
Fosol: The result of going over would be a DbUpdateException as they would attempt to insert a city with a code with a length greater than 4 characters.  I'm hesitant to limit the use here, as it would indicate we probably need to modify the DB to accept more characters.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
Fosol: Nice catch.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
Fosol: correct

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: I use underscores between words in vars to denote inline scoped variables.  But I can go either way on this.

PR #52, opened 2020-03-04
PIMS-969 Authenticated Guest Page skeleton.

devinleighsmith:
The authenticated guest page is only accessible if:
1) the user logs in (protected behind private route).
2) the user has no granted roles or has just been activated.

The auth page is just a placeholder for now. The story to add the form is PIMS-970

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,59 +1,63 @@
-import React, { Component } from 'react';
+import React, { useEffect, useState } from 'react';
 import { Container, Row } from 'react-bootstrap';
 import './App.scss';
-import { BrowserRouter as Router, Route } from 'react-router-dom';
+import { BrowserRouter as Router, Route, Redirect } from 'react-router-dom';
+import { useDispatch, useSelector } from 'react-redux'
 import MapView from './pages/MapView';
+import GuestAccessPage from './pages/GuestAccessPage';
 import PrivateRoute from './PrivateRoute';
 import Login from './pages/Login';
 import Header from './components/navigation/Header';
 import Footer from './components/navigation/Footer';
-import { withKeycloak, ReactKeycloakInjectedProps } from '@react-keycloak/web';
 import { Spinner } from 'react-bootstrap';
-import { KeycloakInstance } from 'keycloak-js';
-import configureStore from 'configureStore';
+import { RootState } from 'reducers/rootReducer';
+import configureStore from "configureStore";
+import { useKeycloak } from '@react-keycloak/web';
+import { getActivateUserAction } from 'actionCreators/authActionCreator';
 import '../node_modules/bootstrap/dist/css/bootstrap.min.css';
+import { IGenericNetworkAction } from 'actions/genericActions';
 
 export const store = configureStore();
-//import Debug from './components/Debug/Debug';
 
-interface IState {
-  keycloak?: KeycloakInstance<'native'>;
-  keycloakInitialized: boolean;
-  keycloakUserLoaded: boolean;
-}
+const App = () => {
+  const { keycloak } = useKeycloak();
+  const [keycloakUserLoaded, setkeycloakUserLoaded] = useState({});
+  const dispatch = useDispatch();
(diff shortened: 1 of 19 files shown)


frontend/src/App.tsx
```diff
@@ -1,55 +1,63 @@
...
-              path="/mapview"
-              component={MapView}
-            />
-          </Row>
-          <Footer></Footer>
-        </Container>
-      </Router>
-    ) : <Spinner animation="border"></Spinner>;
+  const GoToGuestPage = () => {
+    if ((activated && activated.status === 201) || !keycloak?.realmAccess?.roles?.length) {
```
Fosol: I don't think this works, but I could be incorrect.  I think by default a user might have roles from keycloak, but they aren't roles for pims.  Need to double check this.

frontend/src/actionCreators/authActionCreator.tsx
```diff
@@ -0,0 +1,25 @@
+import { showLoading, hideLoading } from "react-redux-loading-bar";
+import { request, success, error } from "actions/genericActions";
+import * as reducerTypes from "constants/reducerTypes";
+import * as API from "constants/API";
+import { ENVIRONMENT } from "constants/environment";
+import CustomAxios from "customAxios";
+import { AxiosResponse } from "axios";
+import { createRequestHeader } from "utils/RequestHeaders";
+
+export const getActivateUserAction = () => (dispatch: Function) => {
```
Fosol: Might only want to send this request if the user doesn't have any roles.

frontend/src/App.tsx
devinleighsmith: We discussed this IRL. Due to some incoming changes related to a GUEST role, this will need to change anyways. Possibly we'll just check for the presence of an agency instead. As a result we'll leave this for now.

frontend/src/actionCreators/authActionCreator.tsx
devinleighsmith: Will leave this for now do to the negligible performance impact.

frontend/src/constants/reducerTypes.tsx
```diff
@@ -8,3 +8,6 @@ export const GET_PARCEL_DETAIL = 'GET_PARCEL_DETAIL';
 // Lookup Codes
 export const LOOKUP_CODE = 'lookupCode';
 export const GET_LOOKUP_CODES = 'GET_LOOKUP_CODES';
+
+// Auth
+export const POST_ACTIVATE_USER = "activateUser";
```
asanchezr: I'm curious to learn how the automatic linting didn't kick in here as one of the rules we have is to use single quotes instead of double quotes. That's something the linter is supposed to auto-fix for us before we are able to commit changes. Did you run `npm install` on your local copy? Note that my PR added new dependencies...

frontend/src/reducers/rootReducer.tsx
```diff
@@ -2,10 +2,19 @@ import { combineReducers } from 'redux';
 import parcelsReducer from 'reducers/parcelsReducer';
 import lookupCodeReducer from 'reducers/lookupCodeReducer';
 import * as reducerTypes from 'constants/reducerTypes';
+import networkReducer from "./networkReducer";
```
asanchezr: We should use single quotes here as per our JS styleguide. This should be auto-fixed by the linter thought...

frontend/src/pages/GuestAccessPage.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return <Container fluid={true}>
+    <Row>
+      <Col>
+        <h1>Welcome to PIMS!</h1>
+      </Col>
+    </Row>
+  </Container >
+}
+
+export default GuestAccessPage;
```
asanchezr: I like you are using **pages** (or container components). Nice!

frontend/src/constants/reducerTypes.tsx
devinleighsmith: I ran npm ci. perhaps merged changes are exempt?

PR #55, opened 2020-03-06
Pims 970

devinleighsmith:
Added role lookup.
Converted lookups to use string ids instead of ints (because of guid based entities).
Fixed post access request service.
Updated access request entity/configuration.
Minor test updates.
Added adapter for lookupdropdown for use with formik
Added post to Access Request service to frontend.
Added basic formik form to make an Access Request.


backend/dal/Migrations/20200306023735_AccessRequestRoleAgency.cs
```diff
@@ -0,0 +1,194 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class AccessRequestRoleAgency : Migration
```
Fosol: I think at this point we need to merge these migrations.  We're not supporting production yet, so I would recommend recreating the initial migration.  We can refresh the DB.

backend/dal/Services/Concrete/LookupService.cs
```diff
@@ -62,6 +62,11 @@ public IEnumerable<Province> GetProvincesNoTracking()
         {
             return this.Context.Provinces.AsNoTracking().OrderBy(a => a.Name).ToArray();
         }
+
+        public IEnumerable<Role> GetRolesNoTracking()
```
Fosol: Add comments for intellisense

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -72,8 +74,18 @@ public User Activate()
         /// <returns></returns>
         public AccessRequest AddAccessRequest(AccessRequest request)
         {
-            if (request == null) throw new ArgumentNullException();
+            if (request == null || request.Agencies == null || request.Roles == null) throw new ArgumentNullException();
```
Fosol: `throw new ArgumentNullException(nameof(request));`

backend/entities/AccessRequest.cs
```diff
@@ -44,14 +44,14 @@ public class AccessRequest : BaseEntity
         /// get - the list of agencies that the user is requesting to be added to.
         /// </summary>
         /// <returns></returns>
-        public ICollection<Agency> Agencies { get; set; } = new List<Agency>();
+        public ICollection<AccessRequestAgency> Agencies { get; set; } = new List<AccessRequestAgency>();
```
Fosol: Don't bother including `set;`.  It's not a desired way to interact with entity collections.

backend/entities/AccessRequest.cs
```diff
@@ -44,14 +44,14 @@ public class AccessRequest : BaseEntity
...
 
         /// <summary>
         /// get - the list of roles this user is requesting.
         /// </summary>
         /// <typeparam name="UserRole"></typeparam>
         /// <returns></returns>
-        public ICollection<Role> Roles { get; set; } = new List<Role>();
+        public ICollection<AccessRequestRole> Roles { get; set; } = new List<AccessRequestRole>();
```
Fosol: Don't bother including `set;`.  It's not a desired way to interact with entity collections.

backend/entities/LookupEntity.cs
```diff
@@ -29,6 +31,7 @@ public abstract class LookupEntity : BaseEntity
         /// get/set - The sort order of the lookup item.
         /// </summary>
         /// <value></value>
+        [NotMapped]
```
Fosol: Why is this not mapped?

backend/entities/Role.cs
```diff
@@ -6,33 +6,21 @@ namespace Pims.Dal.Entities
     /// <summary>
     /// Role class, provides an entity for the datamodel to manage roles.
     /// </summary>
-    public class Role : BaseEntity
+    public class Role : LookupEntity
```
Fosol: This will result in a `Code` property in the `Role`.  I don't think this is the desired implementation of roles.

backend/dal/Migrations/20200306023735_AccessRequestRoleAgency.cs
devinleighsmith: Sure, I can do that.

backend/entities/Role.cs
devinleighsmith: No, it won't. It is extending LookupEntity not CodeEntity. LookupEntity doesn't have codes.

backend/entities/LookupEntity.cs
devinleighsmith: I'll add this when I regenerate the migration.

backend/entities/Role.cs
Fosol: Ah my mistake!

backend/dal/Migrations/20200306222350_AccessRequest.cs
```diff
@@ -0,0 +1,222 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class AccessRequest : Migration
```
Fosol: You just merged the AccessRequest migration correct?

devinleighsmith: yes.

PR #57, opened 2020-03-08
PIMS-938 Imagery Layer integration

asanchezr:
- add `BasemapToggle` component
- docker tweaks for better local development


docker-compose.yml
```diff
@@ -79,7 +79,7 @@ services:
     env_file: ./frontend/.env
     environment:
       - NODE_ENV=development
-      - KEYCLOAK_URL=http://host.docker.internal:8080/auth/admin/realms/quartech/users
```
asanchezr: We are not using this env var in the frontend so I removed it

docker-compose.yml
```diff
@@ -79,7 +79,7 @@ services:
...
+      - API_URL=http://host.docker.internal:5000
```
asanchezr: The **API_URL** was hardcoded in `setupProxy.js`. I made it an env var to support the use case of developing locally on the frontend while targeting a remote API; e.g. pims-dev.pathfinder.gov.bc.ca/api

frontend/public/basemaps.json
```diff
@@ -0,0 +1,16 @@
...
+      "thumbnail": "streets.jpg"
+    },
+    {
+      "name": "Satellite",
+      "url": "https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}",
+      "attribution": "Tiles &copy; Esri &mdash; Source: Esri, DigitalGlobe, GeoEye, Earthstar Geographics, CNES/Airbus DS, USDA, USGS, AeroGRID, IGN, and the GIS User Community",
+      "thumbnail": "satellite.jpg"
+    }
+  ]
+}
```
asanchezr: I put this configuration in an external JSON file because I felt it would allow us to overwrite if needed with configurations specific to different environments (DEV, TEST, PROD) if needed. This follows the same pattern of the `keycloack.json` file.

frontend/src/components/maps/BasemapToggle.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+type BasemapToggleProps = {
+  baseLayers: BaseLayer[];
+  onToggle?: (e: BasemapToggleEvent) => void;
+};
+
+const BasemapToggle: React.FC<BasemapToggleProps> = props => {
+  // TODO: fade out when map is panning/zooming
+  const [updating] = useState(false);
```
asanchezr: As per the **TODO** comment this to signal a future enhancement where the map widgets would kind of fade out when you drag/zoom around the map. This is not part of this story and as such I left it out but wanted to put the initial placeholder code for that

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -65,12 +69,26 @@ const Map: React.FC<MapProps> = props => {
...
   }, [agencyId, propertyClassificationId]);
+
+  // fetch GIS base layers configuration from /public folder
+  useEffect(() => {
+    axios.get('basemaps.json').then(result => {
+      setBaseLayers(result.data?.basemaps);
+      setActiveBasemap(result.data?.basemaps?.[0]);
+    });
+  }, []);
+
```
asanchezr: I made the choice here to load the **basemaps.json** dynamically via a fetch request (with axios) so we have the ability to override the json configuration upon deployment to Openshift via config maps (like we have done with keycloack.json). 

I believe React/webpack support loading json files via `import data from basemaps.json` but that would mean the json file is included in the JS bundle and we wouldn't be able to override/change it without a code change and re-deploy.

frontend/src/setupProxy.js
```diff
@@ -4,7 +4,7 @@ module.exports = function(app) {
   app.use(
     '/api',
     proxy({
-      target: 'http://backend/',
+      target: process.env.API_URL || 'http://backend/',
```
asanchezr: I made this a bit more flexible by using an ENV var (if provided)

PR #58, opened 2020-03-09
PIMS-971

devinleighsmith:
Admin page.
Auth convenenience methods and role based route filtering.
Add navigation to app nav bar drop down.


frontend/package.json
```diff
@@ -58,7 +58,7 @@
     "lint-staged": "10.0.8",
     "prettier": "1.19.1",
     "react-test-renderer": "16.12.0",
-    "redux-mock-store": "1.5.4"
+    "redux-mock-store": "^1.5.4"
```
asanchezr: Hi, redux-mock-store latest version is already 1.5.4 (as of today) so why would put the "^" in front of it? As we saw with some of our dependencies, they usually break the Docker build (npm install) when they are not locked down to exact versions. 

By freezing these dependencies we want to achieve repeatable deployments and make sure that every developer is testing on the very same codebase.

For updating dependencies, there are a few free tools out there that we could explore to automate the update process. I like how [renovate](https://github.com/renovatebot/renovate) or [dependabot](https://dependabot.com/#how-it-works) (acquired by github) issue PRs with the updates and gives the team a chance to review whenever there's a new dependency that needs updating.

frontend/src/PrivateRoute.tsx
```diff
@@ -1,24 +1,31 @@
...
 
 interface IPrivateRouteProps extends RouteProps {
   // tslint:disable-next-line:no-any
   component: any;
+  role?: string;
 }
 
 const PrivateRoute = (props: IPrivateRouteProps) => {
-  const { keycloak } = useKeycloak();
+  const keycloak = useKeycloakWrapper();
```
asanchezr: 👍 I like how you abstracted the keycloak internals!

frontend/src/pages/Login.tsx
```diff
@@ -7,9 +7,12 @@ import './Login.scss';
 import { IGenericNetworkAction } from 'actions/genericActions';
 import { RootState } from 'reducers/rootReducer';
 import { NEW_PIMS_USER } from 'actionCreators/authActionCreator';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import { PROPERTY_VIEW } from 'constants/strings';
 
 const Login = () => {
   const { keycloak } = useKeycloak();
```
asanchezr: Could you please share why are we using both keycloak and it's wrapper together here? Wouldn't the wrapper be able to replace all existing functionality?

devinleighsmith: It certainly can.

frontend/package.json
devinleighsmith: Yes, adding the ^ was not intentional so using one of those tools might help.

PR #60, opened 2020-03-10
PIMS-973 add paginated access request list.

devinleighsmith:


backend/entities/AccessRequest.cs
```diff
@@ -44,14 +44,14 @@ public class AccessRequest : BaseEntity
         /// get - the list of agencies that the user is requesting to be added to.
         /// </summary>
         /// <returns></returns>
-        public ICollection<AccessRequestAgency> Agencies { get; } = new List<AccessRequestAgency>();
+        public ICollection<AccessRequestAgency> Agencies { get; private set; } = new List<AccessRequestAgency>();
```
Fosol: Not sure the value of a `private set` for EF entities.

backend/test/Controllers/Admin/UserControllerTest.cs
```diff
@@ -71,6 +55,25 @@ public UserControllerTest()
             _userController = _helper.CreateAdminUserController(user);
             _mapper = _helper.GetService<IMapper>();
             _pimsService = _helper.GetService<Mock<IPimsAdminService>>();
+
+            _expectedAccessRequest.Agencies.Add(new AccessRequestAgency()
```
Fosol: I did the same thing in my PR :P

frontend/src/constants/reducerTypes.tsx
```diff
@@ -10,5 +10,7 @@ export const LOOKUP_CODE = 'lookupCode';
 export const GET_LOOKUP_CODES = 'GET_LOOKUP_CODES';
 
 // Auth
+export const ACCESS_REQUEST = 'accessRequest';
```
Fosol: What are these consts for?

backend/entities/AccessRequest.cs
devinleighsmith: Profile wasn't working without this.

frontend/src/constants/reducerTypes.tsx
devinleighsmith: consistent use throughout the application. This value is needed by the reducer and also by the selector, and action. Using the constants saves me from having to remember the keys these values are saved/retrieved by.

frontend/package.json
```diff
@@ -6,6 +6,7 @@
     "@bcgov/bc-sans": "1.0.1",
     "@react-keycloak/web": "1.0.0-alpha.6",
     "@reduxjs/toolkit": "1.2.3",
+    "@types/react-paginate": "6.2.1",
```
asanchezr: ^ The type reference should be a **devDependency** as it only provides intellisense and type support for TypeScript. Just a minor tweak that can be dealt as tech-debt later

frontend/src/reducers/accessRequestReducer.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const initialState: IAccessRequestState = {
+  pagedAccessRequests: { page: 0, total: 0, quantity: 0, items: [] },
+};
+
+const accessRequestReducer = (state = initialState, action: IStoreAccessRequestsAction) => {
+  switch (action.type) {
+    case actionTypes.STORE_ACCESS_REQUESTS:
+      console.log('storing access request');
```
asanchezr: /nit console.logs are fine for debugging but we should try to take them out out public PRs. We might need to automate this as myself keep forgetting to remove them too!

PR #61, opened 2020-03-10
PIMS 954, 1002, 1018, 1019, 1017

jdtoombs:
UserControllerTests have decided to start giving me some errors today.. they will be included in the next PR - fixing them right now.

### DIFF
--- a/backend/test/Controllers/CodeLookupControllerTest.cs
+++ b/backend/test/Controllers/CodeLookupControllerTest.cs
@@ -13,6 +13,7 @@ using Model = Pims.Api.Models;
 using AutoMapper;
 using Pims.Api.Helpers.Profiles;
 using Pims.Dal;
+using Newtonsoft.Json;
 
 namespace Pims.Api.Test.Controllers
 {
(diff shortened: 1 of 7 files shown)


backend/test/Controllers/CodeLookupControllerTest.cs
```diff
@@ -114,6 +115,55 @@ public void GetRoleCodes()
...
+            JsonResult roleAction = Assert.IsType<JsonResult>(roleResult);
+            JsonResult classificationAction = Assert.IsType<JsonResult>(classificationResult);
+
+            string allResult = JsonConvert.SerializeObject(actionResult.Value);
+            string agenciesResult = JsonConvert.SerializeObject(agencyAction.Value);
+            string rolesResult = JsonConvert.SerializeObject(roleAction.Value);
+            string classificationsResult = JsonConvert.SerializeObject(classificationAction.Value);
+
+            // Removing corresponding []'s as GetAll returns [{one,combined,list}]
+            Assert.StartsWith(agenciesResult.Remove(agenciesResult.Length - 1, 1), allResult);
```
jdtoombs: Maybe should have just combined the results and used one assert here..

backend/test/Controllers/CodeLookupControllerTest.cs
```diff
@@ -13,6 +13,7 @@
 using AutoMapper;
 using Pims.Api.Helpers.Profiles;
 using Pims.Dal;
+using Newtonsoft.Json;
```
Fosol: I'd recommend not using NewtonSoft and use the built in `System.Text.Json`

backend/test/Controllers/CodeLookupControllerTest.cs
```diff
@@ -114,6 +115,55 @@ public void GetRoleCodes()
...
+        public void GetAll()
+        {
+            // Arrange
+            var role = new Entity.Role
+            {
+                Id = Guid.NewGuid(),
+                Name = "Ministry of Health",
+                Description = "The Ministry of Health"
+            };
+            _pimsService.Setup(m => m.Lookup.GetRolesNoTracking()).Returns(new[] { role });
```
Fosol: I'm going to convert all API unit tests in the next sprint to start using better design patterns (see ImportControllerTest).  We need to setup each test properly, right now we're creating one controller/service for all tests.  This can cause undesirable side effects.

frontend/src/components/common/LookupCodeDropdown.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    defaultTitle={'View by Agency \u00A0'}
+    onSelectCode={jest.fn()}
+  />,
+);
+it('Displays item for all agencies given', () => {
+  const items = agencyDropdown.find(Dropdown.Item);
+  expect(items.length).toBe(6);
+});
+
+// Should isDisabled items be hidden from dropdown? Will fail in current implementation
```
devinleighsmith: This is a good question. currently we don't have a mechanism for disabling values in the database, I suggest we address this at that time.

frontend/src/components/common/LookupCodeDropdown.test.tsx
```diff
@@ -0,0 +1,69 @@
+import * as MOCK from 'mocks/filterDataMock';
```
devinleighsmith: I'd like to see a test in here for the 'all' functionality, but that doesn't need to be in this PR. I'll make a note of this.

frontend/src/mocks/filterDataMock.ts
```diff
@@ -0,0 +1,163 @@
+import { ILookupCode } from '../actions/lookupActions';
+import { IParcelDetail, IParcel } from 'actions/parcelsActions';
+import { MapFilterModel } from 'components/maps/leaflet/Map';
+
+export const SELECTEDCLASSIFICATION = {
```
Fosol: Would be preferable if all the mocked data was JSON

frontend/src/components/maps/MapFilterBar.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+  <MapFilterBar
+    agencyLookupCodes={MOCK.AGENCIES}
+    propertyClassifications={MOCK.CLASSIFICATIONS}
+    onSelectAgency={handleAgencyChanged}
+    onSelectPropertyClassification={handlePropertyClassificationChanged}
+  />,
+);
+
+// Could change as project progresses
+it('Currently has 2 drop downs, one for agency one for classifications', () => {
```
devinleighsmith: I'd be tempted to let the snapshot handle tests like this in the future, but feel free to leave this in.

jdtoombs: Thought about that afterwards, I agree. Will let snapshots handle issues like that from now on!

frontend/src/pages/GuestAccessPage.test.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
devinleighsmith: I think I'd like to see tests covering formik's validation logic, but let's leave that for when this form is finalized.

backend/test/Controllers/CodeLookupControllerTest.cs
asanchezr: Agree we need to have a fresh instance on every test

backend/test/Controllers/CodeLookupControllerTest.cs
```diff
@@ -114,6 +115,55 @@ public void GetRoleCodes()
...
+
+            string allResult = JsonConvert.SerializeObject(actionResult.Value);
+            string agenciesResult = JsonConvert.SerializeObject(agencyAction.Value);
+            string rolesResult = JsonConvert.SerializeObject(roleAction.Value);
+            string classificationsResult = JsonConvert.SerializeObject(classificationAction.Value);
+
+            // Removing corresponding []'s as GetAll returns [{one,combined,list}]
+            Assert.StartsWith(agenciesResult.Remove(agenciesResult.Length - 1, 1), allResult);
+            Assert.Contains(classificationsResult.Substring(1, classificationsResult.Length -2), allResult);
+            Assert.EndsWith(rolesResult.Substring(1), allResult);
```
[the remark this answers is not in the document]
jdtoombs: Ahh yes that makes sense, rookie mistake on my part. Will change it moving forward!

PR #62, opened 2020-03-10
Merging DB migration back to initial.

Fosol:
Updating Import Tool to refresh token.


backend/dal/Migrations/PIMSContextModelSnapshot.cs
```diff
@@ -937,9 +937,6 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("DATETIME2")
                         .HasDefaultValueSql("GETUTCDATE()");
 
-                    b.Property<bool>("IsDisabled")
```
devinleighsmith: Hmm... what happened here?

backend/tools/import/appsettings.json
```diff
@@ -6,9 +6,13 @@
...
     }
   },
+  "Keycloak": {
+    "TokenUrl": "https://sso-dev.pathfinder.gov.bc.ca/auth/realms/quartech/protocol/openid-connect/token",
+    "ClientId": "pims-app"
+  },
   "Api": {
-    "Url": "http://pims-dev.pathfinder.gov.bc.ca/api/tools/import/properties",
-    "Method": "POST"
+    "ImportUrl2": "http://pims-dev.pathfinder.gov.bc.ca/api/tools/import/properties",
```
devinleighsmith: /nit naming

backend/dal/Migrations/PIMSContextModelSnapshot.cs
Fosol: This is auto-generated.

backend/tools/import/appsettings.json
Fosol: Ah forgot I made that change to test something.  Thanks for catching it.

PR #64, opened 2020-03-12
PIMS-1250 Add config and secrets to support health checks and rocketc…

devinleighsmith:
…hat notifications.

Please see comments, as there are some things to think about with this functionality.
This is using this package: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks

### DIFF
--- a/backend/api/Controllers/LookupController.cs
+++ b/backend/api/Controllers/LookupController.cs
@@ -57,7 +57,7 @@ namespace Pims.Api.Controllers
         /// Get all of the role code values
         /// </summary>
         /// <returns></returns>
-        [HttpGet("agencies")]
+        [HttpGet("roles")]
         public IActionResult GetRoles()
         {
             var roleCodes = _mapper.Map<Model.CodeModel[]>(_pimsService.Lookup.GetRolesNoTracking());
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -5,15 +5,18 @@ using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using AutoMapper;
+using HealthChecks.UI.Client;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Tokens;
@@ -92,21 +95,21 @@ namespace Pims.Api
                             {
                                 return Task.CompletedTask;
                             },
-                            OnAuthenticationFailed = context =>
+                        OnAuthenticationFailed = context =>
+                        {
+                            context.NoResult();
+                            context.Response.StatusCode = 500;
+                            context.Response.ContentType = "text/plain";
+                            if (Environment.IsDevelopment())
                             {
-                                context.NoResult();
-                                context.Response.StatusCode = 500;
-                                context.Response.ContentType = "text/plain";
-                                if (Environment.IsDevelopment())
-                                {
-                                    return context.Response.WriteAsync(context.Exception.ToString());
-                                }
-                                return context.Response.WriteAsync("An error occurred processing your authentication.");
-                            },
-                            OnForbidden = context =>
-                            {
-                                return Task.CompletedTask;
+                                return context.Response.WriteAsync(context.Exception.ToString());
                             }
+                            return context.Response.WriteAsync("An error occurred processing your authentication.");
+                        },
+                        OnForbidden = context =>
+                        {
+                            return Task.CompletedTask;
+                        }
                     };
                 });
 
@@ -115,15 +118,15 @@ namespace Pims.Api
                 options.AddPolicy("Administrator", policy => policy.Requirements.Add(new RealmAccessRoleRequirement("administrator")));
             });
 
+            var cs = Configuration.GetConnectionString("PIMS");
+            var builder = new SqlConnectionStringBuilder(cs);
+            var pwd = Configuration["DB_PASSWORD"];
+            if (!String.IsNullOrEmpty(pwd))
+            {
+                builder.Password = pwd;
+            }
             services.AddDbContext<PimsContext>(options =>
             {
-                var cs = Configuration.GetConnectionString("PIMS");
-                var builder = new SqlConnectionStringBuilder(cs);
-                var pwd = Configuration["DB_PASSWORD"];
-                if (!String.IsNullOrEmpty(pwd))
-                {
-                    builder.Password = pwd;
-                }
                 options.UseSqlServer(builder.ConnectionString);
             });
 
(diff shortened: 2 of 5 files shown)


backend/api/Controllers/LookupController.cs
```diff
@@ -57,7 +57,7 @@ public IActionResult GetAgencies()
         /// Get all of the role code values
         /// </summary>
         /// <returns></returns>
-        [HttpGet("agencies")]
```
devinleighsmith: this is an unrelated bugfix.

backend/api/Startup.cs
```diff
@@ -115,15 +118,15 @@ public void ConfigureServices(IServiceCollection services)
                 options.AddPolicy("Administrator", policy => policy.Requirements.Add(new RealmAccessRoleRequirement("administrator")));
```
devinleighsmith: I recommend hiding whitespace changes for this file.

backend/api/Startup.cs
```diff
@@ -115,15 +118,15 @@ public void ConfigureServices(IServiceCollection services)
...
             });
 
+            var cs = Configuration.GetConnectionString("PIMS");
+            var builder = new SqlConnectionStringBuilder(cs);
+            var pwd = Configuration["DB_PASSWORD"];
+            if (!String.IsNullOrEmpty(pwd))
+            {
+                builder.Password = pwd;
+            }
             services.AddDbContext<PimsContext>(options =>
```
devinleighsmith: Moving the logic to get the connection builder string out of  the if

backend/api/Startup.cs
```diff
@@ -138,6 +141,20 @@ public void ConfigureServices(IServiceCollection services)
...
 
             services.AddAutoMapper(typeof(Startup));
+
+            services.AddHealthChecks()
+                .AddCheck("liveliness", () => HealthCheckResult.Healthy())
+                .AddSqlServer(builder.ConnectionString, tags: new[] { "services" });
+            //TODO: Add a health check for keycloak connectivity.
+            services.AddHealthChecksUI(setupSettings: setup =>
+            {
+                setup.AddHealthCheckEndpoint("liveliness", "http://localhost/live");
```
devinleighsmith: While these are hardcoded, I think that is OK since our pims api server would always also be hosting the health check ui interface.

backend/api/Startup.cs
```diff
@@ -138,6 +141,20 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IKeycloakRequestClient, KeycloakRequestClient>();
 
             services.AddAutoMapper(typeof(Startup));
+
+            services.AddHealthChecks()
+                .AddCheck("liveliness", () => HealthCheckResult.Healthy())
```
devinleighsmith: this is just a simple function that always returns true.

backend/api/Startup.cs
```diff
@@ -138,6 +141,20 @@ public void ConfigureServices(IServiceCollection services)
...
+                .AddSqlServer(builder.ConnectionString, tags: new[] { "services" });
```
devinleighsmith: this executes select 1 against the database in the connection string.

backend/api/Startup.cs
```diff
@@ -138,6 +141,20 @@ public void ConfigureServices(IServiceCollection services)
...
+            //TODO: Add a health check for keycloak connectivity.
+            services.AddHealthChecksUI(setupSettings: setup =>
```
devinleighsmith: This adds an interface that provides the following functionality:
1) View the status of health checks within the application.
2) Send notifications in the event that one or more health checks fail, see below for more details.

backend/api/Startup.cs
```diff
@@ -138,6 +141,20 @@ public void ConfigureServices(IServiceCollection services)
...
+            {
+                setup.AddHealthCheckEndpoint("liveliness", "http://localhost/live");
+                setup.AddHealthCheckEndpoint("readiness", "http://localhost/ready");
+                setup.AddWebhookNotification("rocketchat",
+                    uri: Configuration["ROCKETCHAT_HEALTH_HOOK"],
```
devinleighsmith: These are stored as secrets. The two values below could also be hardcoded in theory, but I like knowing which env the healthchecks are failing in.

backend/api/Startup.cs
devinleighsmith: Our openshift pipeline should notify us if the health checks fail during the deployment process, but this provides the extra functionality of notifying us if the health checks fail during normal execution.

devinleighsmith:
If this all works in Openshift I'll add config to our deployment to actually consume the new healthcheck endpoints.

backend/api/Startup.cs
```diff
@@ -115,15 +118,15 @@ public void ConfigureServices(IServiceCollection services)
                 options.AddPolicy("Administrator", policy => policy.Requirements.Add(new RealmAccessRoleRequirement("administrator")));
             });
 
+            var cs = Configuration.GetConnectionString("PIMS");
```
Fosol: I wonder if the connection string is missing whether all this fails in a nice and polite way?

backend/api/Startup.cs
Fosol: Is this an endpoint `/api/liveliness`?

backend/api/Startup.cs
Fosol: What happens if it fails?

backend/api/Startup.cs
Fosol: I'm very curious to see all this working.

backend/api/Startup.cs
Fosol: 1) Make sure all routes are `/api/...`
2) I don't understand the `http://localhost/live` part.  I would think this needs to be configurable, handle https, host and port.

backend/api/Startup.cs
Fosol: When does this get called/triggered?

backend/api/Startup.cs
```diff
@@ -166,10 +183,20 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseAuthorization();
 
             app.UseMiddleware(typeof(LogRequestMiddleware));
-
+            app.UseHealthChecks("/live", new HealthCheckOptions
```
Fosol: Is this now the actual route `/live` rather than `/liveliness`?  I don't understand the differences.  Also routes for the api need to be consistent `/api/...` or we'll have two different paths depending on the caller path.

backend/api/Startup.cs
devinleighsmith: /live is the endpoint that I've configured below.
/live is hosted on our server. however, healthchecksui doesn't know that. HealthChecksUI can be configured to run locally on the same server as the health checks, or remotely. in this case we are running locally, thus the http://localhost. According to the docs just having just having /live works in some cases, but they still have some issues (it didn't work for me within docker).

Https should work, but also shouldn't matter since the connection isn't crossing a network.
Same with being configurable, that shouldn't be necessary unless we plan on having HealthChecksUI not run on the same server. If that is the case then we could create a separate container, and then there are completely different connection parameters that can be used to detect health endpoints on kubernetes.

backend/api/Startup.cs
devinleighsmith: currently just liveliness, but I can add the /api

backend/api/Startup.cs
devinleighsmith: the health check returns unhealthy.

backend/api/Startup.cs
devinleighsmith: Triggered every 10 minutes when a health check is failing (configurable) or once when a failing health check stops failing.

backend/api/Startup.cs
devinleighsmith: I'll test this.

backend/api/Startup.cs
devinleighsmith: liveliness is not the route, it is just the name of the health check.
Step 1. Create health checks and add names/tags to them as necesssary.
Step 2. Map all health checks to endpoints based on the identifiers created in Step 1.

backend/api/Startup.cs
devinleighsmith: I've looked into this a bit more, while I don't want to get hung up on the details too much for this SPIKE, we could configure most of the health checks ui within appsettings.json, for example:

{
  "HealthChecksUI": {
    "HealthChecks": [
      {
        "Name": "HTTP-Api-Basic",
        "Uri": "http://localhost:6457/healthz"
      }
    ],
    "Webhooks": [
      {
        "Name": "",
        "Uri": "",
        "Payload": "",
        "RestoredPayload": ""
      }
    ],
    "EvaluationTimeInSeconds": 10,
    "MinimumSecondsBetweenFailureNotifications": 60
  }
}
we would still need to inject the value of the webhook secret here, but this might be more what you are looking for.

backend/api/Startup.cs
devinleighsmith: Yes, it seems to handle that in a reasonable manner:

image

backend/api/Startup.cs
asanchezr: I also like that we get a nice UI but I'd suggest if you could check it out a bit so it can't get abused to spam our server(s)

PR #67, opened 2020-03-13
PIMS-1183 Filtering sensitive properties.

Fosol:
Fixing API Keycloak configuration bug.
Fixing import tool.


backend/api/Profiles/ParcelProfile.cs
```diff
@@ -12,21 +12,23 @@ public class ParcelProfile : Profile
         #region Constructors
         public ParcelProfile()
         {
-            CreateMap<Entity.Parcel, Models.Parts.ParcelModel>();
+            CreateMap<Entity.Parcel, Models.Parts.ParcelModel>()
+                .ForMember(dest => dest.PID, opt => opt.MapFrom(src => src.ParcelIdentity));
```
asanchezr: Does not impact this PR but some comments/documentation on what makes a `Models.Parts.ParcelModel` different from a `Models.ParcelModel` would go a long way in understanding the naming on the API models

backend/api/Startup.cs
```diff
@@ -56,10 +56,14 @@ public Startup(IConfiguration configuration, IWebHostEnvironment env)
...
-        // This method gets called by the runtime. Use this method to add services to the container.
+        /// <summary>
+        /// This method gets called by the runtime. Use this method to add services to the container.
+        /// </summary>
+        /// <param name="services"></param>
         public void ConfigureServices(IServiceCollection services)
         {
-            services.Configure<Pims.Api.Configuration.KeycloakOptions>(this.Configuration);
+            services.Configure<Pims.Api.Configuration.KeycloakOptions>(this.Configuration.GetSection("Keycloak"));
+            services.Configure<Pims.Dal.PimsOptions>(this.Configuration.GetSection("Pims"));
```
asanchezr: Thanks for adding this

backend/api/Startup.cs
```diff
@@ -95,21 +99,21 @@ public void ConfigureServices(IServiceCollection services)
...
+                                context.NoResult();
+                                context.Response.StatusCode = 500;
+                                context.Response.ContentType = "text/plain";
+                                if (Environment.IsDevelopment())
+                                {
+                                    return context.Response.WriteAsync(context.Exception.ToString());
+                                }
+                                return context.Response.WriteAsync("An error occurred processing your authentication.");
+                            },
+                            OnForbidden = context =>
```
asanchezr: I understand that this is nobody's fault but the last few PRs we seem to be alternating between different spacing here depending on who submits a PR. I'm indifferent to spacing changes like these but let's please agree on a standard that works for everybody so we don't see this all the time.

backend/api/Startup.cs
```diff
@@ -144,20 +148,24 @@ public void ConfigureServices(IServiceCollection services)
...
             {
                 setup.AddHealthCheckEndpoint("liveliness", "http://localhost/live");
                 setup.AddHealthCheckEndpoint("readiness", "http://localhost/ready");
                 setup.AddWebhookNotification("rocketchat",
-                    uri: Configuration["ROCKETCHAT_HEALTH_HOOK"],
-                    payload: Configuration["ROCKETCHAT_HEALTH_PAYLOAD"],
-                    restorePayload: Configuration["ROCKETCHAT_HEALTH_RESTORE"]);
+                    uri : Configuration["ROCKETCHAT_HEALTH_HOOK"],
+                    payload : Configuration["ROCKETCHAT_HEALTH_PAYLOAD"],
+                    restorePayload : Configuration["ROCKETCHAT_HEALTH_RESTORE"]);
```
asanchezr: More spacing changes that we need to agree on

backend/api/Startup.cs
Fosol: This is a vscode issue.  If we can get it to standardize that would be great.

backend/dal/Helpers/Extensions/IdentityExtensions.cs
```diff
@@ -31,7 +32,7 @@ public static Guid GetUserId(this ClaimsPrincipal user)
         public static int? GetAgency(this ClaimsPrincipal user)
         {
             var value = user?.FindFirstValue("agency");
-            return String.IsNullOrWhiteSpace(value) ? null : (int?)Int32.Parse(value);
+            return String.IsNullOrWhiteSpace(value) ? null : (int?) Int32.Parse(value);
```
asanchezr: spacing...

backend/tools/import/Startup.cs
```diff
@@ -65,7 +67,7 @@ public async Task<int> Run(string[] args)
         /// Determine what HTTP method to use.
         /// </summary>
         /// <param name="method"></param>
-        /// <return></return>
+        /// /// <return></return>
```
asanchezr: Too many "///"

backend/tools/import/Startup.cs
```diff
@@ -83,7 +85,7 @@ private static HttpMethod GetMethod(string method)
         }
 
         /// <summary>
-        /// Read the JSON package, iterate through it and send the items to the configured endpoint URL.
+        /// /// Read the JSON package, iterate through it and send the items to the configured endpoint URL.
```
asanchezr: Too many "///"

backend/dal/Helpers/Extensions/EnumerableExtensions.cs
```diff
@@ -35,5 +37,25 @@ public static void ForEach<T>(this IEnumerable<T> items, Action<T> action)
...
+            var list = items.ToList();
+            var remove = list.Where(find);
+
+            foreach (var item in remove)
+            {
+                list.Remove(item);
+            }
+
+            return list;
+        }
```
asanchezr: I'd like to know why we are doing a custom "RemoveAll" here...
I understand that IEnumerable doesn't have it but List does - see [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall?view=netcore-3.1)
We are already converting to a list here so I'd be tempted to go with:
```csharp
var list = items.ToList().RemoveAll(find)
```
Thoughts?

Fosol: Can go either way.
`items.ToList().RemoveAll(find);`
vs
`items.RemoveAll(find)`

asanchezr: On second thoughts, if the list is large, converting to a list could impact performance. There seems to be an IEnumerable extension method already that can help us here:

```csharp
var itemsFiltered = items.SkipWhile(find);
```

See https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skipwhile?view=netcore-3.1#definition

Fosol: That is an immutable solution.  The `RemoveAll(...)` mutates, which is the desired result in this case.

asanchezr: Well, in my opinion this custom RemoveAll doesn't look very efficient. We are first iterating the whole list by doing `ToList()`. Then we are iterating it again with the `.Where()` so that's o(2N) and then we iterate the "remove" collection to remove from the original. 

The worst case scenario here - where ALL parcels need to be removed. We traverse the original list **3 times** - o(3N). That can degrade performance for big lists (ie a lot of parcels)

The "official" SkipWhile would iterate the list only once by asking each time: does this match the supplied predicate function? if yes, skip the element, if not leave it in the returning list

asanchezr: About mutability, looking at the code below that uses your RemoveAll I don't see how it mutates the result....See my comments there

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -82,6 +90,10 @@ public Parcel GetNoTracking(int id)
                 .Include(p => p.Buildings).ThenInclude(b => b.BuildingConstructionType)
                 .Include(p => p.Buildings).ThenInclude(b => b.BuildingPredominateUse)
                 .AsNoTracking().SingleOrDefault(u => u.Id == id);
+
+            var agencies = this.User.GetAgencies();
+            parcel.Buildings.RemoveAll(b => b.IsSensitive && !agencies.Contains(b.AgencyId));
+            return parcel;
```
asanchezr: Please help me understand how the above mutates the `parcel.Buildings` Enumerable

asanchezr: If I follow your code correctly, the "RemoveAll" function takes an Enumerable source: `parcel.Buildings` and return a copy in List form: `items.ToList()`. It will then mutate the list but I don't think that also mutates the underlying Enumerable, does it?

backend/api/Profiles/ParcelProfile.cs
devinleighsmith: Yeah, I actually am not clear on that either.

backend/dal/Security/RoleClaim.cs
```diff
@@ -0,0 +1,29 @@
...
+        PropertyEdit,
+
+        [ClaimName("admin-users")]
+        DisposeRequest,
+
+        [ClaimName("admin-users")]
+        DisposeApprove,
+
+        [ClaimName("admin-users")]
+        SensitiveView
```
asanchezr: Quick question: Do every Role Claim is mapped to the same name "admin-users"? Is this correct? Help me understand

backend/dal/Helpers/Extensions/EnumerableExtensions.cs
Fosol: The function is not performant.  But there isn't a good way to mutate an IEnumerable.  The only thing I could do is check if the original type passed is already a `List`.

That being said there are a number of possible other ways to do this.  I could change the entity and allow for the `Parcel.Buildings` collection to be `set`.  But this isn't ideal in-terms of a EF design.

Either way the extension method itself isn't supposed to be performant.  Nor could it be when mutating an item from the collection.

backend/dal/Services/Concrete/ParcelService.cs
Fosol: I think I've mistakenly mutated the list rather than the ienumerable... good catch

devinleighsmith:
Is the AddAccessRequest_Success passing for you? I'm getting:
    System.InvalidOperationException : Unable to resolve service for type 'Microsoft.Extensions.Options.IOptionsMonitor`1[Pims.Api.Configuration.KeycloakOptions]' while attempting to activate 'Pims.Api.Controllers.UserController'.

backend/dal/Security/RoleClaim.cs
Fosol: These enums simply provide a consistent compile-time safe way to get the string value we use in a Keycloak role.  These roles are converted into Identity Claims.

scripts/gen-env-files.sh
```diff
@@ -46,8 +46,8 @@ POSTGRES_PASSWORD=password
 POSTGRES_DB=geospatial" >> ./database/postgres/.env
 fi
 
-if test -f "./backend/.env"; then
-    echo "./backend/.env exists"
+if test -f "./backend/api/.env"; then
+    echo "./backend/api/.env exists"
```
asanchezr: Please since you are changing this file, please update the connection string below. We are using "pims" now as the database name

backend/api/Profiles/ParcelProfile.cs
Fosol: We should come up with a better way to name them.  Models are often specific to endpoints (although they are often shared by many).

Some models we only want to return specific data.  We don't want to send all information in every endpoint request.

If I make a request for a list of Parcels and I only need to display two of their properties, I shouldn't be receiving all of their properties in the request.  While simpler to understand it doesn't consider network performance and costs associated with the extra data over time.

backend/dal/Helpers/Extensions/EnumerableExtensions.cs
Fosol: I'm going to rewrite this extension to only support ICollection<T>, then it will do what I need it to do.  Which is mutate the collection.

backend/dal/Helpers/Extensions/IdentityExtensions.cs
```diff
@@ -120,6 +121,38 @@ public static bool HasRoles(this ClaimsPrincipal user, params string[] role)
...
+        /// <param name="user"></param>
+        /// <param name="role"></param>
+        /// <returns>True if the user has all of the roles.</returns>
+        public static bool HasRoles(this ClaimsPrincipal user, params RoleClaim[] role)
+        {
+            if (role == null) throw new ArgumentNullException(nameof(role));
+            if (role.Length == 0) throw new ArgumentOutOfRangeException(nameof(role));
+
+            var roles = role.Select(r => r.GetRoleName()).ToArray();
+            var count = user.Claims.Count(c => c.Type == ClaimTypes.Role && roles.Contains(c.Value));
```
devinleighsmith: Is there any benefit to this Count() method over just using All()?

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -38,14 +39,20 @@ public class ParcelService : BaseService<Parcel>, IParcelService
...
             if (neLat != null && neLong != null && swLat != null && swLong != null)
             {
-                query = this.Context.Parcels.AsNoTracking().Where(parcel =>
-                    parcel.Latitude <= neLat &&
-                    parcel.Latitude >= swLat &&
-                    parcel.Longitude <= neLong &&
-                    parcel.Longitude >= swLong);
+                // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
+                query = this.Context.Parcels.AsNoTracking().Where(p =>
+                    (!p.IsSensitive || (viewSensitive && userAgencies.Contains(p.AgencyId))) &&
```
devinleighsmith: Does this support the concept of sub-agencies? Ie, would this allow a user from the ministry of health to see a sensitive island health property?

backend/dal/Helpers/Extensions/IdentityExtensions.cs
Fosol: All() would be a nicer way of writing it.

backend/api/Areas/Admin/Controllers/ParcelController.cs
```diff
@@ -6,15 +6,16 @@
...
 using Pims.Dal.Services.Admin;
-using Microsoft.AspNetCore.Authorization;
 
 namespace Pims.Api.Areas.Admin.Controllers
 {
     /// <summary>
     /// ParcelController class, provides endpoints for managing parcels.
     /// </summary>
-    [Authorize(Roles = "system-administrator")]
+    [HasPermission(Permissions.SystemAdmin)]
```
Fosol: Compile-time safe claim based permission check.

backend/api/Areas/Admin/Controllers/ParcelController.cs
```diff
@@ -127,22 +128,22 @@ public IActionResult AddParcel([FromBody] ParcelModel model)
         {
             var entity = _mapper.Map<Entity.Parcel>(model);
             entity.Evaluations.Add(new Entity.ParcelEvaluation(model.FiscalYear, entity) // TODO: Move this logic to AutoMapper.
-            {
-                EstimatedValue = model.EstimatedValue,
-                AssessedValue = model.AssessedValue,
-                NetBookValue = model.NetBookValue
-            });
+                {
```
Fosol: I hate vscode formatting...

backend/api/Policies/HasPermissionAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Pims.Dal.Security;
+
+namespace Pims.Api.Policies
+{
+    /// <summary>
+    /// HasPermissionAttribute class, provides an attribute that will authorize a users claims to determine if they have permission to perform the requested action.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public class HasPermissionAttribute : TypeFilterAttribute
```
Fosol: Provides a way to decorate our controllers and endpoints to authorize users claims.

backend/api/Policies/PermissionFilter.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.AspNetCore.Mvc.Filters;
+using Pims.Dal.Helpers.Extensions;
+using Pims.Dal.Security;
+
+namespace Pims.Api.Policies
+{
+    /// <summary>
+    /// PermissionFilter class, provides a authorization filter that validates the specified permissions.
+    /// </summary>
+    public class PermissionFilter : IAuthorizationFilter
```
Fosol: Authorization validation filter.

backend/dal/Services/Concrete/ParcelService.cs
Fosol: Yes, we will include a claim that is an array (sub-agencies) called `agencies`.  Which we can use `User.GetAgencies()` to extract.

backend/dal/Security/RoleClaim.cs
Fosol: This has been fixed and is now an enum called `Permissions`

scripts/gen-env-files.sh
Fosol: done

backend/api/Models/Parts/ParcelModel.cs
```diff
@@ -28,13 +30,13 @@ public override bool Equals(object obj)
...
-                   Longitude == other.Longitude &&
-                   Description == other.Description;
+                Id == other.Id &&
+                PID == other.PID &&
+                StatusId == other.StatusId &&
+                ClassificationId == other.ClassificationId &&
+                Latitude == other.Latitude &&
+                Longitude == other.Longitude &&
+                Description == other.Description;
         }
```
asanchezr: Question: shouldn't the new **PIN** property be included in the `Equals` method?

asanchezr: Same for the previous file...

Fosol: probably yes, and why I'm not a fan of this implementation.

Fosol: fixed

PR #68, opened 2020-03-14
PIMS-972 Add user list.

devinleighsmith:
1) Add API to support getting all users for the currently logged-in user.
2) List all users by name, role, agency.
3) Test updates.
4) Refactor common frontend functionality for re-use.


backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -54,9 +55,10 @@ public UserController(ILogger<UserController> logger, IConfiguration configurati
         /// <param name="page"></param>
         /// <param name="quantity"></param>
         /// <param name="sort"></param>
+        /// <param name="userId"></param>
         /// <returns>Paged object with an array of users.</returns>
         [HttpGet("/api/admin/users")]
-        public IActionResult GetUsers(int page = 1, int quantity = 10, string sort = null) // TODO: sort and filter.
+        public IActionResult GetUsers(int page = 1, int quantity = 10, string sort = null, Guid? userId = null) // TODO: sort and filter.
```
Fosol: It looks like you're using the same endpoint for either fetching one user or fetching many.  If so I'd recommend you create a new endpoint for the single user.

Fosol: If you are using the `userId` as a way to fetch the current user's Id, then just use the context `User.GetUserId()`.

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -68,6 +70,20 @@ public UserController(ILogger<UserController> logger, IConfiguration configurati
...
+        /// GET - Returns a paged array of users from the datasource that belong to the same agency (or sub-agency) as the current user.
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        /// <param name="sort"></param>
+        /// <returns>Paged object with an array of users.</returns>
+        [HttpGet("/api/admin/my/users")]
+        public IActionResult GetMyUsers(int page = 1, int quantity = 10, string sort = null) // TODO: sort and filter.
+        { 
+            var userId = this.User.GetUserId();
```
Fosol: I think all you're really interested in is the current user's agencies they belong to.  There is an extension method to do this, as we'll include this information in their identity.  Otherwise we have to constantly scan the DB for it.
`User.GetAgency()` returns their primary agency.
`User.GetAgencies()` returns all the agencies they belong to (including child agencies).
I'd probably lean towards creating a new method in the service layer for this rather than just a filter.  It'll allow us to apply security easier.

backend/dal/Helpers/Extensions/EnumerableExtensions.cs
```diff
@@ -35,5 +36,10 @@ public static void ForEach<T>(this IEnumerable<T> items, Action<T> action)
 
             return item;
         }
+
+        public static IEnumerable<T> Flatten<T>(this IEnumerable<T> e, Func<T, IEnumerable<T>> f)
```
Fosol: Add some comments to explain what's happening.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -36,11 +36,25 @@ public class UserService : BaseService<User>, IUserService
...
         {
+
             this.User.ThrowIfNotAuthorized("system-administrator");
+            var query = this.Context.Users
+                .Include(u => u.Agencies)
+                .ThenInclude(a => a.Agency)
+                .Include(r => r.Roles)
+                .ThenInclude(r => r.Role)
+                .AsNoTracking();
+            if (userId.HasValue) {
```
Fosol: Making this type of request constantly against the DB isn't ideal.  Just use the `User.GetAgencies()` extension method.  We'll include this information in their Claims.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -36,11 +36,25 @@ public class UserService : BaseService<User>, IUserService
         /// <param name="quantity"></param>
         /// <param name="sort"></param>
         /// <returns></returns>
-        public Paged<User> GetNoTracking(int page = 1, int quantity = 10, string sort = null)
+        public Paged<User> GetNoTracking(int page = 1, int quantity = 10, string sort = null, Guid? userId = null)
```
Fosol: I'd remove the `userId` filter argument.  It makes more sense for a specific use-case (i.e. GetMyUsers(...) or GetAgencyUsers(...)).

If we do want to make a generic `GetNoTracking(...)` function then we need to pass filter parameters to it as an filter object or a dictionary.

backend/api/Areas/Admin/Controllers/UserController.cs
devinleighsmith: This is probably confusing, but I'm using that user id as the 'logged in user'. So I'm not getting a single user based on that user id, I'm actually getting all users with agencies 'belonging' to that user id.

Fosol: Ah in that case you don't need it.  If it's the logged in user, we have that in `User.GetUserId()`.

asanchezr:
@devinleighsmith could you please rebase/fix conflicts before I go and review this? thanks!

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -82,9 +83,12 @@ public IActionResult GetUsers(EModel.UserFilter filter)
         /// <returns>Paged object with an array of users.</returns>
         [HttpGet("/api/admin/my/users")]
         public IActionResult GetMyUsers(int page = 1, int quantity = 10, string sort = null) // TODO: sort and filter.
```
Fosol: I'd recommend using the same arguments as the `GetUsers(EModel.UserFilter filter)` possibly.  Then you have the ability to apply additional filters and sorting built in.

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -82,9 +83,12 @@ public IActionResult GetUsers(EModel.UserFilter filter)
...
-        { 
-            var userId = this.User.GetUserId();
-            return GetUsers(page, quantity, sort, userId);
+        {
+            var agencies = this.User.GetAgencies();
+            if (agencies == null || agencies.Length <= 0) return BadRequest("Current user does not belong to an agency");
```
Fosol: Possibly `!(agencies?.Any() ?? false)` 😄

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -82,9 +83,12 @@ public IActionResult GetUsers(EModel.UserFilter filter)
...
+            var filter = new EModel.UserFilter();
```
Fosol: If it's part of the arguments you can just update the `Agencies`.

PR #73, opened 2020-03-17
PIMS-1156

devinleighsmith:
Create common YUP based validators for Parcel, User, Address, Evaluation, and Building objects based on DB schema and basic typing.


frontend/package.json
```diff
@@ -31,7 +31,7 @@
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
     "typescript": "3.7.3",
-    "yup": "0.28.2"
+    "yup": "0.28.3"
```
devinleighsmith: There is a bug in 28.2 that prevents regexs working properly with custom error messages, fixed in 28.3

frontend/src/pages/GuestAccessPage.tsx
```diff
@@ -47,10 +47,10 @@ const GuestAccessPage = () => {
         <Col>
           <Formik
             initialValues={{
-              agency: null,
-              role: null,
+              agency: undefined,
```
devinleighsmith: in Yup undefined better represents initial state.

jdtoombs: I believe my tests check for null state but it's a quick fix - just in case you end up running them. Will change them on my end for next PR!

PR #75, opened 2020-03-18
PIMS-1250: Update local docker/api to run more similarly to openshift.

devinleighsmith:
Update healthchecksui config to match.

Note: This change forces the api to run on port 8080. This already appears to be the case when the api is running on openshift, so it shouldn't change anything there. Let me know if there are concerns - in theory we could just override the appsettings using an openshift configmap, then my change to 8080 wouldn't be necessary.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/dotnet/core/aspnet:3.1 AS base
 WORKDIR /app
-EXPOSE 80 443
+EXPOSE 443 8080
 
 # Copy csproj and restore as distinct layers
 FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build
--- a/backend/api/Program.cs
+++ b/backend/api/Program.cs
@@ -27,6 +27,7 @@ namespace Pims.Api
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
             .UseContentRoot(Directory.GetCurrentDirectory())
+            .UseUrls("http://*:8080")
             .UseStartup<Startup>();
     }
 }
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -173,12 +173,12 @@ namespace Pims.Api
 
             app.UseAuthentication();
             app.UseAuthorization();
-            app.UseHealthChecks("/api/live", new HealthCheckOptions
+            app.UseHealthChecks("/api/live", 8080, new HealthCheckOptions
             {
                 Predicate = r => r.Name.Contains("liveliness"),
                 ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
             });
-            app.UseHealthChecks("/api/ready", new HealthCheckOptions
+            app.UseHealthChecks("/api/ready", 8080, new HealthCheckOptions
             {
                 Predicate = r => r.Tags.Contains("services"),
                 ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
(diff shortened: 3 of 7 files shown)


backend/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/dotnet/core/aspnet:3.1 AS base
 WORKDIR /app
-EXPOSE 80 443
+EXPOSE 443 8080
```
devinleighsmith: Switch docker to use 8080 instead of 80 to be consistent with the api Openshift pod. This means that I don't have to test one setup on docker and a different setup on Openshift.

backend/api/Program.cs
```diff
@@ -27,6 +27,7 @@ public static void Main(string[] args)
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
             .UseContentRoot(Directory.GetCurrentDirectory())
+            .UseUrls("http://*:8080")
```
devinleighsmith: This forces the app to run on 8080. Not strictly required, but does allow me to apply one healthcheck config to both local and openshift.

backend/api/Startup.cs
```diff
@@ -173,20 +173,31 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
                 ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
             });
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers();
-                endpoints.MapHealthChecksUI();
+                endpoints.MapHealthChecksUI(
+                    setup =>
+                    {
+                        setup.UIPath = "/api/healthchecks-ui"; // this is ui path in your browser
```
devinleighsmith: Manually swap the base bath of the UI and dependent resources.

backend/Dockerfile
asanchezr: Thanks for doing this. To align more with Openshift we should consider that in Openshift containers run as **non-root** so that means that docker can not expose "priviledged ports" (< 1024). Please consider removing the 443 port from here as well, or replace with with say 8443 (that's what I've seen done on other images)

asanchezr: I would also be more inclined to make the PORT change via a supported env var within the Dockerfile itself to avoid having to hard-code the port (8080) into the source code - we should at least read it from configuration/environment on the NET Core side

See https://github.com/redhat-developer/s2i-dotnetcore/blob/master/3.1/runtime/Dockerfile#L38-L40 for how the Openshift images do it:

```Dockerfile

EXPOSE 8080

## By default, ASP.NET Core runs on port 5000. We configure it to match
## the container port.
ENV ASPNETCORE_URLS=http://*:8080

...
```

backend/api/Program.cs
asanchezr: See my previous comment. By using an env var we could remove this line. Thoughts?

backend/api/Startup.cs
```diff
@@ -173,20 +173,31 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
             app.UseAuthentication();
             app.UseAuthorization();
-            app.UseHealthChecks("/api/live", new HealthCheckOptions
+            app.UseHealthChecks("/api/live", 8080, new HealthCheckOptions
             {
                 Predicate = r => r.Name.Contains("liveliness"),
                 ResponseWriter = UIResponseWriter.WriteHealthCheckUIResponse
             });
-            app.UseHealthChecks("/api/ready", new HealthCheckOptions
+            app.UseHealthChecks("/api/ready", 8080, new HealthCheckOptions
```
asanchezr: I'd love to see the PORT value come from configuration so we don't tie ourselves to 8080 if things change on Openshift or we decide to use another NET Core image

PR #76, opened 2020-03-18
PIMS-980 Adding PUT /api/keycloak/users endpoints.

Fosol:
This is a big one!  I'm not fully done, as I still have unit tests, code clean up, documentation and a fix for automapper issues.

There are now a few endpoints that integrate PIMS with Keycloak.  This will allow us to update the role and agency the user belongs to.  The ones for this sprint that are important are;

- PUT /api/keycloak/users/{id}
- GET /api/keycloak/users
- GET /api/keycloak/users/{id}

Please note that until I can get our keycloak initial seed script working the db is out of sync.  Which means you'll have to manually do it for now.  I did create a few endpoints to help that may get you 90% of the way there.

- POST /api/keycloak/groups/sync
- POST /api/keycloak/users/sync/{id}


backend/api/Areas/Admin/Controllers/RoleController.cs
```diff
@@ -56,13 +56,13 @@ public RoleController(ILogger<RoleController> logger, IConfiguration configurati
         /// </summary>
         /// <returns>Paged object with an array of roles.</returns>
         [HttpGet("/api/admin/roles")]
-        public IActionResult GetRoles(int page = 1, int quantity = 10, string sort = null) // TODO: sort and filter.
+        public IActionResult GetRoles(int page = 1, int quantity = 10) // TODO: sort and filter.
```
Fosol: Will add search back in later.

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -146,7 +146,7 @@ public IActionResult UpdateUser([FromBody] Model.UserModel model)
         /// <param name="model">The user model.</param>
         /// <returns>The user who was deleted.</returns>
         [HttpDelete]
-        public IActionResult DeleteUser([FromBody] Model.UserModel model)
+        public ActionResult DeleteUser([FromBody] Model.UserModel model)
```
Fosol: I will need to change this back to `IActionResult`

backend/test/Helpers/TestHelper.cs
```diff
@@ -103,12 +104,17 @@ public static IMapper CreateMapper()
...
+                cfg.AddProfile(new AdminProfiles.AgencyProfile());
+                cfg.AddProfile(new AdminProfiles.UserProfile());
+                cfg.AddProfile(new AdminProfiles.RoleProfile());
+                cfg.AddProfile(new KeycloakProfiles.BaseProfile());
+                cfg.AddProfile(new KeycloakProfiles.UserProfile());
+                cfg.AddProfile(new KeycloakProfiles.GroupProfile());
+                cfg.AddProfile(new KeycloakProfiles.AgencyProfile());
             });
-            mapperConfig.AssertConfigurationIsValid();
+            // mapperConfig.AssertConfigurationIsValid(); // TODO: Fix this.
```
Fosol: Temporary until I resolve unit testing issues and automapper.

backend/keycloak/Pims.Keycloak.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
Fosol: The new Keycloak Admin API toolkit

backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -135,7 +136,6 @@ public IActionResult UpdateUser([FromBody] Model.UserModel model)
             var entity = _mapper.Map<Entity.User>(model);
             var updatedEntity = _pimsAdminService.User.Update(entity);
 
-            if (updatedEntity == null) return BadRequest("Item does not exist");
```
devinleighsmith: intentional?

backend/api/Areas/Admin/Profiles/UserProfile.cs
```diff
@@ -18,12 +20,6 @@ public UserProfile()
 
             CreateMap<Model.UserModel, Entity.User>()
                 .ForMember(dest => dest.IsDisabled, opt => opt.Ignore());
-
-            CreateMap<Model.RoleModel, Entity.UserRole>()
```
devinleighsmith: I think I'm using this for the accessrequest api.

backend/api/Areas/Keycloak/Controllers/GroupController.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        #region Endpoints
+        /// <summary>
+        /// Sync keycloak groups into PIMS roles.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost("sync")]
+        public async Task<IActionResult> SyncGroupsAsync()
+        {
+            var gcount = await _keycloakAdmin.GetGroupCountAsync();
```
devinleighsmith: presumably at some point we'll need error handling for this.

backend/api/Areas/Keycloak/Controllers/GroupController.cs
```diff
@@ -0,0 +1,165 @@
...
+                    else
+                    {
+                        _mapper.Map(kgroup, erole);
+                        _pimsAdminService.Role.UpdateOne(erole);
+                    }
+
+                    groups.Add(kgroup);
+                }
+
+                _pimsAdminService.CommitTransaction();
```
devinleighsmith: it feels weird to be managing transactions in a controller, is the long term plan to move these to a KeycloakService?

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -1,96 +1,261 @@
...
+        {
+            var kuser = await _keycloakAdmin.GetUserAsync(id) ?? throw new KeyNotFoundException();
+            var kgroups = await _keycloakAdmin.GetUserGroupsAsync(id);
+
+            var euser = _pimsAdminService.User.Find(id);
+            if (euser == null)
+            {
+                // The user does not exist in PIMS, it needs to be added.
+                euser = _mapper.Map<Entity.User>(kuser);
+                euser.DisplayName = $"{kuser.LastName}, {kuser.FirstName}";
```
devinleighsmith: normally I'd say we should put these in automapper, but I imagine you aren't interested in hearing that right now.

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -1,96 +1,261 @@
...
+                foreach (var group in kgroups)
+                {
+                    var erole = _pimsAdminService.Role.Find(group.Id);
```
devinleighsmith: I think I'd rather have the service do these kinds of checks, but that could be tech debt.

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -1,96 +1,261 @@
...
+            {
+                // The user exists in PIMS, it only needs to be updated.
+                var roles = euser?.Roles.ToArray();
+                _mapper.Map(kuser, euser);
+                foreach (var group in kgroups)
+                {
+                    var erole = _pimsAdminService.Role.Find(group.Id);
+
+                    // If the role doesn't exist, create it.
+                    if (erole == null)
```
devinleighsmith: this provides the rationale for my previous comment, the controllers may end up needing to do this a lot if they are responsible for checking the role status and selectively adding/updating.

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -1,96 +1,261 @@
...
 
         /// <summary>
         /// Update the user for the specified 'id'.
+        /// If the user does not exist in Keycloak or PIMS return a 400-BadRequest.
         /// </summary>
+        /// <exception type="KeyNotFoundException">The user does not exist in Keycloak or PIMS.</exception>
         /// <returns></returns>
         [HttpPut("{id}")]
-        public async Task<IActionResult> UpdateUserAsync(string id, [FromBody] Model.UserModel model)
+        public async Task<IActionResult> UpdateUserAsync(Guid id, [FromBody] Model.Update.UserModel model)
```
devinleighsmith: I'm starting to wonder if it is better to have a custom model for each operation, or just one model that contains more fields then necessary. Obviously this isn't perfect as the backend would end up ignoring some fields sent in by the frontend, but it would cut done on the number of Objects/Mappers that we need.

devinleighsmith: At some point we'll probably want to be able to roll back this entire operation if part of it fails.

backend/api/Areas/Keycloak/Models/AgencyModel.cs
```diff
@@ -0,0 +1,60 @@
+using System;
```
devinleighsmith: Is this just to get around automapper or is there another reason this exists? it seems odd to me that this class exists but has no relationship to the existing AgencyModel.

backend/api/Models/Update/BaseModel.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Pims.Api.Models.Update
```
devinleighsmith: Remind me what the benefit is of having all of the different 'flavours' of these model classes? We already have a BaseModel, this version just gets rid of a few fields. Could we just ignore those extra fields, or does that cause issues?

backend/api/Models/UserModel.cs
```diff
@@ -20,11 +19,9 @@ public class UserModel : BaseModel, IEquatable<UserModel>
         public string LastName { get; set; }
 
         public string Email { get; set; }
+        #endregion
 
-        public IEnumerable<AgencyModel> Agencies { get; set; }
```
devinleighsmith: I think the frontend is depending on these fields right now.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -108,6 +110,21 @@ public AccessRequest AddAccessRequest(AccessRequest request)
             this.Context.CommitTransaction();
             return request;
         }
+
+        /// <summary>
+        /// Get an array of agency IDs for the specified 'userId'.
+        /// This only returns the first two layers (direct parents, their immediate children).
```
devinleighsmith: why only the two layers?

backend/api/Areas/Keycloak/Controllers/GroupController.cs
asanchezr: Agree, controllers should be thin and delegate to business logic to services.

backend/api/Areas/Keycloak/Controllers/UserController.cs
Fosol: Ideally yes, not sure how to best approach this.  We can't wrap two separate data-sources in a transaction.  Which means rollback would be manual and require us to just update the successful parts with original values.  Which also could fail and then leave us in a awkward state.  The complexity involved is a little too high for the value in our case.  It is just easier to resolve whatever caused the failure and then update the user again.

backend/api/Areas/Keycloak/Models/AgencyModel.cs
Fosol: While duplication of models isn't great, always remember that models are specific to their endpoint(s).  It's one of the reasons some design implementations don't use models and only use dynamic objects (in the effort to remove boilerplate hard to keep up-to-date models).

The models in the `/Areas/Keycloak/Models` are for the `Keycloak` controller and are or maybe tightly coupled with the logic related to integrating Keycloak and PIMS.

backend/api/Models/Update/BaseModel.cs
Fosol: There are a number of approaches to reducing duplication of models.  One of them is just to reuse as much as possible and let extra fields just get passed around.  This is a leaky model and results in an outwardly ugly api.  When we add Swagger our models would have properties that are irrelevant.

The downside the a cleaner outward api regrettably results in an uglier inward api.  Essentially creating models for every endpoint or groups of endpoints.  I lean towards this for the follow reasons (even though I hate the duplicate too)

- All endpoints (groups of endpoints) are self contained and won't contaminate or break each other when a model is changed
- API input and output is valid and relevant to the endpoints
- Models are lightweight and generally don't contain any logic, thus are easy enough to maintain.

The main downside is that when an entity is changed it **may** require updating multiple models to support the change.  While this is a risk, it is mitigated by the following;

- Unit tests should catch any breaking changes
- We don't always want to change the shape of all endpoint input/output when an entity changes

backend/api/Models/UserModel.cs
Fosol: I'll see if I can add them back in.

backend/dal/Services/Concrete/UserService.cs
Fosol: This is a design decision.  I design the db to support more for future proofing, but the business requirements only demand two layers ATM.  The performance implications for more are exponential.

backend/api/Models/UserModel.cs
Fosol: Had to update automapper profiles to allow for them to come back.

backend/api/Profiles/RoleProfile.cs
```diff
@@ -13,17 +14,26 @@ public RoleProfile()
...
-                .ForMember(src => src.Code, opt => opt.MapFrom(dest => dest.Name));
+                .ForMember(dest => dest.Id, opt => opt.MapFrom(src => src.Id.ToString()))
+                .ForMember(dest => dest.Code, opt => opt.MapFrom(src => src.Name));
 
             CreateMap<RoleModel, Entity.Role>()
-                .ForMember(src => src.Id, opt => opt.MapFrom(dest => Guid.Parse(dest.Id)))
+                .ForMember(dest => dest.Id, opt => opt.MapFrom(src => Guid.Parse(src.Id)))
                 .IncludeBase<CodeModel, Entity.LookupEntity>();
+
+            CreateMap<Entity.UserRole, RoleModel>()
```
Fosol: Had to add this to solve automapper errors occurring during initialization.

backend/api/Areas/Admin/Profiles/RoleProfile.cs
```diff
@@ -1,6 +1,5 @@
 using AutoMapper;
 using Pims.Api.Models;
-using Pims.Api.Profiles.Extensions;
```
devinleighsmith: nice, I was wondering about why these were added but didn't think complaining about them added much value. Nice to see that they have been removed.

backend/api/Profiles/RoleProfile.cs
devinleighsmith: thanks for this.

PR #78, opened 2020-03-19
add readiness/liveness probes to deploy config.

devinleighsmith:


openshift/templates/pims-api/deploy.yaml
```diff
@@ -140,26 +140,26 @@ objects:
...
-              #   initialDelaySeconds: 10
-              #   timeoutSeconds: 1
-              #   periodSeconds: 10
-              #   successThreshold: 1
-              #   failureThreshold: 3
+              readinessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /api/ready
+                  port: 8080
```
asanchezr: Could you please change it to use `port: ${{APP_PORT}}` instead of hardcoding it to 8080 to keep template generic?

Please note the double braces -> `${{APP_PORT}}` 

This replaces with the "number" representation (e.g. port: 8080) vs the string value (port: "8080")

openshift/templates/pims-api/deploy.yaml
```diff
@@ -140,26 +140,26 @@ objects:
...
+                initialDelaySeconds: 1
+                periodSeconds: 10
+                successThreshold: 1
+                timeoutSeconds: 3
+              livenessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /api/live
+                  port: 8080
+                  scheme: HTTP
```
Fosol: You probably want to add all three values to default configuration.

PR #80, opened 2020-03-20
Add support for SonarCloud code analysis

asanchezr:
- Configure SonarCloud scanning (available for the frontend only)
- Include additional configuration for pathfinder SonarQube server


.gitignore
```diff
@@ -4,3 +4,5 @@ bash.exe.stackdump
 .vs
 *.orig
 *.param
+.scannerwork/
+.sonarqube/
```
asanchezr: These are temporary folders created by `sonnar-scanner` that we don't want in source control

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "3.1.100"
+    "version": "3.1.200"
```
Fosol: Are we upgrading .net core to 3.1.2?  Should we upgrade all the projects?

global.json
```diff
@@ -1,5 +1,5 @@
...
   }
 }
```
asanchezr: I had to change this to `3.1.200` after an update to VS 2019 which included a new minor revision of NET Core SDK. Otherwise `dotnet` commands would not work...

Questions for the team and @Fosol specifically: 
- Is there any reason to keep this file in source control? 
- Can it be removed safely?

I read somewhere that if not present `dotnet` will use "latest" by default but wanted to double check.

Fosol: The file is not "required" according to blogs, but it was actually required to fix bugs in vscode.  Without it omni-sharp is unable to parse C# code correctly.

global.json
asanchezr: See my comments below on this file

global.json
asanchezr: Okay good to know. Can you confirm it still needed? All I did was update my local VS 2019 and after that I didn't have 3.1.100 anymore so it must be a drop in replacement.

I don't believe we need to update all backend projects to use this specific version. They should still be fine as this is still 3.1 not 3.2

asanchezr:
I removed `global.json` from my local repo and was able to successfully compile and debug the backend project with VS Code (see screenshot below). Everything seems functional - breakpoints work, health checks work , etc. The image below shows the DB health check is failing because I didn't start the DB for this test.

image

image

global.json
```diff
@@ -1,5 +0,0 @@
-{
-  "sdk": {
-    "version": "3.1.100"
```
Fosol: If I remove this file my intellisense stops working in vscode.

PR #81, opened 2020-03-20
PIMS-981 Creating Pims Keycloak Integration Services

Fosol:
The PIMS and Keycloak integration library has been extracted.
Fixing import bug
I've decided to keep the naming convention PIMS specific (i.e. Roles, Claims).


backend/Dockerfile
```diff
@@ -10,8 +10,10 @@ COPY api/*.csproj api/
 COPY entities/*.csproj entities/
```
Fosol: For some reason now when I build the docker container it fails unless I first delete all the `bin` and `obj` folders of all of the projects...

backend/api/Areas/Keycloak/Controllers/RoleController.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Pims.Api.Areas.Keycloak.Controllers
+{
+    /// <summary>
+    /// RoleController class, provides endpoints for managing the integration between keycloak groups and PIMS roles.
+    /// </summary>
+    [HasPermission(Permissions.AdminRoles)]
+    [ApiController]
+    [Area("keycloak")]
+    [Route("/api/[area]/roles")]
+    public class RoleController : ControllerBase
```
Fosol: Replaced `GroupController` with `RoleController`.  This means the endpoint changed to `/api/keycloak/roles/`

backend/core/Helpers/ExceptionHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+    public static class ExceptionHelper
+    {
+        /// <summary>
+        /// Provides a simple try+catch wrapper to set a variable.
+        /// </summary>
+        /// <param name="getValue"></param>
+        /// <param name="setter"></param>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public static T HandleKeyNotFound<T>(Func<T> getter, Func<T> setter)
```
Fosol: Just syntactical sugar so that we don't need ugly try+catch in various places to get/set values.

backend/tools/import/ImportOptions.cs
```diff
@@ -26,6 +26,36 @@ public class ImportOptions
         /// </summary>
         /// <value></value>
         public int Quantity { get; set; } = 100;
+
+        /// <summary>
+        /// get/set - The number of items to skip in the imported JSON array.
+        /// </summary>
+        /// <value></value>
+        public int Skip { get; set; } = 0;
```
Fosol: More configuration options to control the importer.

backend/Dockerfile
devinleighsmith: probably need to add those binaries to the dockerignore.

**/bin
**/obj

asanchezr: What Devin said

asanchezr: We already have the following on our dockerignore but that only matches a /bin folder under backend. 

```
[Bb]in/
[Oo]bj/
[Oo]ut/
```

We need

```
**/[Bb]in/
**/[Oo]bj/
**/[Oo]ut/
```

Fosol:
I've updated the `.dockerignore` file and it appears to resolve the issue.

Fosol:
Check out the updated unit tests.  The **Arrange** section is now independent and simple.  Essentially three statements;
- Choose the user executing the task (I'm going to simplify this step in the future)
- Create a `TestHelper`
- Create your `Controller`

It is possible to create a global `TestHelper` and it may be preferable in some cases.  You just need to be aware that once you create a controller, you will not be able to add additional services.

backend/core/Helpers/ExceptionHelper.cs
devinleighsmith: Nice.

PR #82, opened 2020-03-23
PIMS-141 add role based access to the App Nav Bar.

devinleighsmith:
(using PIMS roles, not keycloak roles).

This update also includes changes to handle getting roles into the frontend from keycloak.
Needed to tweak Christine's UI to support a responsive layout (too many links to display on mobile).

PLEASE DO NOT MERGE THIS as we are about to demo.

### DIFF
--- a/auth/keycloak/config/realm-export.json
+++ b/auth/keycloak/config/realm-export.json
@@ -1102,6 +1102,22 @@
       "authenticationFlowBindingOverrides": {},
       "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": -1,
+      "protocolMappers": [
+        {
+          "id": "737b4e24-9d20-4d6d-bc4f-cef73675109b",
+          "name": "Groups",
+          "protocol": "openid-connect",
+          "protocolMapper": "oidc-group-membership-mapper",
+          "consentRequired": false,
+          "config": {
+            "full.path": "false",
+            "id.token.claim": "false",
+            "access.token.claim": "true",
+            "claim.name": "Groups",
+            "userinfo.token.claim": "true"
+          }
+        }
+      ],
       "defaultClientScopes": [
         "web-origins",
         "role_list",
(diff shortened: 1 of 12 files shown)


auth/keycloak/config/realm-export.json
```diff
@@ -1102,6 +1102,22 @@
       "authenticationFlowBindingOverrides": {},
       "fullScopeAllowed": true,
       "nodeReRegistrationTimeout": -1,
+      "protocolMappers": [
```
Fosol: What's for?  Is it result in the token including Groups?

devinleighsmith: This allows keycloak to return roles(groups) into the userinfo object by default. This means that the keycloak library that we are using throughout the application has this information in context.

Fosol: Might need to do this in DEV.  I'd also be careful with linking Groups to front-end functionality.  The groups are designed to be arbitrary.  Functionality is only concerned with 'Keycloak:Roles/PIMS:Claims'.

Fosol: Don't need to change this PR.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,22 +1,27 @@
...
+  const hasClaim = (claim?: string) => {
+    return !claim || keycloak?.hasResourceRole(claim);
   };
   const hasRole = (role?: string) => {
-    return !role || keycloak?.realmAccess?.roles.includes(role);
+    return !role || (keycloak?.userInfo as any)?.Groups?.includes(role);
+  };
+  const roles = () => {
+    return (keycloak?.userInfo as any)?.Groups;
   };
```
asanchezr: One minor observation: We should return here a "copy" of the original "Groups" array instead of a reference to it. Someone with enough coding knowledge could exploit this to add the admin role to the groups and subsequent checks for "isAdmin" would succeed...

I think it would be safer to replace with the following;

```typescript
  const roles = () => {
    return [...(keycloak?.userInfo as any)?.Groups];
  }; 
```

Thoughts?

PR #85, opened 2020-03-24
Pims 976

devinleighsmith:


backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -153,11 +153,25 @@ public IActionResult DeleteUser([FromBody] Model.UserModel model)
             return new JsonResult(model);
         }
 
+        /// <summary>
+        /// Update an access request, generally to grant/deny it.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost("/api/admin/access/request/")]
```
Fosol: This should be a PUT as it's for updating not creating a new one.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+        /// Updates the specified access request in the datasource. if the request is granted, update the associated user as well.
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        /// <returns></returns>
+        public AccessRequest Update(AccessRequest entity)
+        {
+            if(entity.UserId == null)
+            {
+                throw new ArgumentNullException("userId");
```
Fosol: `nameof(entity.UserId)`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+            }
+            entity.ThrowIfNull(nameof(entity));
```
Fosol: If `entity` is null then this line needs to be first.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+            this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin, Permissions.AgencyAdmin);
```
Fosol: I would change this to just `Permissions.AdminUsers`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+
+            var accessRequest = this.Context.AccessRequests
+                .Include(p => p.Agencies)
+                .ThenInclude(p => p.Agency)
+                .Include(p => p.Roles)
+                .ThenInclude(p => p.Role)
+                .Include(p => p.User)
+                .FirstOrDefault(a => a.Id == entity.Id) ?? throw new KeyNotFoundException();
+            if(!accessRequest.IsGranted != true && entity.IsGranted == true && accessRequest.UserId.HasValue)
```
Fosol: Any reason for the double negative?  `accessRequest.IsGranted == true`.  No reason to double check if `accessRequest.UserId.HasValue` as you've checked earlier already.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+            {
+                User user = Get(accessRequest.UserId.Value);
+                entity.Agencies.ForEach((accessRequestAgency) =>
+                {
+                    if (!user.Agencies.Select(a => a.AgencyId).Contains(accessRequestAgency.AgencyId))
```
Fosol: Change to `!user.Agencies.Any(a => a.AgencyId == accessRequestAgency.AgencyId)`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+                    {
+                        user.Agencies.Add(new UserAgency()
+                        {
+                            User = user,
+                            Agency = this.Context.Agencies.Find(accessRequestAgency.AgencyId)
```
Fosol: No real value in fetching the agency here.  Just set the `AgencyId=accessRequestAgency.AgencyId`.  The result should be the same with one less trip to the DB.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+                        });
+                    }
+                });
+                entity.Roles.ForEach((accessRequestRole) =>
+                {
+                    if (!user.Roles.Select(r => r.RoleId).Contains(accessRequestRole.RoleId))
```
Fosol: Change to `!user.Roles.Any(r => r.RoleId == accessRequestRole.RoleId)`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+                    {
+                        user.Roles.Add(new UserRole()
+                        {
+                            User = user,
+                            Role = this.Context.Roles.Find(accessRequestRole.RoleId)
```
Fosol: No value in fetching the role here.  Just set the `RoleId=accessRequestAgency.RoleId`.  The result should be the same with one less trip to the DB.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+                        });
+                    }
+                });
+                Update(user);
```
Fosol: This update to the user will occur in its own transaction.  You need to use `UpdateOne(user)`.  

Think I need a better naming convention for those methods...

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -150,6 +150,65 @@ public override void Remove(User entity)
...
+            }
+
+            this.Context.Entry(accessRequest).CurrentValues.SetValues(entity);
+            accessRequest.UpdatedById = this.User.GetUserId();
+            accessRequest.UpdatedOn = DateTime.UtcNow;
+            this.Context.Set<AccessRequest>().Update(accessRequest);
```
Fosol: I'd change to `this.Context.AccessRequests.Update(accessRequest)`.  No need to use the generic method here, just adds additional reflection calls within EF.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -167,12 +226,10 @@ public Paged<AccessRequest> GetAccessRequestsNoTracking(int page = 1, int quanti
                 .Include(p => p.Roles)
                 .ThenInclude(p => p.Role)
                 .Include(p => p.User)
-                .AsNoTracking();
+                .AsNoTracking()
+                .Where(ar => ar.User != null); // Access Requests with no user cannot be granted.
```
Fosol: We shouldn't be saving access requests without users.  Makes me wonder if the additional filter here is needed.

backend/dal/Services/Admin/Concrete/UserService.cs
devinleighsmith: doesn't work for primatives.

backend/dal/Services/Admin/Concrete/UserService.cs
devinleighsmith: I originally wasn't enforcing the non-null user, missed updating the trickle down logic after I started enforcing it.

PR #88, opened 2020-03-27
PIMS-1220, PIMS-1221, PIMS-1222

devinleighsmith:
Automap Evaluations
Move logic from controller to DAL

### DIFF
--- a/backend/api/Controllers/ParcelController.cs
+++ b/backend/api/Controllers/ParcelController.cs
@@ -67,6 +67,7 @@ namespace Pims.Api.Controllers
         /// <returns></returns>
         [HttpPost]
         [HasPermission(Permissions.PropertyView)]
+        [Route("/api/parcels/filter")]
         public IActionResult GetParcels([FromBody]ParcelFilter filter)
         {
             filter.ThrowBadRequestIfNull($"The request must include a filter.");
@@ -101,14 +102,6 @@ namespace Pims.Api.Controllers
         {
             var entity = _mapper.Map<Entity.Parcel>(model);
             var userId = this.User.GetUserId();
-
-            foreach (var building in model.Buildings)
-            {
-                // We only allow adding buildings at this point.  Can't include an existing one.
-                var b_entity = _mapper.Map<Entity.Building>(building);
-                b_entity.CreatedById = userId;
-                entity.Buildings.Add(b_entity);
-            }
 
             _pimsService.Parcel.Add(entity);
             var parcel = _mapper.Map<ParcelModel>(entity);
--- a/backend/dal/PIMSContext.cs
+++ b/backend/dal/PIMSContext.cs
@@ -1,3 +1,7 @@
+using System;
+using System.Linq;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
 using Microsoft.EntityFrameworkCore;
 using Pims.Dal.Configuration;
 using Pims.Dal.Entities;
@@ -27,7 +31,9 @@ namespace Pims.Dal
         public DbSet<Province> Provinces { get; set; }
         public DbSet<User> Users { get; set; }
         public DbSet<Role> Roles { get; set; }
-        public DbSet<Claim> Claims { get; set; }
+        public DbSet<Pims.Dal.Entities.Claim> Claims { get; set; }
+
+        private readonly IHttpContextAccessor _httpContextAccessor;
         #endregion
 
         #region Constructors
@@ -45,9 +51,9 @@ namespace Pims.Dal
         /// </summary>
         /// <param name="options"></param>
         /// <returns></returns>
-        public PimsContext(DbContextOptions<PimsContext> options) : base(options)
+        public PimsContext(DbContextOptions<PimsContext> options, IHttpContextAccessor httpContextAccessor = null) : base(options)
         {
-
+            _httpContextAccessor = httpContextAccessor;
         }
         #endregion
 
@@ -101,6 +107,31 @@ namespace Pims.Dal
             // }
 
             base.OnModelCreating(modelBuilder);
+        }
+
+        public override int SaveChanges()
+        {
+            // get entries that are being Added or Updated
+            var modifiedEntries = ChangeTracker.Entries()
+                    .Where(x => (x.State == EntityState.Added || x.State == EntityState.Modified));
+            var userId = _httpContextAccessor.HttpContext.User.FindFirst(ClaimTypes.NameIdentifier).Value;
+            if (userId != null)
+            {
+                foreach (var entry in modifiedEntries)
+                {
+                    var entity = entry.Entity as BaseEntity;
+
+                    if (entry.State == EntityState.Added)
+                    {
+                        entity.CreatedById = new Guid(userId);
+                        entity.CreatedOn = DateTime.UtcNow;
+                    }
+                    entity.UpdatedById = new Guid(userId);
+                    entity.UpdatedOn = DateTime.UtcNow;
+                }
+            }
+
+            return base.SaveChanges();
         }
 
         /// <summary>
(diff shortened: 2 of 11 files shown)


backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -200,25 +201,106 @@ public Parcel Add(Parcel parcel)
...
+            if (existingParcel.AgencyId != parcel.AgencyId) throw new NotAuthorizedException("Parcel cannot be transferred to the specified agency.");
 
             this.Context.Parcels.ThrowIfNotUnique(parcel);
 
-            this.Context.Entry(entity).CurrentValues.SetValues(parcel);
-            entity.UpdatedById = this.User.GetUserId();
-            entity.UpdatedOn = DateTime.UtcNow;
-
-            this.Context.Parcels.Update(entity);
+            //Add/Update a parcel and all child collections
```
devinleighsmith: this entire 60ish block could be replaced by:
context.Update(rootEntity);
context.SaveChanges();
which handles "Upserting" the parent and all child entities, supporting mixing and matching adding new entities and updating existing entities. The problem is that "Upserting" does not support deletes. My original thought was that I could messily handle the deletes by loading the existing parcel and then disposing so I could use Context.Update. However, the concept of detaching entities with their children is not supported as of this time. https://github.com/dotnet/efcore/issues/18406

Only solution at this time is to not handle any kinds of delete operations in this update function.

backend/api/Controllers/ParcelController.cs
```diff
@@ -67,6 +67,7 @@ public IActionResult GetParcels()
         /// <returns></returns>
         [HttpPost]
         [HasPermission(Permissions.PropertyView)]
+        [Route("/api/parcels/filter")]
```
devinleighsmith: Can't have two posts with the same route. @Fosol let me know if this is a breaking change.

Fosol: Should be fine, these aren't used yet.

backend/dal/PIMSContext.cs
```diff
@@ -103,6 +109,31 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             base.OnModelCreating(modelBuilder);
         }
 
+        public override int SaveChanges()
+        {
+            // get entries that are being Added or Updated
+            var modifiedEntries = ChangeTracker.Entries()
+                    .Where(x => (x.State == EntityState.Added || x.State == EntityState.Modified));
+            var userId = _httpContextAccessor.HttpContext.User.FindFirst(ClaimTypes.NameIdentifier).Value;
```
Fosol: Do you have access to the extension method `User.GetUserId()`?

backend/dal/PIMSContext.cs
```diff
@@ -103,6 +109,31 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                foreach (var entry in modifiedEntries)
+                {
+                    var entity = entry.Entity as BaseEntity;
+
+                    if (entry.State == EntityState.Added)
+                    {
+                        entity.CreatedById = new Guid(userId);
+                        entity.CreatedOn = DateTime.UtcNow;
+                    }
+                    entity.UpdatedById = new Guid(userId);
```
Fosol: Don't do this if it's a `EntityState.Delete` as it will most likely confuse EF with related properties.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -200,25 +201,106 @@ public Parcel Add(Parcel parcel)
...
         public Parcel Update(Parcel parcel)
         {
-            parcel.ThrowIfNotAllowedToEdit(nameof(parcel), this.User, Permissions.PropertyEdit);
+            var existingParcel = this.Context.Parcels
+                .Include(p => p.Agency)
+                .Include(p => p.Address)
+                .Include(p => p.Evaluations)
+                .Include(p => p.Buildings).ThenInclude(b => b.Evaluations)
+                .Include(p => p.Buildings).ThenInclude(b => b.Address)
+                .SingleOrDefault(u => u.Id == parcel.Id) ?? throw new KeyNotFoundException();
```
Fosol: Probably should change to `FirstOrDefault(...)`

backend/dal/Services/Concrete/ParcelService.cs
Fosol: I think not supporting delete here makes the most sense presently.

backend/dal/PIMSContext.cs
devinleighsmith: oh, right. let me see.

backend/dal/PIMSContext.cs
devinleighsmith: sure, updating.

backend/dal/Services/Concrete/ParcelService.cs
devinleighsmith: Hmm, I don't know, I kind of like the documented behaviour of this when we are searching for a unique key:

Returns the only element of a sequence, or a default value if the sequence is empty; this method throws an exception if there is more than one element in the sequence.

seems appropriate since there should never be more than one match.

asanchezr:
Couple of comments re: process. 

- The title of this PR is PIMS-976 but I can't seem to find that issue in JIRA. A more descriptive title would help me understand what I should be reviewing
- The PR "source" branch is named PIMS-1220 which doesn't match the issue number of the PR title so I'm a bit lost what this PR is about. PIMS-1220 is about single parcel GET and I see a lot of changes related to creating, updating and deleting parcels...

backend/dal/Services/Concrete/ParcelService.cs
Fosol: At the end of the day it doesn't change much here.  As the LINQ statement becomes an SQL statement and if it returns more than one it would throw an error.  However since you're making a request for the primary key there never will be more than one or none returned.  So using `Single` technically is doing more work than is required.

backend/api/Controllers/ParcelController.cs
asanchezr: I understand that there can be no two POST on the same route. But with this change it is now confusing to the API consumer how to filter parcels. There's now two (different) routes to do it:

- GET `/api/parcels?field1=value1&field2=value2`
- POST `/api/parcels/filter/` with a body of  `{ field1: value1, ...}`

I know the POST is there for convenience but I rather remove the [HttpPost] attribute and let it be a GET only operation. Note that I'm not saying we discard the whole `GetParcels(filter)` function, just not exposing it via HTTP

asanchezr: There's a lot of debate out there on what is "right", I only ask we remain consistent, ideas?

- https://stackoverflow.com/questions/19637459/rest-api-using-post-instead-of-get
- https://stackoverflow.com/questions/46585/when-do-you-use-post-and-when-do-you-use-get
- https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/

devinleighsmith:
@asanchezr Agreed. I messed up the commit message on this one, I'm not in any rush so I'll correct it.

Also I ended up doing all three API stories at the same time. Theoretically I'd rather have submitted a PR for each one separately. However, realistically I updated all three at the same time as
1) I was using the API outputs to test the inputs, so as I updated one API I often found/resolved issues in the others.
2) There were some foundational changes required by all three APIs.

I'm also really annoyed that Github doesn't allow me to respond directly to your comment above. Why Github, Why?

backend/api/Controllers/ParcelController.cs
devinleighsmith: I think this is outside the scope of this review since I think you had the same issue with the GET/POST before I was forced to rename it. Feel free to discuss that here but I'm not planning on changing anything at this time.

PR #89, opened 2020-03-27
move access request logic to keycloak controller/service. Update to c…

devinleighsmith:
…all keycloak update user method.

### DIFF
--- a/backend/api/Areas/Keycloak/Controllers/UserController.cs
+++ b/backend/api/Areas/Keycloak/Controllers/UserController.cs
@@ -107,6 +107,20 @@ namespace Pims.Api.Areas.Keycloak.Controllers
 
             return new JsonResult(result);
         }
+
+        /// <summary>
+        /// Update an access request, generally to grant/deny it.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPut("access/request")]
+        public async Task<IActionResult> UpdateAccessRequests(Pims.Api.Models.AccessRequestModel updateModel)
+        {
+            var entity = _mapper.Map<Entity.AccessRequest>(updateModel);
+            var updatedEntity = await _keycloakService.UpdateAccessRequest(entity);
+
+            var user = _mapper.Map<Pims.Api.Models.AccessRequestModel>(updatedEntity);
+            return new JsonResult(user);
+        }
         #endregion
     }
 }
--- a/backend/api/Areas/Keycloak/Controllers/UserController.cs
+++ b/backend/api/Areas/Keycloak/Controllers/UserController.cs
@@ -113,10 +113,10 @@ namespace Pims.Api.Areas.Keycloak.Controllers
         /// </summary>
         /// <returns></returns>
         [HttpPut("access/request")]
-        public async Task<IActionResult> UpdateAccessRequests(Pims.Api.Models.AccessRequestModel updateModel)
+        public async Task<IActionResult> UpdateAccessRequestAsync(Pims.Api.Models.AccessRequestModel updateModel)
         {
             var entity = _mapper.Map<Entity.AccessRequest>(updateModel);
-            var updatedEntity = await _keycloakService.UpdateAccessRequest(entity);
+            var updatedEntity = await _keycloakService.UpdateAccessRequestAsync(entity);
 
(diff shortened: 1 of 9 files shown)


backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -160,6 +162,51 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
...
             return euser;
         }
+
+        /// <summary>
+        /// Updates the specified access request in the datasource. if the request is granted, update the associated user as well.
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        /// <returns></returns>
+        public async Task<Entity.AccessRequest> UpdateAccessRequest(Entity.AccessRequest entity)
```
Fosol: Rename to `UpdateAccessRequestAsync` to keep in alignment with other async pattern.

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -107,6 +107,20 @@ public async Task<IActionResult> UpdateUserAsync(Guid id, [FromBody] Model.Updat
 
             return new JsonResult(result);
         }
+
+        /// <summary>
+        /// Update an access request, generally to grant/deny it.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPut("access/request")]
+        public async Task<IActionResult> UpdateAccessRequests(Pims.Api.Models.AccessRequestModel updateModel)
```
Fosol: Rename to `UpdateAccessRequestsAsync` to keep in alignment with other async naming pattern.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -151,59 +151,36 @@ public override void Remove(User entity)
...
-            this.Context.Entry(accessRequest).CurrentValues.SetValues(entity);
-            accessRequest.UpdatedById = this.User.GetUserId();
-            accessRequest.UpdatedOn = DateTime.UtcNow;
-            this.Context.AccessRequests.Update(accessRequest);
-
-            this.Context.CommitTransaction();
-            return accessRequest;
+                .AsNoTracking()
+                .Where(ar => ar.Id == id)
+                .SingleOrDefault();
```
Fosol: Change to `FirstOrDefault()`, although I'm not certain it's a performance gain in SQL.

Fosol: I suspect you should also add `?? throw KeyNotFoundException();` as well.

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -160,6 +162,51 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
...
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        /// <returns></returns>
+        public async Task<Entity.AccessRequest> UpdateAccessRequestAsync(Entity.AccessRequest entity)
+        {
+            entity.ThrowIfNull(nameof(entity));
+            entity.ThrowIfNull(nameof(entity.UserId));
+
+            this._user.ThrowIfNotAuthorized(Permissions.AdminUsers, Permissions.AgencyAdmin);
+            var accessRequest = _pimsAdminService.User.GetAccessRequestNoTracking(entity.Id);
+            if (!accessRequest.IsGranted != true && entity.IsGranted == true)
```
asanchezr: /nit: Please clean up the logic here. Double negations are harder to understand.

Isn't `if (!accessRequest.IsGranted != true && entity.IsGranted == true)` the same as `if (accessRequest.IsGranted == true && entity.IsGranted == true)` ?

If the property can be null I'd rather do `var value = accessRequest.IsGranted ?? false` and then check the value?

devinleighsmith: should actually be accessRequest.IsGranted != true, not sure how this works as is. Will update in my next PR.

PR #92, opened 2020-03-30
PIMS-1229 DeleteParcel unit tests

Fosol:


backend/test/Controllers/ParcelControllerTest.cs
```diff
@@ -49,28 +49,6 @@ public void DeleteParcel_Success()
             Assert.Equal(mapper.Map<Model.ParcelModel>(parcel), actualParcel);
             service.Verify(m => m.Parcel.Remove(It.IsAny<Entity.Parcel>()), Times.Once());
         }
-
-        [Fact]
-        public void DeleteParcel_NotAuthorized()
```
Fosol: Invalid test.

PR #96, opened 2020-03-31
PIMS-1284, PIMS-1227 Resolving AutoMapper issues

Fosol:
In an effort to resolve AutoMapper challenges along side EF Core Concurrency issues, I have forced us to use unique models for each controller.  The only shared models will be the `BaseModel` and `CodeModel` for now.

I have tested a number of scenarios, but still need to do more testing to confirm nothing is broken.  Please let me know if you discover anything?

This means all Models and AutoMapper Profiles will be found in their own namespace that matches the Controller.

All Controllers now only should return untracked entities.  I still have to update a few scenarios.

Don't use the EF.Update(entity) or EF.Add(entity) methods unless you are extremely careful with what you are doing.  Using these will result in either ignored concurrency, or they will edit tables they should not be touching.


backend/api/Areas/Admin/Controllers/RoleController.cs
```diff
@@ -105,7 +105,7 @@ public IActionResult AddRole([FromBody] Model.RoleModel model)
             _pimsAdminService.Role.Add(entity);
             var role = _mapper.Map<Model.RoleModel>(entity);
 
-            return new CreatedAtActionResult(nameof(GetRole), nameof(RoleController), new { id = role.Id }, role);
+            return CreatedAtAction(nameof(GetRole), new { id = role.Id }, role);
```
Fosol: Fixed a bug.

backend/api/Pims.Api.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8" ?>
+<?xml version="1.0" encoding="utf-8"?>
```
Fosol: Organizing the solution for Visual Studio.
I've given up on vscode to work on the API.  Way too many issues with it.

backend/api/Properties/launchSettings.json
```diff
@@ -3,25 +3,25 @@
     "windowsAuthentication": false,
     "anonymousAuthentication": true,
     "iisExpress": {
-      "applicationUrl": "http://localhost:53591/",
+      "applicationUrl": "http://localhost:5000/",
```
Fosol: Launch in Visual Studio locally.

backend/api/Areas/Keycloak/Profiles/User/AgencyProfile.cs
```diff
@@ -13,6 +12,9 @@ public class AgencyProfile : Profile
         #region Constructors
         public AgencyProfile()
         {
+            CreateMap<Entity.Agency, Entity.Agency>()
```
devinleighsmith: odd.

backend/api/Areas/Keycloak/Profiles/User/Resolvers/UpdateRoleToEntityResolver.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+using System.Linq;
+using AutoMapper;
+using Entity = Pims.Dal.Entities;
+using Model = Pims.Api.Areas.Keycloak.Models.User;
+
+namespace Pims.Api.Areas.Keycloak.Profiles.User.Resolvers
+{
+    /// <summary>
+    /// UpdateRoleToEntityResolver class, provides automapper configuration to convert entiy agencies to model agencies.
```
devinleighsmith: should be converting entity roles to model roles I believe.

backend/api/Areas/Keycloak/Profiles/User/AgencyProfile.cs
asanchezr: ^ See? that's what happens when I have to review 200+ files. After the 20th my attention span drops exponentially LOL

Good catch Devin

Fosol: I agree, but I was testing how we could copy values from one entity to another with AutoMapper.  I think I ended up not doing it.  Probably can remove this.

backend/api/Areas/Keycloak/Profiles/User/Resolvers/UpdateRoleToEntityResolver.cs
Fosol: This copies a Role Model into an Entity.

Fosol:
The majority of the effort (files) were models or profiles.  Other than creating 20 PRs it didn't make a lot of sense to break them up.

backend/dal/Helpers/Extensions/DbContextExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <typeparam name="T"></typeparam>
+        /// <param name="context"></param>
+        /// <param name="source"></param>
+        public static void SetOriginalRowVersion<T>(this DbContext context, T source) where T : BaseEntity
+        {
+            context.Entry(source).OriginalValues[nameof(BaseEntity.RowVersion)] = source.RowVersion;
+        }
+
+        /// <summary>
+        /// Detach the entity from the context.
```
devinleighsmith, line 23: might be worth noting that this only detaches the top-level entity.

backend/api/Areas/Keycloak/Profiles/User/AgencyProfile.cs
Fosol: This was required for the Keycloak User management.

backend/dal/Services/Admin/Concrete/BaseService`.cs
```diff
@@ -46,7 +34,9 @@ public virtual ET Find(params object[] keyValues)
         {
             this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin, Permissions.AgencyAdmin);
 
-            return this.Context.Set<ET>().Find(keyValues);
+            var result = this.Context.Set<ET>().Find(keyValues);
+            this.Context.Entry(result).State = Microsoft.EntityFrameworkCore.EntityState.Detached; // Force detach so that outside the DAL the DB cannot be manipulated.
```
devinleighsmith, line 38: I think this only detaches the top level entity.

Fosol: Good point.  I should update the comment.

devinleighsmith: I also experimented with manually parsing all entities in context and removing them one-by-one. This threw a number of weird errors for me, which I believe is because removing entities from context in this manner is not a supported workflow:
https://github.com/dotnet/efcore/issues/18406#issuecomment-543984411

Officially the way to remove all tracking is to dispose of the active dbcontext. This won't always work for us because we get our dbcontext injected into the controller action, so I'm not sure what the best way forward is.

Fosol: The best way would probably be not passing Entities outside of the DAL.  But I don't want to do that atm.

PR #97, opened 2020-04-01
Fix backend docker build

asanchezr:


Fosol:
I don’t understand including the import tool in the docker image?

asanchezr:
> I don’t understand including the import tool in the docker image?

Is either that or you removing the Import tool from the `.sln` file. 
Check your last PR where you added it to the solution (I guess so you could see it on VS). 
I am happy to instead remove the project from the solution (or actually revert it to what it was before before all the extra "cruft")

Let me know what you prefer

asanchezr:
All I wanted was to unblock the docker build. I agree the import tool shouldn't be on the docker image. However, we only copy the `api.dll` over not everything so it "should" be okay. I still prefer we take out the import project from the solution altogether

Fosol:
Take it out of the solution.

asanchezr:
Discarded changes to docker image; instead `Pims.sln` was reverted to exclude import tool.

PR #99, opened 2020-04-01
PIMS-1226, PIMS-1227, PIMS-1228, PIMS-1229, PIMS-1230

Fosol:
We shouldn't be adding code to models to support unit tests.  I've removed `IEquatable` from every model.  If at some point the application itself requires the ability to compare two models I would suggest creating a custom `IEqualityComparer`, but it won't hurt to use `IEquatable` for those specific occurrences.

Additionally I have created two custom IEqualityComparer's (`ShallowPropertyCompare`, and `DeepPropertyCompare`).  These provided the basic tests that we were using `IEquatable` for, without all the overhead.

Sorry again for the large PR, but it doesn't make sense to do 10 PRs for the same effort.

### DIFF
--- a/backend/Pims.sln
+++ b/backend/Pims.sln
@@ -9,13 +9,38 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Pims.Api.Test", "test\Pims.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Pims.Dal", "dal\Pims.Dal.csproj", "{6DFFF5E1-1B87-403B-99D0-A9E03D8A8EB3}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Pims.Dal.Entities", "entities\Pims.Dal.Entities.csproj", "{1C724CD5-CD24-46CD-835A-A83F673F97B5}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Pims.Dal.Entities", "entities\Pims.Dal.Entities.csproj", "{1C724CD5-CD24-46CD-835A-A83F673F97B5}"
 EndProject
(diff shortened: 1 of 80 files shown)


backend/Pims.sln
```diff
@@ -9,13 +9,38 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Pims.Api.Test", "test\Pims.
 EndProject
```
Fosol, line 9: Solution file updated but without including the Import project.

backend/test/Helpers/PrincipalHelper.cs
```diff
@@ -15,12 +15,12 @@ public static class PrincipalHelper
...
         /// <param name="role"></param>
         /// <returns></returns>
-        public static ClaimsPrincipal CreateForRole(string role)
+        public static ClaimsPrincipal CreateForRole(string role = null)
         {
             var user = new ClaimsPrincipal(new ClaimsIdentity(new System.Security.Claims.Claim[]
             {
                 new System.Security.Claims.Claim (ClaimTypes.NameIdentifier, Guid.NewGuid ().ToString ()),
-                    new System.Security.Claims.Claim (ClaimTypes.Role, role)
+                    new System.Security.Claims.Claim (ClaimTypes.Role, role ?? "none")
```
Fosol, line 23: Can now mimic an authenticated user who is not granted access yet.

backend/test/Pims.Api.Test.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.InMemory" Version="3.1.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
     <PackageReference Include="Moq" Version="4.13.1" />
+    <PackageReference Include="NExpect" Version="1.0.168" />
```
Fosol, line 13: Added NExpect for some nice asserts if anyone is interested.

Fosol:
Models are now truly light weight.  Just a bunch of properties and easier to maintain.

PR #101, opened 2020-04-01
PIMS-599 Map Filter

asanchezr:
- add building blocks for frontend forms
- align frontend service calls with backend models
- add icon button component


frontend/src/components/common/form/Form.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/**
+ * Formik-connected form control that provides thin wrappers for React-Bootstrap form controls;
+ * e.g. <Form.Row>, <Form.Group>, <Form.Control>, etc.
+ */
+const Form = ({ className, children, ...rest }: FormProps) => {
+  const { handleSubmit } = useFormikContext();
+  return (
+    <FormBase noValidate className={className} onSubmit={handleSubmit} {...(rest as any)}>
```
Fosol, line 20: Can you just go `{...any}`?

frontend/src/components/common/form/Select.tsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
```
Fosol, line 1: Does this support dynamic list?

frontend/src/utils/utils.ts
```diff
@@ -0,0 +1,27 @@
+export const classSet = (input?: any): string => {
```
Fosol, line 1: Can you add some documentation `jsdoc` to these?

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
```diff
@@ -39,8 +39,11 @@ describe('fetchParcels action creator', () => {
       neLongitude: 2,
       swLatitude: 3,
       swLongitude: 4,
-      agencyId: null,
-      propertyClassificationId: null,
+      address: null,
```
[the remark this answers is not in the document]
asanchezr: as far as I know `address` is a string and that's what the frontend sends. Strings can be `null` in which case the API call looks like: `/api/parcels?address=&otherparams...`

frontend/src/components/common/form/DisplayError.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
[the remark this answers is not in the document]
asanchezr: yes. the idea is that you describe the "shape" of your form via `Yup` (an schema library) and `<Formik>` (the form library chosen for the frontend) will attach error messages based on the form schema defined. We can go into more details on our next dev training session.

frontend/src/components/common/form/Form.tsx
asanchezr: Nope. The `as any` is to "turn off" the types inferred by typescript. This is a common pattern in React where creating a custom typing when you are "extending" a base class (like here we are wrapping the react-bootstrap <Form>). It's fine as in JavaScript the types are removed and it ends up being `{...rest}` which is what we want

frontend/src/utils/utils.ts
asanchezr: tech-debt yes

frontend/src/components/common/form/Select.tsx
asanchezr: The select supports passing it an `options` array so yes that's a dynamic list. See the usage in `MapFilterBar.tsx`

```tsx
<Select
    field="classificationId"
    placeholder="View by Classification"
    options={classifications}
```

where the `classifications` array comes from a previous call to the API.
Hope that answers your question...

PR #103, opened 2020-04-01
PIMS-1010 Adding unit tests

Fosol:


backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -116,9 +116,6 @@ public IActionResult GetMyUsers(EModel.UserFilter filter)
         public IActionResult GetUser(Guid id)
         {
             var entity = _pimsAdminService.User.Get(id);
-
-            if (entity == null) return NoContent();
```
Fosol, line 120: Not needed as the Get will throw a KeyNotFoundException which is handled by our middlelayer which will then return a BadRequest.

backend/api/Controllers/UserController.cs
```diff
@@ -90,15 +83,15 @@ public IActionResult AddAccessRequest([FromBody] Model.AccessRequestModel access
         {
             if (accessRequestModel == null || accessRequestModel.Agencies == null || accessRequestModel.Roles == null)
             {
-                return BadRequest("Invalid access request specified");
+                throw new BadRequestException("Invalid access request specified");
```
Fosol, line 86: We want our middleware to handle responses so we get consistent JSON messages.

backend/test/Controllers/Admin/UserControllerTest.cs
```diff
@@ -19,66 +19,9 @@ namespace PimsApi.Test.Admin.Controllers
     [Trait("group", "user")]
     public class UserControllerTest
     {
-        #region Variables
-        private static readonly int AGENCY_ID = 2;
-        private static readonly Guid ROLE_ID = Guid.NewGuid();
-        private static readonly Guid USER_ID = Guid.NewGuid();
-        private static readonly Guid ACCCESS_REQUEST_ID = Guid.NewGuid();
-        private readonly Entity.AccessRequest _expectedAccessRequest = new Entity.AccessRequest()
```
Fosol, line 27: All of this setup has been extracted out to helpers.

backend/test/Helpers/ControllerHelper.cs
```diff
@@ -105,12 +105,23 @@ public static void MockConstructorArguments<T>(this TestHelper helper, params ob
...
             foreach (var carg in cargs)
             {
-                var gmake = gmock.MakeGenericType(carg.ParameterType);
-                var mockObjectProp = gmock.GetCachedProperties().FirstOrDefault(p => p.Name == nameof(Mock.Object) && !p.PropertyType.IsGenericParameter) ?? throw new InvalidOperationException($"The mocked type '{type.Name}' was unable to determine the correct 'Object' property.");
+                if (helper.Services.Any(s => s.ServiceType == carg.ParameterType)) continue;
 
                 // If an 'args' type matches, use it for the mock.
                 var arg = args.FirstOrDefault(a => a.GetType() == carg.ParameterType);
-                var mock = arg != null ? Activator.CreateInstance(gmake, arg) : Activator.CreateInstance(gmake);
+                if (arg == null) arg = args.FirstOrDefault(a => carg.ParameterType.IsAssignableFrom(a.GetType()));
```
Fosol, line 112: Fixing issue where passing in args wasn't being added to ServicesCollection.

backend/test/Helpers/TestHelper.cs
```diff
@@ -51,11 +51,12 @@ static TestHelper()
         /// </summary>
         public TestHelper()
         {
-            var config = new Mock<IConfiguration>();
-            this.AddSingleton(config);
-            this.AddSingleton(config.Object);
-            var keycloakOptions = new Mock<IOptionsMonitor<Keycloak.Configuration.KeycloakOptions>>();
-            this.AddSingleton(keycloakOptions.Object);
+            //var config = new Mock<IConfiguration>();
```
Fosol, line 54: No longer required.  Will clean up later.

backend/api/Areas/Admin/Controllers/UserController.cs
asanchezr: Perhaps consider leaving a comment here so we don't add this logic back?

backend/api/Controllers/UserController.cs
asanchezr: See my previous comment about leaving a comment for other devs who might need to modify/maintain this area?

backend/test/Controllers/Admin/UserControllerTest.cs
```diff
@@ -139,47 +86,71 @@ public void GetAccessRequests_PageMax_Success()
...
-            Assert.Equal(mapper.Map<Model.UserModel[]>(expectedUsers), actualAccessRequest.Items, new DeepPropertyCompare());
+            var actualResult = Assert.IsType<Entity.Models.Paged<Model.UserModel>>(actionResult.Value);
+            Assert.Equal(mapper.Map<Model.UserModel[]>(users), actualResult.Items, new DeepPropertyCompare());
             service.Verify(m => m.User.Get(It.IsAny<Entity.Models.UserFilter>()), Times.Once());
         }
         #endregion
 
+        #region GetUser
+        [Fact]
+        public void GetUser()
```
asanchezr, line 133: It seems the pattern for tests are to name them `*_Success` but this one doesn't follow that naming...

PR #104, opened 2020-04-02
PIMS-1407 GET - Properties (Parcels + Buildings)

asanchezr:
- replace `/api/parcels` calls with `/api/properties`
- accommodate new type IBuilding. Add building popup
- refactor map view to support buildings
- fix API calls
- fix failing unit tests


frontend/src/constants/API.tsx
```diff
@@ -33,7 +33,7 @@ export const PARCELS = (params: IParcelListParams | null) =>
 export interface IParcelDetailParams {
   id: number;
 }
-export const PARCEL_DETAIL = (params: IParcelDetailParams) => `/parcels/${params.id}`;
+export const PARCEL_DETAIL = (params: IParcelDetailParams) => `/properties/${params.id}`;
```
Fosol, line 36: An endpoint does not exist for `/properties/{id}`.  You'll need to make a request to the correct endpoint based on the `PropertyTypeId`.

Parcel = `/parcels/{id}`
Building = `/buildings/{id}`

`/properties` returns an array of *property* objects.
```
{
  propertyTypeId: 0 or 1
  ...
}
```
Parcel = 0
Building = 1

asanchezr: yeah I haven't made the changes yet. I just rebased and force pushed my branch. I will remove the **DRAFT PR** label when this is ready for review and ask the team to review. Until then this is in draft mode

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -28,19 +28,19 @@ type MapProps = {
   lat: number;
   lng: number;
   zoom: number;
-  parcels: IParcel[];
+  parcels: IProperty[];
```
Fosol, line 31: I wonder if it makes more sense to update the naming conventions here and elsewhere.  If it’s a `property` object then use `properties` not `parcels`.

asanchezr: I thought about that and initially did it on my branch but then I saw Devin's PR and we will have **massive** conflicts which I'd like to avoid if possible.
We can file a tech-debt item to rename this across the frontend when the code is more stable after all the merges.

asanchezr:
@Fosol @devinleighsmith @jdtoombs the PR is ready for review. Have a look when you get a chance.
Thanks!

frontend/src/actions/parcelsActions.tsx
```diff
@@ -2,18 +2,19 @@ import * as ActionTypes from 'constants/actionTypes';
 
 //Parcel List API action
 
-export interface IParcel {
+export interface IProperty {
   id: number;
+  propertyTypeId: 0 | 1; // 0 = Parcel, 1 = Building
```
devinleighsmith, line 7: This can be tech dept, but I'd like to move these to named constants or enum at some point.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -99,6 +99,34 @@ const Map: React.FC<MapProps> = props => {
     });
   }, []);
 
+  // we need to namespace the keys as IDs are not enough here.
```
devinleighsmith, line 102: I like this refactor, easier to read for me at least.

frontend/src/actions/parcelsActions.tsx
asanchezr: Agree - we should create a custom `type PropertyType = 0 | 1`. I didn't go with `number` to be explicit about the fact that we only support 0 or 1 at the moment and avoid bugs where the propTypeId is set to "3"

PR #105, opened 2020-04-02
PIMS-599 Adding BuildingController.

Fosol:
Fixing Import bug.
Changing FE to user `/properties` endpoint instead of `/parcels`.  This requires frontend to be updated to make a request for `/parcels/{id}` or `/buildings/{id}` based on the `property.propertyTypeId`

0 = Parcel
1 = Building

### DIFF
--- a/frontend/src/constants/API.tsx
+++ b/frontend/src/constants/API.tsx
@@ -29,11 +29,16 @@ export interface IParcelListParams {
   maxLandArea: number | null;
 }
 export const PARCELS = (params: IParcelListParams | null) =>
-  params ? `/parcels?${queryString.stringify(params)}` : '/parcels'; // get filtered properties or all if not specified.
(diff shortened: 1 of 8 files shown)


frontend/src/constants/API.tsx
```diff
@@ -29,12 +29,17 @@ export interface IParcelListParams {
...
 }
 export const PARCELS = (params: IParcelListParams | null) =>
-  params ? `/parcels?${queryString.stringify(params)}` : '/parcels'; // get filtered properties or all if not specified.
+  params ? `/properties?${queryString.stringify(params)}` : '/properties'; // get filtered properties or all if not specified.
 export interface IParcelDetailParams {
   id: number;
 }
 export const PARCEL_DETAIL = (params: IParcelDetailParams) => `/parcels/${params.id}`;
 
+export interface IBuildingDetailParams {
```
Fosol, line 38: The logic for calling the `/buildings/{id}` endpoint needs to be added to the map.

PR #107, opened 2020-04-02
PIMS-1339 Including sub-agencies in filter

Fosol:
When making a request for properties with a parent agency it makes more sense to also include the sub-agencies.

This isn't technically a bug, but a user-experience issue.


Fosol:
Complete the prior PR before this one.

PR #110, opened 2020-04-06
Fixing docker build issue

Fosol:
Change to unit tests broke docker-compose.  This will fix it.

Also debugging why Swagger doesn't resolve in DEV.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -9,7 +9,9 @@ WORKDIR /src
 COPY *.sln .
 COPY api/*.csproj api/
 COPY entities/*.csproj entities/
-COPY test/*.csproj test/
+COPY tests/core/*.csproj tests/core/
+COPY tests/unit/api/*.csproj tests/unit/api/
+COPY tests/unit/dal/*.csproj tests/unit/dal/
 COPY dal/*.csproj dal/
 COPY dal.keycloak/*.csproj dal.keycloak/
 COPY keycloak/*.csproj keycloak/
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -18,6 +18,7 @@ using Microsoft.Extensions.Diagnostics.HealthChecks;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 using Microsoft.IdentityModel.Tokens;
+using Microsoft.OpenApi.Models;
 using Pims.Api.Helpers.Authorization;
 using Pims.Api.Helpers.Middleware;
 using Pims.Api.Helpers.Routes.Constraints;
@@ -236,15 +237,15 @@ namespace Pims.Api
 
             app.UseSwagger(options =>
             {
-                options.RouteTemplate = "/swagger/{documentname}/swagger.json";
+                options.RouteTemplate = this.Configuration.GetValue<string>("Swagger:RouteTemplate");
             });
             app.UseSwaggerUI(options =>
             {
                 foreach (var description in provider.ApiVersionDescriptions)
                 {
-                    options.SwaggerEndpoint($"/swagger/{description.GroupName}/swagger.json", description.GroupName);
+                    options.SwaggerEndpoint(String.Format(this.Configuration.GetValue<string>("Swagger:EndpointPath"), description.GroupName), description.GroupName);
                 }
-                options.RoutePrefix = "swagger";
+                options.RoutePrefix = this.Configuration.GetValue<string>("Swagger:RoutePrefix");
             });
 
             app.UseMiddleware(typeof(ErrorHandlingMiddleware));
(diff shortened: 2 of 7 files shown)


backend/api/Startup.cs
```diff
@@ -236,15 +237,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
 
             app.UseSwagger(options =>
             {
-                options.RouteTemplate = "/swagger/{documentname}/swagger.json";
+                options.RouteTemplate = "/api-docs/swagger/{documentname}/swagger.json";
```
Fosol, line 240: The original route doesn't appear to work on OpenShift.  Trying to see if a different route will work.

backend/Dockerfile
```diff
@@ -9,7 +9,9 @@ WORKDIR /src
 COPY *.sln .
 COPY api/*.csproj api/
 COPY entities/*.csproj entities/
-COPY test/*.csproj test/
+COPY tests/core/*.csproj tests/core/
```
Fosol, line 12: Wish there was a simple way to copy all projects in all sub-directories.

backend/api/Startup.cs
asanchezr: Please change to `"/api/swagger/{documentname}/swagger.json"`

We don't have a reverse proxy rule for "api-docs" (yet) but we do proxy everything for `/api`

backend/api/Startup.cs
```diff
@@ -236,15 +237,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
...
             });
             app.UseSwaggerUI(options =>
             {
                 foreach (var description in provider.ApiVersionDescriptions)
                 {
-                    options.SwaggerEndpoint($"/swagger/{description.GroupName}/swagger.json", description.GroupName);
+                    options.SwaggerEndpoint($"/api-docs/swagger/{description.GroupName}/swagger.json", description.GroupName);
                 }
-                options.RoutePrefix = "swagger";
+                options.RoutePrefix = "api-docs";
```
asanchezr, line 248: You should try changing the route prefix to `api` - we don't have "api-docs" in our reverse proxy in DEV

As per this [article](http://eatcodelive.com/2017/05/19/change-default-swagger-route-in-an-asp-net-core-web-api/), the changes should be:

**app.UseSwagger() changes:**

```csharp
 app.UseSwagger(c =>
 {
      c.RouteTemplate = "api/swagger/{documentName}/swagger.json";
 });
```
**app.UseSwaggerUI() changes:**

```csharp
 app.UseSwaggerUI(c =>
 {
      c.SwaggerEndpoint($"/api/swagger/{description.GroupName}/swagger.json", description.GroupName);
      c.RoutePrefix = "api/swagger";
 });
```

backend/Dockerfile
asanchezr: One suggestion - perhaps we should create an additional solution `*.sln` for the testing projects? That way we would be able to clean up the build solution and wouldn't need to be breaking docker builds everytime we refactor the tests. Thoughts?

Fosol: Without them being part of the solution they wouldn't be visible to the current solution.

backend/api/Startup.cs
Fosol: `/api/swagger/...` doesn't work.  I've tested it already.  I think I'm going to instead create config vars for all these settings.

asanchezr: Sure. Let me know when you find something that fixes it and I will re review

backend/Dockerfile
asanchezr: That was the idea. Having one solution for code and one solution for tests. Up to you what you prefer. Just remember to run it I. Docker after every refactoring of tests

backend/api/Startup.cs
```diff
@@ -277,13 +278,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVers
...
                     setup =>
                     {
-                        setup.UIPath = "/healthchecks-ui"; // this is ui path in your browser
-                        // setup.ApiPath = "/api";
-                        setup.ResourcesPath = "/ui/resources";
-                        setup.WebhookPath = "/hooks";
-                        setup.UseRelativeResourcesPath = false;
-                        setup.UseRelativeApiPath = false;
-                        setup.UseRelativeWebhookPath = false;
+                        setup.UIPath = this.Configuration.GetValue<string>("HealthChecksUI:UIPath"); // this is ui path in your browser
```
Fosol, line 281: I suspect all of this is not required if it's in the configuration, but I can't find any good documentation.

backend/api/appsettings.json
```diff
@@ -1,67 +1,81 @@
...
-      "Secret": "[USE SECRETS]",
-      "Client": "pims-service-account"
+    "BaseUrl": "/api",
+    "HealthChecks": {
+        "Port": "8080",
+        "LivePath": "/api/health/live",
+        "ReadyPath": "/api/health/ready"
+    },
+    "HealthChecksUI": {
+        "UIPath": "/api-health/healthchecks-ui",
```
Fosol, line 9: I'd like it to be `/api-health` and create a route for this in OpenShift.  That way we can extract the UI into its own container.

asanchezr: Please note that `https://pims-dev.pathfinder.gov.bc.ca/api-health/*` does not exist (currently). 
Our reverse proxy only allows root (`"/"`) and `/api`. Maybe that is why the healthcheck UI is failing to load...

Fosol: I have added `/api-health` to DEV for testing.

asanchezr: On your idea if running the healthcheck UI from a separate container - I like it but maybe it will require some research. Currently that UI is being served by the ASP.NET Core web server (Kestrel). I don't know if the creators of the health-check library made it possible to extract the UI.

asanchezr: In the end, having just the health-check library (without UI) seems enough to me. The UI looks nice but always felt like an add-on/nice to have. Also, there are other (free) options out there that are popular at the Lab; e.g. [Uptime Robot](https://uptimerobot.com/)

PR #115, opened 2020-04-07
PIMS-1214, PIMS-1154, add/update parcel form logic

devinleighsmith:
Also includes:
-Test fixes.
-Error Boundary.


backend/api/Controllers/ParcelController.cs
```diff
@@ -129,12 +129,12 @@ public IActionResult AddParcel([FromBody] Model.ParcelModel model)
...
         /// <param name="model"></param>
         /// <returns></returns>
-        [HttpPut("{id}")]
+        [HttpPut()]
         [HasPermission(Permissions.PropertyEdit)]
         [Produces("application/json")]
         [ProducesResponseType(typeof(Model.ParcelModel), 200)]
         [SwaggerOperation(Tags = new[] { "parcel" })]
-        public IActionResult UpdateParcel(int id, [FromBody] Model.ParcelModel model)
+        public IActionResult UpdateParcel([FromBody] Model.ParcelModel model)
```
Fosol, line 137: I changed this to standardize our routes `/parcels/{id}`.  This was based on our discussion.  I know it isn't technically required in-terms of implementation, but I do agree it would best to keep them standardized.

backend/api/Profiles/Parcel/ParcelEvaluationProfile.cs
```diff
@@ -19,7 +19,7 @@ public ParcelEvaluationProfile()
                 .IncludeBase<Entity.BaseEntity, Pims.Api.Models.BaseModel>();
 
             CreateMap<Model.ParcelEvaluationModel, Entity.ParcelEvaluation>()
-                .ForMember(dest => dest.ParcelId, opt => opt.Ignore())
+                .ForMember(dest => dest.ParcelId, opt => opt.MapFrom(src => src.PropertyId))
```
Fosol, line 22: Odd to keep the property name as `PropertyId` for the `Parcel`?

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -226,8 +226,6 @@ public Parcel Update(Parcel parcel)
             // Do not allow switching agencies through this method.
             if (existingParcel.AgencyId != parcel.AgencyId) throw new NotAuthorizedException("Parcel cannot be transferred to the specified agency.");
 
-            this.Context.Parcels.ThrowIfNotUnique(parcel);
```
Fosol, line 229: Any reason to remove this?  The purpose is to handle the fact our unique identification relies on two fields.

backend/api/Controllers/ParcelController.cs
```diff
@@ -129,12 +129,12 @@ public IActionResult AddParcel([FromBody] Model.ParcelModel model)
         /// <param name="id"></param>
         /// <param name="model"></param>
         /// <returns></returns>
-        [HttpPut("{id}")]
+        [HttpPut()]
```
asanchezr, line 132: Question: is there an specific reason why the `id` param was removed here? I know it is also supplied as part of the POST body (and might seem counter-intuitive) but that's how Rest UPDATE operations are modelled:

`PUT /api/whatever-entities/5` vs `PUT /api/whatever-entities/`

frontend/src/actions/parcelsActions.tsx
```diff
@@ -22,48 +22,74 @@ export const storeParcelsAction = (parcelList: IProperty[]): IStoreParcelsAction
...
+  buildingConstructionTypeId: number | any;
+  buildingPredominateUse: string | undefined;
+  buildingPredominateUseId: string | any;
+  buildingOccupantType: string | undefined;
+  buildingOccupantTypeId: number | any;
+  leaseExpiry?: string;
+  occupantName: string;
+  transferLeaseOnSale: boolean;
   buildingTenancy: string;
+  buildingNetBookValue: number;
```
Fosol, line 51: Odd you have this value but none of the others related to evaluation?

frontend/src/mocks/filterDataMock.ts
```diff
@@ -93,6 +93,13 @@ export const mockDetails = [
   {
```
Fosol, line 93: Is this being used by our unit tests?

frontend/src/pages/EditUserPage.tsx
```diff
@@ -88,9 +88,9 @@ const EditUserPage = (props: IUserDetailParams) => {
 
   checkRoles =
     user.roles.length > 0 ? (
-      <Select field="role" placeholder={user.roles.find(x => x).name} options={[]} />
+      <Select field="role" placeholder={user.roles[0].name} options={[]} />
```
Fosol, line 91: Did you want this hardcoded to only one role?

frontend/src/mocks/filterDataMock.ts
jdtoombs: Some of my older tests use this

frontend/src/pages/EditUserPage.tsx
jdtoombs: For now since we only allow the user to have one role or agency (on the FE) I had set it up to display one role. I originally had it so it would just push the new role/agency onto the existing ones but since the front end only supports one at a time I changed it

frontend/src/actionCreators/parcelsActionCreator.tsx
```diff
@@ -64,18 +65,34 @@ export const fetchPropertyDetail = (id: number, propertyTypeId: 0 | 1) => (dispa
...
+      dispatch(fetchParcelDetail(response.data));
       dispatch(hideLoading());
     })
     .catch((axiosError: AxiosError) =>
       dispatch(error(actionTypes.ADD_PARCEL, axiosError?.response?.status, axiosError)),
     )
     .finally(() => dispatch(hideLoading()));
 };
+
+export const updateParcel = (parcel: IParcel) => (dispatch: Function) => {
```
asanchezr, line 84: If I followed this code correctly we now have two, different, `IParcel` interfaces (one for create, one for update)...
Any chance you could try to find a way to rename one of them to avoid confusing? 
I wish I could provide a suggestion but I'm drawing a blank here...
> _There are only two hard things in Computer Science: cache invalidation and naming things._

frontend/src/actions/parcelsActions.tsx
```diff
@@ -22,48 +22,74 @@ export const storeParcelsAction = (parcelList: IProperty[]): IStoreParcelsAction
...
   description: string;
   address: IAddress;
-  latitude: number;
-  longitude: number;
-  localId: string;
-  constructionType: string;
-  postal: string;
-  buildingFloorCount: number;
-  buildingPredominateUse: string;
+  latitude: number | any;
```
asanchezr, line 38: I see a few of these `number | any` typings. Why use any here? Are we trying to express a nullable number here? If not, why not just using `any`?

frontend/src/actions/parcelsActions.tsx
```diff
@@ -22,48 +22,74 @@ export const storeParcelsAction = (parcelList: IProperty[]): IStoreParcelsAction
...
   pid: string;
-  latitude: number;
-  longitude: number;
-  propertyStatus: string;
-  classification: string;
+  pin: string;
+  latitude: number | any;
+  longitude: number | any;
+  statusId: number | any;
+  propertyStatus: string | undefined;
```
asanchezr, line 72: `propertyStatus?: string` denotes an optional property and I think conveys the intent a bit better than `string | undefined` (which is the same)

frontend/src/components/common/ErrorModal.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react';
+import { Modal, Row, Col, Button, Container } from 'react-bootstrap';
+
+import { IGenericNetworkAction } from 'actions/genericActions';
+import { useHistory } from 'react-router-dom';
+
+const ErrorModal = (props: any) => {
```
asanchezr, line 7: I like this! Wrapping all of our components into some sort of "error message" middleware/container.
However, the concept of "error boundaries" is kind of new in React. Could you please provide some JSDoc to help other devs understand the need for this? Thanks!

frontend/src/reducers/LeafletMouseSlice.tsx
```diff
@@ -0,0 +1,22 @@
+import { LeafletMouseEvent } from 'leaflet';
+import { createSlice, createAction } from '@reduxjs/toolkit';
+
+export const saveClickLatLng = createAction<LeafletMouseEvent>('saveClickLatLng');
+export const clearClickLatLng = createAction('clearLatLng');
+const leafletMouseSlice = createSlice({
```
asanchezr, line 6: Looking forward to hear your experience with redux-toolkit...
Could you please add some JSDoc here as it is a different approach to app state management? Thanks!

backend/api/Controllers/ParcelController.cs
devinleighsmith: Took it out because it was being ignored by the service, ie. PUT/5 with a body of { id: 10 } would target the entity with id of 10.

backend/api/Profiles/Parcel/ParcelEvaluationProfile.cs
devinleighsmith: Yeah, the FE treats building evaluations and parcel evaluations as the same object, which uses property id instead of parcel/building id. The actual type is inferred based on the object structure.

backend/dal/Services/Concrete/ParcelService.cs
devinleighsmith: Yep, unique isn't possible when doing updates.

frontend/src/actions/parcelsActions.tsx
devinleighsmith: Hmm, I thought this came from the merge, I'll double check and make sure its being used.

frontend/src/actions/parcelsActions.tsx
devinleighsmith: Yeah, this is functionally equivalent to any, but I have a slight preference for this because it shows the intended type of the field even if I'm required to make this an any to get the desired behaviour out of formik.

backend/dal/Services/Concrete/ParcelService.cs
Fosol: We'll need to fix this to work then.  Otherwise they could update the PID or PIN to a value that already exists.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -20,8 +20,22 @@ public static class ParcelExtensions
         /// <exception type="DbUpdateException">The PID and PIN must be unique.</exception>
         public static void ThrowIfNotUnique(this DbSet<Entity.Parcel> parcels, Entity.Parcel parcel)
         {
-            var alreadyExists = parcels.Any(p => p.Id != parcel.Id && (parcel.PID > 0 && p.PID == parcel.PID) || (parcel.PIN != null && p.PIN == parcel.PIN));
+            var alreadyExists = parcels.Any(p => p.Id != parcel.Id && ((parcel.PID > 0 && p.PID == parcel.PID) || (parcel.PIN != null && p.PIN == parcel.PIN)));
```
devinleighsmith, line 23: This fixes a bug, previously the last || would be evaluated independently

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -20,8 +20,22 @@ public static class ParcelExtensions
...
         }
+
+        /// <summary>
+        /// Make a query to determine if the parcel PID and PIN have been updated on a pre-existing parcel
+        /// - once created the unique PID or PIN of a parcel should not be updated.
+        /// </summary>
+        /// <param name="parcels"></param>
+        /// <param name="parcel"></param>
+        /// <exception type="DbUpdateException">The PID and PIN may not be updated.</exception>
+        public static void ThrowIfPidPinUpdated(this DbSet<Entity.Parcel> parcels, Entity.Parcel parcel)
```
devinleighsmith, line 34: Made this into a separate function for performance reasons.

frontend/src/actionCreators/parcelsActionCreator.tsx
```diff
@@ -12,19 +12,24 @@ import { createRequestHeader } from 'utils/RequestHeaders';
...
-    .then((response: AxiosResponse) => {
-      dispatch(success(actionTypes.GET_PARCELS));
-      dispatch(parcelsActions.storeParcelsAction(response.data));
-      dispatch(hideLoading());
-    })
-    .catch((axiosError: AxiosError) =>
-      dispatch(error(actionTypes.GET_PARCELS, axiosError?.response?.status, axiosError)),
-    )
-    .finally(() => dispatch(hideLoading()));
+  if (
```
devinleighsmith, line 15: we shouldn't be making parcels requests for single points.

devinleighsmith:
@Fosol need you to clear your requested changes before merge.

PR #117, opened 2020-04-08
PIMS-1507 Adding Parcel.Municipality filter

Fosol:
Adding Parcel.Municipality filter
Updating DB Migration (Requires DB Refresh).


backend/dal/Configuration/ParcelConfiguration.cs
```diff
@@ -33,7 +33,7 @@ public override void Configure(EntityTypeBuilder<Parcel> builder)
             builder.HasOne(m => m.Agency).WithMany(m => m.Parcels).HasForeignKey(m => m.AgencyId).OnDelete(DeleteBehavior.ClientSetNull);
 
             builder.HasIndex(m => new { m.PID, m.PIN }).IsUnique(); // This will allow for Crown Land to set ParcelId=0 and PIN=#######.
-            builder.HasIndex(m => new { m.Latitude, m.Longitude, m.StatusId, m.IsSensitive, m.AgencyId, m.ClassificationId, m.LandArea });
+            builder.HasIndex(m => new { m.Latitude, m.Longitude, m.StatusId, m.IsSensitive, m.AgencyId, m.ClassificationId, m.LandArea, m.Municipality, m.Zoning, m.ZoningPotential, m.Description });
```
Fosol, line 36: Updating index to improve searching.

backend/dal/Services/Admin/Concrete/ParcelService.cs
```diff
@@ -70,6 +69,18 @@ public Paged<Parcel> Get(int page, int quantity, ParcelFilter filter = null)
             if (!String.IsNullOrWhiteSpace(filter.Address))
                 query = query.Where(p => EF.Functions.Like(p.Address.Address1, $"%{filter.Address}%") || EF.Functions.Like(p.Address.City.Name, $"%{filter.Address}%"));
 
+            if (!String.IsNullOrWhiteSpace(filter.Description))
```
Fosol, line 72: We can now search for `Description, Municipality, Zoning, ZoningPotential`

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -92,6 +92,18 @@ public IEnumerable<Parcel> Get(ParcelFilter filter)
                 query = query.Where(p => agencies.Contains(p.AgencyId));
             }
 
+            if (!String.IsNullOrWhiteSpace(filter.Description))
```
Fosol, line 95: We can now search for Description, Municipality, Zoning, ZoningPotential

backend/entities/Models/Paged`.cs
```diff
@@ -7,7 +8,7 @@ namespace Pims.Dal.Entities.Models
     /// Paged generic class, provides a structure to hold a single page of items.
     /// </summary>
     /// <typeparam name="TModel"></typeparam>
-    public class Paged<TModel>
+    public class Paged<TModel> : ICollection<TModel>
```
Fosol, line 11: We can now enumerate through a page.

backend/tests/unit/api/Controllers/Admin/ParcelControllerTest.cs
```diff
@@ -31,64 +53,9 @@ public ParcelControllerTest()
 
         #region Tests
         #region GetParcels
-
-        [Fact]
-        public void GetParcels_FilterLongitude_Success()
```
Fosol, line 36: Moved these all into a `Theory`

backend/tests/unit/dal/Services/ParcelServiceTest.cs
```diff
@@ -129,6 +219,16 @@ public void Get()
             Assert.NotNull(result);
             Assert.Equal(EntityState.Detached, context.Entry(result).State);
             Assert.Equal(parcel, result, new ShallowPropertyCompare());
+            Assert.NotNull(parcel.Address);
```
Fosol, line 222: Started confirming the query result to ensure `Include(...)` was implemented.  Not 100% sure InMemoryDatabase will do this correctly.

docs/README.md
```diff
@@ -8,21 +8,21 @@ The PIMS technical architecture stack continues to grow as features and enhancem
...
-| <img src="./images/logo-react.svg" width="25">      | Front-end application | app  | React/Node      | 16.12.0/10 | UI for PIMS                |
-| <img src="./images/logo-dotnetcore.svg" width="25"> | Back-end interface    | api  | .NET Core       | 3.1        | API for data access        |
-| <img src="./images/logo-mssql.svg" width="25">      | Backend database      |  db  | MSSQL           | 2019       | Inventory datasource       |
-| <img src="./images/logo-KeyCloak.png" width="25">   | Authentication        | auth | KeyCloak        |            | Authenticate users         |
+| <img src="./images/logo-react.svg" width="25">      | Front-end application |  app  | React/Node      | 16.12.0/10 | UI for PIMS                |
+| <img src="./images/logo-dotnetcore.svg" width="25"> | Back-end interface    |  api  | .NET Core       | 3.1        | API for data access        |
+| <img src="./images/logo-mssql.svg" width="25">      | Backend database      |  db   | MSSQL           | 2019       | Inventory datasource       |
+| <img src="./images/logo-keycloak.png" width="25">   | Authentication        | auth  | KeyCloak        |            | Authenticate users         |
```
Fosol, line 17: Fixing documentation finally.

backend/dal/Services/Admin/Concrete/ParcelService.cs
asanchezr: The following do "start-with" searches. Wouldn't it make more sense to switch to a "contains" search - especially for the `Description` field. I think all should be switched, thoughts?

backend/dal/Services/Concrete/ParcelService.cs
asanchezr: Same comment as above.

backend/entities/Models/ParcelFilter.cs
```diff
@@ -11,6 +11,16 @@ namespace Pims.Dal.Entities.Models
...
+        /// <summary>
+        /// get/set - The page number.
+        /// </summary>
+        public int Page { get; set; } = 1;
+
+        /// <summary>
+        /// get/set - The quantity of parcels to return in a single request.
+        /// </summary>
+        public int Quantity { get; set; } = 10;
+
```
asanchezr, line 23: Ideally I'd like to see this in a base class or interface or there will be lot's of copy paste (and associated manual errors) if/when we start supporting paginating other resources/entities in the system - `api/users` comes to mind (not 100% sure what the name is for the users endpoint but we will need to paginate it at some point). 
I'm sure we will run into other large lists of entities as PIMS grows - thoughts?

backend/dal/Services/Admin/Concrete/ParcelService.cs
devinleighsmith: As a user I generally expect searches to use contains instead of startswith.

backend/entities/Models/ParcelFilter.cs
Fosol: The next PR I'm working on does this.

backend/dal/Services/Admin/Concrete/ParcelService.cs
Fosol: I had chosen `starts-with` for the performance advantage and thought it actually made more sense for most of these filters (Municipality, Zoning, ZoningPotential), but it should be a `contains` for Description.  I'll probably just make them all `contains` for now for simplicity.

PR #120, opened 2020-04-09
Pims 1384 - autosave

devinleighsmith:
Add formik persist with basic encryption.
Move jwt token storage from localstorage to redux.

### DIFF
--- a/frontend/src/actionCreators/parcelsActionCreator.tsx
+++ b/frontend/src/actionCreators/parcelsActionCreator.tsx
@@ -13,8 +13,9 @@ export const fetchParcels = (parcelBounds: API.IParcelListParams | null) => (
   dispatch: Function,
 ) => {
   if (
-    parcelBounds?.neLatitude != parcelBounds?.swLatitude &&
-    parcelBounds.neLongitude != parcelBounds.swLongitude
+    !parcelBounds ||
+    (parcelBounds?.neLatitude != parcelBounds?.swLatitude &&
+      parcelBounds.neLongitude != parcelBounds.swLongitude)
   ) {
     dispatch(request(actionTypes.GET_PARCELS));
     dispatch(showLoading());
@@ -80,9 +81,10 @@ export const createParcel = (parcel: IParcel) => (dispatch: Function) => {
       dispatch(fetchParcelDetail(response.data));
       dispatch(hideLoading());
     })
-    .catch((axiosError: AxiosError) =>
-      dispatch(error(actionTypes.ADD_PARCEL, axiosError?.response?.status, axiosError)),
-    )
+    .catch((axiosError: AxiosError) => {
+      dispatch(error(actionTypes.ADD_PARCEL, axiosError?.response?.status, axiosError));
+      throw axiosError;
+    })
     .finally(() => dispatch(hideLoading()));
 };
 
@@ -96,8 +98,9 @@ export const updateParcel = (parcel: IParcel) => (dispatch: Function) => {
       dispatch(fetchParcelDetail(response.data));
       dispatch(hideLoading());
     })
-    .catch((axiosError: AxiosError) =>
-      dispatch(error(actionTypes.UPDATE_PARCEL, axiosError?.response?.status, axiosError)),
-    )
+    .catch((axiosError: AxiosError) => {
+      dispatch(error(actionTypes.UPDATE_PARCEL, axiosError?.response?.status, axiosError));
+      throw axiosError;
+    })
     .finally(() => dispatch(hideLoading()));
 };
--- a/frontend/src/components/common/FormikPersist.tsx
+++ b/frontend/src/components/common/FormikPersist.tsx
@@ -1,0 +1,64 @@
+import * as React from 'react';
+import { FormikProps, connect } from 'formik';
+import debounce from 'lodash/debounce';
+import isEqual from 'react-fast-compare';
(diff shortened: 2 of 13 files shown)


frontend/src/actionCreators/parcelsActionCreator.tsx
```diff
@@ -80,9 +81,10 @@ export const createParcel = (parcel: IParcel) => (dispatch: Function) => {
       dispatch(fetchParcelDetail(response.data));
       dispatch(hideLoading());
     })
-    .catch((axiosError: AxiosError) =>
```
devinleighsmith, line 83: propagate any exceptions.

frontend/src/components/common/FormikPersist.tsx
```diff
@@ -0,0 +1,64 @@
...
+    debounce: 300,
+  };
+
+  //Encrypt the formdata before saving.
+  //TODO: in the event the security of this data becomes a concern, this data should be persisted server-side.
+  saveForm = debounce((data: FormikProps<{}>) => {
+    if (this.props.isSessionStorage) {
+      window.sessionStorage.setItem(
+        this.props.name,
+        AES.encrypt(JSON.stringify(data), this.props.secret).toString(),
```
devinleighsmith, line 27: This level of security obtainable here is entirely based on the strength of the key. For now I'm just using the JWT subject to encode this data. We could replace this with something more secure in the future if the security of this formik data becomes a concern.

Also, it would be easy to wipe all saved data in the event that the isSensitive field is ticked, which would be the easiest option.

frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -19,8 +19,10 @@ import BuildingForm from './subforms/BuildingForm';
 import EvaluationForm from './subforms/EvaluationForm';
 import LandForm from './subforms/LandForm';
 import PidPinForm from './subforms/PidPinForm';
+import { render, getByText, fireEvent, act, wait } from '@testing-library/react';
 
 Enzyme.configure({ adapter: new Adapter() });
+jest.mock('lodash/debounce', () => jest.fn(fn => fn));
```
devinleighsmith, line 25: even with fake timers jest can't figure out how to trigger lodash debounce properly.

frontend/src/forms/ParcelDetailForm.tsx
```diff
@@ -16,22 +16,22 @@ import { Form } from 'components/common/form';
...
-  agencyId: number | undefined;
+  secret: string;
+  agencyId?: number;
   disabled?: boolean;
   updateLatLng: Function;
 }
 const ParcelDetailForm = (props: ParcelPropertyProps) => {
   const dispatch = useDispatch();
   const history = useHistory();
-  if (!props.agencyId) {
```
devinleighsmith, line 31: this is handled in the parent.

frontend/src/forms/ParcelDetailForm.tsx
```diff
@@ -93,18 +94,36 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
             initialValues={initialValues}
             validationSchema={ParcelSchema}
             enableReinitialize
-            onSubmit={(values, { setSubmitting }) => {
+            onSubmit={(values, { setSubmitting, resetForm, setStatus, setErrors }) => {
+              let response: any;
               if (!values.id) {
-                dispatch(createParcel(values));
+                response = dispatch(createParcel(values));
```
devinleighsmith, line 100: handling the response here instead of using the network reducer because I want all of the formik data in context.

frontend/src/pages/SubmitProperty.tsx
```diff
@@ -32,16 +32,12 @@ const SubmitProperty = (props: any) => {
     state => state.parcel.parcelDetail as IParcelDetail,
   );
   const properties = useSelector<RootState, IProperty[]>(state => state.parcel.parcels);
-  const addParcelRequest = useSelector<RootState, IGenericNetworkAction>(
```
devinleighsmith, line 35: this logic and the removed logic below were moved to the form where formik values are still in context.

frontend/src/utils/KeycloakEventHandler.tsx
```diff
@@ -1,14 +1,16 @@
 import { KeycloakEventHandler } from '@react-keycloak/web';
 import { KeycloakInstance } from 'keycloak-js';
+import { store } from 'App';
+import { saveJwt, clearJwt } from 'reducers/JwtSlice';
 
 const getKeycloakEventHandler = (keycloak: KeycloakInstance) => {
   const keycloakEventHandler: KeycloakEventHandler = (event, error) => {
     if (event === 'onAuthSuccess') {
-      localStorage.setItem('jwt', keycloak.token!!);
```
devinleighsmith, line 7: based on some reading I was doing, it seems that local storage is the least secure of three JWT storage options( localstorage, in-memory, cookie). changing this to be in-memory doesn't change the current behavior of the app in my testing but makes XSS attacks harder.

ref: https://hasura.io/blog/best-practices-of-using-jwt-with-graphql/

frontend/src/components/common/FormikPersist.tsx
Fosol: This is probably overkill as all this is client-side, but I love it anyway lol.  Cool feature.

PR #121, opened 2020-04-10
Improve PIMS context factory and database documentation

asanchezr:


backend/dal/PIMSContextFactory.cs
```diff
@@ -45,27 +45,44 @@ public PimsContextFactory()
...
                 .SetBasePath(Directory.GetCurrentDirectory())
-                .AddEnvironmentVariables()
-                .AddJsonFile("connectionstrings.json", optional : true, reloadOnChange : true)
-                .AddJsonFile($"connectionstrings.{environment}.json", optional : true, reloadOnChange : true)
-                .AddUserSecrets<PimsContext>()
-                .Build();
+                .AddJsonFile("connectionstrings.json", optional: true, reloadOnChange: true)
+                .AddJsonFile($"connectionstrings.{environment}.json", optional: true, reloadOnChange: true);
+
+            if (!this.isProduction(environment))
```
Fosol, line 62: This isn’t absolutely required as the way secrets works wouldn’t even be relevant in the production environment.  But it doesn’t hurt to have the check.  However there is a built in extension method for the environment name that should be used rather than a custom one.

backend/dal/PIMSContextFactory.cs
```diff
@@ -45,27 +45,44 @@ public PimsContextFactory()
...
             };
 
             var optionsBuilder = new DbContextOptionsBuilder<PimsContext>();
-            // optionsBuilder.UseNpgsql (cs, opts => opts.CommandTimeout ((int) TimeSpan.FromMinutes (10).TotalSeconds));
-            optionsBuilder.UseSqlServer(builder.ConnectionString, opts => opts.CommandTimeout((int) TimeSpan.FromMinutes(10).TotalSeconds));
+            optionsBuilder.UseSqlServer(sqlBuilder.ConnectionString, opts => opts.CommandTimeout((int)TimeSpan.FromMinutes(10).TotalSeconds));
             return new PimsContext(optionsBuilder.Options);
         }
+
+        private bool isProduction(string env)
```
Fosol, line 82: Look for the existing extension method rather than creating a custom one.  I believe there is one already.

backend/docs/DATABASE.md
```diff
@@ -17,25 +17,30 @@ Refer to the CLI documentation [here](https://docs.microsoft.com/en-us/ef/core/m
 
 ## Database Migration Management
 
-The DB is setup and configured through Entity Framework Code-First processes. All dotnet ef commands must be run from the /dal directory.
+The database is setup and configured through Entity Framework Code-First processes. To use the Entity Framework CLI you will need to:
 
-To use the Entity Framework CLI you will need to install the **.NET SDK version**, **dotnet-ef tool** and add a `connectionstrings.json` configuration file to /dal. This connection string can also be included in a .env file in the /dal folder.
+- Install the **.NET SDK version** (download links below), 
+- Install **dotnet-ef tool** and,
+- Add a `connectionstrings.json` configuration file in the `/backend/dal` folder
```
Fosol, line 24: The `connectionstrings.json` file is optional.  I’d reword to highlight this with “Optionally add”.

backend/docs/DATABASE.md
```diff
@@ -45,12 +50,14 @@ You may create a `connectionstrings.json` configuration file within the `/backen
 }
 ```
 
-The default `port` for MSSQL is 1433, but set it to the same value used in the `docker-compose.yaml` configuration file.
+The default `port` for MS-SQL is 1433, but set it to the same value used in the `docker-compose.yaml` configuration file.
 The `database name` should be the same value using the in the database `.env` file.
```
Fosol, line 54: “should be the same value used in the database...”

backend/docs/DATABASE.md
asanchezr: Thanks for the suggestion, I can change it to "optionally" but first wanted to double check the use case with you:
- These instructions are for someone coming to the `/backend/dal/` folder to run migrations on their local (container) DB.

So in the above case - according to what I've read - when you run `dotnet ef database update`, the SDK will look for the `PimsContextFactory` that we created. Given that the `DAL` project is a class library it **doesn't** have an `appsettings.json` with a default connection string defined...

So, my question - if there's no `appsettings.json` and (as per instructions if we changed them above) you don't add either a `connectionstring.json` or `.env` file - where would `dotnet ef ...` find the connection information? 

How would `dotnet` know to go up one folder and then into `/api` to look for connection information?

backend/dal/PIMSContextFactory.cs
asanchezr: I'm okay removing the check for "not Production" - just not sure whether a miss-configured app would be able to read from the secret json file in production.
All code I have seen always "wraps" the `UseSecrets` call with an `if (env.isDevelopment)` so it makes me wonder if secrets **don't work** in production why wrap them in an `if`?

Could it be that secret manager always writes to the json file and it's up to the developer to not use it in production? - or is it more like MSFT put in "smarts" into the secret manager so it returns `null` when in production? I really don't know what the guidance is...

All I see in the docs is "use Secret Manager in development and Azure Vault in Production". 

Again, I'm fine removing the check

backend/dal/PIMSContextFactory.cs
asanchezr: I did look up the extension method you mention but this Factory class is not part of a web application - instead is being called by the `dotnet ef` cli. 

I tried to find a way to get a reference to the `IWebHostEnvironment` env which is the one that supports the extension method, and could not find a way to inject it into the class given that there's no host created... I could be wrong here, if you can find a way to inject the `env` class then we can drop the custom method. Please let me know if you know a way.

backend/dal/PIMSContextFactory.cs
Fosol: Nah leave it in.  It just wouldn't be relevant in a production environment due to all the other required elements (file/configuration).  No different than when the other optional files don't exist.

backend/dal/PIMSContextFactory.cs
Fosol: I looks like the extension method only exists for `HostingEnvironment`, which would require additional effort to setup and configure.  I'd recommend moving your function to an extension method in our `Pims.Core` library.  Also rename with PascalCase `IsProduction()`.

backend/docs/DATABASE.md
Fosol: They would have to select one of the configuration options, but no single one is required.
- appsettings.config
- connectionstrings.config
- .env
- environment variables

asanchezr: Good, I've modified the docs to account for your comments

backend/dal/PIMSContextFactory.cs
asanchezr: I've made the changes, thanks for your suggestions!

PR #123, opened 2020-04-11
Replacing Automapper with Mapster!

Fosol:
The amount of effort to configure and debug Automapper was disproportional with what we are using it for.  I have converted our code base over to Mapster, which is apparently faster and fare more intuitive.  I'll never go back to Automapper.

Due to the size and nature of this PR I suspect I will introduce bugs, but after fixing various issues, all unit-tests pass.  Note that this required fixing various threading issues related to testing and the in-memory database.

Let me know if you find any issues and I'll resolve them quickly.


backend/api/Areas/Admin/Mapping/Parcel/AddressMap.cs
```diff
@@ -0,0 +1,38 @@
+using Mapster;
```
Fosol, line 1: Instead of `Profile` we now use `IRegister` to identify classes that will provide the mapping configuration.

backend/api/Areas/Admin/Mapping/Parcel/AddressMap.cs
```diff
@@ -0,0 +1,38 @@
...
+using Entity = Pims.Dal.Entities;
+
+namespace Pims.Api.Areas.Admin.Mapping.Parcel
+{
+    public class AddressMap : IRegister
+    {
+        public void Register(TypeAdapterConfig config)
+        {
+            config.NewConfig<Entity.Address, Model.AddressModel>()
+                .IgnoreNonMapped(true)
```
Fosol, line 13: I couldn't find a way to globally set this for Dependency Injection regrettably.  So every new configuration map should ignore non-mapped properties to ensure we don't introduce change between versions.

Fosol: Amazingly and intuitively Mapster would map nearly every property without individual configuration, but I think it makes more sense for maintenance over-time to individually configure them.

backend/api/Areas/Admin/Models/Parcel/ParcelModel.cs
```diff
@@ -10,7 +10,7 @@ public class ParcelModel : Model.BaseModel
 
         public string PID { get; set; }
 
-        public string PIN { get; set; }
+        public int? PIN { get; set; }
```
Fosol, line 13: Fixing possible future bugs here so that the type is correct.  I'm not aware of the PIN formatting, so we may need to make it a string in the database.

backend/api/Areas/Tools/Models/Import/BuildingEvaluationModel.cs
```diff
@@ -3,7 +3,7 @@ namespace Pims.Api.Areas.Tools.Models.Import
     public class BuildingEvaluationModel : Pims.Api.Models.BaseModel
     {
         #region Properties
-        public int PropertyId { get; set; }
+        public int BuildingId { get; set; }
```
Fosol, line 6: Note I made these match the database column names.  I don't know if this will breach @devinleighsmith your Property Detail View Page.  Additionally most of this is going to be updated/changed this sprint because we are changing how we implemented evaluations.

backend/api/Areas/Tools/Models/Import/ParcelEvaluationModel.cs
```diff
@@ -3,7 +3,7 @@ namespace Pims.Api.Areas.Tools.Models.Import
     public class ParcelEvaluationModel : Pims.Api.Models.BaseModel
     {
         #region Properties
-        public int PropertyId { get; set; }
+        public int ParcelId { get; set; }
```
Fosol, line 6: Note I made these match the database column names.  I don't know if this will breach @devinleighsmith your Property Detail View Page.  Additionally most of this is going to be updated/changed this sprint because we are changing how we implemented evaluations.

backend/api/Controllers/LookupController.cs
```diff
@@ -90,29 +91,31 @@ public IActionResult GetPropertyClassifications()
         /// <returns></returns>
         [HttpGet("all")]
         [Produces("application/json")]
-        [ProducesResponseType(typeof(IEnumerable<Model.CodeModel>), 200)]
+        [ProducesResponseType(typeof(IEnumerable<Model.CodeModel<object>>), 200)]
```
Fosol, line 94: Quite a few changes here to support generics.  I'm not sure if this will cause issues in the frontend.  Primarily it means that the `id` property will be of different types based on the object.  I may need to convert it back to a string through the mapping if it causes issues.

backend/api/Controllers/ParcelController.cs
```diff
@@ -134,6 +134,7 @@ public IActionResult AddParcel([FromBody] Model.ParcelModel model)
         [Produces("application/json")]
         [ProducesResponseType(typeof(Model.ParcelModel), 200)]
         [SwaggerOperation(Tags = new[] { "parcel" })]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "To support standardized routes (/update/{id})")]
         public IActionResult UpdateParcel(int id, [FromBody] Model.ParcelModel model)
```
Fosol, line 138: Keeping the `id` for the route path throws a warning because we don't presently use it.  I've suppressed the warning.

backend/api/Controllers/ParcelController.cs
```diff
@@ -155,6 +156,7 @@ public IActionResult UpdateParcel(int id, [FromBody] Model.ParcelModel model)
         [Produces("application/json")]
         [ProducesResponseType(typeof(Model.ParcelModel), 200)]
         [SwaggerOperation(Tags = new[] { "parcel" })]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "To support standardized routes (/delete/{id})")]
```
Fosol, line 159: Keeping the `id` for the route path throws a warning because we don't presently use it.  I've suppressed the warning.

backend/api/Helpers/Mapping/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using MapsterMapper;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Reflection;
+
+namespace Pims.Api.Helpers.Mapping
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddMapster(this IServiceCollection services, Action<TypeAdapterConfig> options = null)
```
Fosol, line 11: Makes the `Startup` class code cleaner.  This allows us to use DI for the `IMapper` object.

backend/api/Models/Building/BuildingEvaluationModel.cs
```diff
@@ -3,7 +3,7 @@ namespace Pims.Api.Models.Building
     public class BuildingEvaluationModel : BaseModel
     {
         #region Properties
-        public int PropertyId { get; set; }
+        public int BuildingId { get; set; }
```
Fosol, line 6: Note I made these match the database column names. I don't know if this will breach @devinleighsmith your Property Detail View Page. Additionally most of this is going to be updated/changed this sprint because we are changing how we implemented evaluations.

backend/api/Models/CodeModel.cs
```diff
@@ -1,46 +1,17 @@
 namespace Pims.Api.Models
 {
     /// <summary>
-    /// CodeModel class, provides a model that represents a code lookup item.
+    /// CodeModel class, provides a model that represents a code code item.
     /// </summary>
-    public class CodeModel : BaseModel
+    /// <typeparam name="TKey"></typeparam>
+    public class CodeModel<TKey> : LookupModel<TKey>
```
Fosol, line 7: Note we're using generics now.

backend/api/Models/LookupModel.cs
```diff
@@ -0,0 +1,41 @@
+namespace Pims.Api.Models
+{
+    /// <summary>
+    /// LookupModel class, provides a model that represents a code lookup item.
+    /// </summary>
+    /// <typeparam name="TKey"></typeparam>
+    public class LookupModel<TKey> : BaseModel
```
Fosol, line 7: Note we're using generics now.

backend/api/Models/Parcel/BuildingEvaluationModel.cs
```diff
@@ -3,7 +3,7 @@ namespace Pims.Api.Models.Parcel
     public class BuildingEvaluationModel : BaseModel
     {
         #region Properties
-        public int PropertyId { get; set; }
+        public int BuildingId { get; set; }
```
Fosol, line 6: Note I made these match the database column names. I don't know if this will breach @devinleighsmith your Property Detail View Page. Additionally most of this is going to be updated/changed this sprint because we are changing how we implemented evaluations.

backend/api/Models/Parcel/ParcelEvaluationModel.cs
```diff
@@ -3,7 +3,7 @@ namespace Pims.Api.Models.Parcel
     public class ParcelEvaluationModel : BaseModel
     {
         #region Properties
-        public int PropertyId { get; set; }
+        public int ParcelId { get; set; }
```
Fosol, line 6: Note I made these match the database column names. I don't know if this will breach @devinleighsmith your Property Detail View Page. Additionally most of this is going to be updated/changed this sprint because we are changing how we implemented evaluations.

backend/api/Models/Parcel/ParcelModel.cs
```diff
@@ -9,7 +9,7 @@ public class ParcelModel : BaseModel
 
         public string PID { get; set; }
 
-        public string PIN { get; set; }
+        public int? PIN { get; set; }
```
Fosol, line 12: Updated to match the database column type.  Might need to change database to a string as I don't know the format used by the client.

backend/api/Pims.Api.csproj
```diff
@@ -10,14 +10,26 @@
     <NoWarn>$(NoWarn);1591</NoWarn>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Remove="Areas\Admin\Profiles\**" />
```
Fosol, line 13: Need to clean these up now that they're truly deleted.

backend/tests/core/DatabaseHelper.cs
```diff
@@ -16,54 +18,30 @@ public static class DatabaseHelper
         /// Uses an InMemoryDatabase instead of relational.
         /// </summary>
         /// <param name="helper"></param>
+        /// <param name="dbName"></param>
         /// <param name="permission"></param>
         /// <param name="ensureDeleted"></param>
         /// <returns></returns>
-        public static PimsContext CreatePimsContext(this TestHelper helper, Permissions permission, bool ensureDeleted = false)
+        public static PimsContext CreatePimsContext(this TestHelper helper, string dbName, Permissions permission, bool ensureDeleted = false)
```
Fosol, line 25: Needed to support multi-threading, which requires the database name to be unique for each test.

backend/tests/core/Pims.Core.Test.csproj
```diff
@@ -2,10 +2,14 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IsTestProject>false</IsTestProject>
+    <OutputType>Library</OutputType>
```
Fosol, line 6: Required to fix the vscode `dotnet test` bug.

backend/api/Areas/Tools/Models/Import/BuildingEvaluationModel.cs
devinleighsmith: It probably will. I'm fine with taking this and making the required adjustments on my side.

backend/api/Models/User/AccessRequestRoleModel.cs
```diff
@@ -1,11 +1,12 @@
+using System;
 using System.Collections.Generic;
 
 namespace Pims.Api.Models.User
 {
     /// <summary>
     /// AccessRequestRoleModel class, provides a model that represents a role attached to an access request.
     /// </summary>
-    public class AccessRequestRoleModel : CodeModel
+    public class AccessRequestRoleModel : CodeModel<Guid> // TODO: Why does this inherit from CodeModel?
```
devinleighsmith, line 9: I'm pretty sure it doesn't need to.

PR #124, opened 2020-04-12
PIMS-1350 Add ParcelMap boundaries to Landing Page

asanchezr:
image

### DIFF
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -177,13 +177,14 @@ const Map: React.FC<MapProps> = props => {
           closePopupOnClick={!props.disabled}
         >
           {activeBasemap && (
-            <TileLayer attribution={activeBasemap.attribution} url={activeBasemap.url} />
+            <TileLayer attribution={activeBasemap.attribution} url={activeBasemap.url} zIndex={0} />
           )}
           <WMSTileLayer
             url="https://openmaps.gov.bc.ca/geo/pub/WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW/ows?"
             layers="pub:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW"
             transparent={true}
             format="image/png"
+            zIndex={10}
           />
           {properties && properties.map(renderMarker)}
           {selectedProperty && renderPopup(selectedProperty)}
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -38,18 +38,27 @@ type MapProps = {
   onViewportChanged?: (e: MapViewportChangeEvent) => void;
   onMapClick?: (e: LeafletMouseEvent) => void;
   disableMapFilterBar?: boolean;
-  disabled?: boolean;
+  interactive?: boolean;
+  showParcelBoundaries?: boolean;
 };
 
-const Map: React.FC<MapProps> = props => {
-  // props
-  const {
-    properties,
-    selectedProperty,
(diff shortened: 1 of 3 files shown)


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -39,18 +38,27 @@ type MapProps = {
   onViewportChanged?: (e: MapViewportChangeEvent) => void;
   onMapClick?: (e: LeafletMouseEvent) => void;
   disableMapFilterBar?: boolean;
-  disabled?: boolean;
+  interactive?: boolean;
+  showParcelBoundaries?: boolean;
```
asanchezr, line 42: - Renamed `disabled` to `interactive` - and flipped the associated logic.
- Added a toggle to show/hide the boundaries as the `Map` component is now used in various places; e.g. Landing Page and Submit Property form

frontend/src/pages/SubmitProperty.tsx
```diff
@@ -92,6 +92,7 @@ const SubmitProperty = (props: any) => {
         <MapView
           disableMapFilterBar={true}
           disabled={!!props?.match?.params?.id}
+          showParcelBoundaries={false}
```
asanchezr, line 95: - Hiding the boundaries by default when submitting a property. This is subject to change after feedback from PO.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -3,9 +3,8 @@ import './Map.scss';
 import React, { useRef, useState, useEffect } from 'react';
 import axios from 'axios';
 import { LatLngBounds, LeafletMouseEvent } from 'leaflet';
-import { Map as LeafletMap, TileLayer, Marker, Popup } from 'react-leaflet';
-import { IParcel, IParcelDetail, IProperty, IPropertyDetail } from 'actions/parcelsActions';
-import { ParcelPopupView } from '../ParcelPopupView';
+import { Map as LeafletMap, TileLayer, Marker, Popup, WMSTileLayer } from 'react-leaflet';
+import { IProperty, IPropertyDetail } from 'actions/parcelsActions';
```
Fosol, line 7: Wonder if they should be renamed to `actions/propertyActions`?

frontend/src/pages/MapView.tsx
```diff
@@ -2,11 +2,7 @@ import React, { useEffect } from 'react';
 import Map, { MapViewportChangeEvent } from '../components/maps/leaflet/Map';
 import './MapView.scss';
 import { getFetchLookupCodeAction } from 'actionCreators/lookupCodeActionCreator';
-import {
-  fetchParcels,
-  fetchPropertyDetail,
-  fetchParcelDetail,
-} from 'actionCreators/parcelsActionCreator';
+import { fetchParcels, fetchPropertyDetail } from 'actionCreators/parcelsActionCreator';
```
Fosol, line 5: Should this be also renamed to `fetchProperties, fetchPropertyDetail` and `actionCreators/propertyActionCreator`?

frontend/src/components/maps/leaflet/Map.tsx
asanchezr: Agree, but given that most frontend code will be referencing these files I don't want to do it in this PR. I want to check with Devin whether he liked using `redux-toolkit` for defining our redux code (with less boilerplate and within a single file, instead of 3 files spread out in the solution)
If `redux-toolkit` is a go, I say we file a tech-debt item to refactor our state management code and **then** we would rename this as part of that work. What do you think?

frontend/src/pages/MapView.tsx
asanchezr: See mu previous comment

PR #129, opened 2020-04-15
PIMS-1517 PIMS-1520 add municipality field and tests.

devinleighsmith:


frontend/src/App.tsx
```diff
@@ -44,7 +44,7 @@ const App = () => {
   }, []);
 
   const isInitialized = () => {
-    return !keycloak?.authenticated || keycloak?.userInfo || keycloakUserLoaded;
+    return keycloak && keycloakUserLoaded;
```
asanchezr, line 47: I guess this fixes the login issues in DEV?

frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -86,103 +87,199 @@ const parcelDetailForm = (
...
+    ) => {
+      const input = container.querySelector(`${type}[name="${name}"]`);
+      await wait(() => {
+        if (type === 'input') {
+          fireEvent.change(input!, {
+            target: {
+              value: value,
+            },
+          });
+        } else {
```
asanchezr, line 130: I wonder if the `else` branch is needed here since they both execute the same `fireEvent,change()` code. Am I missing something subtle here?

frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -86,103 +87,199 @@ const parcelDetailForm = (
...
+      landArea: 1234,
+      statusId: 1,
+      classificationId: 1,
+      isSensitive: false,
+      latitude: 0,
+      longitude: 0,
+      evaluations: [],
+      buildings: [],
+    };
+    const fillInput = async (
```
asanchezr, line 116: I like this helper and feel it could be useful in other form tests (eg ParcelFilterMap) - any chance you could extract this to a test helper class/module of it's own?

Fosol: I'd second this!  If we can make it generic enough to reuse.

frontend/src/forms/__snapshots__/ParcelDetailForm.test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ParcelDetailForm renders correctly 1`] = `
+exports[`ParcelDetailForm ParcelDetailForm renders correctly 1`] = `
```
Fosol, line 3: Any reason for the duplicate "ParcelDetailForm"?

frontend/src/reducers/parcelsReducer.spec.tsx
```diff
@@ -52,6 +52,7 @@ const defaultParcelDetail: IParcelDetail = {
     zoning: '20',
     zoningPotential: '21',
     agencyId: '22',
+    municipality: '23',
```
Fosol, line 55: It's currently a free-form string value, not a number.

frontend/src/forms/__snapshots__/ParcelDetailForm.test.tsx.snap
asanchezr: This is auto-generated by React so don't know what changed here...

frontend/src/reducers/parcelsReducer.spec.tsx
asanchezr: Please note the above is the string "23", not a number so it complies. :)

asanchezr:
@devinleighsmith - This PR has been approved and is ready to merge but I'd like a response on my question about two branches of an IF statement running the same code please.

frontend/src/App.tsx
devinleighsmith: unrelated, the change above shouldn't affect behaviour, it just remove some redundant checks.

frontend/src/forms/ParcelDetailForm.test.tsx
devinleighsmith: No, I was messing around with this and had an alternate branch at some point, I'll move this into a helper class and fix the if/else in my next PR.

PR #131, opened 2020-04-15
PIMS-1510 Landing Page - Municipality Filter

asanchezr:
- Added new field to form
- Updated relevant unit tests

### DIFF
--- a/frontend/src/pages/MapView.tsx
+++ b/frontend/src/pages/MapView.tsx
@@ -20,6 +20,7 @@ const parcelBounds: IParcelListParams = {
   swLatitude: 48.43,
   swLongitude: -123.37,
   address: null,
+  municipality: null,
   agencies: null,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/MapView.tsx
```diff
@@ -72,6 +74,7 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
       swLatitude: sw.lat,
       swLongitude: sw.lng,
       address: address ?? null,
+      municipality: municipality ?? null,
```
Fosol, line 77: Is this so the value is `null` but not `undefined`?

asanchezr: Yeah. That's so we always send the field along in the query string like `?municipality=`. 

If `undefined` I believe nothing is sent across

PR #134, opened 2020-04-16
PIMS-1550 Ensure that user is loaded on page load and if logged in

goemen:
I used a React Context that exposes the ready state of keycloak, and userInfo. This allows us to always accurate userInfo properties like Groups when checking for roles. Can extend the userInfo object and use it for more advanced permission checks.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -24,13 +24,13 @@ import SubmitProperty from 'pages/SubmitProperty';
 import LoadingBar from 'react-redux-loading-bar';
 import ErrorBoundary from 'react-error-boundary';
 import ErrorModal from 'components/common/ErrorModal';
+import { AuthStateContext, IAuthState } from 'contexts/authStateContext';
 
 export const store = configureStore();
 
 const App = () => {
   const keycloakWrapper = useKeycloakWrapper();
   const keycloak = keycloakWrapper.obj;
-  const [keycloakUserLoaded, setkeycloakUserLoaded] = useState({});
   const dispatch = useDispatch();
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


frontend/src/App.tsx
```diff
@@ -24,58 +24,63 @@ import SubmitProperty from 'pages/SubmitProperty';
...
     }
-    keycloak?.loadUserInfo().then(() => {
-      setkeycloakUserLoaded(true);
-    });
   }, []);
 
-  const isInitialized = () => {
-    return keycloak && keycloakUserLoaded;
-  };
+  return (
```
Fosol, line 43: Does this allow for an unauthorized user (a true guest) have access to public routes?

devinleighsmith: From the if statement it looks like we'll get the following behaviour:
authenticated + userinfo = display app (this is our happy path)
not authenticated + no userinfo = display app (this is our not-logged-in user path)
authenticated +  no userinfo = display loading bar (presumably this is because the userinfo request is still loading)
not authenticated + userinfo = display app? This is a weird case, not sure if it could ever come up, but it doesn't seem right. I think the userinfo request will fail if the user is not authenticated.

as a result, could we simplify this to:
(!context.ready || (context.ready && context.authenticated && !context.userInfo))

frontend/src/contexts/authStateContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ready: false,
+});
+
+export const AuthStateContextProvider = (props: { children?: any }) => {
+  const keycloak = useKeycloakWrapper();
+  const [userInfo, setUserInfo] = React.useState<any>(null);
+  const keyCloakReady: boolean = useSelector<RootState, boolean>(state => state.keyCloakReady);
+
+  React.useEffect(() => {
+    const loadUserInfo = async () => {
```
devinleighsmith, line 24: I like the addition of keycloak readiness tracking in this provider. I'm not sure if I see the benefit of tracking the userInfo here as well as in the keycloak provider. Is the intent for this provider to track all of the keycloak data instead of the existing keycloak provider?

frontend/src/App.tsx
goemen: @devinleighsmith you right, a result of moving code around. I will make the change and test it

frontend/src/contexts/authStateContext.tsx
goemen: We don't have to track user info. But I realized that the useKeycloakWrapper does not always have the roles because the userInfo is not loaded. Just something to think about.

goemen:
- [ ]  Removed the tracking of user info.

- [ ] Using the keycloak ready state, and checking if the user info is available when authenticated to determine if the Auth State is ready

frontend/src/contexts/authStateContext.tsx
devinleighsmith: I don't think its a problem to track it in more places, just wanted to make sure I understand.

frontend/src/contexts/authStateContext.tsx
```diff
@@ -32,9 +29,10 @@ export const AuthStateContextProvider = (props: { children?: any }) => {
   return (
     <AuthStateContext.Provider
       value={{
-        authenticated: keycloak.obj?.authenticated,
-        ready: keyCloakReady,
-        userInfo,
+        // if user info is not available when authenticated, then the auth state is not ready
```
devinleighsmith, line 32: Cool, moving this makes more sense to me.

frontend/src/constants/reducerTypes.tsx
```diff
@@ -22,6 +22,7 @@ export const GET_REQUEST_ACCESS = 'getRequestAccess';
 export const GET_USERS = 'getUsers';
 
 export const JWT = 'jwt';
+export const KEYCLOAK_READY = 'keyCloakReady';
```
asanchezr, line 25: /nit: perhaps change value to `'keycloakReady'` ?

frontend/src/contexts/authStateContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const AuthStateContext = React.createContext<IAuthState>({
+  ready: false,
+});
+
+export const AuthStateContextProvider = (props: { children?: any }) => {
+  const keycloak = useKeycloakWrapper();
+  const [userInfo, setUserInfo] = React.useState<any>(null);
+  const keyCloakReady: boolean = useSelector<RootState, boolean>(state => state.keyCloakReady);
```
asanchezr, line 18: /nit: please rename to `keycloakReady` to remain consistent with existing naming

frontend/src/contexts/authStateContext.tsx
asanchezr: I also like that we are using React `Context` for this state as it need to be "globally" available to most (if not all) of our components.

What I'd like to suggest is to try to refactor the Keycloak hook and this Context into one solution/provider to simplify how we handle things. Right now this context needs to know about keycloak internals anyway, and then the app components need to be aware of this context "ready" value and also the "keycloak" object returned from the hook. I rather have everything under one provider - either the hook or the context.

This shouldn't block this PR from merging. We can tackle that later as tech-debt.

Good work!

PR #135, opened 2020-04-16
PIMS-1343,PIMS-1344,PIMS-1383

devinleighsmith:
Only display a single view/update button on parcel popup detail depending on user agency ownership.

### DIFF
--- a/frontend/src/actionCreators/parcelsActionCreator.spec.tsx
+++ b/frontend/src/actionCreators/parcelsActionCreator.spec.tsx
@@ -27,9 +27,9 @@ describe('fetchParcels action creator', () => {
     const mockResponse = { data: { success: true } };
     mockAxios.onGet(url).reply(200, mockResponse);
     return fetchParcels(null)(dispatch).then(() => {
-      expect(requestSpy).toHaveBeenCalledTimes(0);
-      expect(successSpy).toHaveBeenCalledTimes(0);
-      expect(dispatch).toHaveBeenCalledTimes(0);
+      expect(requestSpy).toHaveBeenCalledTimes(1);
+      expect(successSpy).toHaveBeenCalledTimes(1);
+      expect(dispatch).toHaveBeenCalledTimes(6);
     });
   });
 
--- a/frontend/src/components/maps/ParcelPopupView.tsx
+++ b/frontend/src/components/maps/ParcelPopupView.tsx
@@ -6,6 +6,7 @@ import { Label } from 'components/common/Label';
 import './ParcelPopupView.scss';
 import { Link } from 'react-router-dom';
 import { EvaluationKeys } from '../../constants/evaluationKeys';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 
 export interface IParcelDetailProps {
   parcel: IParcel | null;
@@ -14,6 +15,7 @@ export interface IParcelDetailProps {
 
 export const ParcelPopupView = (props: IParcelDetailProps | null) => {
   const parcelDetail: IParcel | null | undefined = props?.parcel;
+  const keycloak = useKeycloakWrapper();
 
   return (
     <Container className="parcelPopup" fluid={true}>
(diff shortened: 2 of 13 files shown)


frontend/src/actionCreators/parcelsActionCreator.spec.tsx
```diff
@@ -27,9 +27,9 @@ describe('fetchParcels action creator', () => {
     const mockResponse = { data: { success: true } };
     mockAxios.onGet(url).reply(200, mockResponse);
     return fetchParcels(null)(dispatch).then(() => {
-      expect(requestSpy).toHaveBeenCalledTimes(0);
-      expect(successSpy).toHaveBeenCalledTimes(0);
-      expect(dispatch).toHaveBeenCalledTimes(0);
+      expect(requestSpy).toHaveBeenCalledTimes(1);
```
Fosol, line 30: I'm not sure this is correct.  I had to change them to 0 when `null` is passed to `fetchParcels(...)` for the test to pass.  But I could be doing something wrong.

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -44,13 +46,12 @@ const mockDetails: IParcelDetail = {
     isSensitive: false,
     evaluations: [
       {
-        assessedValue: 1000000,
-        estimatedValue: 0,
-        fiscalYear: 2019,
-        netBookValue: 0,
-        appraisedValue: 0,
+        date: '2019',
+        key: '',
```
Fosol, line 50: Did you want to test an empty string?  I thought I created enums that you could use.

frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    latitude: 48,
+    longitude: 123,
+    propertyStatus: 'active',
+    classification: 'Core Operational',
+    description: 'test',
+    isSensitive: false,
+    evaluations: [
+      {
+        date: '2019',
+        key: '',
```
Fosol, line 49: Did you want to test an empty string?  I thought I created enums that you could use.

frontend/src/actionCreators/parcelsActionCreator.spec.tsx
devinleighsmith: The two scenarios that we support are 
1) null boundaries
or
2) boundaries where the area is bigger then a point
So this should have a request, success, etc.

frontend/src/components/maps/leaflet/MapProperties.test.tsx
devinleighsmith: This is just to make this test compile, I'll write actual tests for this change when I start working on the parcelDetailForm updates.

frontend/src/components/maps/ParcelPopupView.tsx
```diff
@@ -57,14 +59,13 @@ export const ParcelPopupView = (props: IParcelDetailProps | null) => {
...
-                Update
-              </Link>
-            )}
+            {parcelDetail?.id &&
+              !props?.disabled &&
+              (!keycloak.hasAgency(parcelDetail?.agencyId) ? (
+                <Link to={`/submitProperty/${parcelDetail?.id}?disabled=true`}>View</Link>
+              ) : (
+                <Link to={`/submitProperty/${parcelDetail?.id}`}>Update</Link>
+              ))}
```
asanchezr, line 68: The logic here seems a bit hard to follow to me. 
What would you think of maybe having two booleans `canView`, `canUpdate` and then JSX looks a bit cleaner IMO?

```jsx
{canView && (<Link ... />)}
{canUpdate && (<Link ... />)}
```

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -72,6 +72,11 @@ const Map: React.FC<MapProps> = ({
   const [baseLayers, setBaseLayers] = useState<BaseLayer[]>([]);
   const [activeBasemap, setActiveBasemap] = useState<BaseLayer | null>(null);
 
+  //do not jump to map coordinates if we have an existing map but no parcel details.
+  if (mapRef.current && !selectedProperty?.parcelDetail) {
+    lat = (mapRef.current.props.center as Array<number>)[0];
+    lng = (mapRef.current.props.center as Array<number>)[1];
+  }
```
asanchezr, line 79: I'm okay if this PR gets merged as it has already been approved but the above doesn't seem to fix the jumping around on the map. I will post a video shortly.

asanchezr:
@devinleighsmith I still see the map jumping around bug. No need to block this PR, just need to be handled properly in another PR

video

frontend/src/components/maps/leaflet/Map.tsx
devinleighsmith: We talked about this in rocketchat, this PR fixes two issues:
1) The map moving when the popup is closed.
2) The map only opening popups once.

This does not fix the issue assigned to Goeme PIMS-1549 (and isn't intended to).

PR #136, opened 2020-04-17
PIMS-1421 Updating Request Access UI

Fosol:
Fixing database migration - AccessRequest (please note this requires another DB refresh).

The authenticated guest user can now submit and update their access request!

image

There is one unit test related to parcels that I can't resolve.  Additionally locally Jest throws a lot of errors that I also can't debug.


backend/api/Controllers/UserController.cs
```diff
@@ -70,32 +72,96 @@ public async Task<IActionResult> UserInfoAsync()
...
+        public IActionResult GetAccessRequest()
         {
-            if (accessRequestModel == null || accessRequestModel.Agencies == null || accessRequestModel.Roles == null)
+            var accessRequest = _userService.GetAccessRequest();
+            if (accessRequest == null) return NoContent();
+            return new JsonResult(_mapper.Map<Model.AccessRequestModel>(accessRequest));
+        }
+
+        /// <summary>
+        /// Get the most recent access request for the current user.
```
Fosol, line 93: Wrong description, I'll have to update later.

backend/api/Controllers/UserController.cs
```diff
@@ -70,32 +72,96 @@ public async Task<IActionResult> UserInfoAsync()
...
+        /// Provides a way for a user to update their access request.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPut("access/requests/{id}")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.AccessRequestModel), 200)]
+        [ProducesResponseType(typeof(Models.ErrorResponseModel), 400)]
+        [ProducesResponseType(typeof(Models.ErrorResponseModel), 403)]
+        [SwaggerOperation(Tags = new[] { "user" })]
+        public IActionResult UpdateAccessRequest(int id, [FromBody] Model.AccessRequestModel model)
```
Fosol, line 146: We now allow a user to update their request until it become disabled.

backend/api/Helpers/Mapping/ServiceCollectionExtensions.cs
```diff
@@ -10,7 +10,7 @@ public static class ServiceCollectionExtensions
     {
         public static IServiceCollection AddMapster(this IServiceCollection services, Action<TypeAdapterConfig> options = null)
         {
-            var config = new TypeAdapterConfig();
+            var config = TypeAdapterConfig.GlobalSettings;
```
Fosol, line 13: We can now globally configure Mapster.

backend/api/Startup.cs
```diff
@@ -80,7 +80,7 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddMapster(options =>
             {
-                TypeAdapterConfig.GlobalSettings.Default.IgnoreNonMapped(true);
+                options.Default.IgnoreNonMapped(true);
```
Fosol, line 83: By default every property needs to be mapped.  This will allow us to control model versions.  Without it if we added a property to a model it would automatically include it in the output.

backend/dal/Migrations/Initial/PostDeploy/01-Agencies.sql
```diff
@@ -72,7 +72,7 @@ INSERT INTO dbo.[Agencies] (
     11
     , 7
     , 'ADM'
-    , 'Actiing Deputy Minister'
+    , 'Acting Deputy Minister'
```
Fosol, line 75: Spelling mistake

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -82,18 +83,63 @@ public User Activate()
...
+            var accessRequest = this.Context.AccessRequests
+                .Include(a => a.Agencies)
+                .ThenInclude(a => a.Agency)
+                .Include(a => a.Roles)
+                .ThenInclude(r => r.Role)
+                .Include(a => a.User)
+                .AsNoTracking()
+                .FirstOrDefault(a => a.Id == id) ?? throw new KeyNotFoundException();
+            var userId = this.User.GetUserId();
+            if (accessRequest.UserId != userId) throw new NotAuthorizedException();
```
Fosol, line 122: Not allowed to requests someone else's access request.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -107,6 +153,61 @@ public AccessRequest AddAccessRequest(AccessRequest request)
...
+
+            // Add agencies and roles if required.
+            var addAgencies = request.Agencies.Except(entity.Agencies, new AccessRequestAgencyAgencyIdComparer());
+            addAgencies.ForEach(a => entity.Agencies.Add(a));
+
+            var addRoles = request.Roles.Except(entity.Roles, new AccessRequestRoleRoleIdComparer());
+            addRoles.ForEach(r => entity.Roles.Add(r));
+
+            // Copy values into entity.
+            this.Context.Entry(entity).CurrentValues.SetValues(request);
```
Fosol, line 201: The `entity` is being tracked so we need to update it with the untracked `request`.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -107,6 +153,61 @@ public AccessRequest AddAccessRequest(AccessRequest request)
...
+            this.Context.SetOriginalRowVersion(entity);
```
Fosol, line 202: I find this annoying, but concurrency won't work unless we manually do this.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -107,6 +153,61 @@ public AccessRequest AddAccessRequest(AccessRequest request)
...
+
+            this.Context.AccessRequests.Update(entity);
+            this.Context.CommitTransaction();
+            this.Context.Entry(request).CurrentValues.SetValues(entity);
```
Fosol, line 206: I want to mutate the `request` as it is our common pattern everywhere else.  If I don't do this the `request` will not have the latest `rowVersion` and as such the subsequent request to update will fail.

backend/entities/AccessRequest.cs
```diff
@@ -13,7 +13,7 @@ public class AccessRequest : BaseEntity
         /// get/set - The primary key IDENTITY.
         /// </summary>
         /// <value></value>
-        public Guid Id { get; set; }
+        public int Id { get; set; }
```
Fosol, line 16: Guid's are generally bad primary keys as they will cause defragmentation and need the index to be rebuilt constantly.

backend/entities/AccessRequest.cs
```diff
@@ -28,18 +28,24 @@ public class AccessRequest : BaseEntity
         public bool? IsGranted { get; set; }
 
         /// <summary>
-        /// Foreign key to User
+        /// get/set - Foreign key to User
         /// </summary>
         /// <returns></returns>
-        public Guid? UserId { get; set; }
+        public Guid UserId { get; set; } // TODO: This shouldn't be nullable.
```
Fosol, line 34: An access request must be related to a user.  Need to remove the comment.

backend/tests/core/Pims.Core.Test.csproj
```diff
@@ -13,6 +13,10 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.InMemory" Version="3.1.3" />
     <PackageReference Include="Moq" Version="4.13.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1">
```
Fosol, line 16: Fixing issue where VS complains about this assembly because it references `xunit`.

backend/tests/unit/api/Controllers/PropertyControllerTest.cs
```diff
@@ -22,6 +23,46 @@ namespace Pims.Api.Test.Controllers
     public class PropertyControllerTest
     {
         #region Variables
+        public static IEnumerable<object[]> AllPropertiesFilters = new List<object[]>()
```
Fosol, line 26: Theories are so much easier to add new tests and get better coverage!

frontend/src/actionCreators/accessRequestActionCreator.tsx
```diff
@@ -0,0 +1,129 @@
...
+} from 'actions/accessRequestActions';
+import { IAccessRequest } from 'interfaces/accessRequests';
+import * as actionTypes from 'constants/actionTypes';
+import { ENVIRONMENT } from 'constants/environment';
+import CustomAxios from 'customAxios';
+import { AxiosResponse, AxiosError } from 'axios';
+import { createRequestHeader } from 'utils/RequestHeaders';
+
+/**
+ * Converts the 'values' object into an IAccessRequest object
```
Fosol, line 18: Can we start using `tsdoc` everywhere?  I find it very difficult to navigate through our current code without it.

frontend/src/actionCreators/usersActionCreator.tsx
```diff
@@ -25,75 +24,6 @@ export const getActivateUserAction = () => (dispatch: Function) => {
     .finally(() => dispatch(hideLoading()));
 };
 
-export const toAccessRequest = (values: any): adminActions.IAccessRequest => {
```
Fosol, line 28: Moved these to their own action creator file.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -12,27 +19,42 @@ function useKeycloakWrapper() {
   const hasClaim = (claim?: string) => {
     return !claim || keycloak?.hasResourceRole(claim);
   };
+  const userInfo = keycloak?.userInfo as IUserInfo;
```
Fosol, line 22: @devinleighsmith @goemen @asanchezr Please review this merged file.  I'm not certain it's correct.

frontend/src/interfaces/accessRequests.tsx
```diff
@@ -0,0 +1,39 @@
+export interface IUserInfo {
```
Fosol, line 1: I was finding it rather random where interfaces were being stored for data models.  Maybe it was just my lack of understanding, but I thought it might be easier if they were in their own location and perhaps had a similar structure as those in the backend.  Please let me know your thoughts on this?

frontend/src/pages/AccessRequestPage.scss
```diff
@@ -0,0 +1,60 @@
...
+  padding-top: 15px;
+  padding-bottom: 10px;
+}
+
+.navBar {
+  font-weight: 700;
+  text-align: left;
+}
+
+div.required {
```
Fosol, line 53: I had to override the default styling because they make the label text red.  Which is not desirable or based on UI design.  Perhaps the root styling needs to be updated?

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
+import React, { useEffect } from 'react';
```
Fosol, line 1: This file is a merge of Devin's original access request page and Jonny's edit user page, with a little chaos from me added into the mix.  Thanks guys for your hard work which allowed me to get this working.

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+            initialValues={initialValues}
+            validationSchema={AccessRequestSchema}
+            onSubmit={(values, { setSubmitting }) => {
+              dispatch(getSubmitAccessRequestAction(toAccessRequest(values)));
+              setSubmitting(false);
+            }}
+          >
+            {props => (
+              <Form className="userInfo">
+                {(!request?.isFetching && request?.status === 201) || request?.status === 200 ? (
```
Fosol, line 137: We need a better way to do this.  My styling is horribly ugly too.

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  <div className="success">
+                    <p>Your request has been submitted.</p>
+                  </div>
+                ) : null}
+
+                {initialValues.id !== 0 ? (
```
Fosol, line 143: Styling is horrible here too.

adr/database.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Decision
+
+Originally the database generated for the SWU was with PostgreSQL, after further consideration it made more sense to tightly couple both MS-SQL with .NET Core.
+This will give us better performance and tighter integration with Entity Framework Core.
+It was decided to create a linux docker container to host the MS-SQL 2019 database.
+
+## Consequences
+
+There are no known negative consequences related to the PIMS project by making this decision.
```
devinleighsmith, line 23: technically it makes our life harder when it comes to openshift backups - no existing lab project we can use.

backend/api/Mapping/BaseMap.cs
```diff
@@ -9,13 +9,11 @@ public class BaseMap : IRegister
         public void Register(TypeAdapterConfig config)
         {
             config.NewConfig<Entity.BaseEntity, Models.BaseModel>()
-                .IgnoreNullValues(true)
```
devinleighsmith, line 12: what was the motivation for this change?

Fosol: A mistake!  Thanks for catching it!

Fosol: I've now added it to the global settings in the startup configuration.

backend/dal/Services/Concrete/UserService.cs
devinleighsmith: Does this do anything though? you mutate the request value but don't return it. Aren't we expecting that a consumer will use the returned entity rather then a mutated parameter?

adr/database.md
Fosol: I've updated with a comment to reflect this.

backend/dal/Services/Concrete/UserService.cs
Fosol: This was bugging me with our current implementation.  I generally feel we shouldn't mutate objects if we can avoid it, however in the majority (possibly all) our current uses we expect the entity to be mutated.  

This particular case was complicated by the fact that we need to pull the latest data from the DB to determine what has changed.  EF will get confused if we try and track both entities, which means I choose to manually mutate the `request` so that it wouldn't confuse the common use throughout the backend.

To stop mutating entities passed in would be a lot of effort and I'm not sure we gain much by doing that.

Fosol: To be clear, if I don't mutate then the `rowVersion` will not be updated.  In most of our endpoints we don't necessarily use the returned entity.  Which will result in the wrong data being returned.

frontend/src/actionCreators/accessRequestActionCreator.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return {
+    id: values.id,
+    userId: values.userId,
+    user: {
+      id: values.userId,
+      username: values.user.username,
+      email: values.user.email,
+      position: values.user.position,
+    },
+    agencies: [{ id: parseInt(values.agency) }],
```
devinleighsmith, line 32: do you want any handling here for NaN?

Fosol: Hmmm, good idea.

frontend/src/actions/adminActions.tsx
```diff
@@ -50,16 +40,23 @@ export interface IRole {
   id?: number;
 }
 export interface IAccessRequest {
```
devinleighsmith, line 42: probably make sense to move this out as well.

frontend/src/interfaces/accessRequests.tsx
devinleighsmith: That works for me, although I don't think I agree with userInfo being part of AccessRequests.

frontend/src/pages/AccessRequestPage.scss
devinleighsmith: It's red for the UI design on the parcel detail form, my be a disconnect between Aiden/Christine.

frontend/src/pages/AccessRequestPage.tsx
devinleighsmith: Using an alert would be preferable to a <p>
https://react-bootstrap.github.io/components/alerts/

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * The AccessRequestPage provides a way to new authenticated users to submit a request
+ * that associates them with a specific agency and a role within the agency.
+ * If they have an active access request already submitted, it will allow them to update it until it has been approved or disabled.
+ * If their prior request was disabled they will then be able to submit a new request.
+ */
+const AccessRequestPage = () => {
+  const keycloakWrapper = useKeycloakWrapper();
+  const keycloak = keycloakWrapper.obj;
+  let userInfo = keycloak?.userInfo as IUserInfo;
```
devinleighsmith, line 38: this can probably be a const right?

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+    state => (state.network as any)[actionTypes.ADD_REQUEST_ACCESS] as IGenericNetworkAction,
+  );
+  const agencies = _.filter(lookupCodes, (lookupCode: ILookupCode) => {
+    return lookupCode.type === API.AGENCY_CODE_SET_NAME;
+  });
+  const roles = _.filter(lookupCodes, (lookupCode: ILookupCode) => {
+    return lookupCode.type === API.ROLE_CODE_SET_NAME;
+  });
+
+  let checkAgencies: {} | null | undefined;
```
devinleighsmith, line 62: Seems like you could just use: *any* here.

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+  let checkRoles: {} | null | undefined;
+
+  let accessRequest = data?.accessRequest;
```
devinleighsmith, line 65: possibly also a const?

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+  let initialValues: IAccessRequestForm = {
```
devinleighsmith, line 66: const

frontend/src/pages/AccessRequestPage.tsx
devinleighsmith: could add some utility functions to request to incapsulate the fetching check and the expected statuses:
request.isCompleteWithStatus(200, 201) or similar.

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+      firstName: userInfo?.firstName,
+      lastName: userInfo?.lastName,
+      email: userInfo?.email,
+      position: accessRequest?.user?.position ?? userInfo?.position ?? '',
+    },
+    agencies: accessRequest?.agencies ?? [],
+    isGranted: accessRequest?.isGranted ?? false,
+    roles: accessRequest?.roles ?? [],
+    note: accessRequest?.note ?? '',
+    agency: accessRequest?.agencies?.find(x => x !== undefined).id,
```
devinleighsmith, line 82: could this just be x => x ?
(using the truthiness of x).

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+    role: accessRequest?.roles?.find(x => x !== undefined).id,
+    rowVersion: accessRequest?.rowVersion,
+  };
+
+  const mapLookupCode = (code: ILookupCode, defaultId: number | string | null): SelectOption => ({
```
devinleighsmith, line 87: you may be able to use the mapLookupCode from 'utils' instead of this.

frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+                <TextArea
+                  label="Notes"
+                  field="note"
+                  placeholder="Please specify why you need access to this site."
+                />
+
+                <p>
+                  By clicking request, you agree to our{' '}
+                  <a href="https://www2.gov.bc.ca//gov/content/home/disclaimer">
```
devinleighsmith, line 205: Might be good to store these urls in case they are reused, such as in strings.tsx

frontend/src/reducers/accessRequestReducer.tsx
```diff
@@ -19,6 +28,13 @@ const accessRequestReducer = (state = initialState, action: IStoreAccessRequests
           items: [...action.pagedAccessRequests.items],
         },
       };
+    case actionTypes.STORE_ACCESS_REQUEST:
+      return {
+        ...state,
+        accessRequest: {
+          ...action.accessRequest,
```
devinleighsmith, line 35: you may need to explicitly copy all levels of nested data (ie, what I did with the array above).
https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

or use immer, which is used in the slices:
https://github.com/immerjs/immer

devinleighsmith:
Leaving an approval as I don't see any showstoppers but would like to see some of the comments get addressed before merge.

frontend/src/interfaces/accessRequests.tsx
Fosol: moved

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

frontend/src/actions/adminActions.tsx
Fosol: done

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

frontend/src/pages/AccessRequestPage.tsx
Fosol: done

PR #137, opened 2020-04-17
PIMS-1549 Using a redux controlled zoom 

goemen:
Using a redux controlled zoom to stop the map from defaulting to zoom 14 when the page rerenders


auth/keycloak/config/hs_err_pid385.log
```diff
@@ -0,0 +1,1325 @@
+#
```
Fosol, line 1: I doubt we want this in source.

goemen: Oooops!!, will remove it. And yes the zoom level will be saved in redux

asanchezr:
Before we merge this PR, I'd like to take a quick look locally. I think there might be a simpler way to fix this than introducing additional global state for the map zoom level

PR #141, opened 2020-04-20
PIMS-1419 Updating navigation.

Fosol:
Adding missing packages for unit-testing.

Note: A number of unit-tests are failing.  I'm not certain they're related to my changes.


frontend/package.json
```diff
@@ -60,8 +60,10 @@
     "@types/redux-logger": "3.0.7",
     "@types/redux-mock-store": "1.0.2",
     "axios-mock-adapter": "1.17.0",
+    "canvas": "^2.6.1",
```
Fosol, line 63: I keep getting errors locally when `npm test`.  Adding it seems to solve some of the errors.  Not sure it's required though.

frontend/package.json
```diff
@@ -60,8 +60,10 @@
...
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.2",
+    "enzyme-to-json": "^3.4.4",
```
Fosol, line 66: Is used in one of our tests.

frontend/src/App.tsx
```diff
@@ -59,9 +60,14 @@ const App = () => {
                     <Route path="/login" component={Login}></Route>
                     <PrivateRoute path="/accessdenied" component={AccessDenied}></PrivateRoute>
                     <PrivateRoute
-                      path="/admin"
```
Fosol, line 62: Replacing the administration dashboard for now with two routes that go directly to our pages.

frontend/src/PrivateRoute.tsx
```diff
@@ -5,9 +5,14 @@ import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 interface IPrivateRouteProps extends RouteProps {
   // tslint:disable-next-line:no-any
   component: any;
-  role?: string;
+  role?: string | Array<string>;
```
Fosol, line 8: Can now test either option.

frontend/src/components/navigation/AppNavBar.tsx
```diff
@@ -30,47 +30,48 @@ function AppNavBar() {
...
-    } else if (keycloak.isAdmin) {
-      return (
-        <Navbar.Collapse className="links">
-          <Nav.Link onClick={() => history.push('/admin/requests')}>Access Requests</Nav.Link>
-          <Nav.Link onClick={() => history.push('/admin/users')}>User Management</Nav.Link>
-        </Navbar.Collapse>
-      );
-    }
-  };
+  const canAdmin = keycloak.isAdmin ? (
```
Fosol, line 33: Cleaned this up, but we need to plan our navigation better.

frontend/src/components/navigation/Header.tsx
```diff
@@ -22,6 +22,9 @@ const Header = () => {
     errors.forEach(error => dispatch(clear(error.name)));
     setShow(false);
   };
+  const url = `${window.location.protocol}//${window.location.hostname}${
```
Fosol, line 25: Clicking the image should go to the home page

frontend/src/components/navigation/__snapshots__/Header.test.tsx.snap
```diff
@@ -24,7 +24,7 @@ exports[`header renders correctly 1`] = `
           >
             <a
               className="navbar-brand"
-              href="https://gov.bc.ca"
+              href="http://localhostnull"
```
Fosol, line 27: Ah this is wrong.

asanchezr, line 27: As you can see in this snapshot your code above should not be concatenating `null` to the string. It will get "null" appended!

frontend/src/components/navigation/Header.tsx
```diff
@@ -22,6 +22,9 @@ const Header = () => {
...
+    window.location.port ? ':' + window.location.port : null
+  }`;
```
asanchezr, line 27: Please refactor the above. It is adding "null" to the URL if no port is provided

frontend/src/constants/claims.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Claims enum, provides a list of permissions that govern what actions are available to an authenticated user.
+ */
+export enum Claims {
```
Fosol, line 4: Enum rather than basic strings

frontend/src/constants/roles.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Roles enum, provide the available role names that a user can belong to.
+ */
+export enum Roles {
```
Fosol, line 4: Enum rather than basic strings

frontend/src/constants/strings.tsx
```diff
@@ -32,24 +32,6 @@ export const DEFAULT_PAGE = '1';
 export const DEFAULT_PER_PAGE = '25';
 export const DEFAULT_DASHBOARD_PARAMS = '?page=1&per_page=25';
 
-// roles
-export const SYSTEM_ADMINISTRATOR = 'System Administrator';
```
Fosol, line 36: Moved to enums

frontend/src/components/navigation/AppNavBar.tsx
```diff
@@ -30,47 +30,48 @@ function AppNavBar() {
...
     <Navbar className="map-nav" expand="lg">
       <Navbar.Toggle aria-controls="collapse" className="navbar-dark" />
       <Nav>
-        {getRoleBasedLinks()}
+        <Navbar.Collapse className="links">
+          {canAdmin}
+          {canEditProperties}
+          {canPropertyView}
+          {canDisposeRequest}
+          {canDisposeApprove}
```
asanchezr, line 69: I like this refactor! thanks for cleaning it up

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,60 +1,140 @@
 import { useKeycloak } from '@react-keycloak/web';
```
Fosol, line 1: Documentation and strong type control

frontend/src/PrivateRoute.tsx
asanchezr, line 8: I thought we had a single role per user? "single role, single agency" wasn't it?

frontend/src/PrivateRoute.tsx
```diff
@@ -5,9 +5,14 @@ import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
...
+  claim?: string | Array<string>;
```
asanchezr, line 9: I'd rather have this property be renamed to `claims` and always make it an array. It's cleaner and supports the single claim via `claims = [ 'single-claim' ]`

frontend/src/PrivateRoute.tsx
Fosol: We enforce one role and one agency per user in the frontend.  However if a user belongs to a parent agency, they will have an array of agencies.

The code above is to allow a role to use it.  Which means more than one role could use the route.

frontend/package.json
asanchezr: I would prefer to dig a bit deeper into why this is required before adding it. Which specific test is failing?

frontend/package.json
asanchezr: good catch!

frontend/src/PrivateRoute.tsx
Fosol: Not sure I agree entirely.  The way it is allows for overriding.  No different than what we'd do in languages that support overriding natively.

frontend/package.json
Fosol: I don't actually know for certain, as I'm having a hard time reading it all.
image

asanchezr: That is because you are not running Jest in verbose mode. See my comment on RC:
```sh
npm test -- --verbose
```

frontend/src/App.tsx
asanchezr: what's the reason for this?

frontend/src/PrivateRoute.tsx
asanchezr: not really, by having the code support string or array of strings you will have to check whether it is a string or an array. I don't see the need for the added complexity

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,60 +1,140 @@
...
 export interface IUserInfo {
+  displayName?: string;
   username: string;
   name?: string;
   preferred_username?: string;
   firstName?: string;
   lastName?: string;
   email: string;
-  Groups: any[];
+  groups: string[];
```
asanchezr, line 15: What comes out of the `userinfo` endpoint is actually "Groups" (case-sensitive). If you are going to rename in the frontend, also provide instructions on how to change the Groups property in Keycloak

frontend/src/App.tsx
Fosol: The dashboard doesn't need to exist at this point in time.  It also looks horrible and has zero functionality presently.  It will be cleaner to just go directly to the pages that have functionality.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,60 +1,140 @@
...
+   * Determine if the user has the specified 'claim'
+   * @param claim - The name of the claim
+   */
+  const hasClaim = (claim?: string | Array<string>): boolean => {
+    return (
+      !claim ||
+      (typeof claim === 'string'
+        ? userInfo?.roles?.includes(claim)
+        : claim.some(c => userInfo?.roles?.includes(c)))
+    );
```
asanchezr, line 59: The code above that has to check string or array is exactly what I alluded earlier. In my opinion always checking for the array version would lead to less bugs in the future. Also, less confusing to people learning the claims system in the front-end

frontend/src/hooks/useKeycloakWrapper.tsx
Fosol: This is configured by Keycloak and it will be 'groups'

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,60 +1,140 @@
...
+  /**
+   * Determine if the user belongs to the specified 'role'
+   * @param role - The role name or an array of role name
+   */
+  const hasRole = (role?: string | Array<string>): boolean => {
+    return (
+      !role ||
+      (typeof role === 'string'
+        ? userInfo?.groups?.includes(role)
+        : role.some(r => userInfo?.groups?.includes(r)))
```
asanchezr, line 71: Same things with roles

frontend/src/PrivateRoute.tsx
Fosol: The more common use is a single role/claim.  The code has already been update to check.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -1,60 +1,140 @@
...
+    return !agency || userInfo?.agencies?.includes(agency);
   };
-  const roles = () => {
-    return userInfo?.Groups ? [...userInfo?.Groups] : [];
+
+  /**
+   * Return an array of roles the user belongs to
+   */
+  const roles = (): Array<string> => {
+    return userInfo?.groups ? [...userInfo?.groups] : [];
```
asanchezr, line 87: Make sure you check for case-sensitivity here. Again, what I've seen in the JSON response is `obj.Groups`

frontend/src/pages/Login.test.tsx
```diff
@@ -51,7 +51,7 @@ test('login renders correctly', () => {
 
 test('authenticated users are redirected to the mapview', () => {
   (useKeycloak as jest.Mock).mockReturnValue({
-    keycloak: { authenticated: true, userInfo: { Groups: ['System Administrator'] } },
+    keycloak: { authenticated: true, userInfo: { groups: ['System Administrator'] } },
```
asanchezr, line 54: Once more check the name

frontend/src/hooks/useKeycloakWrapper.tsx
Fosol: I have no issues if we want to enforce `[ "role" ]` if it's truly simpler for use.  I just don't think it is simpler for use.  The complexity is already resolved within the function and now provides easier use of the function everywhere else.

frontend/src/App.tsx
asanchezr: Sure, just know that I'm refactoring the routing and will likely bring a different routing later. It's fine for now

frontend/src/hooks/useKeycloakWrapper.tsx
asanchezr: well , then please also export the current configuration or make scripts that update our local environments. The current config on my local is Groups

frontend/src/hooks/useKeycloakWrapper.tsx
asanchezr: It seems we have differing opinions which is okay. You are the author of the code so I'm fine with it. We can always revisit later if it causes issues

frontend/src/hooks/useKeycloakWrapper.tsx
Fosol: Ah yes I need to update the local keycloak!

asanchezr: The realm export in source control is way outdated. Before making further changes that break everyone else's setup please update the realm.json here

frontend/src/hooks/useKeycloakWrapper.tsx
Fosol: All I'm saying is we should hear the team on this.  I'm of the opinion that this is beneficial for use, your are saying it increases complexity and therefore increases future bug opportunities.

frontend/package.json
devinleighsmith: If you do end up adding this, make sure to remove the ^

frontend/package.json
devinleighsmith: same as above, remove ^

frontend/src/components/navigation/Header.tsx
devinleighsmith: this will force refresh the app, you likely want history.push here instead.

frontend/src/pages/Administration.tsx
```diff
@@ -16,7 +16,7 @@ const Administration = () => {
     <Container fluid={true}>
```
devinleighsmith, line 16: If you've removed this from the app routing just delete this file, can always get it back from source control if we want it.

devinleighsmith: Same for the corresponding tests and snapshot.

frontend/src/components/navigation/__snapshots__/Header.test.tsx.snap
Fosol: fixed

frontend/package.json
Fosol: fixed

frontend/src/pages/Administration.tsx
Fosol: fixed

PR #143, opened 2020-04-20
PIMS-1419 Fixing secure routes

Fosol:
Unauthenticated users cannot go to /access/request route.
Authenticated users without admin roles cannot go to /admin routes.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -58,7 +58,7 @@ const App = () => {
                   {keycloak?.authenticated ? <AppNavBar /> : null}
                   <Col style={{ padding: 0 }}>
                     <Route path="/login" component={Login}></Route>
-                    <PrivateRoute path="/accessdenied" component={AccessDenied}></PrivateRoute>
+                    <Route path="/forbidden" component={AccessDenied}></Route>
                     <PrivateRoute
                       path="/admin/users"
                       component={ManageUsers}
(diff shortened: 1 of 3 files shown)


frontend/src/App.tsx
```diff
@@ -58,7 +58,7 @@ const App = () => {
                   {keycloak?.authenticated ? <AppNavBar /> : null}
                   <Col style={{ padding: 0 }}>
                     <Route path="/login" component={Login}></Route>
-                    <PrivateRoute path="/accessdenied" component={AccessDenied}></PrivateRoute>
+                    <Route path="/forbidden" component={AccessDenied}></Route>
```
Fosol, line 61: I prefer single words for URL path segments

Fosol: Anyone can be forbidden if they don't have permission.  Additionally if it becomes a PrivateRoute it causes an infinite loop.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -52,7 +52,7 @@ function useKeycloakWrapper(): IKeycloak {
    */
   const hasClaim = (claim?: string | Array<string>): boolean => {
     return (
-      !claim ||
+      claim !== undefined &&
```
Fosol, line 55: Corrected logic.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -65,7 +65,7 @@ function useKeycloakWrapper(): IKeycloak {
    */
   const hasRole = (role?: string | Array<string>): boolean => {
     return (
-      !role ||
+      role !== undefined &&
```
Fosol, line 68: Corrected logic.

frontend/src/hooks/useKeycloakWrapper.tsx
asanchezr, line 55: I like this change but please also check for `null`. Since you are using the strict equality here `!==` we are not checking for nulls. 

As is, this code will error out if null is passed as `null !== undefined && null.some()` will throw

Fosol: Good catch!  I’ll update

PR #145, opened 2020-04-21
Pims 1454

devinleighsmith:
replace array based evaluations with table.


frontend/src/forms/subforms/EvaluationForm.tsx
```diff
@@ -1,90 +1,188 @@
...
+
+/**
+ * Extend IEvaluation, add the year of the evaluation (regardless of whether the date is set.)
+ */
+interface IFormEvaluation extends IEvaluation {
+  /** the year placeholder of this evaluation */
+  year?: number;
+}
+const NUMBER_OF_EVALUATIONS_PER_PAGE = 2;
+const NUMBER_OF_GENERATED_EVALUATIONS = 20;
```
Fosol, line 30: Weren't you going to set the furthest back date instead?  Otherwise in the future they won't see original values entered.

frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -69,22 +69,15 @@ const history = createMemoryHistory();
...
-  //   useSelector: (state: RootState, action: IPropertyDetail) => ({
-  //     propertyDetail: {
-  //       parcelDetail: {
-  //         latitude: 48.43,
-  //         longitude: -123.37,
-  //       },
-  //     },
-  //   }),
-  // }));
+xit('SubmitProperty renders a disabled form if agency does not match current user.', () => {
```
Fosol, line 80: Is there a way to get this test working again?

frontend/src/reducers/parcelsReducer.spec.tsx
```diff
@@ -29,6 +29,7 @@ const defaultParcelDetail: IParcelDetail = {
     longitude: 1,
     latitude: 2,
     id: 3,
+    fiscals: [],
```
Fosol, line 32: Does it also needs evaluations?

frontend/src/utils/YupSchema.tsx
```diff
@@ -48,23 +49,12 @@ export const Address = Yup.object().shape({
...
-    .required('Required'),
-  assessedValue: Yup.number()
-    .min(0, 'Assessed Value must be a positive number')
-    .required('Required'),
-  netBookValue: Yup.number()
-    .min(0, 'Net Book Value must be a positive number')
-    .required('Required'),
+  year: Yup.number(),
+  date: Yup.string().required('Required'),
+  key: Yup.string().nullable(),
```
Fosol, line 56: Keys are not nullable.  How are you ensuring they are set to the correct enum value?

backend/api/Mapping/Parcel/ParcelBuildingMap.cs
```diff
@@ -12,8 +12,10 @@ public void Register(TypeAdapterConfig config)
         {
             config.NewConfig<Entity.Building, Model.ParcelBuildingModel>()
                 .IgnoreNonMapped(true)
+                .EnableNonPublicMembers(true)
```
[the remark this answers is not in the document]
devinleighsmith: The sub collections were not mapping at all when only the getter was available.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -241,7 +241,87 @@ public Parcel Update(Parcel parcel)
...
+                    if (existingBuilding == null)
+                    {
+                        existingParcel.Buildings.Add(building);
+                    }
+                    else
+                    {
+                        this.Context.Entry(existingBuilding).CurrentValues.SetValues(building);
+                        this.Context.Entry(existingBuilding.Address).CurrentValues.SetValues(building.Address);
+
+                        foreach (var buildingEvaluation in building.Evaluations)
```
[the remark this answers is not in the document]
devinleighsmith: Yes

backend/entities/Building.cs
```diff
@@ -88,13 +88,13 @@ public class Building : Property
         /// get - A collection of evaluations for this building.
         /// </summary>
         /// <typeparam name="BuildingEvaluation"></typeparam>
-        public ICollection<BuildingEvaluation> Evaluations { get; } = new List<BuildingEvaluation> ();
+        public ICollection<BuildingEvaluation> Evaluations { get; private set; } = new List<BuildingEvaluation> ();
```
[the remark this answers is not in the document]
devinleighsmith: This was added directly in response to these collections not being mapped.

frontend/src/components/common/FormikDatePicker.tsx
```diff
@@ -1,19 +1,41 @@
...
-      {...field}
-      {...props}
-      calendarIcon={null}
-      selected={(field.value && new Date(field.value)) || null}
-      onChange={(val: any) => {
-        setFieldValue(field.name, val);
-      }}
-    />
+    <FormGroup>
+      <span className={invalid}></span>
```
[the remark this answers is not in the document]
devinleighsmith: This allows the datepicker to follow the same format as the inputs - and reuse the same css.

frontend/src/components/navigation/AppNavBar.tsx
```diff
@@ -43,9 +43,7 @@ function AppNavBar() {
     <Nav.Link>View Projects</Nav.Link>
   ) : null;
 
-  const canEditProperties = keycloak.hasClaim(Claims.PROPERTY_EDIT) ? (
-    <ManagePropertyDropDown />
-  ) : null;
+  const canEditProperties = true ? <ManagePropertyDropDown /> : null;
```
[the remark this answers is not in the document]
devinleighsmith: I didn't change this myself, must have been introduced by the merge, I'll fix this.

frontend/src/forms/subforms/EvaluationForm.tsx
devinleighsmith: In my memory there wasn't a resolution to that conversation (ie. a specific date) to lookback to (or even a specific number of years to lookback to.

frontend/src/pages/SubmitProperty.test.tsx
devinleighsmith: Yes, not planning on doing that for this review.

frontend/src/reducers/parcelsReducer.spec.tsx
devinleighsmith: it already had evaluations. more importantly this auto-merge doesn't make sense given a previous change, will remove this line.

frontend/src/utils/YupSchema.tsx
devinleighsmith: Currently none of the selects are explicitly validated in Yup (mostly because most of them require an API call to get the list of accepted values).

Specific to this field, I'd argue that it should be removed entirely. The reason being that yup schema validation failures prevent a form from being submitted, and this field isn't visible to the user. So yup should be used to enforce user facing validation - other validation needs to be handled by other mechanisms.

frontend/src/components/navigation/AppNavBar.tsx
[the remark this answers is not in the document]
devinleighsmith: sigh, right, I remember doing this, shouldn't have been committed.

frontend/src/forms/subforms/EvaluationForm.tsx
Fosol: This will work for us now.

frontend/src/components/common/FormikDatePicker.tsx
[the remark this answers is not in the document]
asanchezr: My only comment is on why this wasn't placed under the common form folder as it seems a common form element. Of course we can move later. That's not a blocker

PR #146, opened 2020-04-21
Misc fixes

asanchezr:
* Improve keycloak container builds
* Fix keycloak README
* Proper null checks

### DIFF
--- a/auth/keycloak/README.md
+++ b/auth/keycloak/README.md
@@ -60,7 +60,7 @@ $ docker exec -it keycloak bash
 $ /opt/jboss/keycloak/bin/standalone.sh \
   -Djboss.socket.binding.port-offset=100 \
   -Dkeycloak.migration.action=import \
-  -Dkeycloak.profile.feature.scripts=enabled
+  -Dkeycloak.profile.feature.scripts=enabled \
   -Dkeycloak.profile.feature.upload_scripts=enabled \
   -Dkeycloak.migration.provider=singleFile \
   -Dkeycloak.migration.file=/tmp/realm-export.json
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
     env_file:
       - auth/keycloak/.env
     volumes:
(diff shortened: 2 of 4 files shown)


auth/keycloak/README.md
```diff
@@ -60,7 +60,7 @@ $ docker exec -it keycloak bash
 $ /opt/jboss/keycloak/bin/standalone.sh \
   -Djboss.socket.binding.port-offset=100 \
   -Dkeycloak.migration.action=import \
-  -Dkeycloak.profile.feature.scripts=enabled
+  -Dkeycloak.profile.feature.scripts=enabled \
```
Fosol, line 63: Thanks for catching this

docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
     env_file:
       - auth/keycloak/.env
     volumes:
-      - ./auth/keycloak/config:/tmp
+      - ./auth/keycloak/config/realm-export.json:/tmp/realm-export.json
```
Fosol, line 27: Is it only the one file that needs to be copied?

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -53,7 +53,7 @@ function useKeycloakWrapper(): IKeycloak {
   const hasClaim = (claim?: string | Array<string>): boolean => {
     return (
       claim !== undefined &&
-      claim != null &&
+      claim !== null &&
```
Fosol, line 56: There is no actual value in changing this, but it doesn't hurt either.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -67,7 +67,7 @@ function useKeycloakWrapper(): IKeycloak {
   const hasRole = (role?: string | Array<string>): boolean => {
     return (
       role !== undefined &&
-      role != null &&
+      role !== null &&
```
Fosol, line 70: There is no actual value in changing this, but it doesn't hurt either.

PR #147, opened 2020-04-21
PIMS-1419 Navigation tests

Fosol:


frontend/src/components/navigation/AppNavBar.test.tsx
```diff
@@ -0,0 +1,325 @@
+import React from 'react';
```
Fosol, line 1: Is there a better way to organize tests?  This is getting to be a long file.

jdtoombs: Not that I know of, but you can group tests by functionality if you want. 
Ex) 
```
describe('nav bar link functionality', () => {
  it('related test 1', () => {   
  }
  it('related test 2', () => {   
  }
})
```

PR #148, opened 2020-04-21
PIMS-1419 Fixing navigation

Fosol:


frontend/src/App.tsx
```diff
@@ -73,9 +73,17 @@ const App = () => {
                       path="/access/request"
                       component={AccessRequestPage}
                     ></PrivateRoute>
-                    <PrivateRoute path="/mapView" component={MapView} />
-                    <PrivateRoute path="/submitProperty/:id?" component={SubmitProperty} />
-                    <PrivateRoute path="/edituser" component={EditUserPage} />
+                    <PrivateRoute path="/mapView" component={MapView} role={Claims.PROPERTY_VIEW} />
```
Fosol, line 76: Only allowed to view the map if the user has the claim property-view.

frontend/src/App.tsx
```diff
@@ -73,9 +73,17 @@ const App = () => {
...
+                    <PrivateRoute
+                      path="/submitProperty/:id?"
+                      component={SubmitProperty}
+                      claim={Claims.PROPERTY_ADD}
```
Fosol, line 80: Only allowed to add properties if user has the claim property-add.

frontend/src/App.tsx
```diff
@@ -73,9 +73,17 @@ const App = () => {
...
+                    />
+                    <PrivateRoute
+                      path="/admin/user"
+                      component={EditUserPage}
+                      claim={Claims.ADMIN_USERS}
```
Fosol, line 85: Only allowed to edit users if the user has the claim admin-users.

frontend/src/PrivateRoute.tsx
```diff
@@ -21,7 +21,11 @@ const PrivateRoute = (props: IPrivateRouteProps) => {
       {...rest}
       render={props => {
         if (!!keycloak.obj?.authenticated) {
-          if (keycloak.hasRole(rest.role) || keycloak.hasClaim(rest.claim)) {
+          if (
```
Fosol, line 24: This will allow us to use PrivateRoutes for authenticated users without any special role/claim.

frontend/src/App.tsx
devinleighsmith: shouldn't this be claim={Claims.PROPERTY_VIEW} ?

devinleighsmith: @Fosol Will approve after you fix this.

Fosol: Fixed

PR #149, opened 2020-04-21
PIMS-1419 Navigation tests

Fosol:
Something messed up with my branching so I've created a new PR and updated the organization of the tests.

Please note that a lot of tests are no longer working after merging.

### DIFF
--- a/frontend/src/components/navigation/AppNavBar.tsx
+++ b/frontend/src/components/navigation/AppNavBar.tsx
@@ -7,50 +7,11 @@ import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 import { Claims } from 'constants/claims';
 
 /**
- * Nav bar with with user-related functionality.
+ * Nav bar with with role-based functionality.
  */
 function AppNavBar() {
   const keycloak = useKeycloakWrapper();
   const history = useHistory();
-
-  const SubmitProperty = keycloak.hasClaim(Claims.PROPERTY_ADD) ? (
-    <NavDropdown.Item onClick={() => history.push('/submitProperty')}>
-      Submit Property
-    </NavDropdown.Item>
-  ) : null;
-
-  const PropertyDropdown = keycloak.hasClaim(Claims.PROPERTY_VIEW) ? (
-    <NavDropdown title="Manage Property" id="manage-property-dropdown">
-      {SubmitProperty}
-      <NavDropdown.Item>View Inventory</NavDropdown.Item>
-    </NavDropdown>
-  ) : null;
-
-  const AdminDropdown = keycloak.isAdmin ? (
-    <NavDropdown title="Administration" id="administration">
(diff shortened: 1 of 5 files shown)


frontend/src/components/navigation/AppNavBar.tsx
```diff
@@ -62,11 +58,11 @@ function AppNavBar() {
       <Navbar.Toggle aria-controls="collapse" className="navbar-dark" />
       <Nav>
         <Navbar.Collapse className="links">
-          {canAdmin}
-          {canEditProperties}
-          {canPropertyView}
-          {canDisposeRequest}
-          {canDisposeApprove}
+          {AdminDropdown}
```
asanchezr, line 61: As you are creating sub-conponents it would be a bit cleaner to use the JSX syntax here - `<AdminDropdown/>`

Fosol: Do we really want to make these sub-components?  They're essentially just logic to determine which nav elements are displayed.

To make them sub-components I'd be creating separate files in the components correct?

devinleighsmith: @Fosol see here: https://reactjs.org/docs/introducing-jsx.html

asanchezr: Another way is to turn the above into "render" functions. That's another React pattern I've seen around. E.g.

```jsx
const renderAdminDropdown = () => {
  if (you_have_permission) {
    return <NavLink ... />
  } else {
    return null;
  }
}
``` 
And then in the main JSX body use:

```diff
- {AdminDropdown}
+ { renderAdminDropdown() } <-- note the function call
```
Both approaches are acceptable - I personally like the JSX components better

Fosol: What's the best way to create these sub-components?  If I do the following,
```
  function SubmitPropertyNav() {
    return keycloak.hasClaim(Claims.PROPERTY_ADD) ? (
      <NavDropdown.Item onClick={() => history.push('/submitProperty')}>
        Submit Property
      </NavDropdown.Item>
    ) : null;
  }
```
Then it needs to belong in the parent `AppNavBar()` component, as it needs access to `keycloak` and `history` objects.

If add them to the function as arguments it apparently isn't ideal.  The recommendation is to add a `props` argument to the sub-components.  But then I think I still need to manually pass them.

frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -77,7 +77,7 @@ const store = mockStore({
   },
 });
 
-xit('SubmitProperty renders a disabled form if agency does not match current user.', () => {
+it('SubmitProperty renders a disabled form if agency does not match current user.', () => {
```
asanchezr, line 80: Did you fix this test? I believe it was disabled in a previous PR...

PR #150, opened 2020-04-21
PIMS-1401,1402,1464 add RAEG/SPP to buildings.

devinleighsmith:
Also misc lint and test fixes.


frontend/src/actions/parcelsActions.tsx
```diff
@@ -33,6 +33,7 @@ export interface IAddress {
 export interface IBuilding {
   id: number;
   localId: string;
+  projectNumber: string;
```
Fosol, line 36: It can be null

frontend/src/contexts/authStateContext.tsx
```diff
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useKeycloak } from '@react-keycloak/web';
```
Fosol, line 2: I was fairly certain when I removed this something broke.  I didn't understand why as it doesn't appear to be used in the file.

frontend/src/components/maps/BuildingPopupView.tsx
```diff
@@ -48,6 +48,12 @@ export const BuildingPopupView: React.FC<IBuildingDetailProps> = ({ building })
                 <Label>Predominate Use: </Label>
                 {building?.buildingPredominateUse}
               </ListGroup.Item>
+              {building.projectNumber && (
```
asanchezr, line 51: Maybe `building?. projectNumber` ? To match the surrounding code?

frontend/src/actions/parcelsActions.tsx
```diff
@@ -70,6 +71,7 @@ export interface IParcel {
   id: number | any;
   pid: string | undefined;
   pin: number | any;
+  projectNumber: string;
```
Fosol, line 74: Does this one need to be null too?

devinleighsmith: will add that to my fiscal changes.

PR #152, opened 2020-04-23
PIMS-1427 Adding property endpoints

Fosol:
This PR is to support the new Property List View.

**Important**

- Requires refresh DB Migration
- Updated dependencies in backend solution

**Bugs**

- Fixing paged endpoints
- Fixing financial conversion issue (float to decimal)

**Development**

- Updated Keycloak local development documentation.
- We now have our first database View
- API endpoints added; 
  - GET /api/properties/page
  - POST /api/properties/page/filter
  - GET /api/parcels/page
  - POST /api/parcels/page/filter
  - GET /api/buildings/page
  - POST /api/buildings/page/filter

**Testing**

- Unit tests have been improved
- Adding exclude attribute to remove classes from test coverage


backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -200,7 +200,7 @@ private Entity.City GetOrCreateCity(string name)
                 var code = new string(name.GetFirstLetterOfEachWord(true).Take(4).ToArray());
                 var last = new string(name.Split(" ", StringSplitOptions.RemoveEmptyEntries).Last().Take(4).ToArray());
                 if (last.Length >= 4) last = last.Substring(1, 3);
-                else if (last.Length < 4) last = last.Substring(1, last.Length - 1);
+                else if (last.Length < 4) last = last[1..];
```
devinleighsmith, line 203: whoah, that is some neat syntax

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -23,5 +27,105 @@ public static void ThrowIfNotUnique(this DbSet<Entity.Building> buildings, Entit
...
+                query = query.Where(b => b.BuildingConstructionTypeId == filter.ConstructionTypeId);
+            if (filter.PredominateUseId.HasValue)
+                query = query.Where(b => b.BuildingPredominateUseId == filter.PredominateUseId);
+            if (filter.FloorCount.HasValue)
+                query = query.Where(b => b.BuildingFloorCount == filter.FloorCount);
+            if (!String.IsNullOrWhiteSpace(filter.Tenancy))
+                query = query.Where(b => EF.Functions.Like(b.BuildingTenancy, $"%{filter.Tenancy}%"));
+
+            if (!String.IsNullOrWhiteSpace(filter.Address)) // TODO: Parse the address information by City, Postal, etc.
+                query = query.Where(b => EF.Functions.Like(b.Address.Address1, $"%{filter.Address}%") || EF.Functions.Like(b.Address.City.Name, $"%{filter.Address}%"));
```
devinleighsmith, line 84: Should this also be parsing the city name using the address query field?

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -37,5 +41,104 @@ public static void ThrowIfPidPinUpdated(this DbSet<Entity.Parcel> parcels, Entit
...
         }
+
+        /// <summary>
+        /// Generate a query for the specified 'filter'.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="user"></param>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public static IQueryable<Entity.Parcel> GenerateQuery(this PimsContext context, ClaimsPrincipal user, Entity.Models.ParcelFilter filter)
```
devinleighsmith, line 52: What do you think about trying to move/group some of these query components to common functions? for example lat/lon and evaluations are probably going to work the same between multiple different filters. Not required for this review, just thinking.

backend/dal/Migrations/Initial/PostDeploy/01-ViewProperties.sql
```diff
@@ -0,0 +1,112 @@
+-- This script provides a way to union both parcels and buildings into a single result of properties.
+CREATE VIEW View_Properties AS
```
devinleighsmith, line 2: ah, ok. When I saw the configuration I was confused but this makes sense now.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
Fosol: I did already start that process by pulling them into extension methods.  But yes I like going one step further (when I find time).

backend/dal/Helpers/Extensions/BuildingExtensions.cs
Fosol: I believe so, otherwise "Victoria" won't return much.

backend/api/Controllers/PropertyController.cs
```diff
@@ -88,5 +86,45 @@ public IActionResult GetProperties([FromBody]PropertyFilterModel filter)
...
+        [HasPermission(Permissions.PropertyView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(PageModel<PropertyModel>), 200)]
+        [SwaggerOperation(Tags = new[] { "property" })]
+        public IActionResult GetPropertiesPage([FromBody]PropertyFilterModel filter)
+        {
+            filter.ThrowBadRequestIfNull($"The request must include a filter.");
+            if (!filter.IsValid()) throw new BadRequestException("Property filter must contain valid values.");
+
+            var page = _pimsService.Property.Get((AllPropertyFilter)filter);
```
asanchezr, line 123: minor naming comment - shouldn't the above be named `service.Property.GetPage(...)` for consistency with the other "page" methods?

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+                query = query.Where(p => p.BuildingConstructionTypeId == filter.ConstructionTypeId);
+            if (filter.PredominateUseId.HasValue)
+                query = query.Where(p => p.BuildingPredominateUseId == filter.PredominateUseId);
+            if (filter.FloorCount.HasValue)
+                query = query.Where(p => p.BuildingFloorCount == filter.FloorCount);
+            if (!String.IsNullOrWhiteSpace(filter.Tenancy))
+                query = query.Where(p => EF.Functions.Like(p.BuildingTenancy, $"%{filter.Tenancy}%"));
+
+            if (!String.IsNullOrWhiteSpace(filter.Address)) // TODO: Parse the address information by City, Postal, etc.
+                query = query.Where(p => EF.Functions.Like(p.Address, $"%{filter.Address}%") || EF.Functions.Like(p.City, $"%{filter.Address}%"));
```
asanchezr, line 77: I'm not an expert on the backend so I might be wrong but shouldn't the above be:

`query = query.Where(p => EF.Functions.Like(p.Address.[Address1], $"%{filter.Address}%")`

_[Address1] added by me_

devinleighsmith: yeah, this is what I was getting to in my comment as well, I thought the address and city were discrete fields.

Fosol: I think you're right about the `Address.Address` comment.

backend/dal/Migrations/Initial/PostDeploy/01-ViewProperties.sql
asanchezr: I like this view. Would you consider refactoring `PropertiesController.GetProperties()` to use this view or is there anything that would prevent that? Doesn't have to be in the PR, something to think about.

Note that I'm referring to the original `GetProperties` not this new one that is paginated. The original one makes two calls to the database and concatenates the result. Seems like the view would probably offer better performance?

backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -0,0 +1,146 @@
...
+                    .Include(p => p.Address)
+                    .ThenInclude(a => a.Province);
+
+            var total = query.Count();
+            var items = query.Select(p => new Property(p)
+            {
+                Status = p.Status.Name,
+                Classification = p.Classification.Name,
+                AddressId = p.AddressId,
+                Address = $"{p.Address.Address1} {p.Address.Address2}",
```
asanchezr, line 55: This is quite minor but if `Address.Address2` is empty (which is quite common as most addresses only have one line) there would be an extra space at the end...

Maybe checking for not-empty before appending `Address2`?

backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -0,0 +1,146 @@
...
+            var total = query.Count();
+            var items = query.Select(b => new Property(b)
+            {
+                PID = b.Parcel.PID,
+                PIN = b.Parcel.PIN,
+                Status = b.Status.Name,
+                Classification = b.Classification.Name,
+                Agency = b.Agency.Name,
+                AddressId = b.AddressId,
+                Address = $"{b.Address.Address1} {b.Address.Address2}",
```
asanchezr, line 96: Same comment here about extra `space`

backend/dal/Services/IPropertyService.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Pims.Dal.Services
+{
+    /// <summary>
+    /// IPropertyService interface, provides functions to interact with properties within the datasource.
+    /// </summary>
+    public interface IPropertyService
+    {
+        Paged<Property> Get(AllPropertyFilter filter);
+        Paged<Property> Get(ParcelFilter filter);
+        Paged<Property> Get(BuildingFilter filter);
```
asanchezr, line 13: I think I mentioned before but what would you think of naming these `GetPage` to follow the convention from the other services?

backend/tests/unit/dal/Services/PropertyServiceTest.cs
```diff
@@ -0,0 +1,306 @@
...
+                new object[] { new AllPropertyFilter() { Description = "Description" }, 10 },
+                new object[] { new AllPropertyFilter() { Municipality = "Municipality" }, 1 },
+                new object[] { new AllPropertyFilter() { Tenancy = "BuildingTenancy" }, 1 },
+                new object[] { new AllPropertyFilter() { Zoning = "Zoning" }, 10 },
+                new object[] { new AllPropertyFilter() { ZoningPotential = "ZoningPotential" }, 10 },
+                new object[] { new AllPropertyFilter() { ConstructionTypeId = 2 }, 1 },
+                new object[] { new AllPropertyFilter() { PredominateUseId = 2 }, 1 },
+                new object[] { new AllPropertyFilter() { MinRentableArea = 100 }, 1 },
+                new object[] { new AllPropertyFilter() { MinRentableArea = 50, MaxRentableArea = 50 }, 1 },
+            };
```
asanchezr, line 66: I love all the unit tests covering the filtering functionality. 
Given that this PR is mostly about paginated requests - would you consider adding a couple tests that cover pagination? 

Something like -  "there's 10 properties, if I set `filter.quantity = 5` and `filter.page = 2`, I should see the last five?"

backend/dal/Helpers/Extensions/PropertyExtensions.cs
Fosol: The view flattens the values into Address.

backend/dal/Services/Concrete/PropertyService.cs
Fosol: I can't do the trim or the query will not run in the database regrettably.

backend/dal/Services/Concrete/PropertyService.cs
Fosol: I can't do the trim or the query will not run in the database regrettably.

backend/dal/Services/Concrete/PropertyService.cs
asanchezr: okay no big deal anyways

backend/dal/Helpers/Extensions/PropertyExtensions.cs
asanchezr: I see, given that both Devin and I were kind of confused by the change in behavior would you consider dropping a comment there for future reference? thanks

Fosol: I can add the comment in the next PR.  This specific query is against the new View (of which the property Address is flattened slightly).

backend/tests/unit/dal/Services/PropertyServiceTest.cs
Fosol: That is correct

PR #157, opened 2020-04-26
PIMS-1431

devinleighsmith:
Add Update button to go from view to update.
Major Performance improvements for large forms.
Manage state of parcel detail.
Add support for fiscals
Add positive feedback to validated form fields.
Validate evaluations/fiscals separately.
Rework css.
Add fiscal/evaluation summation component.
Correct clearing behaviour in date inputs.
Test Corrections.


frontend/src/components/maps/__snapshots__/MapFilterBar.test.tsx.snap
```diff
@@ -22,7 +22,7 @@ exports[`MapFilterBar renders correctly 1`] = `
             className="bar-item col"
           >
             <div
-              className="form-group"
+              className="  form-group"
```
Fosol, line 25: Why the empty space?

frontend/src/pages/SubmitProperty.tsx
```diff
@@ -29,27 +36,37 @@ const SubmitProperty = (props: any) => {
...
+    !keycloak.hasAgency(cachedParcelDetail?.agencyId)
   ) {
-    parsed.disabled = 'true'; //if the user doesn't belong to this properties agency, display a read only view.
+    setFormDisabled(true); //if the user doesn't belong to this properties agency, display a read only view.
   }
-  if (!keycloak.agencyId && !parsed.disabled) {
+  if (!keycloak.agencyId && !formDisabled) {
     throw Error('You must belong to an agency to submit properties');
   } else if (!keycloak.obj?.subject) {
     throw Error('Keycloak subject missing');
```
Fosol, line 53: This error probably isn’t helpful if visible to a user.

frontend/src/actions/parcelsActions.tsx
```diff
@@ -56,22 +56,22 @@ export interface IBuilding {
...
+  fiscalYear?: number | any;
   key: string | any;
   value: number | any;
 }
 
 export interface IEvaluation {
-  date: Date | any;
+  date?: Date | any;
   key: string | any;
   value: number | any;
```
asanchezr, line 67: I have seen a number of union types like this in the codebase and would like to know what are we trying to achieve with `number | any` or `Date | any` or `blah | any`? 

As stated [here](https://github.com/microsoft/TypeScript/issues/18568#issuecomment-330612009) - "a union with any will always be any."

Why would we want to have a "key" in `IEvaluation` that is any? shouldn't it always be a string? Please help me understand

frontend/src/utils/utils.tsx
```diff
@@ -0,0 +1,72 @@
+import { ILookupCode } from 'actions/lookupActions';
```
asanchezr, line 1: Not a biggie but ideally `*.tsx` should be reserved to components that output JSX markup. This file doesn't produce any JSX (yet) so I renamed to `*.ts` in a previous PR. I'm not against of everything being *.tsx - just want to ask the team what they prefer.

frontend/src/components/common/form/FastDatePicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+        selected={(value && new Date(value)) || null}
+        onBlur={handleBlur}
+        disabled={disabled}
+        {...rest}
+        onChange={(val: any) => {
+          setFieldValue(
+            field,
+            val
+              ? moment(val)
+                  .add('day', 1)
```
asanchezr, line 81: Could you please explain why we need to add an extra day here?

Also, please note that the `add` syntax above is [deprecated](https://momentjs.com/docs/#/manipulating/add/) - could you please use the current one? thanks
```js
moment().add('seconds', 1); // Deprecated in 2.8.0
moment().add(1, 'seconds');
```

frontend/src/components/common/form/FastInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+      registerField(field, { validate: undefined });
+      return () => {
+        unregisterField(field);
+      };
+    }, []);
+
+    return (
+      <Form.Group
+        controlId={`input-${field}`}
+        className={[!!required ? 'required' : '', outerClassName ?? ''].join(' ')}
```
asanchezr, line 76: The above logic has the unintended consequence to add a leading `space` to the **className** property when required is false -> `required blah` vs `[space] blah`. That's due to the use of `join(' ')`. 

We now see that effect in the changed snapshots as Jeremy pointed out.

Given that adding/removing CSS classes is a common concern of components as we build out frontend, would you consider using a library like [classnames](https://www.npmjs.com/package/classnames) ? It's is quite popular with over 5M downloads/week and there are plenty of resources/guides out there on how to use.

https://www.pluralsight.com/guides/applying-classes-conditionally-react

I've seen this in few places but will not add the same comment over and over. However, please fix or comment here on a solution suitable for all instances.

I'm okay if we push this change to a future PR as tech-debt

frontend/src/components/maps/__snapshots__/MapFilterBar.test.tsx.snap
asanchezr: I replied in a previous comment, hopefully it makes sense

frontend/src/forms/subforms/AddressForm.tsx
```diff
@@ -7,10 +7,11 @@ import { ILookupCode } from 'actions/lookupActions';
 import { ILookupCodeState } from 'reducers/lookupCodeReducer';
 import _ from 'lodash';
 import * as API from 'constants/API';
-import { Form, Input, Select } from 'components/common/form';
+import { Form, FastSelect } from 'components/common/form';
 import { mapLookupCode } from 'utils';
 import { Col } from 'react-bootstrap';
 import { IAddress } from 'actions/parcelsActions';
+import { FastInput } from 'components/common/form/FastInput';
```
asanchezr, line 14: /nit: please add he new "fast" components to the default exports in `/common/forms/index.ts` so it can be exported along with the others above on line 10

frontend/src/actions/parcelsActions.tsx
```diff
@@ -56,22 +56,22 @@ export interface IBuilding {
 }
 
 export interface IFiscal {
-  fiscalYear: number | any;
+  fiscalYear?: number | any;
```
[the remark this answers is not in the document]
devinleighsmith: I could actually make this required and use partials where necessary. Original motivation was my merging of fiscal and evaluation data for the financial component.

frontend/src/pages/SubmitProperty.tsx
devinleighsmith: I'm open to suggestions for a change in wording.

frontend/src/actions/parcelsActions.tsx
devinleighsmith: Didn't we talk about this? I suppose this specific case may be slightly different. Key in particular doesn't need this. However, the other fields that you mention above are used by formik, and therefore need to support their actual values as well as '' (React forms will 'untrack' input fields with values set to null or undefined).

I don't know the best way to represent that to be honest, but I preferred number | any over "any" as it shows the intended type. Could also go with number | string really.

frontend/src/utils/utils.tsx
devinleighsmith: I have no strong preferences, only did this to resolve some warnings I was seeing.

frontend/src/forms/subforms/AddressForm.tsx
devinleighsmith: I did add it there, but clearly didn't update all of these imports.

frontend/src/components/common/form/FastInput.tsx
devinleighsmith: sounds good to me.

frontend/src/components/common/form/FastDatePicker.tsx
devinleighsmith: I'll look into this, this may have been some code added during troubleshooting.

frontend/src/actions/parcelsActions.tsx
asanchezr: I don't recall that conversation but maybe I wasn't present or something. 

If the use case you are trying to support is `blah | <empty_string>` to avoid issues with forms, perhaps a better typing is `blah | string` (except for string which would be, well, a string). Like the date field above would be better typed as `Date | string` capturing the `''`` from empty fields and the Date value when it has a value and it gets parsed.

In any case, my opinion is that typescript should be there to help, not to slow us down. I would suggest we leave as-is for now and revisit after the POC demo, unless the typing `blah | string` is all we need and it is easy to change - I just don't want to cause unnecessary overhead at the moment

PR #158, opened 2020-04-27
PIMS-1417 Implement functionality to manage Pending Access Requests

goemen:
Used ag-grid library to show a list of pending access requests,


backend/api/Areas/Keycloak/Models/User/AccessRequestModel.cs
```diff
@@ -1,14 +1,14 @@
...
 
 namespace Pims.Api.Areas.Keycloak.Models.User
 {
     public class AccessRequestModel : Api.Models.BaseModel
     {
         #region Properties
         public int Id { get; set; }
-        public bool? IsGranted { get; set; }
-        public bool IsDisabled { get; set; }
+        public AccessRequestStatus? Status { get; set; }
```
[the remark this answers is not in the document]
goemen: Should never be null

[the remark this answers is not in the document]
Fosol: Then it shouldn't be nullable, change `AccessRequestStatus?` to `AccessRequestStatus`

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -180,7 +180,7 @@ public async Task<Entity.AccessRequest> UpdateAccessRequestAsync(Entity.AccessRe
 
             this._user.ThrowIfNotAuthorized(Permissions.AdminUsers, Permissions.AgencyAdmin);
             var accessRequest = _pimsAdminService.User.GetAccessRequest(entity.Id);
-            if (!accessRequest.IsGranted != true && entity.IsGranted == true)
+            if (accessRequest.Status != Entity.AccessRequestStatus.APPROVED && entity.Status == Entity.AccessRequestStatus.APPROVED)
```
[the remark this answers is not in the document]
goemen: This code will execute only if status from the database is not APPROVED and the next state is set to APPROVE (admin intends to approve the access request)

I can change this code so that it checks if the current status is ON_HOLD and the next state is APPROVED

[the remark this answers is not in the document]
Fosol: So the current logic allows them to go from Denied to Approved?  If so, does everything work correctly on the management page?  I think if it works, then we keep it as you have it.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -122,6 +122,24 @@ public AccessRequest GetAccessRequest(int id)
             if (accessRequest.UserId != userId) throw new NotAuthorizedException();
             return accessRequest;
         }
+        /// <summary>
+        /// Delete an access request
+        /// </summary>
+        /// <param name="id">The id of the request</param>
+        /// <returns> True = if deleted</returns>
+        public void DeleteAccessRequest(int id)
```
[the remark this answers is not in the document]
goemen: ++

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -34,10 +36,11 @@ public class UserController : ControllerBase
         /// </summary>
         /// <param name="keycloakService"></param>
         /// <param name="mapper"></param>
-        public UserController(IPimsKeycloakService keycloakService, IMapper mapper)
+        public UserController(IPimsKeycloakService keycloakService, IMapper mapper, IUserService userService)
```
devinleighsmith, line 39: @Fosol why are we seeing this warning 4 times?

backend/api/Areas/Keycloak/Controllers/UserController.cs
```diff
@@ -34,10 +36,11 @@ public class UserController : ControllerBase
...
         {
             _keycloakService = keycloakService;
             _mapper = mapper;
+            _userService = userService;
```
devinleighsmith, line 43: Am I missing something? it seems like you are adding userService to this controller but then not using it.

backend/api/Areas/Keycloak/Controllers/UserController.cs
Fosol: Currently we're seeing this warning during `npm test` as a `warning`.  I don't fully understand it.  It has something to do with the order of state added.

frontend/src/actionCreators/accessRequestActionCreator.tsx
```diff
@@ -89,12 +95,14 @@ export const getSubmitAccessRequestAction = (accessRequest: IAccessRequest) => (
 export const getSubmitAdminAccessRequestAction = (accessRequest: IAccessRequest) => (
   dispatch: Function,
 ) => {
-  dispatch(request(actionTypes.UPDATE_REQUEST_ACCESS_ADMIN));
```
devinleighsmith, line 92: removing this will prevent the isFetching state from being set on the network reducer. This isn't strictly required but is something that all of the other api actions are doing.

frontend/src/actionCreators/accessRequestActionCreator.tsx
```diff
@@ -114,16 +122,50 @@ export const getAccessRequestsAction = (params: API.IPaginateAccessRequests) =>
...
+  type: actionTypes.SELECT_REQUEST_ACCESS_ADMIN,
+  selections,
+});
+
+export const getAccessRequestsSortAction = (sorting: ISort[] = []): ISortAccessRequestsAction => ({
+  type: actionTypes.SORT_REQUEST_ACCESS_ADMIN,
+  sorting,
+});
+
+export const getAccessRequestsDeleteAction = (id: number) => (dispatch: Dispatch<any>) => {
```
devinleighsmith, line 157: same comment here about dispatching the action to set isFetching.

This will need updating anyways when you make the api change to start passing an access request during deletion.

frontend/src/constants/API.tsx
```diff
@@ -68,5 +66,6 @@ export const ACTIVATE_USER = () => `/auth/activate`; // get filtered properties
 export const REQUEST_ACCESS = (id?: number) => `/users/access/requests${id ? '/' + id : ''}`; //request access url.
 export const REQUEST_ACCESS_ADMIN = () => `/keycloak/users/access/request`; //request access admin url.
 export const REQUEST_ACCESS_LIST = (params: IPaginateAccessRequests) =>
-  `/admin/users/access/requests/?${queryString.stringify(params)}`; // get paged access requests
+  `/admin/access/request//?${queryString.stringify(params)}`; // get paged access requests
+export const REQUEST_ACCESS_DELETE = (id: number) => `/admin/access/request/${id}`; // get paged access requests
```
devinleighsmith, line 70: the comment here is incorrect.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -51,6 +51,7 @@ function useKeycloakWrapper(): IKeycloak {
    * @param claim - The name of the claim
    */
   const hasClaim = (claim?: string | Array<string>): boolean => {
+    userInfo.roles = keycloak?.realmAccess?.roles || [];
```
devinleighsmith, line 54: Hmm... what is the motivation for this? I think userInfo should contain roles already (via a mapping in keycloak). Am I correct @Fosol ?

goemen: My bad, I was only checking something and forgot to revert the change

frontend/src/pages/admin/access/ManageAccessRequests.tsx
```diff
@@ -0,0 +1,138 @@
...
+      });
+    }
+  }, [gridApi]);
+
+  const updateSelectionList = () => {
+    setSelectedRequests(gridApi?.getSelectedNodes().map(x => x.data) || []);
+    dispatch(getAccessRequestsSelectAction(gridApi?.getSelectedNodes().map(x => x.id) || []));
+  };
+
+  const setFilterModel = () => {
```
devinleighsmith, line 85: Is there an advantage to storing the filter/sort/selections in redux instead of local component state?

My first thought is that it would allow external components to modify the filter/sort/selections of this form directly, is that something that we want to support?

I don't see a major disadvantage to doing this besides performance, so I don't mind keeping this in even if these values will only ever be tracked by this component.

frontend/src/pages/admin/access/components/AccessRequestActions.tsx
```diff
@@ -0,0 +1,55 @@
+import { Dropdown } from 'react-bootstrap';
```
devinleighsmith, line 1: I like what you did here, this is neat.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -170,6 +170,8 @@ public AccessRequest UpdateAccessRequest(AccessRequest entity)
             entity.UpdatedById = this.User.GetUserId(); // TODO: No longer needed.
             entity.UpdatedOn = DateTime.UtcNow;
             this.Context.Entry(accessRequest).CurrentValues.SetValues(entity);
+            accessRequest.UserId = accessRequest.User.Id;
+            Context.Entry(accessRequest).State = EntityState.Modified;
```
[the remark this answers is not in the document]
goemen: It is, without the commit does not update the database

frontend/src/pages/admin/access/constants/constants.ts
```diff
@@ -0,0 +1,58 @@
+import { ColDef } from 'ag-grid-community';
+import { IAccessRequest } from 'interfaces';
+import { AccessStatusDisplay, AccessStatusDisplayMapper } from 'constants/accessStatus';
+
+export const columnDefinitions = (): ColDef[] => {
+  let items: ColDef[] = [
```
devinleighsmith, line 6: does it make sense to shield params.data from NPE's here? I think the component itself is making sure not to display the grid if params.data isn't populated. Perhaps just documenting that params.data is required is good enough.

backend/api/Areas/Keycloak/Controllers/UserController.cs
goemen: you are right, should be removed

frontend/src/pages/admin/access/ManageAccessRequests.tsx
goemen: When the component re-renders after updating the status of an access request, the table loses its state. I moved the table state to redux to maintain the table state when re-rendering

frontend/src/pages/admin/access/constants/constants.ts
goemen: You are right about that, no need to shield it

frontend/src/pages/admin/access/ManageAccessRequests.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  useEffect(() => {
+    if (gridApi) {
+      gridApi.setFilterModel(gridFilterModel || {});
+      gridApi.setSortModel(gridSortModel || []);
+      gridRowSelections.forEach(x => {
+        gridApi.getRowNode(x)?.setSelectedInitialValue(true);
+      });
+    }
+  }, [gridApi]);
```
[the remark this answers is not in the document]
goemen: Now I think about it, this is not necessary. If we were passing the params from the url

backend/api/Areas/Admin/Controllers/AccessRequestController.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <summary>
+        /// Delete an access requests
+        /// </summary>
+        /// <param name="id"></param>
+        [HttpDelete("{id}")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.AccessRequestModel), 200)]
+        [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "admin-access-requests" })]
+        public IActionResult Delete(int id)
```
Fosol, line 76: You should pass the AccessRequest you want to delete to be able to perform concurrency checking.  Include both the id and accessRequest as arguments.

goemen: I am not sure if this is the right place for that. It is recommended that the HttpDelete method should not have a body but instead pass the id of the object

Fosol: HTTP DELETE Methods generally do not have bodies.  However they can.  If you don't pass the `RowVersion` then you can't check concurrency.  Which is bad.

There are workarounds, which is you could pass the `RowVersion` as a query string parameter.  However then you will run into issues with encoding because `RowVersion` can contain invalid  URL characters.  Handling a `RowVersion` in the query string is extremely fragile and difficult to support all scenarios.  Additionally the issues only shows periodically and is very hard to debug if you don't know what you're looking for.

The solution is to use the HTTP DELETE with a body of the object you want to delete.  Just like if you wanted to update an object, you would pass it in the body.  It's a tragedy that deleting objects wasn't thought through fully when they described the HTTP DELETE method.

asanchezr: @goemen we kinda had this discussion before you joined the team. I agree with you on usage of HTTP DELETE best practices but also understand the need to check for concurrency as stated by @Fosol 

As where to put the `RowVersion` an acceptable place would be a custom header `X-RowVersion: <guid>`. However, that would require extra work on the server and we are now on the critical path to our PoC.

Maybe we could revisit this after the demo? For now I suggest we leave as-is

goemen: I understand that. The other thing is that we are using Axios for making requests from the frontend and Axios delete does not have a body arg.

backend/api/Areas/Admin/Models/User/AccessRequestUserModel.cs
```diff
@@ -43,6 +43,12 @@ public class AccessRequestUserModel : Pims.Api.Models.BaseModel
         /// </summary>
         /// <value></value>
         public string Email { get; set; }
+        
+        /// <summary>
+        /// get/set - The username.
+        /// </summary>
+        /// <value></value>
+        public string UserName { get; set; }
```
asanchezr, line 51: This comes down to preference but what I see all around the web is "username" in JSON, which would map to "Username" here. That would also simplify the mapping above where the Entity field is named "Username" too

backend/api/Areas/Keycloak/Mapping/User/UserMap.cs
```diff
@@ -23,6 +26,15 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Roles, src => src.Roles)
                 .Inherits<Entity.BaseEntity, Api.Models.BaseModel>();
 
+            config.NewConfig<Entity.User, KModel.UserModel>()
+                .Map(dest => dest.Id, src => src.Id)
+                .Map(dest => dest.Username, src => src.Username)
```
asanchezr, line 31: Here we map to "Username" but in the other mapper you mapped to another model with "UserName" (upper-case N)
I'd prefer some consistent naming across all of out models if possible. Would you be open to rename the other model to match what's already in place?

Fosol: It should be `Username`, as it is not the User's Name, but their unique identifier Username

backend/api/Areas/Admin/Controllers/AccessRequestController.cs
Fosol: Don't use their default `delete` function.  Use their `request` function, set the method = 'delete` and add a `body`.

asanchezr:
@goemen Now the unit tests are failing. I believe @Fosol fixed those on a recent PR.
Perhaps you need to rebase your branch to get the latest changes into this PR?

frontend/.gitignore
```diff
@@ -22,3 +22,4 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+src/pages/__snapshots__
```
Fosol, line 25: I thought we wanted snapshots in source?

asanchezr: yeah, we want them in git, please revert this line

frontend/package.json
```diff
@@ -69,6 +71,8 @@
     "eslint-config-prettier": "6.7.0",
     "eslint-plugin-prettier": "3.1.1",
     "husky": "4.2.3",
+    "jest-environment-jsdom-fourteen": "^1.0.1",
+    "lint-staged": "10.0.8",
```
asanchezr, line 75: I removed `lint-staged` in a previous PR, maybe this got in here as a merge conflict? In any case, it should be removed

frontend/package.json
```diff
@@ -69,6 +71,8 @@
     "eslint-config-prettier": "6.7.0",
     "eslint-plugin-prettier": "3.1.1",
     "husky": "4.2.3",
+    "jest-environment-jsdom-fourteen": "^1.0.1",
```
asanchezr, line 74: Could you please comment why we need to include this? I checked the version of `react-scripts` we are using and it bundles this dependency so it would seem we shouldn't need to include it here as well. Unless there's something else?

goemen: testing with ag-grid requires this library to run

asanchezr: As stated [here](https://github.com/facebook/create-react-app/issues/7491#issuecomment-520157683) all you would need to do to enable `jest-environment-jsdom-fourteen` is:

```
//package.json
"test": "react-scripts test --env=jsdom-fourteen"
```

I see you did that below. Could you please try removing the dependency from our **package.json** and see if it works with the test? If possible I'd try to avoid bringing extra dependencies. Thanks!

goemen: You, sir, are right, the dependency is not needed

backend/dal/Services/IUserService.cs
```diff
@@ -14,6 +14,7 @@ public interface IUserService
         IEnumerable<int> GetAgencies(Guid userId);
         AccessRequest GetAccessRequest();
         AccessRequest GetAccessRequest(int id);
+        AccessRequest DeleteAccessRequest(AccessRequest accessRequest);
```
Fosol, line 17: Technically the service doesn't need to return the object deleted, only the controller does.  We'll add it to tech debt.

PR #159, opened 2020-04-27
PIMS-1625, and adding asserts for fiscals/evaluations

jdtoombs:
Just wanted to throw a draft up for now to review my 'fix' for the Update/View bug on building markers. Let me know if there is a different approach I should be taking, thanks!

### DIFF
--- a/backend/api/Areas/Admin/Models/Parcel/ParcelModel.cs
+++ b/backend/api/Areas/Admin/Models/Parcel/ParcelModel.cs
@@ -6,7 +6,7 @@ namespace Pims.Api.Areas.Admin.Models.Parcel
     {
         #region Properties
         public string PID { get; set; }
-
+        
         public int? PIN { get; set; }
 
         public float LandArea { get; set; }
--- a/frontend/src/components/maps/BuildingPopupView.tsx
+++ b/frontend/src/components/maps/BuildingPopupView.tsx
@@ -6,15 +6,21 @@ import { IBuilding } from 'actions/parcelsActions';
 import { Alert, Row, Col } from 'react-bootstrap';
 import { Label } from 'components/common/Label';
 import { EvaluationKeys } from '../../constants/evaluationKeys';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import { Link } from 'react-router-dom';
 
 export interface IBuildingDetailProps {
   building: IBuilding | null;
+  disabled?: boolean;
 }
 
-export const BuildingPopupView: React.FC<IBuildingDetailProps> = ({ building }) => {
+export const BuildingPopupView: React.FC<IBuildingDetailProps> = (props: IBuildingDetailProps) => {
+  const keycloak = useKeycloakWrapper();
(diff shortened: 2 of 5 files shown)


backend/api/Areas/Admin/Models/Parcel/ParcelModel.cs
```diff
@@ -6,7 +6,7 @@ public class ParcelModel : PropertyModel
     {
         #region Properties
         public string PID { get; set; }
-
+
```
Fosol, line 9: Adding spaces?

frontend/src/components/maps/BuildingPopupView.tsx
```diff
@@ -31,31 +37,39 @@ export const BuildingPopupView: React.FC<IBuildingDetailProps> = ({ building })
...
-                  {building?.projectNumber}
+                  {buildingDetail?.projectNumber}
                 </ListGroup.Item>
               )}
             </ListGroup>
           </Col>
+          <Col>
+            {!props?.disabled &&
+              (!keycloak.hasAgency(buildingDetail?.agencyId) ? (
+                <Link to={`/submitProperty/${buildingDetail?.parcelId}?disabled=true`}>View</Link>
```
Fosol, line 68: I'm curious about the syntax here.  Are we changing the url to include `?disabled=true`?  If so, that probably isn't the most ideal way to control the view vs edit mode.

Also please change the route from `/submitProperty` to `/property`.

jdtoombs: Not sure which approach we want to take, was just keeping this popup consistent with the ParcelPopupView. But will do re the route!

backend/api/Areas/Admin/Models/Parcel/ParcelModel.cs
jdtoombs: I deleted a line by accident then reverted it back to the original spacing

frontend/src/components/maps/BuildingPopupView.tsx
asanchezr: I've seen this and the other popup where "disabled" means "view" - I'd like to suggest that we use "mode" instead. What does the team think?

`mode = "view"` -> view-mode form
`mode = "edit"` -> edit-mode

this would support future modes (if they become a requirement)

`mode = "admin-mode-that-can-impersonate-users"` -> admin-mode form

jdtoombs: I like it

asanchezr: I'm okay with the URL having `?mode=edit` because I would expect the component (or route or somewhere) would check that the user has the appropriate permission to display the form. If user cannot edit, then I would show the view mode. If user cannot even view properties (we have a permission/claim for that as well), then it would go to the "permission-denied" page

Fosol: I would prefer `mode` as well.

PR #160, opened 2020-04-27
PIMS-1551 Adding GitHub Actions to frontend

Fosol:
Ignoring unit-tests that fail.  We will need to resolve these in future PRs.

Please note that once this is merged all future PRs will run both frontend and backend tests automatically.  If they fail, they should block the PR from being merged.

There are a number of issues with our current unit tests;
- Not mocking axios
- Date not being mocked for snapshots
- Possibly invalid use of `useEffect(...)`
- React `act(...)` complexity


.github/workflows/api-dotnetcore.yml
```diff
@@ -2,48 +2,46 @@ name: API (.NET Core)
...
+          dotnet-version: 3.1.101
+      - name: Install dependencies
+        run: dotnet restore
+        working-directory: ${{env.working-directory}}
+      - name: Build
+        run: dotnet build --configuration Release --no-restore
+        working-directory: ${{env.working-directory}}
+      - name: Test
+        run: dotnet test --no-restore --verbosity normal
+        working-directory: ${{env.working-directory}}
```
asanchezr, line 47: I'd be tempted to bundle "dotnet test" with the build action above as we are building the API twice. Just a suggestion, the actions seems to run fast anyways

backend/docs/GITHUB-ACTIONS.md
```diff
@@ -0,0 +1,59 @@
+# GitHub Actions
+
+GitHub Actions provide a CI/CD workflow abilities configured through YAML that will allow us to build and test our code before a Pull Request is created and merged.
```
asanchezr, line 3: /nit: the actions run after a PR is created but before they get merged

frontend/src/components/common/FormikDatePicker.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { FunctionComponent } from 'react';
```
asanchezr, line 1: @Fosol did you intend to include this file in your PR? 

@devinleighsmith - didn't you renamed/moved this file?

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -86,6 +86,7 @@ const Map: React.FC<MapProps> = ({
   const lastZoom = useSelector<RootState, number>(state => state.mapViewZoom) ?? zoom;
   useEffect(() => {
     dispatch(resetMapViewZoom());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
asanchezr, line 89: A tech-debt suggestion for later: I remember excluding this rule in our `package.json` to avoid having to add it everywhere explicitly. I followed `create-react-app` documentation but something must not be working. You can see the rule turned off on https://github.com/bcgov/PIMS/blob/dev/frontend/package.json#L92

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -178,48 +208,39 @@ describe('MapProperties View', () => {
         </BrowserRouter>
       </Provider>,
     );
+    await act(() => (promise as unknown) as Promise<void>);
```
asanchezr, line 211: the typing above `(promise as unknown) as Promise<void>)` seems a bit weird

What's wrong with `promise as Promise<void>` ?

.github/workflows/api-dotnetcore.yml
Fosol: I think you’re right.  I separated them originally so we could more clearly identify what failed but I don’t think that is necessary

backend/docs/GITHUB-ACTIONS.md
Fosol: I will have to look into seeing if we can pray to rocket chat on a failure at some point

frontend/src/components/common/FormikDatePicker.tsx
Fosol: I think this was a merger mistake.  I’ll double check

frontend/src/components/maps/leaflet/Map.tsx
Fosol: I worry we’re doing something wrong because the warming is everywhere.

frontend/src/components/maps/leaflet/MapProperties.test.tsx
Fosol: Typescript is angry with this format.  So that is the fix that I found

frontend/src/components/common/FormikDatePicker.tsx
devinleighsmith: Yes, this looks like the old version.

frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -199,21 +214,7 @@ describe('ParcelDetailForm', () => {
...
-            <ParcelDetailForm secret="test" agencyId={1} parcelDetail={mockDetails[0]} />
-          </Router>
-        </Provider>,
-      )
-      .toJSON();
-    expect(tree).toMatchSnapshot();
-  });
-
-  it('ParcelDetailForm renders view-only correctly', () => {
+  xit('ParcelDetailForm renders view-only correctly', () => {
```
devinleighsmith, line 217: you're ignoring this test here but you added a non-ignored copy of this test above.

devinleighsmith, line 217: it seems odd that the view only doesn't work but the editable version does above.

frontend/src/components/maps/leaflet/MapProperties.test.tsx
asanchezr: sure, I'm just flagging thing that seem odd. to be addressed later. The type `unknown` is quite new to typescript and I don't know the difference with `any` but it feels that `as any` would be better suited here to force the typing. In any case, this works so leave as-is.

frontend/src/components/maps/leaflet/Map.tsx
asanchezr: yeah, something might be wrong somewhere. I doubt we will have the time to debug this given all we want to accomplish. maybe next sprint

frontend/src/forms/ParcelDetailForm.test.tsx
Fosol: The view-only version throws an error with some axios request it would appear.
```js
TypeError: Cannot read property 'body' of null

      66 |       dispatch(hideLoading());
      67 |     })
    > 68 |     .catch(() => dispatch(error(reducerTypes.PUT_USER_DETAIL)))
         |                  ^
      69 |     .finally(() => dispatch(hideLoading()));
      70 | };
```

PR #164, opened 2020-04-28
Change marker colour for building/properties

jdtoombs:
Quick little change for the map UI

- properties are marked with a **green** pin

- buildings are marked with a **blue** pin


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -78,6 +78,27 @@ const Map: React.FC<MapProps> = ({
   const [baseLayers, setBaseLayers] = useState<BaseLayer[]>([]);
   const [activeBasemap, setActiveBasemap] = useState<BaseLayer | null>(null);
 
+  // different markers for building and parcel
```
devinleighsmith, line 81: Can we manually copy these and include them with our build instead of pulling them from the web?

asanchezr: +1

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -165,13 +186,27 @@ const Map: React.FC<MapProps> = ({
...
+      return (
+        <Marker
+          key={generateKey(p)}
+          position={[p.latitude, p.longitude]}
+          onClick={(e: any) => onMarkerClick?.(p)}
+          icon={blueIcon}
+        />
+      );
+    }
+  };
```
asanchezr, line 209: I was going to approve as-is as I'd love to see this change merged sooner than later.
One minor observation: I'd prefer to see a simpler logic where we don't duplicate the Marker logic in the `if/then/else`. A simpler approach would lead to less opportunities for future bugs where one branch of the if/then/else is modified but not the other...
Something like this:

```tsx
const icon = p.propertyTypeId === 0 ? greenIcon : blueIcon;
<Marker
    key={generateKey(p)}
    position={[p.latitude, p.longitude]}
    onClick={(e: any) => onMarkerClick?.(p)}
    icon={icon}
/>

```

jdtoombs: I like this! Will adjust

frontend/src/components/maps/leaflet/Map.tsx
jdtoombs: I agree with this, tried using local paths but currently could not get it working. Will revisit after I get this bug figured out for view/update of buildings

asanchezr: did you try using `require` with the icon-url? Agree we can solve this later, but would expect this work...

```tsx
var greenIcon = new Icon({
    iconUrl: require('assets/images/green.svg'),
   ...
});
```

_Assuming you place the icons under `/src/assets/images`_

jdtoombs: Require was the trick, updating again

PR #166, opened 2020-04-28
PIMS-1318

devinleighsmith:
Add modal popups when loading/saving parcel detail form drafts.

### DIFF
--- a/frontend/src/OnLoadActions.tsx
+++ b/frontend/src/OnLoadActions.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import GenericModal from 'components/common/GenericModal';
+import { useHistory } from 'react-router-dom';
+import { PARCEL_STORAGE_NAME, clearStorage, isStorageInUse } from 'utils/storageUtils';
+
+const OnLoadActions: React.FC = () => {
+  const history = useHistory();
+
+  return (
+    <GenericModal
+      title="Unsaved Draft"
+      message="You have an unsaved Property Draft, would you like to resume editing it?"
+      cancelButtonText="Discard"
+      okButtonText="Resume Editing"
+      display={
+        history.location.pathname !== '/submitProperty' && isStorageInUse(PARCEL_STORAGE_NAME)
+      }
+      handleOk={() => {
+        history.push('/submitProperty?loadDraft=true');
+      }}
+      handleCancel={() => {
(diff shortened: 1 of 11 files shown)


frontend/src/OnLoadActions.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const history = useHistory();
+
+  return (
+    <GenericModal
+      title="Unsaved Draft"
+      message="You have an unsaved Property Draft, would you like to resume editing it?"
+      cancelButtonText="Discard"
+      okButtonText="Resume Editing"
+      display={
+        history.location.pathname !== '/submitProperty' && isStorageInUse(PARCEL_STORAGE_NAME)
```
asanchezr, line 16: quick check - weren't we going to rename `/submitProperty` to `/property` ? or that's in another PR?

Fosol: I asked Jonny to do this originally.

asanchezr:
This branch is showing as out-of-date in GH. Could you please rebase? Thanks

frontend/src/components/common/FormikPersist.tsx
```diff
@@ -11,59 +13,119 @@ export interface PersistProps {
...
   writeOnly?: boolean;
+  loadDraft?: boolean;
+}
+interface PersistState {
+  showLoadDraftDialog: boolean;
+}
+
+// Do not allow stored data to be loaded if the application version does not match.
+interface VersionedStorage {
+  version: string;
```
Fosol, line 24: How do we set this value?

frontend/src/OnLoadActions.tsx
asanchezr: okay

frontend/src/components/common/FormikPersist.tsx
devinleighsmith: Currently I'm using the value set in package.json
All this means is that whenever we bump the version in a frontend release people will lose their drafts.

asanchezr: I like this. Would love to see a comment somewhere for reference when we need it.

devinleighsmith: Fair, will add in my next PR with the building updates.

PR #170, opened 2020-04-29
PIMS-1342 Property Detail form validation.

devinleighsmith:
Contains minor "low hanging fruit" fixes either requested by Michaela or differing from the UI design. There are some outstanding larger fixes that I'll tackle later, time permitting:
1) building classification (requires DB update)
2) parent/child agency display on form (requires API update).

### DIFF
--- a/frontend/src/forms/ParcelDetailForm.tsx
+++ b/frontend/src/forms/ParcelDetailForm.tsx
@@ -100,11 +100,12 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
   };
   //convert all form values to the format accepted by the API.
   const valuesToApiFormat = (values: IFormParcel): IFormParcel => {
-    values.pin = decimalOrEmpty(values.pin);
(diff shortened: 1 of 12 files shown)


frontend/src/forms/ParcelDetailForm.tsx
```diff
@@ -208,20 +206,19 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
...
-                  </Form.Row>
                 </Col>
               </Row>
               <Row noGutters>
                 <Col>
-                  <h3>Land</h3>
                   {<LandForm {...formikProps} disabled={props.disabled}></LandForm>}
+                  <Form.Row className="sumFinancialsForm">
+                    <SumFinancialsForm {...formikProps} />
+                  </Form.Row>
```
asanchezr, line 221: Something to think about, no need to change. Do you know you could use the formik context in your subforms and not require the props to be passed down? That's how I implemented the common form elements if you want to take a look.
Unless there's some gotcha that I'm unaware.

PR #171, opened 2020-04-29
PIMS-1684 Creating Tool to Sync Keycloak with PIMS

Fosol:
## Description

Presently DEV has authorization issues due to Keycloak being out-of-sync with PIMS.  This is blocking a lot of functional testing and is required for a successful PoC.

This tool provides a way to sync Keycloak with PIMS.

## Development

Please note that this requires another DB refresh.

### DIFF
--- a/backend/api/Areas/Admin/Mapping/User/UserMap.cs
+++ b/backend/api/Areas/Admin/Mapping/User/UserMap.cs
@@ -1,16 +1,16 @@
 using Mapster;
 using Model = Pims.Api.Areas.Admin.Models.User;
 using Entity = Pims.Dal.Entities;
+using System.Linq;
+using Pims.Core.Extensions;
 
 namespace Pims.Api.Areas.Admin.Mapping.User
 {
     public class UserMap : IRegister
     {
-
         public void Register(TypeAdapterConfig config)
         {
             config.NewConfig<Entity.User, Model.UserModel>()
-                .IgnoreNonMapped(true)
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.IsDisabled, src => src.IsDisabled)
                 .Map(dest => dest.Username, src => src.Username)
@@ -21,13 +21,11 @@ namespace Pims.Api.Areas.Admin.Mapping.User
                 .Map(dest => dest.LastName, src => src.LastName)
                 .Map(dest => dest.Email, src => src.Email)
                 .Map(dest => dest.Note, src => src.Note)
-                .Map(dest => dest.Agencies, src => src.Agencies)
-                .Map(dest => dest.Roles, src => src.Roles)
+                .Map(dest => dest.Agencies, src => src.Agencies.Select(a => a.Agency))
+                .Map(dest => dest.Roles, src => src.Roles.Select(a => a.Role))
                 .Inherits<Entity.BaseEntity, Api.Models.BaseModel>();
 
-
             config.NewConfig<Model.UserModel, Entity.User>()
-                .IgnoreNonMapped(true)
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.IsDisabled, src => src.IsDisabled)
                 .Map(dest => dest.Username, src => src.Username)
@@ -40,7 +38,15 @@ namespace Pims.Api.Areas.Admin.Mapping.User
                 .Map(dest => dest.Note, src => src.Note)
                 .Map(dest => dest.Agencies, src => src.Agencies)
                 .Map(dest => dest.Roles, src => src.Roles)
+                .AfterMappingInline((m, e) => UpdateUser(m, e))
                 .Inherits<Api.Models.BaseModel, Entity.BaseEntity>();
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "Required for signature")]
+        private void UpdateUser(Model.UserModel model, Entity.User entity)
+        {
+            entity.Agencies.ForEach(a => a.UserId = entity.Id);
+            entity.Roles.ForEach(r => r.UserId = entity.Id);
         }
     }
 }
--- a/backend/api/Startup.cs
+++ b/backend/api/Startup.cs
@@ -36,6 +36,7 @@ using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using Mapster;
 using Pims.Api.Helpers.Mapping;
+using System.Collections.Generic;
 
 namespace Pims.Api
 {
@@ -82,6 +83,12 @@ namespace Pims.Api
             {
                 options.Default.IgnoreNonMapped(true);
                 options.Default.IgnoreNullValues(true);
+                options.AllowImplicitDestinationInheritance = true;
+                options.AllowImplicitSourceInheritance = true;
+                options.Default.UseDestinationValue(member =>
+                    member.SetterModifier == AccessModifier.None &&
+                    member.Type.IsGenericType &&
+                    member.Type.GetGenericTypeDefinition() == typeof(ICollection<>));
             });
             services.Configure<Keycloak.Configuration.KeycloakOptions>(this.Configuration.GetSection("Keycloak"));
             services.Configure<Pims.Dal.PimsOptions>(this.Configuration.GetSection("Pims"));
--- a/backend/dal/Services/Admin/Concrete/ClaimService.cs
+++ b/backend/dal/Services/Admin/Concrete/ClaimService.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Security.Claims;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Logging;
 using Pims.Core.Extensions;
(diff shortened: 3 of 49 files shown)


backend/api/Areas/Admin/Mapping/User/UserMap.cs
```diff
@@ -40,7 +38,15 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Note, src => src.Note)
                 .Map(dest => dest.Agencies, src => src.Agencies)
                 .Map(dest => dest.Roles, src => src.Roles)
+                .AfterMappingInline((m, e) => UpdateUser(m, e))
```
devinleighsmith, line 41: I haven't seen this before, what is this doing?

Fosol: It allows me to set the UserId correctly.  Essentially without it the child objects in a collection don't have access to the parent object.

backend/dal/Services/Admin/Concrete/ClaimService.cs
```diff
@@ -0,0 +1,134 @@
...
+            var claims = this.Context.Claims
+                .Include(r => r.Roles)
+                .Where(r => !exclude.Contains(r.Id));
+            claims.ForEach(r =>
+            {
+                r.Roles.Clear();
+            });
+
+            this.Context.Claims.RemoveRange(claims);
+            var result = this.Context.CommitTransaction();
```
devinleighsmith, line 129: I would appreciate this being typed to an int for clarity instead of a var.

backend/api/Startup.cs
```diff
@@ -82,6 +83,12 @@ public void ConfigureServices(IServiceCollection services)
             {
                 options.Default.IgnoreNonMapped(true);
                 options.Default.IgnoreNullValues(true);
+                options.AllowImplicitDestinationInheritance = true;
+                options.AllowImplicitSourceInheritance = true;
+                options.Default.UseDestinationValue(member =>
```
Fosol, line 88: This allows us to access and update collections without a setter.  @devinleighsmith you won't need to have a `private set;` anymore.

backend/entities/Claim.cs
```diff
@@ -21,6 +21,11 @@ public class Claim : BaseEntity
         /// <value></value>
         public string Name { get; set; }
 
+        /// <summary>
+        /// get/set - The GUID that identifies this Role in Keycloak.
+        /// </summary>
+        public Guid? KeycloakRoleId { get; set; }
```
Fosol, line 27: Using a subordinate property to link with keycloak to free us from some of the issues that integration with keycloak was causing.

backend/entities/Role.cs
```diff
@@ -9,6 +9,11 @@ namespace Pims.Dal.Entities
     public class Role : LookupEntity<Guid>
     {
         #region Properties
+        /// <summary>
+        /// get/set - The GUID that identifies this Group in Keycloak.
+        /// </summary>
+        public Guid? KeycloakGroupId { get; set; }
```
Fosol, line 15: Using a subordinate property to link with keycloak to free us from some of the issues that integration with keycloak was causing.

backend/tools/keycloak/sync/Exceptions/HttpResponseException.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        #region Constructors
+        public HttpResponseException(HttpResponseMessage response, string message = null) : base(message)
+        {
+            this.StatusCode = response.StatusCode;
+            using var stream = response.Content.ReadAsStreamAsync().Result;
+            if (response.Content.Headers.ContentType == new MediaTypeHeaderValue("application/json"))
+            {
+                var results = JsonSerializer.DeserializeAsync<object>(stream).Result;
+                this.Details = JsonSerializer.Serialize(results); // TODO: Not ideal to return JSON as the error.
```
devinleighsmith, line 25: is it weird that we are deserializing the json stream and then serializing it again right afterwords?

backend/dal/Services/Admin/Concrete/ClaimService.cs
Fosol: Not sure what you mean here.  `var` is just a syntax feature that allows us to just inherit the type of the right.  Which is an int.

backend/tools/keycloak/sync/Models/ClaimModel.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+namespace Pims.Tools.Keycloak.Sync.Models
+{
+
```
devinleighsmith, line 5: /nit this one is missing a comment.

backend/tools/keycloak/sync/Exceptions/HttpResponseException.cs
Fosol: It is weird.  It was a quick hack for me to copy paste some code to get the string value out of the steam.  It should just read the stream though.

Fosol: A now I remember what this originated from.  It used to deserialize the JSON object to extract a value from it.  I'll change it to clean it up.

backend/tools/keycloak/sync/Models/ClaimModel.cs
Fosol: They are inconsistent...  I'm coding like a mad man atm...

backend/tools/keycloak/sync/Factory.cs
```diff
@@ -0,0 +1,386 @@
...
+using Pims.Tools.Keycloak.Sync.Extensions;
+using Pims.Tools.Keycloak.Sync.Models;
+using KModel = Pims.Tools.Keycloak.Sync.Models.Keycloak;
+
+namespace Pims.Tools.Keycloak.Sync
+{
+    /// <summary>
+    /// Factory class, provides a way to sync Keycloak roles, groups and users with PIMS.
+    /// </summary>
+    public class Factory : IFactory
```
devinleighsmith, line 22: Could the naming here be more informative?

backend/tools/keycloak/sync/Factory.cs
```diff
@@ -0,0 +1,386 @@
...
+            {
+                var krole = await GetRoleAsync(role);
+                role.KeycloakGroupId = krole.id;
+
+                _logger.LogInformation($"Updating Role '{role.Name}'.");
+                await HandleRequestAsync<RoleModel, RoleModel>(HttpMethod.Put, $"{_options.Api.Uri}/admin/roles/{role.Id}", role);
+            }
+
+            // Fetch all Users in PIMS and update keycloak so they have the same roles and claims.
+            var users = await HandleRequestAsync<PageModel<UserModel>>(HttpMethod.Get, $"{_options.Api.Uri}/admin/users?page=1&quantity=50"); // TODO: Handle paging.
```
devinleighsmith, line 96: So this is fetching the top 50 users in PIMS and updating their roles and claims? I suppose that works for now.

backend/tools/keycloak/sync/Factory.cs
Fosol: Yes but I'm not going to change it for a job that runs once and may never be run again.

backend/tools/keycloak/sync/Factory.cs
Fosol: Yes for a once off critical job for today, it works.

backend/tools/keycloak/sync/Factory.cs
```diff
@@ -0,0 +1,386 @@
...
+            _logger = logger;
+        }
+        #endregion
+
+        #region Methods
+        /// <summary>
+        /// Sync roles, groups and users in keycloak and PIMS.
+        /// </summary>
+        /// <returns></returns>
+        public async Task<int> SyncAsync()
```
devinleighsmith, line 63: I worry a bit about this all being done in one block with no error handling between sections. Is the intent that if any of the steps fail the whole sync should fail?

backend/dal/Services/Admin/Concrete/ClaimService.cs
devinleighsmith: yeah, this is more of a nitpick, it isn't immediately obvious to me what the return type of CommitTransaction is in this case, I'd need to read the doc of the method.

For me that is usually when I decide to use the actual type instead of var.

backend/api/Startup.cs
devinleighsmith: nice

backend/tools/keycloak/sync/Models/ClaimModel.cs
devinleighsmith: yeah, fair. Feel free to tech debt it for later, not much point in commenting code that may be heavily modified/removed in the near term.

backend/tools/keycloak/sync/Factory.cs
Fosol: Any failure will abort, as intended.  But yes I would prefer a more robust solution.  But not today, or for a once run job.

backend/dal/Services/Admin/Concrete/ClaimService.cs
asanchezr: Agree with Devin

PR #175, opened 2020-04-30
PIMS-1585 paginate the buildings on the parcel detail form.

devinleighsmith:
Add pagination to building subform.
Update pagination UI.
Move field formatting logic from submit action to field onBlur.
Fix some inconsistent refresh behaviour.
Add support for building classifications and conditional subagencies.
Add error display to hidden paginated subforms.


frontend/src/components/common/form/FastInput.tsx
```diff
@@ -90,8 +95,14 @@ export const FastInput: React.FC<FastInputProps> = memo(
           isValid={!!touch && !error && value && !disabled}
           value={value}
           placeholder={placeholder}
-          onBlur={handleBlur}
+          onBlur={(e: any) => {
+            if (onBlurFormatter) {
```
devinleighsmith, line 99: This allows individual fields to be responsible for cleaning up their own data, instead of making the form to the cleanup during onSubmit.

frontend/src/components/common/form/FastSelect.tsx
```diff
@@ -117,11 +119,17 @@ export const FastSelect: React.FC<FastSelectProps> = memo(
           disabled={disabled}
           custom={custom}
           isInvalid={!!touch && !!error}
-          onBlur={handleBlur}
+          onBlur={(e: any) => {
```
devinleighsmith, line 122: Same idea here, enforcing number type behaviour means that I don't have to format the values manually during onsubmit.

frontend/src/forms/ParcelDetailForm.tsx
```diff
@@ -103,9 +95,6 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
     values.pin = values?.pin ? parseInt(values.pin) : undefined;
     values.pid = values?.pid ? values.pid : undefined;
     values.statusId = values.statusId ? 1 : 0;
-    values.classificationId = decimalOrEmpty(values.classificationId);
```
devinleighsmith, line 106: removing all of this processing as it is no longer needed.

frontend/src/forms/ParcelDetailForm.tsx
```diff
@@ -152,10 +135,10 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
       };
     });
 
-    const yupErrors = validateYupSchema(values, ParcelSchema).then(
+    const yupErrors: any = validateYupSchema(values, ParcelSchema).then(
       () => {},
       (err: any) => {
-        return { ...yupToFormErrors(err), ...financialErrors };
+        return _.merge(yupToFormErrors(err), financialErrors);
```
devinleighsmith, line 141: both the yup and manual financial validation is required.

frontend/src/forms/subforms/EvaluationForm.tsx
```diff
@@ -112,6 +113,26 @@ export const filterEmptyFinancials = (evaluations: IFinancial[]) =>
...
+ */
+const getPageErrors = (errors: any, nameSpace: any) => {
+  const evaluationErrors = getIn(errors, nameSpace);
+  const errorsPerPage = Object.keys(keyTypes).length * NUMBER_OF_EVALUATIONS_PER_PAGE;
+  return _.uniq(
+    _.reduce(
+      evaluationErrors,
+      (acc: number[], error, index) => {
+        if (error) {
+          acc.push(Math.ceil((parseInt(index) + 1) / errorsPerPage));
```
devinleighsmith, line 127: because there are multiple instances of possible errors per page, we need to map all errors to the page that they are displayed on, so that user's know which paginated pages contain errors.

frontend/src/components/common/form/FastSelect.tsx
Fosol: I prefer this

frontend/src/forms/subforms/AddressForm.tsx
```diff
@@ -64,15 +64,15 @@ const AddressForm = <T extends any>(props: AddressProps & FormikProps<T>) => {
...
             options={cities}
           />
         </Form.Row>
         <Form.Row>
           <Form.Label column md={2}>
             Province
           </Form.Label>
-          <FastSelect
-            formikProps={props}
+          <Select
```
Fosol, line 75: When would I use FastSelect vs Select?

frontend/src/forms/subforms/EvaluationForm.tsx
Fosol: Is it clear to the user if the error is on a different page than the one they're viewing?

frontend/src/forms/subforms/AddressForm.tsx
devinleighsmith: Large forms.

frontend/src/forms/subforms/EvaluationForm.tsx
devinleighsmith: The error message calls out the specific page number(s) that have errors.

frontend/src/forms/subforms/LandForm.tsx
```diff
@@ -117,6 +117,20 @@ const LandForm = <T extends any>(props: LandProps & FormikProps<T>) => {
...
           </Form.Row>
+          {getIn(props.values, withNameSpace('subAgency')) && (
+            <Form.Row>
+              <Form.Label column md={2}>
+                Sub-Agency
+              </Form.Label>
+              <FastInput
+                formikProps={props}
+                disabled={true}
+                type="string"
```
asanchezr, line 128: If not too much trouble I'd prefer this be named "data-type" or something like that to avoid confusion with `<input type="text">`

What do you think?

PR #182, opened 2020-05-03
PIMS-1472 Property List - Filtering

asanchezr:
Basic filtering functionality working. Some bugs still present around filtering by agencies and min/max land area. 

Some bugs are frontend but I noticed the server doesn't send the right results when land area filters are sent to the API. Once this PR is merged we can fixed those.

### DIFF
--- a/frontend/src/features/properties/list/PropertyListView.tsx
+++ b/frontend/src/features/properties/list/PropertyListView.tsx
@@ -1,16 +1,22 @@
 import './PropertyListView.scss';
 
-import React, { useState, useMemo, useRef, useCallback } from 'react';
+import React, { useState, useMemo, useRef, useCallback, useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { Container, Button } from 'react-bootstrap';
 import queryString from 'query-string';
+import _ from 'lodash';
+import * as API from 'constants/API';
 import CustomAxios from 'customAxios';
 import { ENVIRONMENT } from 'constants/environment';
-import { Table } from 'components/Table';
 import { IPagedItems } from 'interfaces';
-import { IPropertyFilter, IProperty } from '.';
+import { decimalOrUndefined } from 'utils';
+import download from 'utils/download';
+import { RootState } from 'reducers/rootReducer';
+import { ILookupCode } from 'actions/lookupActions';
+import { ILookupCodeState } from 'reducers/lookupCodeReducer';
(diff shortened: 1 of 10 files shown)


frontend/src/features/properties/list/columns.ts
```diff
@@ -10,6 +10,10 @@ export const columns: Column<IProperty>[] = [
     Header: 'Sub Agency',
     accessor: 'subAgency',
   },
+  {
+    Header: 'Property Name',
+    accessor: 'description',
+  },
```
asanchezr, line 16: Adding missing **Property Name** column to the table

frontend/src/features/properties/list/interfaces.ts
```diff
@@ -69,7 +69,12 @@ export interface IPropertyFilter {
...
   municipality?: string;
+  projectNumber?: string;
   classificationId?: number;
   statusId?: number;
-  agencies?: number[];
+  agencies?: number | number[];
+  minLandArea?: number;
+  maxLandArea?: number;
+  minLotArea?: number;
+  maxLotArea?: number;
```
asanchezr, line 79: The server-side model accepts several "land area" parameters. I briefly tried both `minLandArea` and `minLotArea` and cannot get proper results. @Fosol maybe you can take a look tomorrow? 

Once this PR gets merged it will become very apparent that entering data in the Min/Max Area Size fields do not return results

Fosol: MinLand and MinLot are the same in the DB.  We don’t actually have a two different values.  Land = Lot.

Fosol: I suspect perhaps the building need to get a LotSize added, but am not certain as it wasn’t expressed this way in the original dataset.

asanchezr: Yeah I don't know either what's going on. I will update the PR so it passes linting and once merged you can take a look at the back end

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -1,97 +1,192 @@
...
+  const {
+    pageIndex,
+    pageSize,
+    // allAgencies,
+    filter: {
+      address,
+      municipality,
+      projectNumber,
+      classificationId,
+      // agencies,
```
Fosol, line 48: Does this mean the agency filter doesn’t work yet?

PR #183, opened 2020-05-03
PIMS-1472 Bug fixes

asanchezr:


frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -34,35 +34,36 @@ const getServerQuery = (state: {
...
   } = state;
 
-  // const parsedAgencies =
-  //   agencies === null || agencies === undefined || agencies === ''
-  //     ? [...allAgencies]
-  //     : [parseInt(agencies, 10)];
+  // show properties for all agencies if none selected in the agency filter
+  let parsedAgencies = [...agencyIds];
+  if (agencies !== null && agencies !== undefined && agencies !== '') {
+    parsedAgencies = [parseInt(agencies, 10)];
```
devinleighsmith, line 57: is this just shorthand syntax for calling parseInt on the first element of agencies and returning that as an array?

asanchezr: Nope. The name is misleading but `agencies` is a single-select right now so it actually holds a string value. However, the new endpoint made for the property list requires an array to be sent.

So the current logic is: 
* if nothing selected (`null || undefined || ""`) we send the whole array of IDs to the server as `number[]`. 
* If the user made a selection, this will be in string form so yes, convert the string to number and send an array with that Id only.

I can add a comment there if it helps

asanchezr: In addition to that, the new autocomplete that Jonny is working on supports only one selection (as long as I could see from what he posted in RocketChat)

Fosol: Why send the whole array if they didn’t select any agency?

asanchezr: because that's what the API is expecting. See my comments in RC where I say that if I send no agencies in the URL I get 400 Bad Request. Ideally I'd prefer not to send **all** agencies but I don't have time to fix the backend ATM so this works around that.

Maybe you could take a look?

asanchezr: Something to do with `Filter.isValid()` in the backend controller for `/api/properties/search/page`

Fosol: The api isn't expecting the agency to be provided.  It's a filter option just like any other field.

Fosol: All it expects is something is provided.  It doesn't like someone asking for the whole DB.

asanchezr: when you load the Property View page the first time there won't be any filter applied and it will make a request to `/api/properties/search/page?page=1&quantity=10` 

The above is not asking for the whole DB just 10 results - still the API responds with 400 Bad Request. Check it in Postman

asanchezr: If you have time to fix/modify the backend I'm more than happy to remove the workaround

PR #184, opened 2020-05-03
Agency auto-complete

jdtoombs:
*Updated 

Adding reusable auto complete component.

Will have to adjust for parent child relations ships when endpoints support this. 

autocomplete06

### DIFF
--- a/frontend/src/components/common/form/AutoCompleteText.tsx
+++ b/frontend/src/components/common/form/AutoCompleteText.tsx
@@ -1,0 +1,72 @@
+import './AutoCompleteText.scss';
+
+import React from 'react';
+import { Form } from 'react-bootstrap';
+import { useState } from 'react';
+import { DisplayError } from './DisplayError';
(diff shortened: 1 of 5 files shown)


frontend/src/components/common/form/AutoCompleteText.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  const [suggestions, setSuggestions] = useState<any>([]);
+  const [text, setText] = useState<any>('');
+
+  const onTextChanged = (e: any) => {
+    const val = e.target.value;
+    let dynamicSuggestions: any[] = [];
+    if (val.length > 0) {
+      const regex = new RegExp(`^${val}`, 'i');
```
devinleighsmith, line 25: I'm not sure we necessarily want the start of the string, my preference would be just a general match.

jdtoombs: fixed!

PR #189, opened 2020-05-04
Fixing bug related to user agency

Fosol:
While testing we discovered some agencies result in throwing an error when adding new properties.

### DIFF
--- a/backend/dal/Helpers/Extensions/IdentityExtensions.cs
+++ b/backend/dal/Helpers/Extensions/IdentityExtensions.cs
@@ -265,8 +265,8 @@ namespace Pims.Dal.Helpers.Extensions
         }
 
         /// <summary>
-        /// Get the top level agency this user belongs to.
-        /// If they don't belong to any agencies return 'null'.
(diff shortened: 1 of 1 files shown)


backend/dal/Helpers/Extensions/IdentityExtensions.cs
```diff
@@ -279,16 +279,13 @@ public static Agency GetAgency(this ClaimsPrincipal user, PimsContext context)
...
-            // This will only at most do two iterations before returning a value.
-            foreach (var agency in agencies)
-            {
-                // If we find a parent agency we will return it.
-                // Otherwise return the parent of the first child agency.
-                if (agency.ParentId == null) return agency;
-                return context.Agencies.Find(agency.ParentId);
-            }
+            // If one of the agencies is a parent, return it.
+            var parentAgency = agencies.FirstOrDefault(a => a.ParentId == null);
```
Fosol, line 283: Had `!=` where it should be `==`.  We want the first parent agency.

PR #193, opened 2020-05-05
User management dashboard pagination, sorting and filtering

goemen:


backend/api/Areas/Admin/Controllers/UserController.cs
```diff
@@ -93,12 +93,6 @@ public IActionResult GetUsers(EModel.UserFilter filter)
         [SwaggerOperation(Tags = new[] { "admin-user" })]
         public IActionResult GetMyUsers(EModel.UserFilter filter)
         {
-            if (!(filter?.Agencies?.Any() ?? false))
-            {
-                filter.Agencies = this.User.GetAgencies();
```
Fosol, line 98: Does this mean that if the filter contains agencies it will use those instead of the users agencies?  If so that would provide an **Agency Administrator** a way to view and edit users not in their agency.  This isn't ideal.  Only **System Administrators** can view and edit all users.

This endpoint should always and only return users that the current user is allowed to view and edit.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -53,32 +54,51 @@ public Paged<User> Get(int page, int quantity)
         public Paged<User> Get(UserFilter filter = null)
         {
             this.User.ThrowIfNotAuthorized(Permissions.AdminUsers);
+            var agencies = !this.User.IsInRole(Permissions.SystemAdmin.GetName())
```
Fosol, line 57: Use the cleaner extension method `HasPermission(Permissions.SystemAdmin)`.

Fosol: If you're going to return all agencies, don't filter on any of them.  It's a far better query.

Fosol: Generally it's best practice to introduce a statement as a positive condition, rather than a negative.  Change to `this.User.HasPermission(Permissions.SystemAdmin) ? positive : negative`

Fosol: You will still need to ensure the user can only view users that their permissions allow.  **System Administrators** can view everyone.  **Agency Administrators** can only view users within their agencies.

frontend/src/constants/API.tsx
```diff
@@ -9,6 +9,17 @@ export interface IPaginateParams {
...
 }
 
+export interface IGetUsersParams extends IPaginateParams {
+  username?: string;
+  firstName?: string;
+  lastName?: string;
+  email?: string;
+  agency?: string;
+  role?: string;
+  isDisabled?: boolean | null;
```
Fosol, line 19: Why would `isDisabled` ever be null?

frontend/src/actionCreators/usersActionCreator.tsx
```diff
@@ -30,7 +30,7 @@ export const getUsersAction = (params: API.IPaginateParams) => (dispatch: Functi
     .post(ENVIRONMENT.apiUrl + API.POST_USERS(), params)
     .then((response: AxiosResponse) => {
       dispatch(success(actionTypes.GET_USERS, response.status));
-      dispatch(adminActions.storeUsers(response.data));
+      dispatch(adminActions.storeUsers({ ...response.data, page: response.data.page - 1 }));
```
asanchezr, line 33: Given that the frontend "page" differs in value from what the server sends (i.e page starts at "0" in the frontend, and at "1" in the backend") - I would suggest we use a different name in the frontend to avoid confusion. 

I faced this in my table code as well and went with `pageIndex = page - 1`

In my opinion this avoids accidental `page = page` thinking which would cause bugs. What do you think?

frontend/src/constants/API.tsx
asanchezr: I agree with @Fosol here. The boolean is already optional by way of `isDisabled?`. So this indicates the resulting object can have 3 states:

```js
isDisabled: true
isDisabled: false
isDisabled: undefined // property not present in the object
```

Why would we need a **fourth** state?

```js
isDisabled: null // present in the object with value "null"
```

frontend/src/pages/admin/users/ManageUsers.test.tsx
```diff
@@ -34,18 +34,51 @@ jest.mock('react-router-dom', () => ({
 const successStore = mockStore({
   [reducerTypes.USERS]: {
     pagedUsers: {
-      page: 1,
+      page: 0,
```
asanchezr, line 37: See my previous comment on naming this `pageIndex` instead

backend/api/Areas/Admin/Controllers/UserController.cs
goemen: The agency id filter is applied in the users services. See changes in the Get users

backend/dal/Services/Admin/Concrete/UserService.cs
goemen: Just to be clear, only sys-admin and agency-admin can get the user list?

frontend/src/actionCreators/usersActionCreator.tsx
goemen: I agree, make sense to me

backend/dal/Services/Admin/Concrete/UserService.cs
Fosol: Correct, but only sys-admins can see everyone.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -53,32 +54,59 @@ public Paged<User> Get(int page, int quantity)
...
-                if (filter.Agencies?.Any() == true)
-                    query = query.Where(u => u.Agencies.Any(a => filter.Agencies.Contains(a.AgencyId)));
-                if (filter.Sort?.Any() == true)
+                    query = query.Where(u => EF.Functions.Like(u.Email, $"%{filter.Email}%"));
+                if (!string.IsNullOrWhiteSpace(filter.Position))
+                    query = query.Where(u => EF.Functions.Like(u.Position, $"%{filter.Position}%"));
+                if (filter.IsDisabled != null)
+                    query = query.Where(u => u.IsDisabled == filter.IsDisabled);
+                if (!string.IsNullOrWhiteSpace(filter.Role))
+                    query = query.Where(u => u.Roles.Any(a => a.Role.Name.ToLower().Contains(filter.Role.ToLower())));
```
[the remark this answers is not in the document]
goemen: `query = query.Where(u => u.Roles.Any(r => EF.Functions.Like(r.Role.Name, $"%{filter.Role}"))); 
`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -53,32 +54,59 @@ public Paged<User> Get(int page, int quantity)
...
+
+            IQueryable<int> agencies = null;
+
+            if (User.HasPermission(Permissions.AgencyAdmin))
+            {
+                agencies = Context.Agencies.Where(a => User.GetAgencies().Contains(a.Id))
+                    .Select(a => a.Id);
+            } else if (User.HasPermission(Permissions.SystemAdmin))
+            {
+                agencies = Context.Agencies.Select(x => x.Id);
```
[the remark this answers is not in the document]
goemen: Not sure if I follow, can we jump in a quick call?

frontend/src/pages/admin/users/ManageUsers.tsx
```diff
@@ -12,55 +11,101 @@ import './ManageUsers.scss';
...
 
 const ManageUsers = () => {
   const dispatch = useDispatch();
-  const MAX_USERS_PER_PAGE = 10;
   const pagedUsers = useSelector<RootState, IPagedItems<IUser>>(state => {
-    return (state.users as IStoreUsersAction).pagedUsers;
+    return (state.users as IUsersState).pagedUsers;
   });
+
+  const pageSize = useSelector<RootState, number>(state => {
```
devinleighsmith, line 25: I'm not sure if there is any benefit to this suggestion, but it seems like you could make a single selector for the top level state.users object and then destructure that into the values that you need, rather then making two selectors here.

frontend/src/pages/admin/users/ManageUsers.tsx
```diff
@@ -12,55 +11,101 @@ import './ManageUsers.scss';
...
+      case 'filterChange':
+      case 'changeRowsPerPage':
+        dispatch(setUsersFilter(state.filterList));
+        dispatch(setUsersPageSize(state.rowsPerPage));
+
+        dispatch(
+          getUsersPaginationAction({
+            page: state.page + 1,
+            quantity: state.rowsPerPage,
+            username: state.filterList[1][0] || '',
```
devinleighsmith, line 65: I'm not a big fan of using a two dimensional array to get values like this, but it may not matter if we are swapping table implementations anyhow.

frontend/src/pages/admin/users/ManageUsers.tsx
```diff
@@ -12,55 +11,101 @@ import './ManageUsers.scss';
...
+      role: u.roles && u.roles.length > 0 ? u.roles[0].name : 'No role attached',
+      agency: u.agencies && u.agencies.length > 0 ? u.agencies[0].name : 'No agency associated',
       position: u.position,
     }),
   );
 
+  if (sort) {
+    userList = sortBy(userList, u => u[sort.sortBy]);
+    if (sort.direction === 'desc') {
+      userList = userList.reverse();
```
devinleighsmith, line 96: This does reverse the order of the list assuming that the sort order of all elements is equal correct?

ie. if I order by position [{position: 'test', name:'doug'}, {position: 'test', name 'frank'}] -> [{position: 'test', name 'frank'}, {position: 'test', name:'doug'}]

frontend/src/reducers/usersReducer.tsx
```diff
@@ -1,16 +1,39 @@
...
+  rowsPerPage: number;
+  filterList: string[][];
+  sort: IUsersSort;
 }
 
 const initialState: IUsersState = {
-  pagedUsers: { page: 0, total: 0, quantity: 0, items: [] },
+  pagedUsers: { page: 1, pageIndex: 0, total: 0, quantity: 0, items: [] },
+  rowsPerPage: MAX_USERS_PER_PAGE,
+  filterList: [[], [], [], [], [], [], [], [], [], []],
```
devinleighsmith, line 24: could this have been implemented as a dictionary instead (an object of keys where each key is an array)? Even if we needed to create a translation function between the dictionary and this format I think that would be my preference.

frontend/src/pages/admin/users/ManageUsers.tsx
goemen: Yeah, this how the grid presents the filter values

frontend/src/pages/admin/users/ManageUsers.tsx
goemen: If you do that it means that even when the other properties change then the reaction to the pageSize change will be triggered even if the pageSize did not change

frontend/src/pages/admin/users/ManageUsers.tsx
goemen: @devinleighsmith I am not sure if I understand. Even when the values for position are the same, the reverse of the sort will give the correct results

frontend/src/reducers/usersReducer.tsx
goemen: I am a little hesitant to invest any more time to clean it 100% because we still have to refactor after deciding on which grid system

frontend/src/reducers/accessRequestReducer.tsx
```diff
@@ -30,7 +30,7 @@ export interface IAccessRequestState {
 }
 
 const initialState: IAccessRequestState = {
-  pagedAccessRequests: { page: 0, total: 0, quantity: 0, items: [] },
+  pagedAccessRequests: { page: 1, pageIndex: 0, total: 0, quantity: 0, items: [] },
```
asanchezr, line 33: I'm not loving that we need to keep track of "two" page instances (one for frontend and one for server). I'm not sure why you need the server page in the frontend as it can be calculated: `page = pageIndex + 1` just before issuing an API request.

This is no blocker and will likely be revisited when we refactor the tables

frontend/src/reducers/usersReducer.tsx
asanchezr: agree that further work should be pushed to the refactor

PR #196, opened 2020-05-06
PIMS-1382 parcel detail form test updates.

devinleighsmith:


frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -97,7 +108,7 @@ describe('ParcelDetailForm', () => {
       agencyId: 1,
       address: {
         line1: 'addressval',
-        cityId: 1,
+        cityId: '1',
```
Fosol, line 111: I'm curious why we use strings here?

frontend/src/pages/admin/access/ManageAccessRequests.test.tsx
```diff
@@ -13,6 +12,7 @@ import * as API from 'constants/API';
 import { ensureGridApiHasBeenSet } from '../../../utils/testUtils';
 import ManageAccessRequests from './ManageAccessRequests';
 import { AgGridReact } from 'ag-grid-react/lib/agGridReact';
+import { act as domAct } from 'react-dom/test-utils';
```
Fosol, line 15: Is there a reason to have an alias?

frontend/src/forms/ParcelDetailForm.test.tsx
asanchezr: Most likely this comes from a `<select>`. Selects always return their value in string form, or so I've seen.

https://www.w3schools.com/tags/tag_option.asp

frontend/src/pages/admin/access/ManageAccessRequests.test.tsx
devinleighsmith: Given that there are two different act functions, (one for react-dom, one from react-test-renderer) i've been consistently calling the import from react-dom domAct so that I can differentiate them (since they are not interchangeable).

PR #197, opened 2020-05-06
PIMS-1418 using material ui theme provider to override styling

goemen:


backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -69,7 +69,7 @@ public Paged<User> Get(UserFilter filter = null)
                 .ThenInclude(r => r.Role)
```
Fosol, line 69: Please also remove lines 60-63

goemen:
@Fosol before I change the query I want to test a theory in dev `if (User.HasPermission(Permissions.AgencyAdmin) && !User.HasPermission(Permissions.SystemAdmin))
            {
                query = query.Where(user => user.Agencies.Any(a => this.User.GetAgencies().Contains(a.AgencyId)));
            }`

goemen:
> @Fosol before I change the query I want to test a theory in dev `if (User.HasPermission(Permissions.AgencyAdmin) && !User.HasPermission(Permissions.SystemAdmin)) { query = query.Where(user => user.Agencies.Any(a => this.User.GetAgencies().Contains(a.AgencyId))); }`

I have tested things locally with that change and its  helping

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -56,22 +56,16 @@ public Paged<User> Get(UserFilter filter = null)
         {
             this.User.ThrowIfNotAuthorized(Permissions.AdminUsers);
 
-            var userAgencies = User.GetAgencies();
-            if (userAgencies == null || !userAgencies.Any())
```
Fosol, line 60: Thanks

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -56,22 +56,16 @@ public Paged<User> Get(UserFilter filter = null)
...
                 .ThenInclude(a => a.Agency)
                 .Include(r => r.Roles)
                 .ThenInclude(r => r.Role)
                 .AsNoTracking();
 
-            if (User.HasPermission(Permissions.AgencyAdmin))
+            if (User.HasPermission(Permissions.AgencyAdmin) && !User.HasPermission(Permissions.SystemAdmin))
             {
-                query = query.Where(user => user.Agencies.Any(a => userAgencies.Contains(a.AgencyId)));
+                query = query.Where(user => user.Agencies.Any(a => this.User.GetAgencies().Contains(a.AgencyId)));
```
Fosol, line 68: It's possible I think for `this.Users.GetAgencies()` to return `null`.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -56,22 +56,16 @@ public Paged<User> Get(UserFilter filter = null)
...
 
-            var userAgencies = User.GetAgencies();
-            if (userAgencies == null || !userAgencies.Any())
-            {
-                throw new NotAuthorizedException("Current user does not belong to an agency");
-            }
-
             var query = this.Context.Users
                 .Include(u => u.Agencies)
                 .ThenInclude(a => a.Agency)
```
Fosol, line 61: I think this will remain an `inner join` which will result in not returning users who are not part of an `agency` or `role`.

goemen: I am seeing users with no agency or role. I want to test a theory in dev. Testing how having both sys-admin and agency-admin affects the results. After my test, I will another PR with the query change.

PR #199, opened 2020-05-07
PIMS-1885 Edit User Page - position, isDisabled and note fixes

goemen:


frontend/src/pages/EditUserPage.tsx
```diff
@@ -146,21 +146,42 @@ const EditUserPage = (props: IEditUserPageProps) => {
             {props => (
               <Form className="userInfo">
                 <Label>IDIR/BCeID</Label>
-                <Input field="username" value={props.values.username} readOnly={true} type="text" />
+                <Input
```
devinleighsmith, line 149: You can also get inputs using container.querySelector(`input[name="username"]`);

PR #201, opened 2020-05-07
PIMS-1677 Adding dependencies to React useEffect

goemen:
The second argument of the useEffect hook is an array of dependencies. When some of the dependencies change, the effect callback will be executed.

**Use cases:**

1. `React.useEffect(callback)` => no dependency array means that the effect callback will be executed every time the component rerenders

2. `React.useEffect(callback, [])` => empty dependency array means that the effect callback will be executed once

3. `React.useEffect(callback, [d1, d2, d3])` => means that the effect callback will be executed every time when one of d1, d2, d3 changes

**Things to watch out for**
An effect can get stuck in a loop if have an effect that causes its dependencies to change.


frontend/src/pages/AccessRequestPage.tsx
```diff
@@ -43,8 +43,7 @@ const AccessRequestPage = () => {
 
   useEffect(() => {
     dispatch(getCurrentAccessRequestAction());
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [dispatch]);
```
Fosol, line 46: Isn't `dispatch` a function?  How does it test if state is changed with it?

PR #214, opened 2020-05-13
Postal code format

jdtoombs:
Postal code will now automatically format to upper case, remove a hyphen and replace with a space, and if there is no space it will add one.

### DIFF
--- a/frontend/src/utils/YupSchema.tsx
+++ b/frontend/src/utils/YupSchema.tsx
@@ -50,7 +50,7 @@ export const Address = Yup.object().shape({
     .nullable(),
   provinceId: Yup.string().required('Required'),
   postal: Yup.string().matches(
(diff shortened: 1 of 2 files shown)


frontend/src/utils/YupSchema.tsx
```diff
@@ -50,7 +50,7 @@ export const Address = Yup.object().shape({
     .nullable(),
   provinceId: Yup.string().required('Required'),
   postal: Yup.string().matches(
-    /^[a-zA-z][0-9][a-zA-z]\s*?[0-9][a-zA-z][0-9]$/,
+    /^[a-zA-z][0-9][a-zA-z]\s*-*[0-9][a-zA-z][0-9]$/,
```
jdtoombs, line 53: Had to add the hyphen and space to the schema because it would run the validation before the input was formatted resulting in invalid validation. So even though the hyphen and space will not be present when the user clicks away the schema still saw it. Please let me know if there is a better way!

PR #215, opened 2020-05-13
PIMS-1690 Replace AG-Grid with React-Table

goemen:


backend/api/Areas/Admin/Controllers/AccessRequestController.cs
```diff
@@ -40,25 +40,35 @@ public class AccessRequestController : Controller
...
+        [SwaggerOperation(Tags = new[] {"admin-access-requests"})]
+        public IActionResult GetPage(int page = 1, int quantity = 10, string sort = null,
+            string searchText = null, string role = null, string agency = null,
+            Entity.AccessRequestStatus? status = Entity.AccessRequestStatus.OnHold)
         {
             if (page < 1) page = 1;
             if (quantity < 1) quantity = 1;
             if (quantity > 20) quantity = 20;
-            var result = _pimsAdminService.User.GetAccessRequests(page, quantity, sort, status);
+            if (status == null) status = Entity.AccessRequestStatus.OnHold;
```
Fosol, line 67: Out of curiosity is this even required as you have a default value in the arguments.  Is this only required because the requests passes an empty param (i.e. `?status=&page=1`)?

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -237,6 +237,42 @@ public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, s
             var accessRequests = query.Skip((page - 1) * quantity).Take(quantity);
             return new Paged<AccessRequest>(accessRequests, page, quantity, query.Count());
         }
+
+        /// <summary>
+        /// Get all the access requests that users have submitted to the system
```
Fosol, line 242: Summary doesn't represent what the function actually does.
`Get all access requests that match the specified 'filter'.`

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -237,6 +237,42 @@ public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, s
...
+                .Include(p => p.Roles)
+                .ThenInclude(p => p.Role)
+                .Include(p => p.User)
+                .AsNoTracking();
+
+            query = query.Where(request => request.Status == filter.Status);
+
+            if (!string.IsNullOrWhiteSpace(filter.Role))
+                query = query.Where(ar => ar.Roles.Any(r =>
+                    EF.Functions.Like(r.Role.Name, $"%{filter.Role}")));
```
Fosol, line 261: Curious why the LIKE begins with % but doesn't end with %.  I'd think it would be more intuitive if had both, or ended with %.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -237,6 +237,42 @@ public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, s
...
+
+            if (!string.IsNullOrWhiteSpace(filter.Agency))
+                query = query.Where(ar => ar.Agencies.Any(a =>
+                    EF.Functions.Like(a.Agency.Name, $"%{filter.Agency}")));
```
Fosol, line 265: Curious why the LIKE begins with % but doesn't end with %.  I'd think it would be more intuitive if had both, or ended with %.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -237,6 +237,42 @@ public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, s
...
+
+            if (!string.IsNullOrWhiteSpace(filter.SearchText))
+            {
+                query = query.Where(ar => EF.Functions.Like(ar.User.Username, $"%{filter.SearchText}"));
```
Fosol, line 269: Curious why the LIKE begins with % but doesn't end with %.  I'd think it would be more intuitive if had both, or ended with %.

backend/entities/Models/AccessRequestFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Get/Set - agency
+        /// </summary>
+        public string Agency { get; set; }
+
+        /// <summary>
+        /// Get/Set - Status
+        /// </summary>
+        public AccessRequestStatus? Status { get; set; }
+
+        public AccessRequestFilter(int page, int quantity, string[] sort, string searchText, string role, string agency,
```
Fosol, line 25: Your implementation here isn't entirely correct for best practices.  You've made the class mutable with properties that can be `set` but you haven't provided a base constructor.  It's an odd combination, even if in our use-case it most likely will never matter.

Most models should have a default base empty constructor.

frontend/package.json
```diff
@@ -26,6 +26,7 @@
     "react": "16.12.0",
     "react-app-polyfill": "1.0.5",
     "react-bootstrap": "1.0.0-beta.16",
+    "react-click-away-listener": "^1.1.0",
```
asanchezr, line 29: /nit: please remove the `^` - we want dependencies locked in to avoid breaking builds when they get updated without our review. thanks!

backend/tests/unit/api/Controllers/Admin/AccessRequestControllerTest.cs
```diff
@@ -40,17 +41,18 @@ public void GetAccessRequests_Success()
             var accessRequest2 = EntityHelper.CreateAccessRequest(2);
             var accessRequests = new[] { accessRequest1, accessRequest2 };
             var paged = new Entity.Models.Paged<Entity.AccessRequest>(accessRequests);
-            service.Setup(m => m.User.GetAccessRequests(It.IsAny<int>(), It.IsAny<int>(), It.IsAny<string>(), It.IsAny<Entity.AccessRequestStatus?>())).Returns(paged);
+
+            service.Setup(m => m.User.GetAccessRequests(It.IsAny<AccessRequestFilter>())).Returns(paged);
 
             // Act
-            var result = controller.GetPage(1, 10);
+            var result = controller.GetPage(1, 10, null, null, null, null, Entity.AccessRequestStatus.OnHold);
```
Fosol, line 48: You don't need to change this now, but here you can see the result of your choice to add the query parameters as arguments on the endpoint.  If you look at other implementations with filters you can see a cleaner approach where it isn't required.

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -237,6 +237,42 @@ public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, s
...
+            }
```
asanchezr, line 270: I'm okay with these filter but would like to ask @Fosol whether it makes sense to have these filter be "start-with". I think some of them might not find data that is there - specially the search text one. 

The user would have to know that they are doing a "start-with" query, instead of a "contains"

Thoughts?

frontend/src/components/menu/index.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { ReactNode } from 'react';
+import { ListGroup, Form } from 'react-bootstrap';
+import classNames from 'classnames';
+import ClickAwayListener from 'react-click-away-listener';
+import './menu.scss';
+
+export interface IMenuItemProps {
```
Fosol, line 7: I think we should start adding documentation to our exports.  It's tech-debt that is quickly adding up.

frontend/src/components/menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+    </ListGroup.Item>
+  );
+};
+
+interface IProps {
+  children?: any;
+  label?: ReactNode;
+  options: IMenuItemProps[];
+  width?: string;
+  filter?: boolean;
```
Fosol, line 33: What does `filter` mean in this context?  Does it mean add an input to allow filter?

frontend/src/constants/actionTypes.tsx
```diff
@@ -27,11 +27,13 @@ export const ADD_REQUEST_ACCESS = 'addRequestAccess';
 export const GET_REQUEST_ACCESS = 'getRequestAccess';
 export const UPDATE_REQUEST_ACCESS_ADMIN = 'updateRequestAccessAdmin';
 export const UPDATE_REQUEST_ACCESS_STATUS_ADMIN = 'updateRequestAccessStatusAdmin';
```
Fosol, line 29: The more I see this file the more I cringe.  I think we need to quickly start using typescript enums.

https://medium.com/@Alserda/readable-ts-enum-actiontypes-in-react-redux-c105fa0a7c9b

frontend/src/components/menu/index.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { ReactNode } from 'react';
+import { ListGroup, Form } from 'react-bootstrap';
+import classNames from 'classnames';
+import ClickAwayListener from 'react-click-away-listener';
+import './menu.scss';
```
asanchezr, line 5: Minor comment on naming. We have adopted to name component files with Uppercase - i.e. `Menu.tsx` and `Menu.scss` vs `menu.tsx`.

We also have the folder start with Uppercase to indicate this is a component.

backend/dal/Services/Admin/Concrete/UserService.cs
goemen: I made a mistake here, changing it to `$"%{filter.SearchText}%"`

frontend/src/hooks/useApi.ts
```diff
@@ -0,0 +1,26 @@
...
+export const useApi = () => {
+  const dispatch = useDispatch();
+  const axios = CustomAxios();
+
+  axios.interceptors.request.use(
+    config => {
+      dispatch(showLoading());
+      return config;
+    },
+    error => dispatch(hideLoading()),
```
Fosol, line 14: I see this extra logic a lot in the codebase.  You can just use a `finally` to `hideLoading()`.

frontend/src/hooks/useApi.ts
```diff
@@ -0,0 +1,26 @@
+import CustomAxios from 'customAxios';
+import { useDispatch } from 'react-redux';
+import { showLoading, hideLoading } from 'react-redux-loading-bar';
+
+export const useApi = () => {
```
Fosol, line 5: I like this hook idea.

frontend/src/components/menu/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface IProps {
+  children?: any;
+  label?: ReactNode;
+  options: IMenuItemProps[];
+  width?: string;
+  filter?: boolean;
+  searchPlaceholder?: string;
+}
+
+export const Menu = (props: IProps) => {
```
asanchezr, line 37: Another minor observation, if you type this as `export const Menu: React.FC<IProps> = (...) => { ... }` you don't have to add the `children` property as the React typing does it for you. It also provides a better typing than `children: any` which can help catch bugs.

frontend/src/pages/admin/access/ManageAccessRequests.tsx
```diff
@@ -1,114 +1,129 @@
+import * as React from 'react';
+import { Container, Row, Col, Form } from 'react-bootstrap';
+import { AccessRequestStatus } from 'constants/accessStatus';
+import { Table } from 'components/Table';
+import { IPaginate, toAccessRequestApiPaginateParams } from 'utils/CommonFunctions';
+import * as API from 'constants/API';
+import { IAccessRequest } from 'interfaces';
+import './ManageAccessRequests.scss';
+import { useDispatch, useSelector } from 'react-redux';
```
Fosol, line 9: The more I see us use redux the more I'm questioning why we're using where we use it.  Does this page share state with other pages?  Or is simply how we're sharing state with sub-components of this larger component.  If it's the latter I suspect we're not entirely designing this correctly.  However I'm still not up-to-speed with everything react/redux and don't have a strong opinion yet.

frontend/src/hooks/useApi.ts
asanchezr, line 5: Thank you for doing this! I've been meaning to add this for a while.
Love it!

frontend/src/pages/admin/access/ManageAccessRequests.scss
```diff
@@ -1,4 +1,48 @@
+$header-height: 72px;
+$footer-height: 46px;
+$navbar-height: 45px;
+$filter-height: 76px;
+$contentheading-height: 60px;
```
asanchezr, line 5: Please check the other places we have this. I think Devin moved this to a common scss file that can be included there... @devinleighsmith ?

frontend/src/hooks/useApi.ts
goemen: I am sad that I did not have the use of it in my table. But yeah, it centralizes the loading control

frontend/src/pages/admin/access/ManageAccessRequests.scss
goemen: I found them

frontend/package.json
asanchezr: Also, since we are replacing `ag-grid` - wouldn't it make sense to remove it from `package.json`? I still see it in lines 11-12 above...

```json
    "ag-grid-community": "23.0.2",
    "ag-grid-react": "23.0.3",
```

frontend/src/pages/admin/access/ManageAccessRequests.scss
```diff
@@ -1,4 +1,48 @@
+$header-height: 72px;
```
devinleighsmith, line 1: if these are the same, you could get these from _variables.scss instead.

backend/dal/Services/Admin/Concrete/UserService.cs
goemen: I put it in both ends, and its all good now

frontend/src/components/menu/index.tsx
goemen: It means hide/show filter box

frontend/src/constants/actionTypes.tsx
goemen: I have other ideas to clean this up. I introduced a new hook call useApi, and we can extend it to have feature-focused actions. I will present my idea to the team on Friday if we have enough time

frontend/src/components/menu/index.tsx
Fosol: I would recommend `showFilter`

backend/api/Areas/Admin/Controllers/AccessRequestController.cs
goemen: Not necessary, I changed the endpoint param to be not nullable

backend/entities/Models/AccessRequestFilter.cs
devinleighsmith: Does he need to add one in addition to the one C# adds by default to all classes?

Fosol: C# doesn't add one by default.  If someone tried to create a new instance of this class it would fail if they didn't provide any arguments.

PR #217, opened 2020-05-14
Postal code format

jdtoombs:


backend/api/Controllers/BuildingController.cs
```diff
@@ -93,6 +93,10 @@ public IActionResult GetBuildings([FromBody]BuildingFilter filter)
         public IActionResult GetBuilding(int id)
         {
             var entity = _pimsService.Building.Get(id);
+            if(entity.Address?.Postal != null)
+            {
+                entity.Address.Postal = entity.Address.Postal.Substring(0,3) + " " + entity.Address.Postal.Substring(3,3);
```
Fosol, line 98: You can do this `entity.Address.Postal.Insert(3, " ")`, and since you are doing it multiple places, add an extension method to `StringExtensions.FormatAsPostal(...)`.

backend/api/Controllers/ParcelController.cs
```diff
@@ -98,6 +98,14 @@ public IActionResult GetParcels([FromBody]ParcelFilter filter)
         public IActionResult GetParcel(int id)
         {
             var entity = _pimsService.Parcel.Get(id);
+            if(entity.Address?.Postal != null)
+            {
+               entity.Address.Postal = entity.Address.Postal.Substring(0,3) + " " + entity.Address.Postal.Substring(3,3);
```
Fosol, line 103: You can do this `entity.Address.Postal.Insert(3, " ")`, and since you are doing it multiple places, add an extension method to `StringExtensions.FormatAsPostal(...)`.

backend/api/Controllers/ParcelController.cs
```diff
@@ -98,6 +98,14 @@ public IActionResult GetParcels([FromBody]ParcelFilter filter)
...
+            }
+            foreach (var item in entity.Buildings)
+            {
+                item.Address.Postal = item.Address.Postal.Substring(0,3) + " " + item.Address.Postal.Substring(3,3);
```
Fosol, line 107: You can do this `entity.Address.Postal.Insert(3, " ")`, and since you are doing it multiple places, add an extension method to `StringExtensions.FormatAsPostal(...)`.

backend/api/Controllers/BuildingController.cs
Fosol: Also since this is the response, move logic to the mapper instead.

backend/api/Controllers/ParcelController.cs
Fosol: Also since this is the response, move logic to the mapper instead.

backend/api/Controllers/ParcelController.cs
Fosol: Also since this is the response, move logic to the mapper instead.

backend/api/Controllers/ParcelController.cs
```diff
@@ -115,8 +123,13 @@ public IActionResult GetParcel(int id)
...
         {
+            model.Address.Postal = model.Address.Postal.Replace(" ", "");
+            foreach(var item in model.Buildings)
+            {
+                item.Address.Postal = item.Address.Postal.Replace(" ", "");
+            }
             var entity = _mapper.Map<Entity.Parcel>(model);
-
+            
             _pimsService.Parcel.Add(entity);
```
Fosol, line 133: Move formatting of postal to service layer instead so that it's handled with any future endpoints.

backend/api/Controllers/ParcelController.cs
```diff
@@ -137,6 +150,11 @@ public IActionResult AddParcel([FromBody] Model.ParcelModel model)
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "To support standardized routes (/update/{id})")]
         public IActionResult UpdateParcel(int id, [FromBody] Model.ParcelModel model)
         {
+            model.Address.Postal = model.Address.Postal.Replace(" ", "");
```
Fosol, line 153: Move formatting of postal to service layer instead so that it's handled with any future endpoints.

frontend/src/forms/subforms/AddressForm.tsx
```diff
@@ -41,6 +41,20 @@ const AddressForm = <T extends any>(props: AddressProps & FormikProps<T>) => {
     return nameSpace ? `${nameSpace}.${fieldName}` : fieldName;
   };
 
+  const postalCodeFormatter = (postal: string) => {
```
Fosol, line 44: I think this may be overthought.  All you really want to do is add a space at the 3 index position and uppercase.  You're not validating in this function as far as I can tell.  If you can find a nice non-mutating solution that inserts a space at the index position 3.
```js
var postal = 'a8a8a8'.split('');
postal.splice(3,0,' '); // mutating however
var formatted = postal.join('').toUpperCase();
```

frontend/src/utils/YupSchema.tsx
```diff
@@ -50,7 +50,7 @@ export const Address = Yup.object().shape({
     .nullable(),
   provinceId: Yup.string().required('Required'),
   postal: Yup.string().matches(
-    /^[a-zA-z][0-9][a-zA-z]\s*?[0-9][a-zA-z][0-9]$/,
+    /^[a-zA-z][0-9][a-zA-z]\s*-*[0-9][a-zA-z][0-9]$/,
```
Fosol, line 53: Odd regex for the center space.  You're saying 0 or more spaces or 0 or more dashes.  I'm fairly certain you just want [\s-]?

jdtoombs: Oh woops yes you're right

frontend/src/forms/subforms/AddressForm.tsx
devinleighsmith: my recommendation would be to use capturing groups to get the first three and last three characters, and then format those as desired. That way you can use a single regex and you don't need any additional processing functions outside of the regex.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

jdtoombs: Gotchya, thanks. Adjusting now

backend/api/Areas/Admin/Mapping/Parcel/AddressMap.cs
```diff
@@ -2,6 +2,7 @@
 using System;
 using Model = Pims.Api.Areas.Admin.Models.Parcel;
 using Entity = Pims.Dal.Entities;
+using Pims.Core.Extensions;
```
devinleighsmith, line 5: yes, this is a much better place for this

frontend/src/forms/subforms/AddressForm.tsx
```diff
@@ -42,17 +42,12 @@ const AddressForm = <T extends any>(props: AddressProps & FormikProps<T>) => {
...
-      postal = postal.substr(0, 3) + ' ' + postal.substr(3, 5);
-      return postal.toUpperCase();
-    } else if (hyphen.test(postal)) {
-      postal = postal.replace('-', ' ');
-      return postal.toUpperCase();
-    } else {
-      return postal.toUpperCase();
+    const regex = /[a-zA-z][0-9][a-zA-z][\s-]?[0-9][a-zA-z][0-9]/;
+    const format = postal.match(regex);
+    if (format !== null) {
```
devinleighsmith, line 47: this works, and I don't think we need to rework this again, but what I meant was something like (pseudocode)
const regex = /([a-zA-z][0-9][a-zA-z])[\s-]?*([0-9][a-zA-z][0-9])/;
const matches = postal.match(regex);
if(format !== null and format.length === 2) {
postal = `${format[0]} ${format[1]}`
}

devinleighsmith: also making comments with backticks in github is a pain.

frontend/src/utils/YupSchema.tsx
```diff
@@ -98,9 +98,7 @@ export const Building = Yup.object().shape({
   buildingFloorCount: Yup.number()
     .min(1, 'Floor Count must be a positive number')
     .required('Required'),
-  buildingTenancy: Yup.string()
-    .max(100, 'Tenancy must be less then 100 characters')
-    .required('Required'),
+  buildingTenancy: Yup.string().max(100, 'Tenancy must be less then 100 characters'),
```
devinleighsmith, line 101: is this not required? I see it is required in the mockups.

frontend/src/forms/subforms/AddressForm.tsx
jdtoombs: Ohh yes that is way better

frontend/src/forms/subforms/BuildingForm.test.tsx
```diff
@@ -122,7 +122,7 @@ describe('sub-form BuildingForm functionality', () => {
       fireEvent.click(submit!);
     });
     const errors = await getAllByText('Required');
-    expect(errors).toHaveLength(11);
+    expect(errors).toHaveLength(10);
```
jdtoombs, line 125: tenancy no longer required

frontend/src/utils/YupSchema.tsx
devinleighsmith: nvm, I see your above comment. Ignore this.

backend/core/Extensions/StringExtensions.cs
```diff
@@ -35,5 +35,10 @@ public static bool IsProduction(this string env)
         {
             return env != null && env.Equals("Production", StringComparison.OrdinalIgnoreCase);
         }
+
+        public static string FormatAsPostal(this string postal)
```
Fosol, line 39: Document your functions next time :D

jdtoombs: Ooh good point, I'll add right now so it will be in the next PR!

PR #218, opened 2020-05-14
PIMS-105 Update doc and provide method to tag images.

devinleighsmith:
provide pipeline assisted method of applying tags to built images.
-Update doc to explain how to check sha on deployed images.

### DIFF
--- a/openshift/pipelines/Jenkinsfile.cicd
+++ b/openshift/pipelines/Jenkinsfile.cicd
@@ -358,6 +358,64 @@ pipeline {
         // """
       }
     }
+    stage("Tag API") {
+      environment {
+        COMP_NAME = "api"
+      }
+      steps {
+        script {
+          openshift.withCluster() {
+            openshift.withProject() {
+              try {
+                timeout(time: 1, unit: "MINUTES") {
+                  def api_tag = input(message: "Enter the desired API tag", parameters: [
+                    string(name: 'API_TAG', defaultValue: '', description: 'Tag to apply to the API image.')
+                  ])
+                  if (api_tag != '') {
+                    def image_name = "${APP_NAME}-${COMP_NAME}"
+                    def hash = cli.getImageTagHash("${image_name}")
+                    echo "[ci/cd]  Create permanent tag ${api_tag} for image ${image_name}@${hash}"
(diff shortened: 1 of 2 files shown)


openshift/pipelines/Jenkinsfile.cicd
```diff
@@ -358,6 +358,64 @@ pipeline {
...
+    stage("Tag API") {
+      environment {
+        COMP_NAME = "api"
+      }
+      steps {
+        script {
+          openshift.withCluster() {
+            openshift.withProject() {
+              try {
+                timeout(time: 1, unit: "MINUTES") {
```
Fosol, line 370: I think this now expects us to enter a tag for all pipelines that use this Jenkinsfile.  Do we want to do this in our other environments?  I'd think for DEV we'd just want to keep using the `dev` tag.

openshift/templates/maintenance/caddy.bc.json
```diff
@@ -0,0 +1,110 @@
+{
```
Fosol, line 1: Please convert to yaml to keep it the same as our other templates and configs.

openshift/templates/maintenance/caddy.dc.json
```diff
@@ -0,0 +1,136 @@
+{
```
Fosol, line 1: Please convert to yaml to keep it the same as our other templates and configs.

openshift/pipelines/Jenkinsfile.cicd
devinleighsmith: Right, let me explain exactly what this is doing.
1) If you don't enter a tag the current functionality of the pipeline is unchanged.
2) If you enter a tag, a new tag will be created in addition to any tags normally created by the pipeline. The new phases is ignored by all future phases of the pipeline, but it is essentially "saved" for historic reference.

openshift/templates/maintenance/caddy.bc.json
devinleighsmith: uhhh, great. These force pushes were supposed to be on a separate branch. let me clean this up.

devinleighsmith: moved all of this out, will add in it's own PR later.

PR #219, opened 2020-05-14
PIMS-1272 MS SQL Image

Fosol:
## Description

Updated the templates to create our persistent database.  We can now run initialization scripts on startup.

When a new database is created the `pims` database will be created.  This resolves the issue where we would have to do this manually in the past.

Additionally a new user is created which will now be used instead of the `sa` account for connection from the API.

## Impact

I will need to refresh all of our database and API instances after this is merged.


database/mssql/entrypoint.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
 # Start SQL Server
-/opt/mssql/bin/sqlservr & /usr/config/setup.sh
+echo "Start MS SQL Server and run scripts"
 
+if ! whoami &> /dev/null; then
```
Fosol, line 6: This is required because of the Microsoft bug where it cannot run `sqlcmd` in an OpenShift container.  OpenShift assigns a random UID user account to the pod, which Microsoft doesn't like.

database/mssql/setup.sh
```diff
@@ -1,22 +1,30 @@
...
+	is_up=$?
+	sleep 5
 done
 
-if [ $STATUS -ne 0 ]; then
-	echo "Error: MSSQL SERVER took more than thirty seconds to start up."
-	exit 1
-fi
+echo "======= MSSQL SERVER STARTED ========"
+# Run every script in /scripts
```
Fosol, line 17: I've kept this section in as it will allow us to add additional scripts overtime easily.

database/mssql/openshift/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM mssql-rhel-server:2019-latest
+
+RUN mkdir -p /usr/config
+WORKDIR /usr/config
+COPY . /usr/config
+
+USER root
+RUN chmod -R g=u /etc/passwd
```
Fosol, line 8: This is required because of the Microsoft bug where it cannot run `sqlcmd` in an OpenShift container.  OpenShift assigns a random UID user account to the pod, which Microsoft doesn't like.

openshift/templates/pims-db/mssql-build.yaml
```diff
@@ -0,0 +1,133 @@
+# This is a build template for .NET Core 3.1
+# It will build a new image from the specified source control repo
+# that will run your application.
+kind: Template
```
Fosol, line 4: We now will use a custom built image for our database so that we can initialize it.

openshift/templates/pims-db/mssql-deploy.yaml
```diff
@@ -149,21 +169,51 @@ parameters:
...
+    displayName: The Microsoft SQL Server Edition
+    description: The edition of the database to run [Developer|Express|Standard|Enterprise|EnterpriseCore]
+    required: true
+    value: Developer
   - name: MSSQL_SA_PASSWORD
     displayName: Microsoft SQL Server Password
     description: Microsoft SQL Server sa password
     from: '[\w]{5}_[\w]{15}!'
     generate: expression
+  - name: TIMEZONE
```
devinleighsmith, line 188: Since we are using DateTimes and not DateTimeOffsets I'm not sure we can reasonably support this:
https://docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime

I think we are restricted to running the DB in local or UTC only, unless none of our datetimes are dependent on timezone (ie. all datetimes are used to track dates or times only).

Fosol: This is unrelated.  The image has an option to set the timezone, I've imply added it as configuration option.  

We can choose whatever date datatype we want when storing dates.

PR #220, opened 2020-05-14
PIMS-1417 Agency admins can approve access request in their own agency

goemen:
Agency administrators can manage access requests to their own agency, and Sys Admins can manage all access requests.

Also, removed ag-grid dependency

### DIFF
--- a/backend/dal/Services/Admin/Concrete/UserService.cs
+++ b/backend/dal/Services/Admin/Concrete/UserService.cs
@@ -221,21 +221,17 @@ namespace Pims.Dal.Services.Admin
         /// <param name="quantity"></param>
         /// <param name="sort"></param>
         /// <param name="status"></param>
-        public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, string sort = null, AccessRequestStatus? status = null)
+        public Paged<AccessRequest> GetAccessRequests(int page = 1, int quantity = 10, string sort = null,
+            AccessRequestStatus status = AccessRequestStatus.OnHold)
         {
-            this.User.ThrowIfNotAuthorized(Permissions.AdminUsers);
+            var sortArray = new string[] { };
+            if (!string.IsNullOrWhiteSpace(sort))
+            {
+                sortArray[0] = sort;
+            }
 
-            var query = this.Context.AccessRequests
-                .Include(p => p.Agencies)
-                .ThenInclude(p => p.Agency)
-                .Include(p => p.Roles)
-                .ThenInclude(p => p.Role)
-                .Include(p => p.User)
-                .AsNoTracking();
-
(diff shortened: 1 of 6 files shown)


backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -254,6 +250,12 @@ public Paged<AccessRequest> GetAccessRequests(AccessRequestFilter filter)
                 .Include(p => p.User)
                 .AsNoTracking();
 
+            var userAgencies = this.User.GetAgencies();
```
devinleighsmith, line 253: so part of this makes sense - a system admin should get all of the access requests regardless of what agency they are in. However, what about agency admins that have no agency? won't this mean that they'll get all access requests as well?

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -221,21 +221,17 @@ public AccessRequest GetAccessRequest(int id)
...
+            var sortArray = new string[] { };
+            if (!string.IsNullOrWhiteSpace(sort))
+            {
+                sortArray[0] = sort;
+            }
 
-            query = query.Where(request => request.Status == status);
-            var accessRequests = query.Skip((page - 1) * quantity).Take(quantity);
-            return new Paged<AccessRequest>(accessRequests, page, quantity, query.Count());
+            var filter = new AccessRequestFilter(page, quantity, sortArray, null, null, null, status);
```
devinleighsmith, line 233: Could the latter 4 parameters default to null? that would let you clean this up a bit.

backend/dal/Services/Admin/Concrete/UserService.cs
devinleighsmith: Looks like the status must be non-null, in that case you could move the three nullable params to the end and default them to null.

backend/dal/Services/Admin/Concrete/UserService.cs
Fosol: You could also use the implementation in other filters where it'll parse the query string instead of manually passing arguments in a specific order.

backend/dal/Services/Admin/Concrete/UserService.cs
goemen: You right about the agency admins with no agencies, however, the chances of that happening is very low. @Fosol what do you think?

Fosol: Users shouldn't be able to get past the Access Request Page without an Agency.  Currently the only way to by-pass if if we manually go to Keycloak and add agencies.

Fosol: So it's very low

PR #225, opened 2020-05-15
Re-enable API code coverage on CI servers

asanchezr:
- Follow recommended `coverlet` approach on CI servers
- Exclude migrations from code coverage
- Restore GitHub action
- Update Jenkins pipeline

### DIFF
--- a/.github/workflows/api-dotnetcore.yml
+++ b/.github/workflows/api-dotnetcore.yml
@@ -31,24 +31,33 @@ jobs:
       - name: Test
         run: dotnet test --no-restore --verbosity normal
         working-directory: ${{env.working-directory}}
-      # - name: Generate code coverage
-      #   run: coverlet ./tests/unit/api/bin/Release/netcoreapp3.1/Pims.Api.Test.dll --target "dotnet" --targetargs "test ./ --no-build" -o "./tests/TestResults/api-coverage.json" --exclude "[*.Test]*" --exclude "[*]*Model" --exclude-by-attribute "CompilerGenerated" -f json
-      #   working-directory: ${{env.working-directory}}
-      # - name: Merge code coverage
-      #   run: coverlet ./tests/unit/dal/bin/Release/netcoreapp3.1/Pims.Dal.Test.dll --target "dotnet" --targetargs "test ./ --no-build" -o "./tests/TestResults/coverage.xml" --exclude "[*.Test]*" --exclude "[*]*Model" --exclude-by-attribute "CompilerGenerated" --merge-with "./tests/TestResults/api-coverage.json" -f opencover
-      #   working-directory: ${{env.working-directory}}
-      # - name: Codecov
-      #   uses: codecov/codecov-action@v1.0.7
-      #   with:
-      #     # User defined upload name. Visible in Codecov UI
-      #     name: PIMS
-      #     # Repository upload token - get it from codecov.io. Required only for private repositories
-      #     token: ${{env.codeCov-token}}
-      #     # Path to coverage file to upload
-      #     file: ${{env.working-directory}}/tests/TestResults/coverage.xml
-      #     # Flag upload to group coverage metrics (e.g. unittests | integration | ui,chrome)
-      #     flags: unittests
-      #     # Environment variables to tag the upload with (e.g. PYTHON | OS,PYTHON)
-      #     env_vars: C#
-      #     # Specify whether or not CI build should fail if Codecov runs into an error during upload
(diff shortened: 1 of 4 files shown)


.github/workflows/api-dotnetcore.yml
```diff
@@ -31,24 +31,33 @@ jobs:
...
+          rm -rf dal/TestResults
+          cd api
+          dotnet test --collect:"XPlat Code Coverage" --settings coverlet.runsettings --no-restore
+          mv TestResults/*/coverage.json ../TestResults/coverage.json
+          cd ..
+          cd dal
+          dotnet test --collect:"XPlat Code Coverage" --settings coverlet.runsettings --no-restore
+          mv TestResults/*/* ../TestResults/
+          cd ..
+          head TestResults/coverage.opencover.xml
```
Fosol, line 48: How does the merge of the two reports occur?

asanchezr: look at the `coverlet.runsettings` under **tests/unit/dal**. The xml has a `<MergeWith>` entry

backend/tests/unit/dal/coverlet.runsettings
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<RunSettings>
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="XPlat code coverage">
+        <Configuration>
+          <Format>json,opencover</Format>
+          <MergeWith>../TestResults/coverage.json</MergeWith>
```
asanchezr, line 8: The merge with the `API` report happens due to this config setting.

**NOTE** - if more test sub-projects are added later, we will need to keep merging with the other ones and execute test serially. Executing tests in parallel will not work with code coverage unfortunately - a restriction of the tool we are using.

.github/workflows/api-dotnetcore.yml
```diff
@@ -31,24 +31,33 @@ jobs:
...
+          mkdir -p TestResults
+          rm -rf api/TestResults
+          rm -rf dal/TestResults
+          cd api
+          dotnet test --collect:"XPlat Code Coverage" --settings coverlet.runsettings --no-restore
+          mv TestResults/*/coverage.json ../TestResults/coverage.json
+          cd ..
+          cd dal
+          dotnet test --collect:"XPlat Code Coverage" --settings coverlet.runsettings --no-restore
+          mv TestResults/*/* ../TestResults/
```
devinleighsmith, line 46: just curious, why does this line move everything from dal to ../TestResults where the api just moves coverage.json?

asanchezr: Good question - the tool we use - `coverlet` - only supports merging their own proprietary format `coverage.json` so that's why I only move that one for the api tests. This however is the "final run" where we want it to also output the coverage in **opencover** format, which is the format expected by most tools (SonarQube etc). So on this step I move them both. I could just move the `coverage.opencover.xml` here as the json file is irrelevant at this point but seemed weird to do so.

I do see your point in that it looks odd and requires you have knowledge of how the tool works. I will change the api to move all files like here for consistency. The opencover xml from api should be overwritten by the final one - I will test this before pushing an update

A comment on the action would also help I think

devinleighsmith: thanks

backend/tests/unit/dal/coverlet.runsettings
devinleighsmith: This is good to know.

.github/workflows/api-dotnetcore.yml
asanchezr: For future reference, if we have N test projects the flow of events would be:

**Pre-conditions:**
- All projects export their individual coverage percents in JSON and OpenCover format
- There's no way to merge OpenCover xmls together (that I could find)
- Common folder "../TestResults" is  git ignored so nothing gets in source control

**Steps:**

- Test-project 1 
  - generate coverage files (without merging)
  - copy results to common folder "../TestResults"
- Test-project 2
  - generate coverage files merging with previous `coverage.json`
  - the previous `coverage.opencoverage.xml` is ignored
  - copy results to common folder "../TestResults" 
...
- Test-project N
  - generate coverage files merging with previous `coverage.json`
  - the previous `coverage.opencoverage.xml` is ignored
  - copy results to common folder "../TestResults" 

The final `coverage.opencover.xml` is the one we want

PR #226, opened 2020-05-15
PIMS-2039 Fixing Postal Code bug

Fosol:
## Description

The formatting of a Postal Code resulted in a regression issue.  This PR resolves it.

### DIFF
--- a/backend/core/Extensions/StringExtensions.cs
+++ b/backend/core/Extensions/StringExtensions.cs
@@ -36,9 +36,16 @@ namespace Pims.Core.Extensions
             return env != null && env.Equals("Production", StringComparison.OrdinalIgnoreCase);
         }
 
+        /// <summary>
+        /// Formats the specified postal value.
+        /// </summary>
+        /// <param name="postal"></param>
+        /// <returns></returns>
         public static string FormatAsPostal(this string postal)
         {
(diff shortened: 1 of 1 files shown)


backend/core/Extensions/StringExtensions.cs
```diff
@@ -36,9 +36,16 @@ public static bool IsProduction(this string env)
...
+        /// Formats the specified postal value.
+        /// </summary>
+        /// <param name="postal"></param>
+        /// <returns></returns>
         public static string FormatAsPostal(this string postal)
         {
-            return postal.Insert(3, " ");
+            if (postal?.Length >= 6)
+                return postal?.ToUpper().Insert(3, " ");
+            return postal?.ToUpper();
```
asanchezr, line 48: I made a comment on the other PR that maybe this should check there's no space at index 3 before adding it as you can grow the number of spaces by calling this function over and over. 

The fix above is a step in the right direction - I still have a quick question: why check for greater than 6 chars? Aren't all postal codes in Canada 6 chars (A1A1A1) ? Wouldn't it be better to check for `postal?.Length == 6`?

Fosol: The >= was just a broad brush solution.  I've updated.

PR #228, opened 2020-05-15
PIMS-1273 add openshift bc and dc for backups, update mssql-deploy dc…

devinleighsmith:
… to use required backup volume.


openshift/templates/backup/backup-build.yaml
```diff
@@ -0,0 +1,72 @@
...
+    description:
+      The name assigned to all of the resources.  Use 'backup-{database name}'
+      depending on your database provider
+    required: true
+    value: backup-mssql
+  - name: GIT_REPO_URL
+    displayName: Git Repo URL
+    description: The URL to your GIT repo.
+    required: true
+    value: https://github.com/devinleighsmith/backup-container.git
```
Fosol, line 49: Might want to change the default value here

openshift/templates/backup/backup-build.yaml
```diff
@@ -0,0 +1,72 @@
...
+kind: Template
+apiVersion: v1
+metadata:
+  name: "${NAME}-build-template"
+  creationTimestamp:
+objects:
+  - kind: ImageStream
+    apiVersion: v1
+    metadata:
+      name: "${NAME}"
```
Fosol, line 11: May want to append to the name, or by accident someone could choose a name that exists for a different configuration.

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
Fosol, line 1: This file has one ugly name.

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
```diff
@@ -0,0 +1,11 @@
...
...mongodb/my_database\n# - mongo=mongodb:27017/my_database\n# - mssql=mssql_server:1433/my_database\n# -----------------------------------------------------------\n# Cron Scheduling:\n# -----------------------------------------------------------\n# List your backup and verification schedule(s) here as well.\n# The schedule(s) must be listed as cron tabs that \n# execute the script in 'scheduled' mode:\n#   - ./backup.sh -s\n#\n# Examples (assuming system's TZ is set to PST):\n# - 0 1 * * * default ./backup.sh -s\n#   - Run a backup at 1am Pacific every day.\n#\n# - 0 4 * * * default ./backup.sh -s -v all\n#   - Verify the most recent backups for all datbases\n#     at 4am Pacific every day.\n# -----------------------------------------------------------\n# Full Example:\n# -----------------------------------------------------------\n# postgres=postgresql:5432/TheOrgBook_Database\n# mongo=mender-mongodb:27017/useradm\n# postgres=wallet-db/tob_issuer\n# mssql=pims-db-dev:1433/pims\n#\n# 0 1 * * * default ./backup.sh -s\n# 0 4 * * * default ./backup.sh -s -v all\n# ============================================================\n0 1 * * * default ./backup.sh -s\nmssql=pims-db-dev:1433/pims"
```
Fosol, line 9: Is the formatting necessary?  It's fairly ugly to edit this way.

openshift/templates/backup/backup.conf
```diff
@@ -0,0 +1,52 @@
...
+# postgres=postgresql:5432/TheOrgBook_Database
+# mongo=mender-mongodb:27017/useradm
+# postgres=wallet-db/tob_issuer
+# mssql=pims-db-dev:1433/pims
+#
+# 0 1 * * * default ./backup.sh -s
+# 0 4 * * * default ./backup.sh -s -v all
+# ============================================================
+0 1 * * * default ./backup.sh -s
+mssql=pims-db-dev:1433/pims
```
Fosol, line 52: Can something like this be replaced by environment variables in the build config or deploy config?

openshift/templates/pims-db/mssql-deploy.yaml
```diff
@@ -90,6 +90,9 @@ objects:
             - name: mssql-data
               persistentVolumeClaim:
                 claimName: ${APP_NAME}-${COMP_NAME}-${ENV_NAME}${ID}-data
+            - name: backup
```
Fosol, line 93: Is this essentially ensuring that when deploying a database it also deploys a backup volume too?  Are persistent volumes within the same pod safe, or can they suffer the same disaster?

openshift/templates/backup/backup-build.yaml
```diff
@@ -0,0 +1,72 @@
...
+    required: false
+    value: "/docker"
+  - name: DOCKER_FILE_PATH
+    displayName: Docker File
+    description:
+      The path and file of the docker file defining the build.  Choose either
+      'Dockerfile' for Postgres builds or 'Dockerfile_Mongo' for MongoDB builds or 'Dockerfile_MSSQL'
+      for MSSQL builds.
+    required: false
+    value: Dockerfile_MSSQL
```
Fosol, line 67: I don't think I see these dockerfiles anywhere?  Should they be part of the PR?

openshift/templates/backup/backup-deploy.overrides.sh
```diff
@@ -0,0 +1,41 @@
+_includeFile=$(type -p overrides.inc)
```
Fosol, line 1: What calls this file?

openshift/templates/backup/backup-build.yaml
asanchezr: Agree

openshift/templates/backup/backup-build.yaml
asanchezr: If I understood correctly, they are coming from a separate "base" repo that has all the logic related to the `backup-container`

https://github.com/BCDevOps/backup-container/blob/master/docker/Dockerfile

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
asanchezr: If you use JSON you are forced to use `\n` for line breaks. YAML provides a nicer, human-readable alternative but I don't know if the backup tool supports that

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
asanchezr: Agree with the poor naming but have a feeling this is what the backup tool expects...
If not, we could rename it but not required for this PR to go thru

openshift/templates/backup/backup-deploy.yaml
```diff
@@ -0,0 +1,432 @@
...
+    required: true
+    value: netapp-file-standard
+  - name: VERIFICATION_VOLUME_MOUNT_PATH
+    displayName: Verification Volume Mount Path
+    description:
+      The path on which to mount the verification volume.  This is used by
+      the database server to contain the database configuration and data files. For
+      Mongo, please use /var/lib/mongodb/data . For MSSQL, please use /var/opt/mssql/data
+    required: true
+    value: "/var/lib/pgsql/data"
```
asanchezr, line 412: Shouldn't we use `/var/opt/mssql/data` here as our default?

openshift/templates/backup/backup-deploy.yaml
```diff
@@ -0,0 +1,432 @@
...
+    value: "/"
+  - name: BACKUP_VOLUME_NAME
+    displayName: Backup Volume Name
+    description:
+      The name of the persistent volume used to store the backups.  Please
+      note, when using the recommended nfs-backup storage class the name of the pvc
+      MUST be taken from the manually provisioned claim; nfs-backup storage MUST be
+      provisioned manually.
+    required: true
+    value: bk-jcxjin-dev-r9c53stkggh6
```
asanchezr, line 361: Ideally I'd like to see this value removed so - in combination with being a required field - forces the user to provide a value for it...

I suspect if someone runs this template and forgets to pass `BACKUP_VOLUME_NAME`, that **bk-jcxjin-dev-r9c53stkggh6** won't be there on their environment to work

openshift/templates/pims-db/mssql-deploy.yaml
```diff
@@ -221,3 +224,8 @@ parameters:
     description: Volume space available for data, e.g. 512Mi, 2Gi
     value: 5Gi
     required: true
+  - name: "BACKUP_VOLUME_NAME"
+    displayName: "Backup Volume Name"
+    description: The name of the persistent volume used to store the backups.  Please note, when using the recommended nfs-backup storage class the name of the pvc MUST be taken from the manually provisioned claim; nfs-backup storage MUST be provisioned manually.
+    required: true
+    value: bk-jcxjin-dev-r9c53stkggh6
```
asanchezr, line 231: Same comment as before - we should clear this value (I think?)

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
devinleighsmith: I'd prefer to avoid renaming files copied from the backup-conf for traceability reasons.

openshift/templates/backup/backup-conf-configmap_DeploymentConfig.json
devinleighsmith: This file is 99% copied from the backup-conf file. Just the very bottom cron/connection string information is something I've added.

openshift/templates/backup/backup.conf
devinleighsmith: this ends up being listed in a config map. I guess you can think of this as a sensible default - we may end up with multiple lines for cron jobs here.

openshift/templates/backup/backup-build.yaml
devinleighsmith: Yes, I think it makes sense to include a minimal amount of files from the backup container repo for tracking purposes - but in general we should use the base repo because that is where new features fixes and updates will show up.

openshift/templates/pims-db/mssql-deploy.yaml
devinleighsmith: Have a read through the description of NFS storage on the backup repo:
https://github.com/BCDevOps/backup-container

openshift/templates/backup/backup-deploy.overrides.sh
devinleighsmith: I think I'lll actually remove these files - I think it makes more sense to keep the generated pims specific file in our repo and leave the scripts in the repo to reduce the risk of our versions of scripts getting out of sync.

openshift/templates/backup/backup-deploy.yaml
devinleighsmith: yep, I'll fix this.

openshift/templates/backup/backup-deploy.yaml
devinleighsmith: fair, I'll remove this.

PR #229, opened 2020-05-15
PIMS-2038, 1838, 1708

jdtoombs:
There was something strange going on with the classification values being passed when adding a new building (it would default to the property's classification). Causing [this problem.](https://pimsteam.atlassian.net/browse/PIMS-2038?atlOrigin=eyJpIjoiMGJlM2E2MTEzOGQ3NDU5MmE1M2VlZGE3NzllNTdiMjUiLCJwIjoiaiJ9). Now the user is defaulted to no value in the building classification and when selected the values pass through correctly.

There is also another bug that the City appears to be auto filled (when adding a new building) but is not actually sending a value. I will address this after I get my remaining stories done.

### DIFF
--- a/frontend/src/forms/ParcelDetailForm.test.tsx
+++ b/frontend/src/forms/ParcelDetailForm.test.tsx
@@ -176,7 +176,7 @@ describe('ParcelDetailForm', () => {
       });
       const errors = getAllByText('Required');
       const idErrors = getAllByText('PID or PIN Required');
-      expect(errors).toHaveLength(9);
+      expect(errors).toHaveLength(8);
       expect(idErrors).toHaveLength(2);
     });
 
--- a/frontend/src/forms/subforms/BuildingForm.tsx
+++ b/frontend/src/forms/subforms/BuildingForm.tsx
@@ -43,6 +43,8 @@ export const defaultBuildingValues: any = {
   buildingConstructionTypeId: '',
   buildingPredominateUse: undefined,
   buildingPredominateUseId: '',
+  classificationId: '',
+  classification: undefined,
   buildingOccupantType: undefined,
   buildingOccupantTypeId: '',
   transferLeaseOnSale: false,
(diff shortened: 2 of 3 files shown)


frontend/src/forms/subforms/BuildingForm.tsx
```diff
@@ -78,21 +80,11 @@ const BuildingForm = <T extends any>(props: BuildingProps & FormikProps<T>) => {
...
 
-  //set the lat/lon of this building, but only if the building lat/lon hasn't been touched and has no value.
-  if (
-    !getIn(props.touched, withNameSpace('latitude')) &&
-    !getIn(props.values, withNameSpace('latitude')) &&
-    props.values.latitude &&
-    props.touched.latitude
-  ) {
+  //set the lat/lon of this building, but only if the building lat/lon has no value.
+  if (!getIn(props.values, withNameSpace('latitude')) && props.values.latitude) {
```
jdtoombs, line 84: Was never reaching these statements when it was checking if the values were not touched. Now it fills the lat long as expected and can update it afterwards if it is not where desired.

frontend/src/forms/ParcelDetailForm.test.tsx
```diff
@@ -176,7 +176,7 @@ describe('ParcelDetailForm', () => {
       });
       const errors = getAllByText('Required');
       const idErrors = getAllByText('PID or PIN Required');
-      expect(errors).toHaveLength(9);
+      expect(errors).toHaveLength(8);
```
jdtoombs, line 179: estimated no longer required

frontend/src/constants/evaluationKeys.tsx
```diff
@@ -1,4 +1,5 @@
 export enum EvaluationKeys {
```
jdtoombs, line 1: Woops this is causing bugs and is supposed to be FiscalKeys - fixing now

PR #231, opened 2020-05-17
Add Makefiles. Update documentation

asanchezr:
This will support more automation of manual tasks. 
In this update I have provided make commands that automate refreshing the database


asanchezr:
> Nice addition to make life better. Did you choose makefile for the syntax benefits? Their main purpose is for compilation.

Yes, their main (original) purpose was for compilation of C++ code but they can execute any shell command. It's been gaining popularity lately and thought the team might want to try it.

My goal would be to get to what Team Common Components have done with their [Makefile](https://github.com/bcgov/openshift-launchpad/blob/develop/Makefile)

Notice how all their commands (docker, oc, etc) can be "abstracted" away and you no longer have to remember specifics about docker - "is it `docker run` or `docker build` or `docker whatever -d`"?

We could replace all that with simple make commands:

```bash
make run # <-- docker-compose up -d
make build # <-- docker-compose build --no-cache
make clean 
# ^^ @docker-compose rm -f -v -s
# @docker volume rm -f frontend-node_modules-data
# ...
```

To me it feels a better dev experience where we unify on make commands and forget about specific syntax of various tools we use. Of course, all this is optional - you can still type those long commands if you so desire or need more control.

I also like that make let's you define "pre-conditions". Take a look at the Makefile I added to the frontend project - `make test` will check if your `node_modules` folder exists, if not it will run `npm ci` for you, then it will run the linter and formatting and then the tests. So to me, before I submit a PR all I have to do in the frontend folder is `make test` and if everything passes I know the PR won't fail the GH actions. Again, all this is optional but makes life easier I think.

Finally, `make` comes pre-installed on macOS and Linux and can easily be installed on Windows via `chocolatey` or `scoop` so I don't think it would be a big ask for devs to install.

PR #233, opened 2020-05-19
PIMS-1782 view link on property list view

asanchezr:


frontend/src/features/properties/list/columns.tsx
```diff
@@ -22,6 +23,7 @@ export const columns: ColumnWithProps<IProperty>[] = [
     Header: 'Agency',
     accessor: 'agencyCode', // accessor is the "key" in the data
     align: 'left',
+    width: 65,
```
devinleighsmith, line 26: I'm not used to seeing static widths on flex based grids, how does this work for resizing?

asanchezr: I believe it provides a percentage-based sizing. Not like 65% but for example, all columns in `react-table` are 150px by default so in this case I'm making this column roughly half-size of the others.

You can try with what's already in DEV and expand/collapse your browser window. Even if columns are listed at 150px they should grow/shrink as appropriate - at least according to react-table docs. If they don't then we would have to look at it but haven't done extensive testing honestly only a couple window resizes

asanchezr: NVM I tested and it is showing the full width, not shrinking or growing. I'll see later if I can make it more responsive but there might not be much since it has lots of data. Most likely will scroll horizontally unless you have a huge screen

devinleighsmith: thanks for checking.

asanchezr:
I have now made the table responsive by using width percentages

responsive-table-video

PR #234, opened 2020-05-19
PIMS-1420 Adding ParcelId to Properties

Fosol:
## Description

Adds the `ParcelId` to the results so that buildings can include a link to their parent parcel.

Note - This will require either refreshing your DB or manually running the SQL script.

### DIFF
--- a/backend/api/Areas/Property/Models/Search/PropertyModel.cs
+++ b/backend/api/Areas/Property/Models/Search/PropertyModel.cs
@@ -2,108 +2,263 @@ using System;
 
 namespace Pims.Api.Areas.Property.Models.Search
 {
+    /// <summary>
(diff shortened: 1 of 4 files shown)


backend/api/Areas/Property/Models/Search/PropertyModel.cs
```diff
@@ -2,108 +2,263 @@
...
+        /// </summary>
         public string Classification { get; set; }
 
+        /// <summary>
+        /// get/set - The GIS latitude location of the property.
+        /// </summary>
         public double Latitude { get; set; }
 
+        /// <summary>
+        /// get/set - The GIS latitude location of the property.
```
devinleighsmith, line 57: /nit longitude

PR #235, opened 2020-05-19
PIMS-1272 run the db migration as a pre-hook container.

devinleighsmith:


.s2i/bin/assemble
```diff
@@ -0,0 +1,266 @@
+#!/bin/bash
```
Fosol, line 1: I want to understand this file more.  I know it's OpenShift's way of creating source-to-image containers.  Why is it in the root?  Shouldn't it exist in the `/api` since it's only related to that?  Why do we need this for this feature of db migration?

openshift/templates/pims-api/deploy.yaml
```diff
@@ -156,6 +156,22 @@ objects:
...
+          intervalSeconds: 1
+          maxSurge: 25%
+          maxUnavailable: 25%
+          pre:
+            execNewPod:
+              command:
+                - bash
+                - "-c"
+                - >-
+                  dotnet ef database update --project=api --context=pimsContext
```
Fosol, line 169: Does this pick up all the configuration for the `api`?

.s2i/bin/assemble
devinleighsmith: Yeah, this is the unfortunate result of MS not creating their images with Openshift in mind. Without this the random uid assigned by openshift doesn't have permissions to execute the dotnet-ef tool installed during the build. See the comment slightly below this for more details. 

I think this could maybe go into /api? I hate how openshift automagically finds and detects these in source, it feels very unsafe to me.

openshift/templates/pims-api/deploy.yaml
devinleighsmith: it ran as I expected in my simple test, not sure what else I can do here for validation.

PR #236, opened 2020-05-20
PIMS-1766 Replace user MUI Datatable with react table

goemen:


frontend/src/components/Table/PageSizeSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Menu options={pageSizeOptions} width="60px">
+      <div style={{ display: 'flex' }}>
+        <span>Show</span>
+        <Form.Control
+          size="sm"
+          value={`${selected}`}
+          type="number"
+          style={{ width: 50, marginLeft: 10, marginRight: 10 }}
+          onChange={() => {}}
```
Fosol, line 35: Why not make `onChange` optional?

frontend/src/components/Table/Table.tsx
```diff
@@ -82,54 +93,117 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
...
+    );
+  };
+
+  const onPageSizeChange = (size: number) => {
+    props.onPageSizeChange && props.onPageSizeChange(size);
+  };
+
+  const renderLoading = () => {
+    return (
+      <div className="table-loading">
```
Fosol, line 126: No need to change, but this looks like a nice thing to allow customization.

PR #239, opened 2020-05-20
Enabling Project Number

jdtoombs:
Project Number now editable on the Parcel Detail Form when the user has the `admin-properties` claim.

Just wanted to do this while I thought of it.

### DIFF
--- a/frontend/src/forms/subforms/PidPinForm.tsx
+++ b/frontend/src/forms/subforms/PidPinForm.tsx
@@ -4,6 +4,8 @@ import { Input, Form } from 'components/common/form';
 import { Col } from 'react-bootstrap';
 import TooltipIcon from 'components/common/TooltipIcon';
 import { PidTooltip, PinTooltip } from 'forms/strings';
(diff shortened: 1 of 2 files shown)


frontend/src/forms/subforms/PidPinForm.tsx
```diff
@@ -33,6 +34,17 @@ const PidPinForm: FunctionComponent<PidPinProps> = (props: PidPinProps) => {
     const { nameSpace } = props;
     return nameSpace ? `${nameSpace}.${fieldName}` : fieldName;
   };
+
+  const keycloak = useKeycloakWrapper();
+
+  const disableProjectNumber = () => {
+    if (keycloak.hasClaim('admin-properties')) {
```
goemen, line 41: I would change this constant to projectNumberDisabled and
`const projectNumberDisabled = !keycloak.hasClaim(Claims.ADMIN_PROPOERTIES)`

jdtoombs: Ooh that is way cleaner, on it.

PR #241, opened 2020-05-21
PIMS-2057, PIMS-2043 Project DB Changes

Fosol:
## Description

Updating documentation

As is the nature of change, the initial seed scripts needed to be updated based on latest requirement.

Additionally we are going to provide a way to dynamically control the project status workflow (to a limited degree).


backend/dal/Configuration/ProjectStatusConfiguration.cs
```diff
@@ -23,6 +23,12 @@ public override void Configure(EntityTypeBuilder<ProjectStatus> builder)
 
             builder.Property(m => m.Description).HasMaxLength(1000);
 
+            builder.Property(m => m.Route).IsRequired();
+            builder.Property(m => m.Route).HasMaxLength(150);
+
+            builder.Property(m => m.Workflow).IsRequired();
```
Fosol, line 29: This property provides us a way to query for a subset of the project status.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -6,70 +6,94 @@ INSERT INTO dbo.[ProjectStatus] (
...
     , 'Draft'
     , 0
     , 'A new draft project that is not ready to submit to apply to be added to the Surplus Property Program.'
+    , '/project/draft'
+    , 'SubmitDisposal'
 ), (
     1
     , 1
-    , 'Properties Selected'
+    , 'Select Properties'
```
Fosol, line 22: Project status updated based on latest requirements and discussion.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -6,70 +6,94 @@ INSERT INTO dbo.[ProjectStatus] (
...
     , [Description]
+    , [Route]
+    , [Workflow]
 ) VALUES (
     0
     , 0
     , 'Draft'
     , 0
     , 'A new draft project that is not ready to submit to apply to be added to the Surplus Property Program.'
+    , '/project/draft'
```
Fosol, line 17: These `routes` will need to exist in some format in the APP.

backend/dal/Migrations/Initial/PostDeploy/01-Tasks.sql
```diff
@@ -4,52 +4,28 @@ SET IDENTITY_INSERT dbo.[Tasks] ON
...
+    , [Name]
     , [IsDisabled]
     , [Description]
     , [IsOptional]
     , [SortOrder]
 ) VALUES (
     1
-    , 'Update Financials'
     , 1
+    , 'Surplus Declaration & Readiness Checklist completed and sent'
```
Fosol, line 16: Tasks updated based on latest requirements and discussion.

backend/entities/Project.cs
```diff
@@ -19,6 +19,16 @@ public class Project : BaseEntity
...
 
+        /// <summary>
+        /// get/set - The foreign key to the owning agency.
+        /// </summary>
+        public int AgencyId { get; set;  }
+
+        /// <summary>
+        /// get/set - The owning agency.
+        /// </summary>
+        public Agency Agency { get; set; }
```
Fosol, line 30: Adding link to agency

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
asanchezr: My initial reaction would be to question why would the backend need to dictate a client-side routing concern? But I'd love to learn more what's the rationale behind this change?

backend/dal/Configuration/ProjectStatusConfiguration.cs
asanchezr: Could you please expand what this property is for? I'm not sure I follow what you mean here... thanks

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
Fosol: It was an option.  Either the APP maintains constants that link to the primary keys for each project status, or the db has a way to configure what each step says and does (limited).  We choose the db configurable way.  Now the APP makes a request to an endpoint to find out what the workflow is and uses it to move from page to page.  It of course isn't fully configurable, but it does allow the APP to do less hard-coding.

backend/dal/Configuration/ProjectStatusConfiguration.cs
Fosol: There are many project statuses, and most likely many more to come.  But we only want to extract the relevant status for a specific workflow (Submit Disposal Project).  This way when creating a new disposal project we can simply make a request for a specific workflow and we'll get back the steps/status that are relevant and will be displayed in the APP.

New Project -> Select Properties -> Update Information -> Required Documents -> ... -> Submit

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
asanchezr: I recognize this is a first stab at designing a "workflow engine" on the backend and am happy to see how it works with this iteration. However, I'd like to propose that we discuss the design at a later time (not too late).

About the APP having to keep constants I'm not sure that's necessary as we already have a way to get lookup values from the server. We could have an endpoint like:

```
GET /workflows/{id}/available-transitions
```

That returns those constants - say for our first workflow "SubmitDisposal" (e.g. id = 1)

```
GET /workflows/{id}/available-transitions

# response

{
  transitions: [
    {
       id: 1,
       name: "Draft",
       order: 1  // first step
    },
    {
      id: 2,
      name: "Select Properties"
      order: 2  // second step
    },
    ....
  ]
}
```

Fosol: Or we can do a nearly identical solution which is what we're doing `ProjectStatus` = a list of valid transitions that a project can go through.

asanchezr: I'm no expert in designing workflow engines but a couple google searches yield good resources:

- https://www.kennethlange.com/how-to-model-workflows-in-rest-apis/
- http://joyent.github.io/node-workflow/workflowapi.html

Fosol: We're not looking for an elaborate workflow engine, simply an array of steps/stages/statuses that we can configure with small details (name, description, order, route).  This reduces the requirement to hard-code primary keys into the APP.

Fosol: Additionally your first link 3rd example is essentially what is here.

Fosol:
@asanchezr Devin and I discussed the option.  At the end of the day the APP can choose to use whatever route it prefers, but this way it doesn't need to maintain it's own library of constants to match pages with status/steps/stages.  All the APP needs to do is have a route that matches the DB value.

PR #243, opened 2020-05-21
PIMS-2044 Disposal Project endpoints

Fosol:
## Description

Basic endpoints for managing projects.

## Endpoints

The following are the new endpoints that provide a way to view, add, update and delete projects.

- GET : /api/projects/disposal/{projectNumber}
- POST : /api/projects/disposal
- PUT: /api/projects/disposal/{projectNumber}
- DELETE : /api/projects/disposal/{projectNumber}

- GET : /api/lookup/project/tier/levels
- GET : /api/lookup/all

The following provide the project status/stages and checklists.
- GET : /api/projects/disposal/workflow
- GET : /api/projects/disposal/tasks

> Please note I have only tested this with Postman and I did not add/remove properties from the project yet.

### DIFF
--- a/backend/api/Areas/Project/Models/Dispose/TaskModel.cs
+++ b/backend/api/Areas/Project/Models/Dispose/TaskModel.cs
@@ -1,0 +1,27 @@
+using Pims.Dal.Entities;
+
+namespace Pims.Api.Areas.Project.Models.Dispose
+{
+    /// <summary>
+    /// TaskModel class, provides a model to represent a task.
+    /// </summary>
+    public class TaskModel : Api.Models.LookupModel<int>
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - The foreign key to the task type.
+        /// </summary>
+        public TaskTypes TaskType { get; set; }
+
+        /// <summary>
+        /// get/set - The task description.
+        /// </summary>
+        public string Description { get; set; }
+
+        /// <summary>
+        /// get/set - Whether the task is optional.
+        /// </summary>
+        public bool IsOptional { get; set; }
+        #endregion
+    }
+}
--- a/backend/api/Mapping/Converters/AgencyConverter.cs
+++ b/backend/api/Mapping/Converters/AgencyConverter.cs
@@ -1,5 +1,3 @@
-using System.Collections.Generic;
-using MapsterMapper;
 using Entity = Pims.Dal.Entities;
 
 
@@ -10,26 +8,26 @@ namespace Pims.Api.Mapping.Converters
     /// </summary>
     public static class AgencyConverter
     {
+        /// <summary>
+        /// Extracts the agency code, or the specified 'agency' is a child it will return the parent code.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns></returns>
         public static string ConvertAgency(Entity.Agency source)
         {
             if (source?.ParentId == null) return source?.Code;
             return source.Parent?.Code;
         }
 
+        /// <summary>
+        /// Extract the sub-agency code.
+        /// If the specified 'agency' is a parent it will return null.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns></returns>
         public static string ConvertSubAgency(Entity.Agency source)
         {
             return source?.ParentId == null ? null : source.Code;
-        }
-
-        public static ICollection<Entity.AccessRequestAgency> ConvertAgency(this IMapper mapper, Entity.AccessRequest destination, Models.User.AccessRequestModel source, IEnumerable<Models.User.AgencyModel> sourceMember)
-        {
-            foreach (var agency in sourceMember)
-            {
-                var accessRequestAgency = mapper.Map<Entity.AccessRequestAgency>(agency);
-                accessRequestAgency.AccessRequestId = source.Id;
-                destination.Agencies.Add(accessRequestAgency);
-            }
-            return destination.Agencies;
         }
     }
 }
--- a/backend/dal/Services/Concrete/ParcelService.cs
+++ b/backend/dal/Services/Concrete/ParcelService.cs
@@ -128,7 +128,7 @@ namespace Pims.Dal.Services
                     (isAdmin || !p.IsSensitive || (viewSensitive && userAgencies.Contains(p.AgencyId)))) ?? throw new KeyNotFoundException();
 
             // Remove any sensitive buildings from the results if the user is not allowed to view them.
-            parcel?.Buildings.RemoveAll(b => b.IsSensitive && !userAgencies.Contains(b.AgencyId));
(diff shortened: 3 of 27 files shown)


backend/api/Mapping/Converters/AgencyConverter.cs
```diff
@@ -10,26 +8,26 @@ namespace Pims.Api.Mapping.Converters
...
+        /// If the specified 'agency' is a parent it will return null.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <returns></returns>
         public static string ConvertSubAgency(Entity.Agency source)
         {
             return source?.ParentId == null ? null : source.Code;
         }
-
-        public static ICollection<Entity.AccessRequestAgency> ConvertAgency(this IMapper mapper, Entity.AccessRequest destination, Models.User.AccessRequestModel source, IEnumerable<Models.User.AgencyModel> sourceMember)
```
Fosol, line 24: Not used or required.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -128,7 +128,7 @@ public Parcel Get(int id)
                     (isAdmin || !p.IsSensitive || (viewSensitive && userAgencies.Contains(p.AgencyId)))) ?? throw new KeyNotFoundException();
 
             // Remove any sensitive buildings from the results if the user is not allowed to view them.
-            parcel?.Buildings.RemoveAll(b => b.IsSensitive && !userAgencies.Contains(b.AgencyId));
+            parcel?.Buildings.RemoveAll(b => !isAdmin && b.IsSensitive && !userAgencies.Contains(b.AgencyId));
```
Fosol, line 131: Corrected issue where an administrator would not be able to see sensitive buildings.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -0,0 +1,285 @@
...
+                .Include(p => p.Properties).ThenInclude(b => b.Building).ThenInclude(b => b.Evaluations)
+                .Include(p => p.Properties).ThenInclude(b => b.Building).ThenInclude(b => b.Fiscals)
+                .AsNoTracking()
+                .FirstOrDefault(p => p.ProjectNumber == projectNumber &&
+                    (isAdmin || userAgencies.Contains(p.AgencyId))) ?? throw new KeyNotFoundException();
+
+            // Remove any sensitive properties from the results if the user is not allowed to view them.
+            if (!viewSensitive)
+            {
+                project?.Properties.RemoveAll(p => p.Parcel.IsSensitive);
```
devinleighsmith, line 85: Does it make sense to remove them or should we throw an error?

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -0,0 +1,285 @@
...
+            this.User.ThrowIfNotAuthorized(Permissions.PropertyAdd);
+
+            var agency = this.User.GetAgency(this.Context) ??
+                throw new NotAuthorizedException("User must belong to an agency before adding projects.");
+
+            project.AgencyId = agency.Id; // Always assign the current user's agency to the project.
+            project.Agency = agency;
+            project.StatusId = 0; // Always start a project as a Draft.
+            project.Status = this.Context.ProjectStatus.Find(project.StatusId);
+            project.TierLevel = this.Context.TierLevels.Find(project.TierLevelId);
```
devinleighsmith, line 108: Do you not have to do something similar for the properties?

backend/dal/Services/Concrete/ProjectService.cs
Fosol: This would only be for the purpose of viewing.  Someone is viewing a project (which they are allowed to do) but they're not allowed to see the sensitive properties.

backend/dal/Services/Concrete/ProjectService.cs
Fosol: Adding properties to projects isn't completed (although it might have basic functionality).

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        /// <summary>
+        /// Update the specified project in the datasource.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        /// <returns></returns>
+        public Project Update(Project project)
+        {
+            project.ThrowIfNotAllowedToEdit(nameof(project), this.User, new[] { Permissions.PropertyEdit, Permissions.AdminProperties });
```
devinleighsmith, line 133: can users without sensitive permissions edit projects with sensitive properties?

Fosol: Hmmm, I suspect we shouldn't allow them in that case.

devinleighsmith: That may be an issue if we update this to support removing properties. Users without the permission will get a filtered list of properties, and when submitting that back to the DB it'll look like the sensitive property was removed.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -0,0 +1,285 @@
...
+            }
+            return project;
+        }
+
+        /// <summary>
+        /// Add the specified project to the datasource.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <returns></returns>
+        public Project Add(Project project)
```
devinleighsmith, line 96: Do we care if a property is in more than one project?

backend/dal/Services/Concrete/ProjectService.cs
Fosol: I think the easiest is just not allow them to edit in this case.

backend/dal/Services/Concrete/ProjectService.cs
Fosol: We do, but there is an issue we'll need to solve which is when a property has been in a prior project (approved/denied/transferred) and now it is part of another.  Which means we don't want a property to be part of multiple 'open' projects.

backend/api/Areas/Project/Models/Dispose/TaskModel.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// get/set - The task description.
+        /// </summary>
+        public string Description { get; set; }
+
+        /// <summary>
+        /// get/set - Whether the task is optional.
+        /// </summary>
+        public bool IsOptional { get; set; }
+        #endregion
```
asanchezr, line 25: Question - `sortOrder` wouldn't be required in the frontend to display the list in the right order?

asanchezr: NVM - I see now that his is the generic task. The project task has sortOrder

Fosol: It probably should still be added to the model but isn’t required atm

PR #244, opened 2020-05-22
PIMS-2070 Adding Search Projects

Fosol:
## Description

Adding way to search for projects and the ability to page through the results of a filter.

- GET : /api/projects/search/page?{filter}
- POST : /api/projects/search/page/filter

> Initial endpoint will most likely need to be enhanced once requirements are completed.

### DIFF
--- a/backend/entities/Views/Property.cs
+++ b/backend/entities/Views/Property.cs
@@ -1,4 +1,5 @@
 using System;
+using System.ComponentModel.DataAnnotations.Schema;
 
 namespace Pims.Dal.Entities.Views
 {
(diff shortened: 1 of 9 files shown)


backend/entities/Views/Property.cs
```diff
@@ -117,6 +118,7 @@ public class Property
         /// <summary>
         /// get/set - The most recent estimate.
         /// </summary>
+        [Column(TypeName = "MONEY")]
```
goemen, line 121: Is it possible to create a constant/enum or something for the values of the TypeName?

PR #245, opened 2020-05-22
PIMS-2077 Project Number Auto-Generation

Fosol:
## Description

When adding new project this will auto-generate the Project Number.
The format of the number is configurable in the appsettings.config.

> This will require another DB refresh.

### DIFF
--- a/backend/api/appsettings.json
+++ b/backend/api/appsettings.json
@@ -53,6 +53,9 @@
   "Pims": {
     "ServiceAccount": {
       "Email": "support@pims.com"
+    },
+    "Project": {
+      "NumberFormat": "SPP-{0:00000}"
     }
   },
   "Keycloak": {
--- a/backend/dal/Configuration/ProjectConfiguration.cs
+++ b/backend/dal/Configuration/ProjectConfiguration.cs
@@ -29,7 +29,6 @@ namespace Pims.Dal.Configuration
             builder.HasOne(m => m.Agency).WithMany().HasForeignKey(m => m.AgencyId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.TierLevel).WithMany().HasForeignKey(m => m.TierLevelId).OnDelete(DeleteBehavior.ClientSetNull);
 
-            builder.HasIndex(m => new { m.Name }).IsUnique();
             builder.HasIndex(m => new { m.Name, m.StatusId, m.TierLevelId, m.AgencyId });
 
             base.Configure(builder);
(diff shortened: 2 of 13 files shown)


backend/api/appsettings.json
```diff
@@ -53,6 +53,9 @@
   "Pims": {
     "ServiceAccount": {
       "Email": "support@pims.com"
+    },
+    "Project": {
+      "NumberFormat": "SPP-{0:00000}"
```
Fosol, line 58: Configuration for project numbers.

backend/dal/Configuration/ProjectConfiguration.cs
```diff
@@ -29,7 +29,6 @@ public override void Configure(EntityTypeBuilder<Project> builder)
             builder.HasOne(m => m.Agency).WithMany().HasForeignKey(m => m.AgencyId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.TierLevel).WithMany().HasForeignKey(m => m.TierLevelId).OnDelete(DeleteBehavior.ClientSetNull);
 
-            builder.HasIndex(m => new { m.Name }).IsUnique();
```
Fosol, line 32: Not required to be unique.

backend/dal/Configuration/ProjectNumberConfiguration.cs
```diff
@@ -0,0 +1,25 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Pims.Dal.Entities;
+
+namespace Pims.Dal.Configuration
+{
+    /// <summary>
+    /// ProjectNumberConfiguration class, provides a way to configure project numbers in the database.
+    ///</summary>
+    public class ProjectNumberConfiguration : BaseEntityConfiguration<ProjectNumber>
```
Fosol, line 10: We use the `ProjectNumber` to provide a way to generate unique numbers for **Projects**.  The reason we don't simply have **Project.Id** is because we wouldn't be able to at the same time generate the **Project.ProjectNumber** (which is the actual primary key and unique value to identify the project.

The downside is we can generate project numbers that get orphaned under some conditions.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectNumbers.sql
```diff
@@ -0,0 +1,3 @@
+PRINT 'Set Identity Seed for ProjectNumbers'
+
+DBCC CHECKIDENT ([ProjectNumbers], RESEED, 10000)
```
Fosol, line 3: Set the initial ProjectNumber.Id value.

PR #249, opened 2020-05-25
PIMS-1841 Confirm approval

goemen:
@devinleighsmith can you chime in here. I can use some guidance

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -37,6 +37,7 @@
     "redux": "4.0.5",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
+    "styled-components": "5.1.0",
     "text-mask-addons": "3.8.0",
     "typescript": "3.7.3",
     "yup": "0.28.3"
@@ -69,6 +70,7 @@
     "@types/redux-logger": "3.0.7",
     "@types/redux-mock-store": "1.0.2",
     "@types/text-mask-addons": "3.8.0",
+    "@types/styled-components": "5.1.0",
     "@types/yup": "0.26.32",
     "axios-mock-adapter": "1.17.0",
     "canvas": "2.6.1",
--- a/frontend/src/components/project/disposal/ApprovalConfirmation.tsx
+++ b/frontend/src/components/project/disposal/ApprovalConfirmation.tsx
@@ -1,0 +1,48 @@
+import * as React from 'react';
+import { Form } from 'react-bootstrap';
+import { Field } from 'formik';
+
+export type OnChangeType = (checked: boolean) => void;
+
+interface IApprovalConfirmationProps {
+  onChange?: OnChangeType;
+  field: string;
+  checked?: boolean;
+  userDisplayName: string;
+}
+
+/**
+ * A component for displaying an approval confirmation checkbox
+ * @component
+ * @example
+ * const onChange = (checked) => console.log('confirmed');
+ * const userDisplayName = 'PIMS User';
+ * const fieldName = 'confirmation';
+ * return (
+ *
+ * <Formik initialValues={{[fieldName]: false}}>
+ *    <Form>
+ *        <ApprovalConfirmationCheckbox field={fieldName} onChange={onChange} userDisplayName={userDisplayName} />
+ *    </Form>
+ * </Formik>
+ * );
+ */
+export const ApprovalConfirmationCheckbox: React.FC<IApprovalConfirmationProps> = ({
+  userDisplayName,
+  field,
+  onChange,
+}) => {
+  const label = `I, ${userDisplayName}, confirm by checking this box that I have received the approval 
+    to request the property be submitted to the Enhanced Referral Program and marketed internally to other Ministry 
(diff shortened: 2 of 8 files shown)


frontend/src/pages/ApprovalConfirmation.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const save = () => {};
+
+  const next = () => {};
+
+  return (
+    <Wrapper>
+      <Title>Approval Confirmation</Title>
+      <ConfirmationCheckbox onChange={checked => setConfirm(checked)} />
+      <StepActions onNext={next} onSave={save} />
```
devinleighsmith, line 63: @goemen I think it makes the most sense if you just write this component independently of the others I'm working on, and then I handle integrating it with my code. As a result, I'm fine with this as-is

goemen: On the integration part, is there any input props or events that this component has to support?

goemen: Like for instance, I have to check if the project was approved by exemption or not, and display different text on the page. How do I make this check?

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -51,49 +51,57 @@ INSERT INTO dbo.[ProjectStatus] (
 ), (
     5
     , 5
+    , 'Approval Confirmation'
```
Fosol, line 54: This isn't correct based on the template provided by Christine and approved by Mikaela.

> Draft > Select Properties > Update Information > Required Documentation > Approval > Submitted

The user will need to confirm approval before the project can move from **Approval** to **Submitted**.  There is no requirement for an additional Project Status.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -51,49 +51,57 @@ INSERT INTO dbo.[ProjectStatus] (
...
-    10
-    , 10
+    11
+    , 11
     , 'Denied'
     , 0
     , 'The project has been denied to be added to the Surplus Property Program.'
     , '/project/denied'
-    , 'ReviewDisposal'
+    , 'SubmitDisposal'
```
Fosol, line 106: This is also incorrect at this point in time.  The template does not provide a **Denied** state on the **Start a Project**

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
goemen: I had a short discussion with @devinleighsmith, it is not clear where the Approval Confirmation step should be

goemen: Ok, looks like it is in Review and Submit

frontend/package.json
```diff
@@ -37,6 +38,7 @@
     "redux": "4.0.5",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
+    "styled-components": "^5.1.0",
```
asanchezr, line 41: Please remove the "^" here

frontend/package.json
```diff
@@ -6,6 +6,7 @@
     "@bcgov/bc-sans": "1.0.1",
     "@react-keycloak/web": "2.1.0",
     "@reduxjs/toolkit": "1.2.3",
+    "@types/styled-components": "^5.1.0",
```
asanchezr, line 9: Please move this typing reference to `devDependencies`. Thanks

frontend/src/components/project/disposal/ApprovalConfirmation.tsx
```diff
@@ -0,0 +1,48 @@
...
+export type OnChangeType = (checked: boolean) => void;
+
+interface IApprovalConfirmationProps {
+  onChange?: OnChangeType;
+  field: string;
+  checked?: boolean;
+  userDisplayName: string;
+}
+
+/**
```
Fosol, line 14: Documentation!  I am so happy to see this!

frontend/src/components/project/disposal/ApprovalConfirmation.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * return (
+ *
+ * <Formik initialValues={{[fieldName]: false}}>
+ *    <Form>
+ *        <ApprovalConfirmationCheckbox field={fieldName} onChange={onChange} userDisplayName={userDisplayName} />
+ *    </Form>
+ * </Formik>
+ * );
+ */
+export const ApprovalConfirmationCheckbox: React.FC<IApprovalConfirmationProps> = ({
```
Fosol, line 30: I wonder if this approval checkbox will need to be reused and could be more generic.  Don't do it now, but I suspect there will be more of them.

goemen: It will be moved to the project disposal feature directory after Devin pulls in my code.

PR #251, opened 2020-05-26
PIMS-1783 UNIT TEST - Property List - Edit Property

asanchezr:
- Refactor Property List to make it more testable
- Add tests for Property List table
- Add Makefiles for local docker development
- Update snapshots


docker-compose.yml
```diff
@@ -95,4 +95,5 @@ networks:
 volumes:
   api-db-data:
   frontend-node-cache:
+    name: pims-frontend-node-cache
```
asanchezr, line 98: The default behaviour in docker-compose is to prefix volumes with the folder name where the code was downloaded. Since each developer can put the code wherever he/she wants it can be difficult to find the name of the docker volumes that are created in a `docker-compose.yml` file.

Recent versions of `docker-compose` added this new option **name** where the name is hard-coded and not prefixed. This was added here so the `Makefile` can delete the node_modules cache when you run `make clean`

frontend/Makefile
```diff
@@ -46,17 +46,15 @@ clean: ## Purges node_modules
...
 	@echo "$(P) Installing dependencies..."
+	@npm ci
 
-build: install ## Builds the frontend
+build: ## Builds the frontend
 	@echo "$(P) Building in release mode..."
 	@npm run build
 
-run: install ## Serves the frontend via webpack-dev-server
+run: ## Serves the frontend via webpack-dev-server
```
asanchezr, line 57: do not force a slow "npm ci" operation before building or running locally. Most devs are on top of their dependencies anyways

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -55,11 +55,11 @@ const mockDetails: IParcelDetail = {
     address: {
       line1: '1234 mock Street',
       line2: 'N/A',
-      city: 'Victoria',
-      province: 'BC',
+      cityId: 0,
+      provinceId: 'BC',
```
asanchezr, line 59: This is unrelated to my changes but it was failing the tests since the API response changed but this test wasn't updated

frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -1,33 +1,192 @@
...
 import thunk from 'redux-thunk';
 import { ILookupCode } from 'actions/lookupActions';
 import * as API from 'constants/API';
 import { Provider } from 'react-redux';
 import * as reducerTypes from 'constants/reducerTypes';
+import service from '../service';
+
+// Set all module functions to jest.fn
+jest.mock('../service');
+const mockedService = service as jest.Mocked<typeof service>;
```
asanchezr, line 17: The approach for this test is to mock only the **service** class/helper. No need to mock axios here...

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -152,12 +148,12 @@ const PropertyListView: React.FC = () => {
       // Only update the data if this is the latest fetch
       if (fetchId === fetchIdRef.current && agencyIds?.length > 0) {
         const query = getServerQuery({ pageIndex, pageSize, filter, agencyIds });
-        const response = await CustomAxios().get<IPagedItems<IProperty>>(getPropertyListUrl(query));
+        const data = await service.getPropertyList(query);
```
asanchezr, line 151: Use a `service` class/helper instead of axios. In my opinion, this level of indirection allows for easier mocking in tests.

frontend/src/features/properties/list/interfaces.ts
```diff
@@ -41,9 +41,9 @@ export interface IProperty {
   // Parcel Properties
   landArea: number;
   landLegalDescription: string;
-  municipality: string;
-  zoning: string;
-  zoningPotential: string;
+  municipality?: string;
+  zoning?: string;
+  zoningPotential?: string;
```
asanchezr, line 46: These are not being returned by the API currently so had to make them optional. If that goes against business rules, the API needs to return empty or null properties in the JSON response

frontend/src/features/properties/service.ts
```diff
@@ -0,0 +1,30 @@
...
+const getPropertyList = async (filter: IPropertyFilter): Promise<IPagedItems<IProperty>> => {
+  const url = API_ENDPOINTS.propertiesSearch(filter);
+  const response = await CustomAxios().get<IPagedItems<IProperty>>(url);
+  return response.data;
+};
+
+// TODO: Refactor later
+const getPropertyReport = async (filter: IPropertyFilter): Promise<any> => {
+  return Promise.reject('Not implemented yet');
+};
```
asanchezr, line 25: The download report uses a different approach of invoking a redux action to trigger the download so I did not refactor that one. Tech debt for later.

docker-compose.yml
Fosol: I like this

frontend/src/components/maps/leaflet/MapProperties.test.tsx
Fosol: Thanks for fixing

frontend/src/features/properties/list/interfaces.ts
Fosol: Currently the API does not return null valued properties

PR #257, opened 2020-05-27
PIMS-2035  Paging Hover (Disable page click if there is only one page result)

goemen:


frontend/src/components/Table/Table.scss
```diff
@@ -1,16 +1,26 @@
+$table-color: #dee2e6;
```
asanchezr, line 1: /nit: I'd love to see this moved to one of our "global" SCSS files - like  `colors.scss`

PR #269, opened 2020-05-28
Initial WIP implementation of project dispose view.

devinleighsmith:
Known Issues, TODOS:
1) Correct lot size display (landarea)
2) Correct sub agency display
3) Track confirmation in the DB.
4) Update table based on Alejandro's improvements.
5) fix table page size behaviour
6) add final success page.
7) Refactor all steps to separate formik component from logic/layout.
8) Finish functionality of review project form.
9) Add yup validation to all steps.
10) Styling pass
11) allow saved properties to be removed using the property list view select
12) Performance/Flickering during step transitions.
13) re-completing old steps should not revert project completing progress.
14) Update to consistent usage of enums instead of constants.
15) migrate some hardcoded values to constants.ts
16) clean up the projectConverter.


devinleighsmith:
Looks like the checks are failing due to code coverage issues. Not sure how to proceed. I can write some quick tests or we can force this through.

devinleighsmith:
Sorry about the size of this one, updates will all be kept small.

Fosol:
@devinleighsmith You can lower the `coverageThreshold` if you want temporarily.  Just edit the `package.json`

backend/api/Areas/Project/Models/Dispose/ProjectModel.cs
```diff
@@ -76,7 +76,7 @@ public class ProjectModel : Api.Models.BaseModel
         /// <summary>
         /// get/set - An array of properties associated with this project.
         /// </summary>
-        public IEnumerable<ProjectPropertyModel> Properties { get; set; } = new List<ProjectPropertyModel>();
+        public IEnumerable<ProjectPropertyModel> ProjectProperties { get; set; } = new List<ProjectPropertyModel>();
```
Fosol, line 79: Why rename?

backend/api/Helpers/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Pims.Api.Helpers.Extensions
+{
+    public static class DateTimeExtensions
+    {
+        public static int ToFiscalYear(this DateTime dateTime)
```
Fosol, line 10: Where is the output required for this?  Currently where fiscal is relevant we only capture the year not the date.  It's also not ideal as it's hard-coded (although I wouldn't change it).

backend/api/Models/Parcel/PropertyModel.cs
```diff
@@ -19,7 +19,7 @@ public class PropertyModel : BaseModel
 
         public int AgencyId { get; set; }
 
-        public string SubAgency { get; set; }
+        public virtual string SubAgency { get; set; }
```
Fosol, line 22: Why virtual?  Are you overriding it somewhere?

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -102,9 +102,19 @@ public Project Get(string projectNumber)
                 .Include(p => p.Properties).ThenInclude(p => p.Parcel)
                 .Include(p => p.Properties).ThenInclude(p => p.Parcel).ThenInclude(p => p.Evaluations)
                 .Include(p => p.Properties).ThenInclude(p => p.Parcel).ThenInclude(p => p.Fiscals)
-                .Include(p => p.Properties).ThenInclude(b => b.Building)
-                .Include(p => p.Properties).ThenInclude(b => b.Building).ThenInclude(b => b.Evaluations)
-                .Include(p => p.Properties).ThenInclude(b => b.Building).ThenInclude(b => b.Fiscals)
+                .Include(p => p.Properties).ThenInclude(p => p.Parcel).ThenInclude(p => p.Classification)
```
Fosol, line 105: Not sure we have a better solution, but we're really pulling down the whole db here.

backend/api/Areas/Project/Models/Dispose/ProjectModel.cs
devinleighsmith: More accurately reflects the object. Properties are actually child entities of projectProperty. Project Property is really just a join entity between projects(buildings or parcels) and the project.

backend/api/Helpers/Extensions/DateTimeExtensions.cs
devinleighsmith: I actually moved this logic from the backend to the frontend so this is unused. The idea is that I need to know the current fiscal when submitting the estimated/netbook from the frontend since the table only includes the value not the fiscalyear.

backend/api/Models/Parcel/PropertyModel.cs
devinleighsmith: yeah, I could copy this entire model as-is into the Project namespace, but I thought I'd leave it and just create copies of the parcelModel and buildingModel and update those as needed.

backend/dal/Services/Concrete/ProjectService.cs
devinleighsmith: shouldn't linq optimize this so that this is just a whole bunch of joins for one ID? we could look at the generated SQL if necessary.

PR #270, opened 2020-05-29
PIMS-1729,  PIMS-1733 Bugfixes

asanchezr:


frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -63,14 +63,14 @@ describe('SubmitProperty', () => {
     const { getByText } = render(
       getSubmitProperty({ match: { params: { id: mockDetails[0]?.id } } }, store),
     );
-    getByText('Update Property');
+    getByText('Property Detail');
```
goemen, line 66: Need to wrap this with `expect`

frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -63,14 +63,14 @@ describe('SubmitProperty', () => {
...
   });
 
   it('SubmitProperty renders edit button if parcel detail loaded.', () => {
     const { getByText } = render(
       getSubmitProperty({ match: { params: { id: mockDetails[0]?.id } } }),
     );
-    getByText('Update Property');
+    getByText('Property Detail');
```
goemen, line 73: Need to wrap this with expect

frontend/src/pages/SubmitProperty.test.tsx
```diff
@@ -88,12 +88,12 @@ describe('SubmitProperty', () => {
...
```
goemen, line 91: Need to wrap this with expect

frontend/src/pages/SubmitProperty.test.tsx
asanchezr: I only changed the expected text in the tests. It was already failing the tests even without `expect` as it couldn't find the element after I changed the title. 

I agree this test is a bit flaky as it reads "SubmitProperty renders edit button if parcel detail loaded" but the test doesn't check for that at all. 

Given that I wasn't the original author of this test I was going to file a tech-debt task to update/improve this tests as several of them seem outdated in the sense that yes, they pass, but do not do what the test description says.

asanchezr: @goemen  Also please remember we agreed as a team to not use the "Request changes" review in GitHub as it blocks the merge entirely. I think we should only use this if the changes would bring down the whole app in production or something else **critical** in nature. Unit tests hardly qualify for that.

If the team feels different let's change out current process but that's what we have agreed so far.

devinleighsmith: Could just update the test description to better match what is being tested in the interim. Can log the tech-debt for adding more tests separately.

asanchezr: I actually like the test descriptions as they cover valid use cases. Is the implementation that is lacking behind. There's already a test to check for the form title - which works as expected. These other tests cover other scenarios but the implementation only re-checks the form title. 
I'll disable those and file a tech-debt item to implement as time permits

PR #280, opened 2020-06-02
PIMS-2063 add/update project unit tests.

devinleighsmith:
Please see comments - Think we may need to rework some existing tests.


backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -600,11 +601,12 @@ public void Update_AsAdmin()
             // Act
             var projectToUpdate = service.Get(project.ProjectNumber);
             projectToUpdate.Description = "A new description";
-            var result = service.Update(projectToUpdate);
+            service.Update(projectToUpdate);
+            var result = service.Get(project.ProjectNumber);
```
devinleighsmith, line 605: Oook, so it looks like returning the entity passed in from the controller is having a negative effect on our existing tests. Essentially, this is what is happening in service methods following this pattern.
1) The test creates a new Project (or parcel or whatever).
2) The object from #1 is used to update the database.
3) The object from #1 is returned, and the tests assert that object #1 is equal to object#1.

The problem is that the assertions in step 3 don't actually test anything. What we want to be testing is that the update function affects the entity within the database, what we are actually testing is that we are getting the same object reference out of an Update method that we put in.

As an example, the current dev version of the Update_UpdateBuilding test still passes if I change the ProjectService.Update method to just be:
```
public Project Update(Project project)
{
    return project;
}
```

In addition to testing issues, the object returned from this (and other) Update Api controllers cannot be trusted, as the data returned may not actually have been persisted. I haven't solved the issue with the API in this PR, but I have changed the tests to call update() and then call get() to retrieve the actual entity from the DB. However, it may make more sense to fix the issue in the API itself.

backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -579,7 +580,7 @@ public void Update()
 
             // Assert
             Assert.NotNull(result);
-            Assert.Equal(projectToUpdate, result);
+            result.Should().BeEquivalentTo(projectToUpdate, options => options.Excluding(o => o.SelectedMemberPath.Contains("Updated")));
```
devinleighsmith, line 583: So, I was looking at these tests and wondering why they were just using Assert.Equal(object1, object2), which just compares the object references, instead of Assert.Equal(object1, object2, new DeepPropertyCompare()). It turns out that a number of these tests crash when using the DeepPropertyCompare. Also, even when the test doesn't crash it is hard to tell what is causing the equality to fail as the assert message isn't helpful and debugging large, nested entities to find the differences isn't a good solution.

I can think of two solutions:
1) stop using equals on large objects, assert on general features of those objects instead (ie, the objects id, type, or number of values in the case of an array)
2) Use the library that I found above, https://fluentassertions.com/ allows deep comparison of nested objects with the ability to exclude/include properties via configuration. This library also gracefully handles circular references. It also provides specific feedback on the exact portions of the objects that failed equality testing.

backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -709,17 +755,27 @@ public void Update_AddBuilding()
...
+            service.Update(projectToUpdate);
+            var result = service.Get(project.ProjectNumber);
 
             // Assert
             Assert.NotNull(result);
-            Assert.Equal(projectToUpdate, result);
             Assert.Single(result.Properties);
+            result.Should().BeEquivalentTo(projectToUpdate, options => options
+            .IgnoringCyclicReferences()
+            .Excluding(x => x.SelectedMemberPath == "Properties[0].Building.Projects")
```
devinleighsmith, line 766: While I like the ability to ignore problematic paths in general, I don't like how messy this makes the test. You can hide these excludes behind a configurationOptions object, but that could result in a lot of overhead for these sorts of highly nested objects. 

In the below example, I like this

backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -709,17 +755,27 @@ public void Update_AddBuilding()
...
+            .Excluding(x => x.SelectedMemberPath == "Properties[0].Building.Parcel")
+            .Excluding(x => x.SelectedMemberPath == "Properties[0].Building.Agency.Parcels")
+            .Excluding(x => x.SelectedMemberPath == "Properties[0].Id")
```
devinleighsmith, line 769: I do like how expressive this line is, it is essentially saying that I expect the original object to be equal to the database object, except for the Id, which makes sense in an Add workflow.

backend/tests/unit/dal/Services/ProjectServiceTest.cs
Fosol: I'm a fan of FluentAssertions.  Using Assert.Equal with a comparer was providing some of the similar benefits of FluentAssertions.

Fosol (approved):

backend/tests/unit/dal/Services/ProjectServiceTest.cs
asanchezr: I also like fluent assertions

asanchezr (approved):

PR #286, opened 2020-06-03
PIMS-2410, 2411, 2397

jdtoombs:
- SRES now able to edit any property
- updating corresponding tests 
- restrict classification options to surplus active and surplus encumbered (still working on unit tests)

### DIFF
--- a/frontend/src/features/projects/dispose/components/columns.tsx
+++ b/frontend/src/features/projects/dispose/components/columns.tsx
@@ -32,12 +32,15 @@ const EditableMoneyCell = (cellInfo: any) => {
 
 const EditableClassificationCell = (cellInfo: any) => {
   const classifications = useCodeLookups().getOptionsByType('PropertyClassification');
+  const filteredClassifications = classifications.filter(
+    (x: any) => x?.label === 'Surplus Active' || x?.label === 'Surplus Encumbered',
+  );
   const context = useFormikContext();
   return (
     <FastSelect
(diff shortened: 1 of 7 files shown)


frontend/src/features/projects/dispose/components/columns.tsx
```diff
@@ -32,12 +32,18 @@ const EditableMoneyCell = (cellInfo: any) => {
 
 const EditableClassificationCell = (cellInfo: any) => {
   const classifications = useCodeLookups().getOptionsByType('PropertyClassification');
+  const filteredClassifications = classifications.filter((x: any) => {
+    if (x.label === 'Surplus Active' || x.label === 'Surplus Encumbered') {
+      return x;
+    }
+    return null;
+  });
   const context = useFormikContext();
```
asanchezr, line 41: Please change this to return a proper boolean value. This works because x is "truthy" and null is "falsy" but can lead to bugs.

Basically, for `.filter()` if the callback function returns true, the current element will be in the resulting array. If it returns false, it won’t be. So there's no need to return the element here, just true or false.

This could be re-written as:

```ts
 const filteredClassifications = classifications.filter((x: any) => x?.label === 'Surplus Active' || x?.label === 'Surplus Encumbered');

```

jdtoombs: Ahh I knew it didn't feel right, thanks for the help

PR #289, opened 2020-06-05
PIMS-2217 Disable Property

goemen:
Editing if within active project

### DIFF
--- a/backend/api/Areas/Project/Mapping/Search/ProjectPropertyMap.cs
+++ b/backend/api/Areas/Project/Mapping/Search/ProjectPropertyMap.cs
@@ -46,6 +46,7 @@ namespace Pims.Api.Areas.Project.Mapping.Search
                     src => src.Building != null ? GetAgencyName(src.Building.Agency) : GetAgencyName(src.Parcel.Agency))
 
                 .Map(dest => dest.LandArea, src => src.Building != null ? GetLandArea(src.Building.Parcel) : GetLandArea(src.Parcel))
+                .Map(dest => dest.ParcelId, src => src.Building != null ? src.Building.Parcel.GetId() : src.Parcel.GetId())
                 .Inherits<BaseEntity, BaseModel>();
 
         }
--- a/backend/api/Models/Parcel/ParcelModel.cs
+++ b/backend/api/Models/Parcel/ParcelModel.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using Pims.Api.Areas.Project.Models.Dispose;
 
 namespace Pims.Api.Models.Parcel
 {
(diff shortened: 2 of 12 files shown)


backend/api/Models/Parcel/ParcelBuildingModel.cs
```diff
@@ -34,6 +34,8 @@ public class ParcelBuildingModel : PropertyModel
 
         public float RentableArea { get; set; }
 
+        public string ProjectNumber { get; set; }
```
Fosol, line 37: Rarely is it good to override an inherited property.

backend/api/Areas/Project/Mapping/Search/ProjectPropertyMap.cs
```diff
@@ -55,6 +56,11 @@ public void Register(TypeAdapterConfig config)
             return parcel?.LandArea;
         }
 
+        private float? GetParcelId(Parcel parcel)
```
Fosol, line 59: These could be extension methods so that they aren't duplicated.

backend/api/Models/Parcel/ParcelModel.cs
```diff
@@ -19,6 +20,8 @@ public class ParcelModel : PropertyModel
 
         public string ZoningPotential { get; set; }
 
+        public  string ProjectNumber { get; set; }
```
Fosol, line 23: Don't override unless there is a specific reason.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -221,6 +223,7 @@ public Parcel Update(Parcel parcel)
                     }
                     else
                     {
+                        CanUpdateOrRemoveProperty(existingBuilding.ProjectNumber, isAdmin);
```
Fosol, line 226: The current way the function is created it will attempt to update all buildings within a parcel.  Which would result in throwing this error, even if they didn't make any changes to the specific building.  This would be very frustrating to a user to get this error when they did not edit the building in question.

Fosol: Additionally it would appear you haven't ensured they can't delete the building lower down within this Update function.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -352,6 +358,14 @@ public void Remove(Parcel parcel)
             this.Context.Parcels.Remove(entity); // TODO: Shouldn't be allowed to permanently delete parcels entirely.
             this.Context.CommitTransaction();
         }
+
+        private static void CanUpdateOrRemoveProperty(string projectNumber, bool isAdmin)
```
Fosol, line 362: This again looks like a reasonable extension method.
`CanUpdateOrRemoveProperty(this ClaimsPrincipal user, Parcel parcl) ...`
`CanUpdateOrRemoveProperty(this ClaimsPrincipal user, Building building) ...`

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -341,11 +348,14 @@ public void Remove(Parcel parcel)
...
 
             this.Context.Entry(entity).CurrentValues.SetValues(parcel);
+            this.User.CanUpdateOrRemoveParcelOrThrow(this.Context, entity);
+
             entity.Buildings.ForEach((building) =>
             {
                 this.Context.BuildingEvaluations.RemoveRange(building.Evaluations);
                 this.Context.BuildingFiscals.RemoveRange(building.Fiscals);
                 this.Context.Buildings.Remove(building);
+                this.User.CanUpdateOrRemoveBuildingOrThrow(this.Context, building);
```
Fosol, line 358: Ideally you would do this before performing the above statements as it short-circuits with it.

PR #290, opened 2020-06-05
PIMS-2158 Adding project claims

Fosol:
## Description

We now have project level claims instead of using property claims for projects.

DAL, API and Nav-menu have been updated.

- Also fixed keycloak user sync


auth/keycloak/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # >docker build -t keycloak .
 # >docker run --env-file=../../.env -d keycloak
 # https://hub.docker.com/r/jboss/keycloak/
-FROM jboss/keycloak:9.0.0
+FROM jboss/keycloak:10.0.2
```
Fosol, line 4: Was hoping upgrade would fix import/export, but it didn't.

backend/dal/Security/Permissions.cs
```diff
@@ -13,39 +13,54 @@ namespace Pims.Dal.Security
     public enum Permissions
     {
         [Display(GroupName = "admin", Name = "system-administrator", Description = "Can administer application settings.")]
-        SystemAdmin,
+        SystemAdmin = 1,
```
Fosol, line 16: Set values so that database doesn't get out of sync if we every reorder them.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -562,16 +599,20 @@ public Project SetStatus(Project project, Workflow workflow, ProjectStatus statu
             switch (toStatus.Code)
             {
                 case ("SU"): // Submit
+                    this.User.ThrowIfNotAuthorized(Permissions.DisposeRequest, "User does not have permission to submit project.");
```
Fosol, line 602: Only users with `dispose-request` claim can submit.  This means the APP needs to be updated to disable the **Submit** button if the user does not have permission.

PR #291, opened 2020-06-05
Unit tests for Classification Options

jdtoombs:
**Ready for re-review**

*Update:

-  ~~separate component `LimitedSelect` which allows you to restrict what options you can select by passing an array of labels you wish to have appear~~

- added the functionality to `FastSelect` no longer a separate component 

- existing value will load with the option attribute hidden so it will not appear in drop down but be visible initially

### DIFF
--- a/frontend/src/constants/classifications.ts
+++ b/frontend/src/constants/classifications.ts
@@ -1,0 +1,6 @@
+export enum CLASSIFICATIONS {
+  CoreOperational = 0,
+  CoreStrategic = 1,
+  SurplusActive = 2,
+  SurplusEncumbered = 3,
+}
--- a/frontend/src/features/projects/dispose/components/columns.tsx
+++ b/frontend/src/features/projects/dispose/components/columns.tsx
@@ -6,7 +6,7 @@ import { CellProps } from 'react-table';
 import { formatMoney, formatNumber } from 'utils';
 import { IProperty } from '../interfaces';
 import { useFormikContext } from 'formik';
-import { FastCurrencyInput, FastSelect, Input } from 'components/common/form';
+import { FastCurrencyInput, Input, FastSelect } from 'components/common/form';
 import useCodeLookups from 'hooks/useLookupCodes';
 import { FaRegTimesCircle } from 'react-icons/fa';
 import _ from 'lodash';
@@ -30,17 +30,15 @@ const EditableMoneyCell = (cellInfo: any) => {
 
 const EditableClassificationCell = (cellInfo: any) => {
   const classifications = useCodeLookups().getOptionsByType('PropertyClassification');
-  const filteredClassifications = classifications.filter(
-    (x: any) => x?.label === 'Surplus Active' || x?.label === 'Surplus Encumbered',
-  );
   const context = useFormikContext();
   return (
     <FastSelect
+      limitLabels={['Surplus Active', 'Surplus Encumbered']}
       formikProps={context}
       type="number"
-      options={filteredClassifications}
+      options={classifications}
       field={`properties.${cellInfo.row.id}.classificationId`}
-    ></FastSelect>
+    />
   );
 };
 
--- a/frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
+++ b/frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
@@ -1,4 +1,5 @@
 import * as Yup from 'yup';
+import { CLASSIFICATIONS } from 'constants/classifications';
 
 export const ApprovalConfirmationStepSchema = Yup.object().shape({
   confirmation: Yup.boolean()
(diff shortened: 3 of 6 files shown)


frontend/src/features/projects/dispose/components/columns.tsx
```diff
@@ -37,6 +37,7 @@ const EditableClassificationCell = (cellInfo: any) => {
   return (
     <FastSelect
       formikProps={context}
+      placeholder="Must selct one"
```
jdtoombs, line 40: To prevent it from defaulting to `Surplus Active`

asanchezr: /nit: fix typo "Must sel(e)ct one"

frontend/src/features/projects/dispose/components/columns.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+it('renders only surplus active and encumbered', () => {
+  const tree = renderer
+    .create(
+      <Provider store={store}>
+        <EditableClassificationCell {...mockCellInfo} />
+      </Provider>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
```
asanchezr, line 39: This test just checks that the rendered component matches the last snapshot on file. If a dev inadvertently changes the logic of the component to display say 'Core Operational' and updates the snapshot this test will pass. I think the test name is a bit misleading as it only checks the `it('renders ok')`

We can leave as-is for now but I'd like to see a "real" test created that checks the business rule here: "only shows surplus active and encumbered". Ideally you would render the component with react-test-library and assert the number of "option" elements or something like that. Thoughts?

frontend/src/features/projects/dispose/components/columns.tsx
jdtoombs: oops! Thank you

frontend/src/features/projects/dispose/components/columns.test.tsx
jdtoombs: Oh yes, that is a good call. I had that originally but started to overthink and thought it was redundant. But the way you worded it there shows me that it is not redundant at all and will catch errors on change! Will change the name of the test and add another before I merge!

frontend/src/features/projects/dispose/components/columns.tsx
jdtoombs: I am going to do some more playing around, this right now overrides the existing and only shows SE/SA

asanchezr: Now that I understand the "full" business requirements which are (correct me if I'm wrong please):

> - Show current assigned value in "read-only" mode
> - Only display SE and SA as new options to select

I think this should warrant a new component that takes an initial "read-only" value object (with label and id to not loose the selection made before) and the allowed options. I wouldn't change the current Select component but create a new one for this specific use case

jdtoombs: @asanchezr , I think that might be best too. The only other ways I was thinking of was to add the optional attribute of `disabled` onto the type `SelectOption` and disable everything that isn't SA/SE. But the problem with that is it still shows them in the list - they are just grayed out. The other way I was thinking was to make the placeholder the previous value..

frontend/src/components/common/form/LimitedSelect.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { memo, useEffect } from 'react';
```
jdtoombs, line 1: There wasn't too many changes to `FastSelect`... I am wondering if it should stay a separate component or add more to `FastSelect`. Not sure what the best practice is!

Fosol: If you're essentially displaying the current value and only allowing them to select specific values, I'd leave it as a single component.  This is a common scenario for dropdown/selects.

frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
```diff
@@ -18,7 +18,11 @@ export const DocumentationStepSchema = Yup.object().shape({
 export const UpdateInfoStepYupSchema = Yup.object().shape({
   properties: Yup.array().of(
     Yup.object().shape({
-      classification: Yup.string().required('Required'),
+      classificationId: Yup.number().test(
+        'is-valid',
+        'Must select Surplus Active or Surplus Encumbered',
+        (val: any) => val === 2 || val === 3,
```
Fosol, line 24: These hardcoded values should be constants maintained in an `enum`

frontend/src/constants/classifications.ts
```diff
@@ -0,0 +1,6 @@
+export enum CLASSIFICATIONS {
+  CoreOperational,
```
Fosol, line 2: You'll need to set the correct int value for each.

jdtoombs: @Fosol it sets them by default to 0, 1, 2, 3 unless otherwise specified

PR #292, opened 2020-06-06
PIMS-2233,2234

devinleighsmith:
Also generally update the project dispose workflow to make it compatible with the backend changes.


backend/api/Startup.cs
```diff
@@ -157,9 +158,10 @@ public void ConfigureServices(IServiceCollection services)
             {
                 builder.Password = pwd;
             }
+            ILoggerFactory debugLoggerFactory = LoggerFactory.Create(builder => { builder.AddDebug(); });
             services.AddDbContext<PimsContext>(options =>
             {
-                options.UseSqlServer(builder.ConnectionString);
+                options.UseSqlServer(builder.ConnectionString).UseLoggerFactory(debugLoggerFactory);
```
devinleighsmith, line 164: This just prints the generated SQL when debugging, really helped me out.

frontend/src/features/projects/dispose/ProjectDisposalSubmitted.tsx
```diff
@@ -17,7 +17,7 @@ export function ProjectDisposalSubmitted() {
   const keycloak = useKeycloakWrapper();
   return (
     <Container className="ProjectDisposalSubmitted">
-      <ColoredWrapper style={{ color: '#2e8540' }}>
+      <ColoredWrapper>
```
devinleighsmith, line 20: fixed from last PR

frontend/src/features/projects/dispose/steps/__snapshots__/ProjectDraftStep.test.tsx.snap
```diff
@@ -77,7 +76,7 @@ exports[`renders correctly 1`] = `
               onBlur={[Function]}
               onChange={[Function]}
               required={true}
-              value=""
+              value="Temp"
```
devinleighsmith, line 79: Johnny will need to correct this as part of PIMS-2389

asanchezr:
A couple backend unit tests are failing that need fixing:

```
[xUnit.net 00:00:12.65]     Pims.Dal.Test.Services.ProjectServiceTest.Update_AddBuilding [FAIL]
[xUnit.net 00:00:12.66]       System.InvalidOperationException : Buildings in a Project cannot be added to another Project.
```

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -248,5 +248,19 @@ public static Entity.Project AddTask(this Entity.Project project, params Entity.
...
+        /// <summary>
+        /// Throw an exception if the passed property is not in the same agency or sub agency as this project.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <param name="property"></param>
+        /// <returns></returns>
+        public static void ThrowIfPropertyNotInProjectAgency(this Entity.Project project, Entity.Property property)
+        {
+            //properties may be in the same agency or sub-agency of a project. A parcel in a parent agency may not be added to a sub-agency project.
+            if (property.AgencyId != project.AgencyId
```
Fosol, line 261: I don't believe this allows a parent agency to sell different sub-agency properties in a single project.

backend/tests/core/Entities/ProjectHelper.cs
```diff
@@ -79,8 +79,9 @@ public static List<Entity.Project> CreateProjects(int startId, int count, Entity
         /// <param name="id"></param>
         /// <param name="agency"></param>
         /// <returns></returns>
-        public static Entity.Project CreateProject(this PimsContext context, int id, Entity.Agency agency)
+        public static Entity.Project CreateProject(this PimsContext context, int id, int agencyId)
```
Fosol, line 82: You probably want to add an override rather than change the function structure here.

backend/tests/core/Entities/ProjectHelper.cs
```diff
@@ -79,8 +79,9 @@ public static List<Entity.Project> CreateProjects(int startId, int count, Entity
...
         {
+            var agency = EntityHelper.CreateAgency(agencyId);
```
Fosol, line 84: As this is a helper with `PimsContext`, you may want to first see if the agency exists.
`context.Agencies.Find(agencyId) ?? EntityHelper.CreateAgency(agencyId)`

backend/tests/core/Entities/ProjectHelper.cs
devinleighsmith: The reason for this change was actually to prevent a stack overflow, the above method calls itself infinitely.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
devinleighsmith: I've tested this scenario. Adding a parcel belonging to a sub-agency of the project's agency does not throw this error.

PR #294, opened 2020-06-08
PIMS-2045 update the review page to allow it to act as read only form.

devinleighsmith:


frontend/src/features/projects/dispose/hooks/useStepForm.tsx
```diff
@@ -44,7 +47,10 @@ const useStepForm = () => {
       });
     return response;
   };
-  return { onSubmit };
+  const canUserEditForm = (projectAgencyId: number) =>
+    (keycloak.hasAgency(projectAgencyId) && keycloak.hasClaim(Claims.PROJECT_EDIT)) ||
+    keycloak.hasClaim(Claims.ADMIN_PROPERTIES);
```
Fosol, line 52: Should be `Claims.ADMIN_PROJECTS`

PR #296, opened 2020-06-09
PIMS 2223 Surplus Property Program Project List View

goemen:


frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -137,33 +167,48 @@ const ProjectListView: React.FC = () => {
     setDeleteId(projectNumber);
   };
 
+  const onRowClick = (row: IProject) => {
+    if (mode === PageMode.APPROVAL) {
+      history.push(`/dispose/projects/assess/properties?projectNumber=${row.projectNumber}`);
+    } else {
+      history.push(
+        `/dispose/projects/${row.status.toLowerCase()}?projectNumber=${row.projectNumber}`,
```
Fosol, line 175: This comment isn't related to this PR, but to the current route paths.

The problem with using the status as a path parameter it makes it impossible for the user to reference the project in a generic way (i.e. `/dispose/projects/SPP-24234`).  The way we're currently doing this would result in bad conditions if the user goes to the route manually and the project is in a different status.  This is very problematic if for example the project is in a status that shouldn't allow for it.

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -137,33 +167,48 @@ const ProjectListView: React.FC = () => {
...
+            </Col>
+            <Col className="bar-item">
+              <Form.Group>
+                <Form.Label>Active Projects:&nbsp;</Form.Label>
+                <Field name="active" label="Active Projects" type="checkbox" />
+              </Form.Group>
+            </Col>
+            <Col className="bar-item">
+              <Form.Group>
+                <Form.Label>My projects:&nbsp;</Form.Label>
```
Fosol, line 199: When I tested this filter for the standard **Project List View** it did not work as expected.  I'll post a bug once I have some formal time to test again.

PR #303, opened 2020-06-11
PIMS-2445 Reducing load on project paging

Fosol:
## Description

A tech-debt decision made earlier to include additional property information in the project search endpoint was now causing memory issues in DEV.

This fix (while still not ideal) dramatically reduces the required data and still provides the summed `NetBook` and `Estimated` values.

This PR will only resolve the memory issue.  Another PR is required to change the UI behaviour to make a separate request when expanding projects to look at their properties.


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -40,43 +40,6 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
                 .Include(p => p.TierLevel)
                 .Include(p => p.Agency)
                 .Include(p => p.Agency).ThenInclude(a => a.Parent)
-                .Include(p => p.Properties)
```
Fosol, line 43: Can't asked to return the whole DB :P

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -58,6 +58,23 @@ public Paged<Project> GetPage(ProjectFilter filter)
                 .Take(filter.Quantity)
                 .ToArray();
 
+            // TODO: Update the View to include the Project.Id so that we can do a group by on that value instead of multiple separate requests.
+            foreach (var project in items)
```
Fosol, line 62: Not ideal, but should resolve our current issue

backend/entities/Project.cs
```diff
@@ -89,6 +90,18 @@ public class Project : BaseEntity
         /// </summary>
         public DateTime? DeniedOn { get; set; }
 
+        /// <summary>
+        /// get/set - The netbook value which is the sum of the properties.
+        /// </summary>
+        [NotMapped]
+        public decimal NetBook { get; set; }
```
Fosol, line 97: Temporary placeholders that won't require a DB refresh.

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -131,14 +131,6 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
         // For now we'll just calculate it.
         setData(
           data.items.map((project: IProject) => {
-            project.sumEstimated = project.properties.reduce(
```
Fosol, line 134: Not required

PR #304, opened 2020-06-11
PIMS-2389 - Don't create project until name is entered

jdtoombs:
test

Now when navigating to **Dispose Properties** it will start at the route `/draft` once the minimum of a name is entered you may click save/next and the project will be created and the project number (draft) will be generated.

Moved around some code in a way that made sense in my mind in order to accomplish this. Please let me know if there is something that looks out of the ordinary.

### DIFF
--- a/frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
+++ b/frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
@@ -1,8 +1,8 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import './ProjectDisposeView.scss';
 import { Container, Spinner } from 'react-bootstrap';
 import { Route, match as Match, useHistory, Redirect } from 'react-router-dom';
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
 import { RootState } from 'reducers/rootReducer';
 import _ from 'lodash';
 import {
@@ -11,6 +11,8 @@ import {
   useStepper,
   StepActions,
   IStatus,
+  IProject,
+  saveProject,
 } from '.';
 import { FormikValues } from 'formik';
 import { IGenericNetworkAction } from 'actions/genericActions';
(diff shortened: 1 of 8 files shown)


frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
```diff
@@ -46,7 +46,6 @@ export const SelectProjectPropertiesStepYupSchema = Yup.object().shape({
 });
 
 export const ProjectDraftStepYupSchema = Yup.object().shape({
-  projectNumber: Yup.string().required('Required'),
```
jdtoombs, line 49: As it is generated this would previously stop the user from hitting save/next after entering a name. Worked before as the project was created right away.

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -76,6 +121,7 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
           />
           {getProjectRequest?.isFetching !== true ? (
             <Container fluid className="step-content">
+              {preDraft ? <ProjectDraftStep formikRef={formikRef} /> : null}
```
jdtoombs, line 124: before the first step it will load the basic form that it will use to create the project in the draft step

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -62,8 +99,16 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
...
     history,
+    dispatch,
+    projectNumber,
   ]);
 
+  if (projectNumber !== null && projectNumber !== undefined && getProjectRequest?.error) {
+    throw Error(`Unable to load project number ${projectNumber}`);
+  }
+
+  projectNumber && preDraft && setPreDraft(false);
```
jdtoombs, line 110: get rid of the original form after the project is created

frontend/src/features/projects/dispose/ProjectDisposeView.tsx
```diff
@@ -1,61 +1,21 @@
...
-          `${match.url}${workflowStatuses[0].route}?projectNumber=${project.projectNumber}`,
-        );
-      });
-    }
-  }, [dispatch, historyReplace, match.url, projectNumber, workflowStatuses]);
-
-  if (projectNumber !== null && projectNumber !== undefined && getProjectRequest?.error) {
-    throw Error(`Unable to load project number ${projectNumber}`);
-  }
-  useEffect(() => {
```
jdtoombs, line 44: originally tied this to state in `ProjectDisposeLayout` but reverted that so I could move this back here to try and keep more of the original structure

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -43,11 +56,35 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
...
+      note: formikRef.current?.values.note,
+      description: formikRef.current?.values.description,
+      properties: [],
+      tierLevelId: 1,
+      statusId: 0,
+      agencyId: 0,
+      tasks: [],
+    };
+    if (!projectNumber) {
+      await (dispatch(createProject(preDraftValues)) as any).then((project: IProject) => {
```
devinleighsmith, line 77: Maybe I'm missing something, but why call create project and then save project?

jdtoombs: Hmm yeah good point.. I'll look into it. Just commented out save project quickly to see if it would work but seemed to break something

PR #305, opened 2020-06-11
Allows simple review form to be saved

devinleighsmith:
Update scss files to use colors.scss where I felt confident.

### DIFF
--- a/frontend/src/features/projects/dispose/components/ReviewApproveActions.tsx
+++ b/frontend/src/features/projects/dispose/components/ReviewApproveActions.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import styled from 'styled-components';
 import { Button } from 'react-bootstrap';
 import { useFormikContext } from 'formik';
-import { ReviewWorkflowStatus } from '../interfaces';
+import { ReviewWorkflowStatus, DisposeWorkflowStatus } from '../interfaces';
 
 const FlexRight = styled.div`
   width: 100%;
@@ -11,16 +11,11 @@ const FlexRight = styled.div`
   margin: 0.5rem 0;
 `;
 
-interface IReviewApproveActionsProps {
-  //onSave: () => void; TODO
-  //onNext: () => void; TODO
-}
-
 /**
  * A component for project review actions
  * @component
  */
-export const ReviewApproveActions: React.FC<IReviewApproveActionsProps> = () => {
+export const ReviewApproveActions: React.FC = () => {
   const { values, submitForm } = useFormikContext<any>();
   return (
     <>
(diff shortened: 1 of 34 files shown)


frontend/src/features/projects/dispose/components/ReviewApproveActions.tsx
```diff
@@ -11,36 +11,37 @@ const FlexRight = styled.div`
...
-export const ReviewApproveActions: React.FC<IReviewApproveActionsProps> = () => {
+export const ReviewApproveActions: React.FC = () => {
   const { values, submitForm } = useFormikContext<any>();
   return (
     <>
       <FlexRight>Approve for Enhanced Referral Process</FlexRight>
       <FlexRight>
         <span>
           <Button
+            disabled={
```
devinleighsmith, line 26: Since there is no transition after this form is submitted, some of the buttons must be disabled based on the project status.

frontend/src/features/projects/dispose/forms/__snapshots__/UpdateInfoForm.test.tsx.snap
```diff
@@ -7,20 +7,6 @@ exports[`Update Info Form Matches Snapshot 1`] = `
   <div
     className="UpdateInfoForm container-fluid"
   >
-    <div
```
devinleighsmith, line 10: This is now hidden by default when no title is supplied.

frontend/src/features/projects/dispose/hooks/useStepForm.tsx
```diff
@@ -50,10 +51,13 @@ const useStepForm = () => {
   const onSubmitReview = (values: any, actions: any) => {
     const apiValues = _.cloneDeep(values);
     let response: any = dispatch(
-      updateWorkflowStatus(apiValues, values.statusCode, 'ACCESS-DISPOSAL'),
+      updateWorkflowStatus(apiValues, values.statusId, 'ACCESS-DISPOSAL'),
     );
     response
-      .then((values: any) => {
+      .then((values: IProject) => {
+        return dispatch(updateProject({ ...apiValues, rowVersion: values.rowVersion }));
```
devinleighsmith, line 58: I don't like this, but currently I don't think there is a way to change a milestone status and call the update API in request.

Fosol: I'm not sure I understand.  For minor status updates or simply updating use the standard `update` endpoint.  If you are going to change a milestone status, then call the `setStatus` endpoint.  Do you mean you want to do both in a single request?

devinleighsmith: So when the user clicks the "Approve" button the frontend must
1) Trigger a milestone status transition to the new status
2) Also save any changes that have been made to the form.

This requires two chained calls to the backend.

PR #307, opened 2020-06-11
PIMS-2044/PIMS-2445 Lazy load properties

goemen:


frontend/src/features/projects/list/properties/columns.tsx
```diff
@@ -50,7 +49,7 @@ export const columns: ColumnWithProps<IProjectProperty>[] = [
   },
   {
     Header: 'Property Name',
-    accessor: 'name',
+    accessor: 'description',
```
Fosol, line 52: You should be able to use the `name` value

PR #317, opened 2020-06-13
PIMS-2453 PIMS-2444, PIMS-2452

devinleighsmith:
Refactor project dispose layout and useStepForm to simplify code and
communication with backend.

### DIFF
--- a/backend/dal/Helpers/Extensions/ProjectExtensions.cs
+++ b/backend/dal/Helpers/Extensions/ProjectExtensions.cs
@@ -46,7 +46,7 @@ namespace Pims.Dal.Helpers.Extensions
             {
                 var statuses = context.Workflows.Where(w => w.Code == "ACCESS-DISPOSAL")
                     .SelectMany(w => w.Status).Select(x => x.StatusId).ToArray();
-                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("RE"));
+                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("AS-I"));
             }
 
             if (!String.IsNullOrWhiteSpace(filter.ProjectNumber))
--- a/backend/dal/Helpers/Extensions/ProjectExtensions.cs
+++ b/backend/dal/Helpers/Extensions/ProjectExtensions.cs
@@ -46,7 +46,7 @@ namespace Pims.Dal.Helpers.Extensions
             {
                 var statuses = context.Workflows.Where(w => w.Code == "ACCESS-DISPOSAL")
                     .SelectMany(w => w.Status).Select(x => x.StatusId).ToArray();
-                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("SU"));
+                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("RE"));
             }
 
             if (!String.IsNullOrWhiteSpace(filter.ProjectNumber))
--- a/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
+++ b/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
@@ -71,7 +71,7 @@ INSERT INTO dbo.[ProjectStatus] (
     , 0
     , 'The project has been submitted for review to be added to the Surplus Property Program.'
     , '/projects/review'
-    , 1
+    , 0
 ), (
     7
     , 6
@@ -81,7 +81,7 @@ INSERT INTO dbo.[ProjectStatus] (
     , 0
     , 'Submitted project property information review.'
     , '/projects/assess/properties'
-    , 0
+    , 1
 ), (
     8
     , 7
--- a/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
+++ b/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
@@ -66,7 +66,7 @@ INSERT INTO dbo.[ProjectStatus] (
     6
     , 5
     , 'Review'
-    , 'Re'
+    , 'RE'
     , 1
     , 0
     , 'The project has been submitted for review to be added to the Surplus Property Program.'
--- a/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
+++ b/backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
@@ -65,21 +65,21 @@ INSERT INTO dbo.[ProjectStatus] (
 ), (
     6
     , 5
-    , 'Submitted'
-    , 'SU'
+    , 'Review'
+    , 'Re'
     , 1
     , 0
     , 'The project has been submitted for review to be added to the Surplus Property Program.'
-    , '/projects/submitted'
+    , '/projects/review'
     , 1
 ), (
     7
     , 6
-    , 'Property Review'
+    , 'Submitted'
     , 'AS-I'
     , 1
     , 0
-    , 'Property information review.'
+    , 'Submitted project property information review.'
     , '/projects/assess/properties'
     , 0
 ), (
--- a/backend/dal/Services/Concrete/ProjectService.cs
+++ b/backend/dal/Services/Concrete/ProjectService.cs
@@ -657,8 +657,12 @@ namespace Pims.Dal.Services
             var toStatus = this.Context.ProjectStatus
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId);
-            if (!fromStatus.ToStatus.Any(s => s.ToStatusId == project.StatusId)) throw new InvalidOperationException($"Invalid project status transitions from '{fromStatus.Name}' to '{toStatus?.Name}'.");
-
+            // TODO: ignore status transitions for ACCESS-DISPOSAL steps for now.
+            var ignoreTransitionStatuses = this.Context.Workflows.Include(w => w.Status).Where(w => w.Code == "ACCESS-DISPOSAL").SelectMany(s => s.Status).Select(s => s.StatusId);
+            if (!ignoreTransitionStatuses.Contains(fromStatusId)) { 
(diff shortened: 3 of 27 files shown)


frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -1,7 +1,7 @@
-import React, { useEffect, useRef, useState } from 'react';
```
jdtoombs, line 1: Thanks for this! Was just working on this on my end and cleaning it up then had a flashback to reviewing this yesterday 🤦 

I was cringing at my old approach while fixing it lol got rid of all the preDraft stuff and the extra form.  But I was stuck on the point where the next button would work; however, it would not mark step one as complete. But this looks like it covers that!!

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -46,7 +46,7 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
             {
                 var statuses = context.Workflows.Where(w => w.Code == "ACCESS-DISPOSAL")
                     .SelectMany(w => w.Status).Select(x => x.StatusId).ToArray();
-                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("SU"));
+                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("RE"));
```
Fosol, line 49: I suspect this is incorrect.  I think this was being used to return projects for the Approval Requests.  It would be returning projects that are in the `ACCESS-DISPOSAL` workflow and those that were submitted `SU`.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -658,8 +658,12 @@ public Project SetStatus(Project project, Workflow workflow)
             var toStatus = this.Context.ProjectStatus
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId);
-            if (!fromStatus.ToStatus.Any(s => s.ToStatusId == project.StatusId)) throw new InvalidOperationException($"Invalid project status transitions from '{fromStatus.Name}' to '{toStatus?.Name}'.");
-
+            // TODO: ignore status transitions for ACCESS-DISPOSAL steps for now.
```
Fosol, line 661: You shouldn't need to do this, just add additional valid transitions to the sql script.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -670,7 +674,7 @@ public Project SetStatus(Project project, Workflow workflow)
             // This could be extracted at some point to a configurable layer, but not required presently.
             switch (toStatus.Code)
             {
-                case ("SU"): // Submit
+                case ("RE"): // Review
```
Fosol, line 677: This is incorrect.  Review should not be a milestone and we only generate a project number when the request is submitted.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -65,21 +65,21 @@ INSERT INTO dbo.[ProjectStatus] (
...
-    , 'Submitted'
-    , 'SU'
+    , 'Review'
+    , 'RE'
     , 1
     , 0
     , 'The project has been submitted for review to be added to the Surplus Property Program.'
-    , '/projects/submitted'
+    , '/projects/review'
     , 1
```
Fosol, line 74: Review is not a milestone.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
```diff
@@ -65,21 +65,21 @@ INSERT INTO dbo.[ProjectStatus] (
...
     , 6
-    , 'Property Review'
+    , 'Submitted'
     , 'AS-I'
     , 1
     , 0
-    , 'Property information review.'
+    , 'Submitted project property information review.'
     , '/projects/assess/properties'
     , 0
```
Fosol, line 84: Submitted is a milestone.

backend/tests/core/Entities/ProjectStatusHelper.cs
```diff
@@ -66,7 +66,7 @@ public static List<Entity.ProjectStatus> CreateDefaultProjectStatuses()
                 new Entity.ProjectStatus("Update Information", "DR-I") { Id = 3, SortOrder = 2, RowVersion = new byte[] { 12, 13, 14 } },
                 new Entity.ProjectStatus("Required Documentation", "DR-D") { Id = 4, SortOrder = 3, RowVersion = new byte[] { 12, 13, 14 } },
                 new Entity.ProjectStatus("Approval", "DR-A") { Id = 5, SortOrder = 4, RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.ProjectStatus("Submitted", "SU") { Id = 6, SortOrder = 5, IsMilestone = true, RowVersion = new byte[] { 12, 13, 14 } },
+                new Entity.ProjectStatus("Reviewed", "RE") { Id = 6, SortOrder = 5, IsMilestone = true, RowVersion = new byte[] { 12, 13, 14 } },
```
Fosol, line 69: Incorrect.

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -44,19 +47,44 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
...
+        let nextStepId = getNextStep(currentStatus)?.id;
+        let workflowStatusCode: string | undefined = undefined;
+        if (nextStepId === undefined) {
+          nextStepId = ReviewWorkflowStatus.PropertyReview;
+          workflowStatusCode = 'ACCESS-DISPOSAL';
         }
+
+        addOrUpdateProject(values, formikRef).then((project: IProject) =>
+          updateProjectStatus(project, nextStepId!, workflowStatusCode).then((project: IProject) =>
+            dispatch(fetchProject(project.projectNumber)),
```
Fosol, line 85: Why do you fetch the project at this point?  You should have an updated copy from the prior actions.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
devinleighsmith: So wait, are you saying that this was wrong before? This change was due to PIMS-2452 which was just to rename the Submitted step to Review.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
devinleighsmith: Same here, the requested change was just to change the name of the step. Was this step also in error?

backend/dal/Helpers/Extensions/ProjectExtensions.cs
devinleighsmith: The old submitted status is now review, so this is only a change in naming. Are you saying this check was incorrect before my change?

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatus.sql
Fosol: The last step for the first workflow would be `Review`, but when they submit the last status is `Submitted`.  Which is a milestone and should start the next workflow for assessment.
There is a requirement to update the SQL scripts to reflect this, as they were an original list and there have been some changes.

There may be a issue we need to think through for the assessment workflow however.  Since `Submitted` is the first status of the assessment workflow, it may require us to handle it a little differently when presenting the data on the single form.  

We can chat through this if you want a call?

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
devinleighsmith: Last time I checked the create/update apis were not correctly returning the updated projects from the database, they were just returning a copy of the passed in object. I brought this up in a code review at the start of this sprint. Has this been fixed?

Fosol: If they're not returning the correct data, they need to be updated.  Should be as simple as
- return correct object
- make sure the right data is included in that object

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -46,7 +46,7 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
             {
                 var statuses = context.Workflows.Where(w => w.Code == "ACCESS-DISPOSAL")
                     .SelectMany(w => w.Status).Select(x => x.StatusId).ToArray();
-                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("SU"));
+                query = query.Where(p => statuses.Contains(p.StatusId) || p.Status.Code.Equals("AS-I"));
```
Fosol, line 49: Not loving the code used here, but it'll do lol.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -669,7 +671,7 @@ public Project SetStatus(Project project, Workflow workflow)
             // This could be extracted at some point to a configurable layer, but not required presently.
             switch (toStatus.Code)
             {
-                case ("SU"): // Submit
+                case ("AS-I"): // Review
```
Fosol, line 674: It's still `submitted` but it's just a comment.

backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -1528,7 +1528,7 @@ public void SetStatus_Submit_NotAuthorizedException()
             var workflows = init.CreateDefaultWorkflows();
             init.SaveChanges();
             init.AddStatusToWorkflow(workflows.First(), init.ProjectStatus.Where(s => s.Id <= 6)).SaveChanges();
-            var submit = init.ProjectStatus.First(s => s.Code == "SU");
+            var submit = init.ProjectStatus.First(s => s.Code == "RE");
```
Fosol, line 1531: Not sure about this, but I can test later.

PR #323, opened 2020-06-15
Remove debugger statement.

devinleighsmith:
Well, this is embarrassing.


backend/dal/Helpers/Extensions/ServiceExtensions.cs
```diff
@@ -14,7 +14,7 @@ namespace Pims.Dal.Helpers.Extensions
     public static class ServiceExtensions
     {
         #region Variables
-        public static readonly string[] EDITABLE_STATUS_CODES = new[] { "DR", "DR-P", "DR-I", "DR-D", "DR-A" };
```
devinleighsmith, line 17: I think this should fix the submission issue on the review page.

PR #327, opened 2020-06-16
PIMS-1522 Swagger API with authentication

goemen:


backend/api/Startup.cs
```diff
@@ -37,7 +37,9 @@
 using Mapster;
 using Pims.Api.Helpers.Mapping;
 using System.Collections.Generic;
+using DocumentFormat.OpenXml.EMMA;
```
asanchezr, line 40: According to https://www.nuget.org/packages/DocumentFormat.OpenXml/ this is used for working with Office Word, Excel, and PowerPoint documents. 

Do we need it to enable HTTPS and JWT tokens on Swagger??

Fosol: Configuring authentication on Swagger does allow for a more complete tool/documentation.

goemen:
reopen

PR #330, opened 2020-06-16
increase min cpu/memory to ensure build is timely.

devinleighsmith:


devinleighsmith:
this change has already been made in tools, this just keeps our source up to date.

PR #332, opened 2020-06-17
PIMS-2175, 2193

jdtoombs:
Display clickable Project Number on property pins that belong to a project. Only visible to owning agency and SRES

Update Assessed Value tables. `ParcelDetailForm` now has `Land`, `Improvements`, and `Total` columns. Building  Assessed Value table remains the same; however, the value entered there is reflected under `Improvments` on the `ParcelDetailForm`


jdtoombs:
Working on some column scaling issues right now.... not quite ready after all

PR #337, opened 2020-06-18
PIMS-2586, 2587

jdtoombs:
Bug 1 - Project number is now read only for SRES until they click the edit button
Bug 2 - Project number no longer disappears when updating a parcel


backend/api/Mapping/Parcel/ParcelMap.cs
```diff
@@ -48,6 +48,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.PID, src => ParcelConverter.ConvertPID(src.PID))
                 .Map(dest => dest.PIN, src => src.PIN)
+                .Map(dest => dest.ProjectNumber, src => src.ProjectNumber)
```
jdtoombs, line 51: have to change this quickly because if it submits "" as a project number they will be viewed as belonging to a project

PR #333, opened 2020-06-18
PIMS-2177 Updating Project Financials

Fosol:
## Description

> An update to the Initial DB Migration

This update is to support and unblock related work on project financials, property financials, notifications and ERP exemption requests.

### Database Updates
PIMS-2177 Adding Project Financials properties
PIMS-1984 Adding Notification entities
PIMS-195 Adding ERP Exemption properties
- Updating Project Status

### Development

There are a few critical bugs that need to be resolved in the APP related to these features.

Project Status now have a `GroupName` that can be used instead of the `Name` property when displaying the current status of a project.  This allows status to share a name that better represents the overall state.

PR #338, opened 2020-06-18
Update frontend to use status codes not IDs.

devinleighsmith:
Also correct bugs in frontend introduced by https://github.com/bcgov/PIMS/pull/333

### DIFF
--- a/backend/dal/Services/Concrete/ProjectService.cs
+++ b/backend/dal/Services/Concrete/ProjectService.cs
@@ -521,6 +521,7 @@ namespace Pims.Dal.Services
             var status = this.Context.ProjectStatus
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId);
+            project.Status = status;
 
             // If the tasks haven't been specified, generate them.
             var taskIds = project.Tasks.Select(t => t.TaskId).ToArray();
--- a/frontend/src/components/common/form/Select.tsx
+++ b/frontend/src/components/common/form/Select.tsx
@@ -121,7 +121,12 @@ export const Select: React.FC<SelectProps> = ({
         multiple={multiple}
         onChange={onSelectChange}
         onBlur={(e: any) => {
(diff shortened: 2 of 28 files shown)


backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -521,6 +521,7 @@ public Project Update(Project project)
             var status = this.Context.ProjectStatus
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId);
+            project.Status = status;
```
devinleighsmith, line 524: Fixes an NPE thrown by project.IsProjectClosed(_options.Project)

frontend/src/components/common/form/Select.tsx
```diff
@@ -121,7 +121,12 @@ export const Select: React.FC<SelectProps> = ({
         multiple={multiple}
         onChange={onSelectChange}
         onBlur={(e: any) => {
-          if (type === 'number' && !isNaN(parseInt(value))) {
+          if (type === 'number' && multiple) {
```
devinleighsmith, line 124: handle numeric multi-selects properly.

frontend/src/features/projects/dispose/hooks/useStepForm.tsx
```diff
@@ -71,6 +71,7 @@ const useStepForm = () => {
       .catch((error: AxiosError) => {
         const msg: string = error?.response?.data?.error ?? error.toString();
         formikRef?.current?.setStatus({ msg });
+        throw Error('axios request failed');
```
devinleighsmith, line 74: Ensures that add/update failures prevent navigation to the next step (hiding the API error).

backend/dal/Services/Concrete/ProjectService.cs
Fosol: Ah essentially manual lazy-load.

PR #340, opened 2020-06-19
Documentation Review Bug

jdtoombs:
Tiny thing I noticed while doing some testing, just wanted to get it in in case it is blocking anyone.


frontend/src/features/projects/dispose/forms/ReviewProjectForm.tsx
```diff
@@ -22,7 +22,7 @@ const ReviewProjectForm = ({ canEdit }: { canEdit: boolean }) => {
   const { values } = useFormikContext<IProject>();
```
jdtoombs, line 22: Found one more little thing, updating

PR #341, opened 2020-06-19
PIMS-2488 Spread pins with the same lat long

asanchezr:
### Changes
- Implemented "spiderfy" positioning algorithm to spread pins with the same lat/long. It resembles spider legs, thus the name (I didn't come up with the name)
- If less than 8 points in a cluster, they are positioned in a circle. If 9 or more, they are placed in a spiral. This option is configurable.
- Added a way to override the default positioning of the map popup so it shows under the right pin when a cluster is "spiderfied"

### Notes
 - This PR also fixes critical map bug **PIMS-2591** causing the app to crash.

video-3


frontend/src/actions/parcelsActions.ts
```diff
@@ -123,11 +123,13 @@ export interface IAddress {
 export interface IParcelDetail {
   propertyTypeId: 0;
   parcelDetail: IParcel | null;
+  position?: [number, number]; // (optional) a way to override the positioning of the map popup
```
Fosol, line 126: This is a `Point` correct?  lat, long

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -203,16 +193,20 @@ const Map: React.FC<MapProps> = ({
...
+    // this is useful when showing "spiderfied" markers belonging to a cluster
+    const latlng = position ?? [parcelDetail.latitude as number, parcelDetail.longitude as number];
     return (
       <Popup
-        position={[parcelDetail.latitude as number, parcelDetail.longitude as number]}
+        position={latlng}
         offset={[0, -25]}
         onClose={() => onMarkerPopupClose?.(item)}
         closeButton={interactive}
+        autoPan={false} // fix for PIMS-2591: infinite loop crash
```
Fosol, line 209: Thanks for solving this

frontend/src/actions/parcelsActions.ts
asanchezr: Yes, typing could be improved

PR #343, opened 2020-06-19
PIMS-2315 (Exemption Submission) 

jdtoombs:


frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
```diff
@@ -49,3 +49,13 @@ export const ProjectDraftStepYupSchema = Yup.object().shape({
   name: Yup.string().required('Required'),
   description: Yup.string(),
 });
+
+export const EnhancedReferalExemptionSchema = Yup.object().shape({
+  exemptionRational: Yup.string().when(
+    'exemptionRequested',
+    (exemptionRequested: boolean, schema: any) =>
+      exemptionRequested
+        ? schema.required('Rationale is required when applying for an exemption.')
```
jdtoombs, line 58: do we want a minimum length for the rationale?

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -58,6 +58,20 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
   const projectNumber = queryString.parse(query).projectNumber;
   const historyReplace = history.replace;
 
+  let exemptionRequest: boolean = formikRef?.current?.values.exemptionRequested;
+  let exemptionRationale: string = formikRef?.current?.values.exemptionRational;
+
+  const [rationaleRequired, setRationaleRequired] = useState(false);
+
+  exemptionRequest &&
```
jdtoombs, line 66: see comments in yup schema

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -177,6 +191,7 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
                   onSave={() => onSave(formikRef)}
                   onNext={onNext}
                   saveDisabled={currentStatus.code === DisposeWorkflowStatus.Approval}
+                  nextDisabled={rationaleRequired}
```
jdtoombs, line 194: `rationaleRequired` only disables the button on save but the schema already kind of covers the functionality that this is trying to achieve.  this just makes it so the user can't physically click the button if they save the form with the exemption checkbox set to true and no rationale. may not be necessary

frontend/src/features/projects/dispose/components/ExemptionRequest.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const { values } = useFormikContext();
+  const checked = getIn(values, exemptionField);
+
+  return (
+    <React.Fragment>
+      <h3>
+        {sectionHeader}
+        <TooltipIcon toolTipId="exemptionTooltip" toolTip={tooltip} />
+      </h3>
+      <Form.Row className="ProjectExemptionRequestheck">
```
devinleighsmith, line 46: ProjectExemptionRequestheck -> ProjectExemptionRequestCheck

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
devinleighsmith: I'd prefer that this logic be moved to the ReviewProjectForm if possible, and then just use schema validation, rather then disabling the button. That way you provide the user feedback why that field is required, otherwise you are just disabling the button with no information as to why.

devinleighsmith: Also, I'm trying to keep this file from getting any more complicated since a lot of critical logic executes here.

frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
devinleighsmith: This seems like a Mikaela question.

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
jdtoombs: Sounds good @devinleighsmith, I'll adjust

frontend/src/features/projects/dispose/components/ExemptionRequest.tsx
jdtoombs: woops thank you

backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -414,6 +414,8 @@ public void Add_Project_SimpleFields()
             project.ProjectNumber = "test-generation-override";
             project.PrivateNote = "private note";
             project.PublicNote = "public note";
+            project.ExemptionRequested = true;
+            project.ExemptionRational = "Providing reasoning for exemption request";
```
asanchezr, line 418: /nit: do we want `ExemptionRational` or `ExemptionRationale`?

frontend/src/features/projects/dispose/components/ExemptionRequest.tsx
```diff
@@ -0,0 +1,66 @@
...
+  rationaleField: string;
+  /** Header used at the start of the component */
+  sectionHeader?: string;
+}
+
+/**
+ * Component used for requesting an ERP exemption while submitting a project.
+ * @param param0
+ */
+export default function ProjectNotes({
```
asanchezr, line 29: /nit: Could you please update the component default export to match the filename - i.e. `export default function ExemptionRequest(...)` ?

frontend/src/features/projects/dispose/interfaces.ts
```diff
@@ -204,6 +206,8 @@ export interface IApiProject {
   properties: IProjectProperty[];
   note: string;
   privateNote: string;
+  exemptionRequested?: boolean;
+  exemptionRational?: string;
```
asanchezr, line 210: This is coming from the server where it is named "exceptionRational" instead of "exceptionRationale" - not a big deal but should be consistent with naming everywhere else.

asanchezr: I had to look up the difference lol
https://writingexplained.org/rational-vs-rationale-difference#:~:text=Rationale%20vs.,reasons%20that%20support%20a%20claim.

jdtoombs: Yeah, I was noticing that too, it is just the API returns exemptionRational but on the frontend I only displayed it with the spelling `rationale` as I think that is the one we want

frontend/src/features/projects/dispose/components/ExemptionRequest.tsx
jdtoombs: oops

frontend/src/features/projects/dispose/interfaces.ts
jdtoombs: Going to adjust these though

backend/tests/unit/dal/Services/ProjectServiceTest.cs
jdtoombs: Fixing now, was just trying to avoid DB refresh but I need to satisfy my OCD lol

frontend/src/features/projects/dispose/forms/disposalYupSchema.ts
Fosol: You can ask Mikaela, but I wouldn't bother validating on an arbitrary length.

PR #347, opened 2020-06-22
PIMS-2548 ERP Workflow Form

devinleighsmith:


backend/dal/Migrations/Initial/PostDeploy/01-WorkflowProjectStatus.sql
```diff
@@ -42,5 +42,8 @@ INSERT INTO dbo.[WorkflowProjectStatus] (
 ), (
     2 -- ACCESS-DISPOSAL
     , 13 -- Denied
+), (
+    2 -- ACCESS-DISPOSAL
```
Fosol, line 46: I'm wondering if there should also be a workflow for ERP?

devinleighsmith: logged as possible tech debt https://pimsteam.atlassian.net/browse/PIMS-2600

PR #349, opened 2020-06-23
PIMS-2186 Automate production database backup before deploy

asanchezr:
Clipboard01


openshift/player.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+ARGV=${@:2}
```
Fosol, line 3: what does this script do in openshift?

asanchezr: It will simplify the pipeline code, improving maintanibility. It also allows you to run the same commands from your command line for debugging etc

PR #350, opened 2020-06-23
PIMS-2472 Project Summary View.

devinleighsmith:


frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -165,10 +172,17 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
...
-    if (row.statusCode === ReviewWorkflowStatus.PropertyReview) {
+    if (
+      row.statusCode === ReviewWorkflowStatus.PropertyReview &&
+      adminOrCanApprove(keycloak, row)
+    ) {
       history.push(`/dispose/projects/assess/properties?projectNumber=${row.projectNumber}`);
     } else if (ReviewWorkflowStatuses.includes(row.statusCode)) {
-      history.push(`/dispose/projects/approved?projectNumber=${row.projectNumber}`);
+      if (keycloak.hasClaim(Claims.ADMIN_PROJECTS)) {
+        history.push(`/dispose/projects/approved?projectNumber=${row.projectNumber}`);
```
Fosol, line 182: How are we handling when users manually return to a page?  Can they make edits when they shouldn't be allowed?

devinleighsmith: Users:
1) Can't get to any of the draft pages at all after submission. Those pages throw errors when manually navigated to after submission
2) Manually navigating back to the project review page will load that page, but none of the buttons are clickable.

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatusTransitions.sql
```diff
@@ -59,4 +59,13 @@ INSERT INTO dbo.[ProjectStatusTransitions] (
 ), (
     11 -- Approved for ERP
     , 14 -- Cancelled
+), (
+    11 -- Approved for ERP
+    , 15 -- On Hold
+), (
+    15 -- On Hold
+    , 12 -- On Hold
```
asanchezr, line 67: Shouldn't it be `12 -- Approved for SPL` ? 
Same with the rest below?

frontend/src/features/projects/dispose/interfaces.ts
```diff
@@ -138,13 +138,13 @@ export interface IProject {
...
-  OnHoldNotificationSentOn?: Date;
-  TransferredWithinGreOn?: Date;
-  ClearanceNotificationSentOn?: Date;
+  initialNotificationSentOn?: Date;
+  thirtyDayNotificationSentOn?: Date;
+  sixtyDayNoficationSentOn?: Date;
+  ninetyDayNotificationSentOn?: Date;
+  onHoldNotificationSentOn?: Date;
+  transferredWithinGreOn?: Date;
+  clearanceNotificationSentOn?: Date;
```
asanchezr, line 147: thanks for fixing this

backend/dal/Migrations/Initial/PostDeploy/01-ProjectStatusTransitions.sql
Fosol: Your comments don't line up

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -333,12 +333,24 @@ public Project Update(Project project)
...
                 .Include(p => p.Responses)
                 .SingleOrDefault(p => p.Id == project.Id) ?? throw new KeyNotFoundException();
-            this.ThrowIfNotAllowedToUpdate(originalProject, _options.Project);
 
             var userAgencies = this.User.GetAgencies();
             var originalAgencyId = (int)this.Context.Entry(originalProject).OriginalValues[nameof(Project.AgencyId)];
             if (!isAdmin && !userAgencies.Contains(originalAgencyId)) throw new NotAuthorizedException("User may not edit projects outside of their agency.");
 
+            //If the user isn't allowed to update the project, just update the notes.
+            if (!this.IsAllowedToUpdate(originalProject, _options.Project) && !project.IsProjectInDraft(_options.Project))
```
Fosol, line 342: For now we'll go with this, but it still has a security hole, as it allows for editing the note after a project is completed.  We'll have to readdress later.  Should add a / TODO

PR #356, opened 2020-06-25
PIMS-2554, 2555 (Exemption Approval)

jdtoombs:
Fixed all the merge conflicts/ updated tweaks, no longer draft pr!


backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -754,6 +754,23 @@ public Project SetStatus(Project project, Workflow workflow)
                         }
                     });
                     break;
+                    case ("AP-EXE"): // Approve for ERP Exemption
```
Fosol, line 757: This looks identical to `case ("AP-ERP")`.  If it is, just allow them both to use the same code.
```
case ("AP-EXE"):
case ("AP-ERP"):
// code here
```

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -41,11 +41,11 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
                 .Include(p => p.Agency).ThenInclude(a => a.Parent)
                 .AsNoTracking();
 
-            if (filter.AccessDisposal.HasValue && filter.AccessDisposal.Value)
+            if (filter.AccessWorkflow.HasValue && filter.AccessWorkflow.Value)
```
Fosol, line 44: We should extract as configurable in the appsettings.json later.

backend/entities/Models/ProjectFilter.cs
```diff
@@ -39,10 +39,10 @@ public class ProjectFilter : PageFilter
         public bool? CreatedByMe { get; set; }
 
         /// <summary>
-        /// get/set - Filter by workflow - ACCESS-DISPOSAL.
+        /// get/set - Filter by workflow - ACCESS-DISPOSAL or ACCESS-EXEMPTION.
         /// </summary>
         /// <value></value>
-        public bool? AccessDisposal { get; set; }
+        public bool? AccessWorkflow { get; set; }
```
Fosol, line 45: I'm just realizing we're using the work `ACCESS` instead of `ASSESS`...  Ugh

jdtoombs: haha i thought it was supposed to be that but then I thought I was missing something!

PR #369, opened 2020-06-29
PIMS-2627 Update project when changing milestone status

Fosol:
## Description

Previously it required submitting to both the `Update` and the `SetStatus` endpoints to update a project and change to a **milestone** status.

This fix allows the `SetStatus` endpoint to also save the project changes.

Additionally it fixes a notification bug.

@devinleighsmith I'll assign you this bug to fix the remaining issues in the front-end.  The **Next** button shouldn't require two AJAX requests anymore.  It should just be able to use the `SetStatus` when the status needs to be changed, otherwise use the `Update` endpoint when the status doesn't change.


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -376,7 +377,210 @@ public static bool IsProjectInDraft(this Entity.Project project, ProjectOptions
...
+                if (existingProperty == null)
+                {
+                    //Todo: Navigation properties on project object were causing concurrency exceptions.
+                    var eproperty = property.PropertyType == Entity.PropertyTypes.Land ? context.Parcels.Find(property.ParcelId) : context.Buildings.Find(property.BuildingId) as Entity.Property;
+                    // Ignore properties that don't exist.
+                    if (eproperty != null)
+                    {
+                        if (property.PropertyType == Entity.PropertyTypes.Land)
+                        {
+                            var existingParcel = context.Parcels
```
devinleighsmith, line 416: Actually, do we need to load the evaluations and fiscals here? it seems like we load those properties into existingParcel but then don't use them.

Fosol: Not sure I understand.  The update function will need to handle any changes made.  Additionally under most conditions it will need to recalculate the project financial values.

PR #371, opened 2020-06-29
PIMS-2623, PIMS-2624, PIMS-2627, PIMS-2629

devinleighsmith:
- Also update review approval form to allow deny with no validation.
- Correct potential race condition with usage of set... and useState().

### DIFF
--- a/frontend/src/features/projects/dispose/components/ApprovalActions.tsx
+++ b/frontend/src/features/projects/dispose/components/ApprovalActions.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import { useFormikContext } from 'formik';
 import { ReviewWorkflowStatus } from '../interfaces';
 import GenericModal from 'components/common/GenericModal';
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 import { DisplayError, Button } from 'components/common/form';
 import { cancellationWarning } from '../strings';
 import { useStepForm } from '..';
@@ -30,6 +30,11 @@ export const ApprovalActions = ({
   const { values, submitForm, validateForm } = useFormikContext<any>();
   const [cancel, setCancel] = useState(false);
   const { noFetchingProjectRequests } = useStepForm();
+  useEffect(() => {
+    if (submitStatusCode !== undefined) {
+      submitForm().then(() => setSubmitStatusCode(undefined));
+    }
+  }, [setSubmitStatusCode, submitForm, submitStatusCode]);
   return (
     <>
       <FlexRight>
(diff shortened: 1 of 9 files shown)


frontend/src/features/projects/dispose/components/ReviewApproveActions.tsx
```diff
@@ -48,7 +53,11 @@ export const ReviewApproveActions = ({
           }
           style={{ marginLeft: 10 }}
           onClick={() => {
-            setApproveERP(true);
+            validateForm().then((errors: any) => {
```
devinleighsmith, line 56: Don't display the approve dialog unless the form is valid.

frontend/src/features/projects/dispose/tabs/EnhancedReferralTab.tsx
```diff
@@ -18,15 +17,13 @@ const EnhancedReferralTab: React.FunctionComponent<IEnhancedReferralTabProps> =
...
 }: IEnhancedReferralTabProps) => {
-  const { submitForm } = useFormikContext();
   return (
     <>
       <AgencyResponseForm isReadOnly={isReadOnly} />
       <EnhancedReferralCompleteForm
         isReadOnly={isReadOnly}
         onClickOnHold={() => {
           setSubmitStatusCode(ReviewWorkflowStatus.OnHold);
-          submitForm();
```
devinleighsmith, line 29: calling this after the set... call above makes the assumption that the value submitStatusCode value will be set before submitForm() is executed. This is not guaranteed. I've refactored this so that the actions watch the value of submitStatusCode via useEffect and call submitForm if necessary.

frontend/src/features/projects/dispose/hooks/useStepForm.tsx
```diff
@@ -49,30 +47,13 @@ const useStepForm = () => {
...
-                  statusId: values.statusId,
-                  rowVersion: values.rowVersion,
-                }),
-              );
-        })
-        .catch((error: any) => {
-          const msg: string = error?.response?.data?.error ?? error.toString();
-          formikRef.current?.setStatus({ msg });
-        });
+        : (dispatch(updateProject(apiValues)) as any)
```
Fosol, line 52: Ah, much better.  I was trying to figure this out this morning, but couldn't.

frontend/src/features/projects/dispose/components/ApprovalActions.tsx
```diff
@@ -30,6 +30,11 @@ export const ApprovalActions = ({
   const { values, submitForm, validateForm } = useFormikContext<any>();
   const [cancel, setCancel] = useState(false);
   const { noFetchingProjectRequests } = useStepForm();
+  useEffect(() => {
```
devinleighsmith, line 33: See my comment at the end of this review. This ensures that the value of submitStatusCode is set correctly before submitForm() is called.

PR #382, opened 2020-07-03
PIMS-2816 DevOps - Add SonarQube to pipeline

asanchezr:
Initial PR that runs SonarQube scans for React/JavaScript in the CI/CD. 
The backend scan will be added in my next PR.

**NOTE** - this PR also fixes a bug in the build pipelines where they are running in "dry run" mode without actually deploying new code.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -87,6 +87,7 @@
     "eslint-config-prettier": "6.7.0",
     "eslint-plugin-prettier": "3.1.1",
     "husky": "4.2.3",
+    "jest-sonar-reporter": "2.0.0",
     "jest-styled-components": "7.0.2",
     "prettier": "1.19.1",
     "pretty-quick": "2.0.1",
@@ -98,7 +99,7 @@
     "build": "cross-env CI=true react-scripts build",
     "test": "cross-env CI=true react-scripts test --env=jsdom-fourteen --verbose --watchAll=false --runInBand",
     "test:watch": "react-scripts test --env=jsdom-fourteen",
-    "coverage": "cross-env CI=true react-scripts test --coverage --env=jsdom-fourteen --watchAll=false --runInBand",
+    "coverage": "cross-env CI=true react-scripts test --coverage --testResultsProcessor jest-sonar-reporter --env=jsdom-fourteen --watchAll=false --runInBand",
     "update-snapshots": "react-scripts test -- --updateSnapshot",
     "eject": "react-scripts eject",
     "lint": "eslint src/ --ext .jsx,.js,.ts,.tsx --max-warnings 0",
@@ -110,7 +111,13 @@
   "jest": {
     "collectCoverageFrom": [
       "src/**/*.{js,jsx,ts,tsx}",
-      "!<rootDir>/node_modules"
+      "!<rootDir>/node_modules/**",
+      "!<rootDir>/coverage/**",
+      "!<rootDir>/public/**",
+      "!<rootDir>/build/**",
+      "!<rootDir>/src/serviceWorker.**",
+      "!<rootDir>/src/setupProxy.*",
+      "!<rootDir>/src/setupTests.*"
     ],
     "coverageThreshold": {
       "global": {
--- a/frontend/sonar-project.properties
+++ b/frontend/sonar-project.properties
@@ -6,15 +6,14 @@ sonar.projectName=PIMS Frontend (React)
 
 # Paths below are relative to the sonar-project.properties file. Defaults to .
 sonar.sources=src
-sonar.exclusions=**/node_modules/**, **/coverage/**, **.spec.**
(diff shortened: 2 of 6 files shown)


frontend/package.json
```diff
@@ -87,6 +87,7 @@
     "eslint-config-prettier": "6.7.0",
     "eslint-plugin-prettier": "3.1.1",
     "husky": "4.2.3",
+    "jest-sonar-reporter": "2.0.0",
```
asanchezr, line 90: We need `jest-sonar-reporter` to pass unit test results to SonarQube; e.g "100 test executed, 80 pass, 20 skipped, etc"

This is different than code coverage, which we already were passing down

frontend/package.json
```diff
@@ -110,7 +111,13 @@
...
     "collectCoverageFrom": [
       "src/**/*.{js,jsx,ts,tsx}",
-      "!<rootDir>/node_modules"
+      "!<rootDir>/node_modules/**",
+      "!<rootDir>/coverage/**",
+      "!<rootDir>/public/**",
+      "!<rootDir>/build/**",
+      "!<rootDir>/src/serviceWorker.**",
+      "!<rootDir>/src/setupProxy.*",
+      "!<rootDir>/src/setupTests.*"
```
asanchezr, line 120: Ignoring things we don't want to collect coverage from

frontend/sonar-project.properties
```diff
@@ -6,15 +6,14 @@ sonar.projectName=PIMS Frontend (React)
...
-sonar.test.inclusions=**.spec.**
+sonar.test.inclusions=src/**/*.spec.**, src/**/*.test.**
 
-# --- TODO: integrate code coverage reports into SonarQube ---
-
-# sonar.javascript.lcov.reportPaths=frontend/coverage/lcov.info
-# sonar.testExecutionReportPaths=frontend/coverage/test-report.xml
+# --- integrate code coverage reports into SonarQube ---
+sonar.javascript.lcov.reportPaths=coverage/lcov.info
+sonar.testExecutionReportPaths=test-report.xml
```
asanchezr, line 16: - `lcov.info` is generated by Jest
- `test-report.xml` is generated by `jest-sonar-reporter`

openshift/pipelines/Jenkinsfile.cicd
```diff
@@ -34,6 +34,9 @@ pipeline {
     MAINTENANCE_DIRECTORY = "maintenance"
     DEVOPS_DIRECTORY = "openshift"
 
+    // SonarQube configuration
+    SONARQUBE_URL_INT = "http://sonarqube:9000"
```
asanchezr, line 38: We don't need the external URL here because our sonarqube instance runs on the same openshift namespace/project `<...>-tools`

openshift/pipelines/Jenkinsfile.cicd
```diff
@@ -197,23 +203,24 @@ pipeline {
           steps {
             script {
               dir(DEVOPS_DIRECTORY) {
-                sh "OC_JOB_NAME=${OC_JOB_NAME} ./player.sh build app-base apply"
-                sh "OC_JOB_NAME=${OC_JOB_NAME} ./player.sh build app apply"
+                sh "OC_JOB_NAME=${OC_JOB_NAME} ./player.sh build app-base -apply"
+                sh "OC_JOB_NAME=${OC_JOB_NAME} ./player.sh build app -apply"
```
asanchezr, line 207: Not doing dry runs anymore

openshift/pipelines/scripts/common.groovy
```diff
@@ -10,24 +10,33 @@ def version = "1.0"
...
   timeout(20) {
     dir('frontend') {
-      sh 'node --version'
-      sh 'npm --version'
       echo 'Installing NPM Dependencies...'
       sh 'npm ci'
-      echo 'Reporting Outdated and Vulnerable Dependencies...'
+      echo 'Reporting Vulnerable Dependencies...'
       sh 'npm audit || true'
-      sh 'npm outdated || true'
```
asanchezr, line 23: We don't do anything with outdated dependencies so I removed this check here. GitHub already provides a `dependabot` that alerts you when you have a vulnerable dependency that needs updating. Not the same but good enough. 
I'm okay to restore this if others think this is of value, but I have not seen that we regularly review and update dependencies

frontend/package.json
Fosol: I'm assuming this is a dev dependency?

openshift/pipelines/scripts/common.groovy
Fosol: I think dependabot is good enough presently.

frontend/package.json
asanchezr: Yes it is

PR #383, opened 2020-07-03
PIMS-2646 Export Projects to Excel

Fosol:
## Description

Adding API endpoints to export projects to CSV and Excel.
- GET /api/reports/projects
- POST /api/reports/projects/filter

Adding API endpoints to export surplus property list to CSV and Excel.  This report returns only projects that are in a status that is linked to the **SPL** workflow.

image

- GET /api/reports/surplus/properties/list
- POST /api/reports/surplus/properties/list/filter

> This work is incomplete, as there are a number of SPL details that need to be clarified

> This PR requires refreshing your DB

## Project Status Changes

I have updated the scripts to include all known status for this sprint, including transitions.  Let me know if you see any issues resulting in this update.

## Database Changes

There a number of partially complete changes to the DB to align with future changes.  The `Project` entity is getting really ugly with all the metadata we're adding (notes, dates, etc).

- There is now a `ProjectNotes` table that will slowly take over adding all these different notes
- There is now a `Project.Metadata` string property that we will use to serialize and deserialize a bunch of random data (i.e. dates).  This is not implemented yet and most likely will not be implemented this sprint.  For now we'll continue adding date columns...
- `ProjectStatus.IsOptional` now provides us a way to identify status that shouldn't be displayed in the stepper, as they are generally one or many optional steps.

## Bugs

On the **ERP Workflow Form** the bottom action buttons are incorrect.


backend/entities/Models/ProjectFilter.cs
```diff
@@ -74,12 +79,14 @@ public ProjectFilter(Dictionary<string, Microsoft.Extensions.Primitives.StringVa
             var filter = new Dictionary<string, Microsoft.Extensions.Primitives.StringValues>(query, StringComparer.OrdinalIgnoreCase);
             this.ProjectNumber = filter.GetStringValue(nameof(this.ProjectNumber));
             this.Name = filter.GetStringValue(nameof(this.Name));
-            this.StatusId = filter.GetIntNullValue(nameof(this.StatusId));
+            this.StatusId = filter.GetIntArrayValue("status");
+            this.StatusId = filter.GetIntArrayValue(nameof(this.StatusId));
```
devinleighsmith, line 83: are these two lines correct?

backend/api/Areas/Project/Models/Workflow/ProjectStatusModel.cs
```diff
@@ -20,6 +20,16 @@ public class ProjectStatusModel : Api.Models.CodeModel<int>
         /// get/set - whether or not this status is a milestone, requiring special transition logic.
         /// </summary>
         public bool IsMilestone { get; set; }
+
```
asanchezr, line 23: This file looks exactly the same as https://github.com/bcgov/PIMS/pull/383/files#diff-fb4b4bc721a7b997ed31135d056133d2 down to the same name. What makes this model different to the other one?

Fosol: models are controller specific to ensure changes don't break other endpoints.  It has the negative side effect where we will have to update multiple models when we want a change to go everywhere.  However this is more desirable than the other.

backend/entities/Models/ProjectFilter.cs
Fosol: They are, I decided it was likely someone would return `status` as the query parameter rather than the more verbose `statusId`.

asanchezr:
@Fosol you will have to fix the test or whatever is failing the GH actions. I re-ran them and they still fail

PR #385, opened 2020-07-03
PIMS-1831 (SRES ability to edit agency)

jdtoombs:


frontend/src/components/common/form/AutoCompleteText.tsx
```diff
@@ -15,6 +15,8 @@ export type IAutoCompleteProps = {
   textVal?: string;
   autoSetting?: string;
   required?: boolean;
+  /** to determine whether to show the code or label defaults to label*/
```
jdtoombs, line 18: so it just shows the abbreviation code when viewing to keep it consistent with the way it was before. also good for the long agency names.

PR #389, opened 2020-07-06
Agency fix, preliminary GRE exemption transfer

jdtoombs:


frontend/src/components/common/form/AutoCompleteText.tsx
```diff
@@ -105,7 +105,7 @@ export const AutoCompleteText: React.FC<IAutoCompleteProps> = ({
         <Form.Control
           autoComplete={autoSetting}
           name={field}
-          value={text}
+          value={text ? text : value}
```
Fosol, line 108: I think you can just use `text ?? value`

jdtoombs: oooh good to know! thanks

PR #394, opened 2020-07-07
PIMS-2722 Not included in SPL

devinleighsmith:


frontend/src/features/projects/dispose/testUtils.ts
```diff
@@ -86,6 +86,102 @@ export const tasks: ITask[] = [
   },
 ];
 
+export const mockWorkflow = [
```
Fosol, line 89: Ha, you created one too.

PR #396, opened 2020-07-08
PIMS-2758 On Hold remove from landing page

goemen:


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -592,5 +592,28 @@ public static void Merge(this Entity.Project originalProject, Entity.Project upd
...
             }
         }
+
+        /// <summary>
+        /// Set Project properties' visibility to other agencies
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="project"></param>
+        /// <param name="visibility"></param>
+        public static void SetProjectPropertiesVisiblity(this PimsContext context, Entity.Project project,  bool visibility)
```
Fosol, line 602: I always like good reusable extension methods

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -594,55 +594,19 @@ public Project SetStatus(Project project, Workflow workflow)
...
-                            p.Building.IsVisibleToOtherAgencies = true;
-                            this.Context.Buildings.Update(p.Building);
-                        }
-                        else if (p.ParcelId.HasValue)
-                        {
-                            p.Parcel.IsVisibleToOtherAgencies = true;
-                            this.Context.Parcels.Update(p.Parcel);
-                        }
-                    });
+                    this.Context.SetProjectPropertiesVisiblity(originalProject, true);
```
Fosol, line 597: I don't think this is actually valid.  This status is for approving an exemption.  At this point the project itself isn't approved.  Secondly it would be rare for exemption to go to ERP.  If they do, this logic would occur when `AP-ERP` occurred.

goemen: Should I make this change? @Fosol

PR #404, opened 2020-07-09
Excel Exports and Preliminary SPP Filter Setup

jdtoombs:
Still need to get the markers to change colour and to change the button for the landing page filter but included the initial setup in this PR.


backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -71,6 +71,8 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
                 query = query.Where(p => EF.Functions.Like(p.ProjectNumber, $"{filter.ProjectNumber}%"));
             if (filter.IgnorePropertiesInProjects == true)
                 query = query.Where(p => p.ProjectNumber == null);
+            if(filter.InSurplusPropertyProgram == true)
+                query = query.Where(p => !String.IsNullOrWhiteSpace(p.ProjectNumber));
```
Fosol, line 75: You will need to test to ensure this query actually sends the correct SQL.  This may actually request all projects from the DB, place them in memory and then check if the `ProjectNumber` value is not null, empty or whitespace.

You can either check **google** to find out if it's a supported function in LINQ EF, or you can output the generated SQL and review.

backend/entities/Models/PropertyFilter.cs
```diff
@@ -191,6 +198,7 @@ public override bool IsValid()
                 || this.SWLongitude.HasValue
                 || !String.IsNullOrWhiteSpace(this.ProjectNumber)
                 || this.IgnorePropertiesInProjects == true
```
Fosol, line 200: Both these boolean checks are probably not ideal.  I think it should be just checking if they not null `.HasValue`.  But it isn't critical for our use case.

frontend/src/components/maps/MapFilterBar.tsx
```diff
@@ -76,6 +79,7 @@ export type MapFilterChangeEvent = {
   classificationId: string;
   minLotSize: string;
   maxLotSize: string;
+  inSurplusPropertyProgram: boolean;
```
Fosol, line 82: Wouldn't this be a nullable boolean?  We only want to include it if it's set to `true`.  Again, probably not required to be changed.

jdtoombs: Oh yes you're right

backend/dal/Helpers/Extensions/PropertyExtensions.cs
devinleighsmith: FYI, a while ago I set up a logger such that all of the SQL gets dumped to the output window of VS WHEN DEBUGGING ONLY if you decide to go that route.

frontend/src/components/maps/MapFilterBar.tsx
```diff
@@ -149,6 +156,11 @@ const MapFilterBar: React.FC<MapFilterProps> = ({
               <span className="mx-2">-</span>
               <Input field="maxLotSize" placeholder="Max Lot Size" />
             </Col>
+            {keycloak.hasClaim(Claims.ADMIN_PROPERTIES) && (
+              <Col className="bar-item flex-grow-0">
+                <SppButton handleErpClick={noop} handleSppClick={noop} />
```
jdtoombs, line 161: placeholder for now so Goeme and I can both work on it as we have some overlap

backend/dal/Helpers/Extensions/PropertyExtensions.cs
jdtoombs: It is supported! 🎉

PR #405, opened 2020-07-10
PIMS-2783 Surplus Property Worflow Form.

devinleighsmith:


backend/dal/Migrations/Initial/PostDeploy/01-Tasks.sql
```diff
@@ -114,6 +114,38 @@ INSERT INTO dbo.[Tasks] (
     , 13
     , 0
     , 2
+), (
+    14
+    , 'Appraisal ordered'
```
Fosol, line 119: Since these are essentially duplicates of the original assessment workflow, are you copying the values over?  I don't think it's necessary for this iteration, but curious.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -602,6 +602,7 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId) ?? throw new KeyNotFoundException();
 
+            IEnumerable<int> incompleteTaskIds = new List<int>();
```
Fosol, line 605: Why pull this out of the `if` block?  I don't think we test for incomplete tasks if the status doesn't change.

Fosol: Additionally if a variable is only for private use there isn't much value in declaring an `IEnumerable<>` and then setting it as a `List<>`.   It just makes it more difficult to use.  Might as well go with `var` then you have all the access to `List` functionality natively.

backend/entities/Project.cs
```diff
@@ -111,6 +111,23 @@ public class Project : BaseEntity
         public string AgencyResponseNote { get; set; }
         #endregion
 
+        #region SPL
+        /// <summary>
+        /// get/set - Note summerizing offers received.
+        /// </summary>
+        public string OffersNote { get; set; }
```
Fosol, line 118: Where you planning to move this to the agency response table later?

backend/entities/Project.cs
```diff
@@ -111,6 +111,23 @@ public class Project : BaseEntity
...
+
+        /// <summary>
+        /// get/set - Text field describing project purchaser.
+        /// </summary>
+        public string Purchaser { get; set; }
+
+        /// <summary>
+        /// get/set - Whether or not the contract for the project is conditional
+        /// </summary>
+        public bool? IsContractConditional { get; set; }
```
Fosol, line 128: Where you planning to move this to the agency response table later?

backend/entities/Project.cs
```diff
@@ -176,6 +193,11 @@ public class Project : BaseEntity
         /// get/set - When the project was disposed.
         /// </summary>
         public DateTime? DisposedOn { get; set; }
+
+        /// <summary>
+        /// get/set - When the project was disposed.
+        /// </summary>
+        public DateTime? OfferAcceptedOn { get; set; }
```
Fosol, line 200: Where you planning to move this to the agency response table later?

backend/entities/Project.cs
```diff
@@ -235,6 +257,11 @@ public class Project : BaseEntity
         /// get/set - Record the interest component.
         /// </summary>
         public decimal InterestComponent { get; set; }
+
+        /// <summary>
+        /// get/set - Amount offered during SPL
+        /// </summary>
+        public decimal OfferAmount { get; set; }
```
Fosol, line 264: Where you planning to move this to the agency response table later?

backend/dal/Migrations/Initial/PostDeploy/01-Tasks.sql
devinleighsmith: Yes, the frontend copies them over when approved for SPL -> Pre-Marketing

backend/dal/Services/Concrete/ProjectService.cs
devinleighsmith: Right, I originally pulled this out because I was using the list in the DIS calculations below but was able to refactor to be more generic. Will remove.

backend/entities/Project.cs
devinleighsmith: Time permitting. Focusing on getting something matching the lofi first.

backend/entities/Project.cs
devinleighsmith: This one no. To me the agency response table is for per-agency data. I'm currently under the impression that only one contract can be accepted at a time, so there wouldn't be a benefit to adding a column for every row.

backend/entities/Project.cs
devinleighsmith: Same here, wasn't planning on adding this if there is only one per project.

backend/entities/Project.cs
devinleighsmith: Also thinking this occurs once per project.

devinleighsmith: this one would make sense to add I think

frontend/src/components/common/form/Check.tsx
```diff
@@ -67,20 +76,43 @@ export const Check: React.FC<CheckProps> = ({
...
-          disabled={disabled}
-          custom={custom}
-          defaultChecked={checked}
-          isInvalid={!!touch && !!error}
-          {...rest}
-          value={checked}
-          placeholder={placeholder}
-          onChange={handleChange}
-        />
+        <>
```
devinleighsmith, line 79: We've had a number of instances of radio buttons being requested for boolean flags in the DB, so I decided to stop fighting it and add generic support for that in this component.

frontend/src/components/common/form/DisplayError.tsx
```diff
@@ -5,15 +5,18 @@ import { useFormikContext, getIn } from 'formik';
...
+  /** Display error even if field hasn't been touched */
+  errorPrompt?: boolean;
 };
 
 /**
  * Formik-connected, thin wrapper around React-Bootstrap to display form validation errors
  */
-export const DisplayError: React.FC<DisplayErrorProps> = ({ field }) => {
-  const { errors } = useFormikContext();
+export const DisplayError: React.FC<DisplayErrorProps> = ({ field, errorPrompt }) => {
```
devinleighsmith, line 15: This was technically a bug, the error shouldn't display unless the field is touched. Added a prop to give backwards compatibility.

frontend/src/constants/reducerTypes.ts
```diff
@@ -36,6 +36,7 @@ export enum ProjectReducers {
   WORKFLOW = 'projectWorkflow',
   WORKFLOW_TASKS = 'projectWorkflowTasks',
   PROJECT = 'project',
-  PROJECT_APPROVAL_TAB = 'projectApprovalTab',
+  ERP_TAB = 'erpTab',
```
devinleighsmith, line 39: separate out the slices for erp/spl since they aren't on the same page anymore.

frontend/src/features/projects/assess/components/ReviewApproveActions.tsx
```diff
@@ -56,6 +56,9 @@ export const ReviewApproveActions = ({
             validateForm().then((errors: any) => {
               if (Object.keys(errors).length === 0) {
                 setApproveERP(true);
+              } else {
```
devinleighsmith, line 59: This is really dumb, but by default the validateForm function does not touch all the fields and so the errors won't display. This change forces those errors to display if any exist but pretending to try and submit the form even though we know it'll fail validation (which touches all the fields).

frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
```diff
@@ -53,7 +52,7 @@ export const validateTasks = (project: IProject) => {
  * {isReadOnly formikRef} formikRef allow remote formik access
  */
 const ReviewApproveStep = ({ formikRef }: IStepProps) => {
-  const { project, goToDisposePath } = useStepper();
+  const { project, goToDisposePath } = useProject();
```
devinleighsmith, line 55: Updating to not use useStepper outside of the initial dispose stepper steps.

frontend/src/features/projects/common/ApprovalTransitionPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+        throw Error('No valid route for current project status');
+      }
+      history.replace(`${project?.status?.route}?projectNumber=${project.projectNumber}`);
+    })
+    .catch(() => {
+      throw Error(`Failed to update status to ${toStatusCode}`);
+    });
+};
+
+const ApprovalTransitionPage: React.FunctionComponent<IApprovalTransitionPageProps> = props => {
```
devinleighsmith, line 26: This page handles transitioning between statuses automatically when a status only has one valid toStatus. Made this a page to try and allow for re-use whenever this situation arises (currently just twice). The user won't be able to navigate to this page with the back/forward buttons due to the use of history.replace.

frontend/src/features/projects/common/ApprovalTransitionPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const toStatus = _.find(workflowStatuses, { code: project?.statusCode })?.toStatus;
+
+  useEffect(() => {
+    if (toStatus === undefined) {
+      dispatch(fetchProjectWorkflow(project?.workflowCode));
+      return;
+    }
+    if (project !== undefined && !isTransitioned) {
+      if (toStatus?.length !== 1) {
+        history.replace(`${project.status?.route}?projectNumber=${project.projectNumber}`);
```
devinleighsmith, line 41: The project is not in a state to allow automatic transitions, so just navigate to the state given by the status route.

frontend/src/features/projects/common/ProjectRouter.tsx
```diff
@@ -32,33 +34,26 @@ const ProjectRouter = ({ location }: { match: Match; location: Location }) => {
...
-    state => (state.network as any)[ProjectActions.GET_PROJECT] as any,
+    state => (state.network as any)[ProjectActions.GET_PROJECT_WORKFLOW] as any,
   );
-  const project: any = useSelector<RootState, IProjectWrapper>(state => state.project).project;
   useEffect(() => {
     if (projectNumber !== null && projectNumber !== undefined) {
       dispatch(clearProject());
-      dispatch(fetchProject(projectNumber as string));
+      (dispatch(fetchProject(projectNumber as string)) as any).then((project: IProject) => {
+        dispatch(fetchProjectWorkflow(project?.workflowCode));
```
devinleighsmith, line 46: We only load (and store in redux) the current workflow based on project status. As a result we need to load the workflow dynamically based on the project status before displaying a component to ensure consistent behaviour.

frontend/src/features/projects/common/ProjectRouter.tsx
```diff
@@ -32,33 +34,26 @@ const ProjectRouter = ({ location }: { match: Match; location: Location }) => {
...
-    }
-  }, [dispatch, project, workflowCode, workflowStatusesLength]);
-
   if (projectNumber !== null && projectNumber !== undefined && getProjectRequest?.error) {
     throw Error(`Unable to load project number ${projectNumber}`);
   }
   return (
     <>
-      {getProjectRequest?.isFetching !== true ? (
+      {getProjectRequest?.isFetching === false && projectNumber === project.projectNumber ? (
```
devinleighsmith, line 56: also ensure that the project that we have stored in redux matches the projectnumber in the path.

frontend/src/features/projects/common/forms/AdditionalPropertyInformationForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface IAdditionalPropertyInformationFormProps {
+  isReadOnly?: boolean;
+}
+
+/**
+ * Surplus Properties information - in addition to standard property information.
+ * @param param0 isReadOnly disable editing
+ */
+const AdditionalPropertyInformationForm = ({
```
devinleighsmith, line 14: This is a common form component used in both the SPL and ERP project information tabs.

frontend/src/features/projects/common/forms/AppraisalCheckListForm.tsx
```diff
@@ -9,14 +9,17 @@ import TooltipIcon from 'components/common/TooltipIcon';
 interface IAppraisalCheckListFormProps {
   className?: string;
   isReadOnly?: boolean;
+  taskStatusCode?: string;
```
devinleighsmith, line 12: since there are multiple copies of the Appraisal tasks, this allows the parent component to specify which one we want (the optional or required variant).

frontend/src/features/projects/common/interfaces.ts
```diff
@@ -125,31 +125,38 @@ export interface IProject {
...
+  disposedOn?: Date | string;
+  offerAcceptedOn?: Date | string;
   netBook?: number;
   assessed?: number;
   estimated?: number;
-  workflowCode?: string;
+  workflowCode?: string | '';
+  offerAmount?: number | '';
+  isContractConditional?: boolean;
+  purchaser?: string | '';
```
devinleighsmith, line 159: This is all to do with formik not allowing controlled fields to have the value of undefined.

frontend/src/features/projects/common/interfaces.ts
```diff
@@ -219,7 +232,8 @@ export interface IProjectProperty {
   building?: IApiProperty;
 }
 
-export const initialValues: any = {
+export const initialValues: IProject = {
```
devinleighsmith, line 235: Now whenever we get a project from the api we start with this project and overwrite all valued fields using ... syntax. This ensures that all fields have valid formik defaults without having to process each field individually, which is verbose and error prone.

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -74,7 +74,8 @@ export const toFlatProject = (project?: IApiProject) => {
     };
     return property;
   });
-  const flatProject: IProject = { ...project, properties: flatProperties };
+  //always copy the project values over initial values, this ensures that formik's requirement of non-undefined fields is satisfied.
+  const flatProject: IProject = { ...initialValues, ...project, properties: flatProperties };
```
devinleighsmith, line 78: This is what I mentioned above. Start with the default empty string values and then overwrite with the actual project values to ensure no undefined values leak into the forms.

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -171,5 +172,10 @@ export const toApiProject = (project: IProject) => {
     exemptionRationale: project.exemptionRationale,
     exemptionRequested: project.exemptionRequested,
   };
+  // convert all empty strings (required by formik) to undefined
+  Object.keys(apiProject).forEach(key => {
```
devinleighsmith, line 176: Of course, going back from app -> api we can't/don't want to write a whole bunch of empty strings back, so clean those up before submission.

frontend/src/features/projects/common/tabs/ProjectInformationTab.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
```
devinleighsmith, line 1: moved this file as it is used by ERP and SPL

frontend/src/features/projects/common/tabs/tabValidation.ts
```diff
@@ -0,0 +1,74 @@
+import { SPPApprovalTabs, IProject, IProjectTask } from '..';
```
devinleighsmith, line 1: This validation logic is also used by ERP and SPL so extracted out to this common file.
The motivation for this logic is that we need a way to validate all of the tabs (since they have required fields) before submission, but also inform the user that there are errors on tabs that are currently not being displayed. the ValidationGroup provides a way of splitting validation out on a per-tab basis, so that the component knows what tab an error belongs to.

frontend/src/features/projects/dispose/ProjectDisposeLayout.test.tsx
```diff
@@ -38,102 +39,6 @@ const loc = {
   hash: '',
 } as Location;
 
-const mockWorkflow = [
```
devinleighsmith, line 41: this is copy pasted into a few test, trying to use the one in testutils instead.

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -110,22 +114,11 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
 
   //If the current route isn't set, set based on the query project status.
   useEffect(() => {
-    if (location.pathname === '/dispose') {
-      if (project.status?.route !== undefined && projectNumber !== undefined) {
-        historyReplace(`/dispose${project.status?.route}?projectNumber=${project.projectNumber}`);
-      } else {
-        dispatch(clearProject());
-      }
+    if (location.pathname === '/dispose' && workflowStatuses?.length > 0) {
```
devinleighsmith, line 117: Due to to the previous change in routing behaviour I was able to simplify this further.

frontend/src/features/projects/dispose/__snapshots__/ProjectDisposeView.test.tsx.snap
```diff
@@ -3,115 +3,5 @@
...
-          onClick={[Function]}
-          type="button"
-        >
-          Save
-        </button>
-      </div>
-    </div>
-  </div>
-</div>
+/>
```
devinleighsmith, line 6: I need to clean this up.

frontend/src/features/projects/dispose/hooks/stepperContext.tsx
```diff
@@ -18,26 +11,17 @@ export const StepperContext = createContext({} as any);
 export const StepContextProvider = (props: { children?: any }) => {
   // Use State to keep the values. Initial values are obtained from StepperContextProvider’s props.
   const [currentStatus, setCurrentStatus] = useState<IStatus>();
+  const [disposeWorkflowStatuses, setDisposeWorkflowStatuses] = useState<IStatus[]>();
   // Make the context object (or array)
-  const stepContext = { currentStatus, setCurrentStatus };
+  const stepContext = { currentStatus, setCurrentStatus, disposeWorkflowStatuses };
```
devinleighsmith, line 16: the initial dispose steps only ever need to load the workflow once, so might as well do that here during initial context setup instead of the hook (which was resulting in multiple api calls in some situations).

frontend/src/features/projects/dispose/hooks/useStepper.test.tsx
```diff
@@ -114,22 +43,19 @@ describe('useStepper hook functionality', () => {
   afterAll(() => {
     jest.clearAllMocks();
   });
-  describe('getNextStep', () => {
+  describe('getNextWorkflowStatus', () => {
```
devinleighsmith, line 46: just testing the function called by the hook.

frontend/src/features/projects/dispose/hooks/useStepper.ts
```diff
@@ -131,37 +128,37 @@ const useStepper = () => {
       project?.statusId !== undefined &&
       project.statusId > 0 &&
       project.statusCode !== ReviewWorkflowStatus.ExemptionReview &&
-      diposeWorkflowStatuses?.length > 0 &&
+      disposeWorkflowStatuses?.length > 0 &&
```
devinleighsmith, line 131: just a typo.

frontend/src/features/projects/dispose/steps/ProjectDraftStep.tsx
```diff
@@ -32,7 +32,6 @@ const ProjectDraftStep = ({ isReadOnly, formikRef }: IStepProps) => {
       <Formik
         initialValues={draftFormValues}
         validationSchema={ProjectDraftStepYupSchema}
-        enableReinitialize={true}
```
devinleighsmith, line 35: Due to some changes in this PR there is no need to re-initialize the form after initial load. This actually simplifies the workflow of these dispose workflows a lot

frontend/src/features/projects/dispose/tabs/DocumentationTab.tsx
```diff
@@ -1,43 +0,0 @@
-import * as React from 'react';
```
devinleighsmith, line 1: The documentation tab should never have been under /dispose

frontend/src/features/projects/erp/components/ApprovalActions.tsx
```diff
@@ -49,7 +51,7 @@ export const ApprovalActions = ({
             showSubmitting
             isSubmitting={!noFetchingProjectRequests}
             onClick={() => {
-              submitForm();
+              (submitDirectly ?? submitForm)();
```
devinleighsmith, line 54: Allow the save action to bypass formik's validation. useful for saving in-progress work without needing to have all required fields filled out.

frontend/src/features/projects/erp/steps/ErpStep.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
```
devinleighsmith, line 1: This is actually just a rename of approvalStep->ErpStep

ApprovalStep made sense when the design was to have the ERP and SPL tabs on the same page, but since that is no longer the case updating the name made sense.

frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -0,0 +1,142 @@
...
+import './SplStep.scss';
+import { DocumentationStepSchema } from 'features/projects/dispose';
+
+const CenterBoldText = styled.div`
+  text-align: center;
+  font-family: 'BCSans-Bold';
+  margin-bottom: 1.5rem;
+`;
+
+const copyAppraisalTasks = (project: IProject): IProject => {
```
devinleighsmith, line 40: Unfortunately we need to duplicate any interactions the user may have had with the the optional appraisal tasks to the required appraisal tasks displayed on this form.

frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -0,0 +1,142 @@
...
+        //only perform validation if the user is attempting to dispose the project.
+        validate={(values: IProject) =>
+          submitStatusCode === ReviewWorkflowStatus.Disposed || submitStatusCode === undefined
+            ? handleValidate(values, splValidationGroups)
+            : Promise.resolve({})
+        }
+      >
+        {({ values }) => (
+          <Form>
+            <StepStatusIcon
```
devinleighsmith, line 111: I'd like to move this out to a layout actually since this is used in a few places.

PR #409, opened 2020-07-11
PIMS-2638 Project Financial Data

Fosol:
## Description

Adding additional financial data to projects.  Updating project reporting (exports to Excel).  Enhancing notifications.

**Corrected bug in DEV that is blocking testing.**

> Requires DB Refresh

> Note - Tools have been moved from `/backend/tools` to `/tools`.  You will also need to update your `.env` configuration files associated with tools from `Keycloak__ClientSecret` to `Auth__Keycloak__Secret`.

### Highlights

- Surplus Property List projects can now be imported into PIMS.
- New Tool - Convert Excel to JSON
- Updated Tool - Import to PIMS now supports projects
- CHES Notifications - Opt-in configuration

### Notifications

Previously by default notifications would go out to their intended recipients.  This has been changed to require you opt-in.  This will reduce risk of sending notifications when testing.

To configure notifications we now have the following options;

| Key | Default | Description |
| --- | --- | --- |
| EmailEnabled | false | When set to false, no email will be sent by the system. |
| EmailAuthorized | false | When set to false, all email is sent to either the `OverrideTo` or to the currently logged in user. |
| OverrideTo | string | A semi-colon separated list of email addresses you want all emails to go to instead of their intended recipients. |
| BccUser | bool | Always BCC any emails the current user activates. |
| AlwaysBcc | string | A semi-colon separated list of email addresses you want all emails to be BCCed to. |

### Reporting

We have to primary **Disposal Project** reports.  

- A generic list of all the projects based on the current filter.
- A specifically formatted SPL project export that should only contain projects in the SPL workflow.  This export is made to match Maggie's data.  Please note that we are still missing some of that data, and don't have a way to manage that data within PIMS.

### New Tool - Convert Excel to JSON

A common activity required to extract, parse and clean data received from Excel files is now made easier.

This tool allows us to map Excel Sheets to JSON formats and generate the results.

#### Features

- Configure mapping of Excel column to JSON property name
- Control exported data types
- Convert data with a few pre-configured functions
- Map data values to clean bad data, or renaming


backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -20,6 +20,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.StatusId, src => src.StatusId)
                 .Map(dest => dest.StatusCode, src => src.Status.Code)
                 .Map(dest => dest.Status, src => src.Status)
+                .Map(dest => dest.RiskId, src => src.RiskId)
```
Fosol, line 23: Risks are used for financial calculations [Complete, Green, Yellow, Red]

backend/api/Areas/Tools/Controllers/ImportController.cs
```diff
@@ -68,6 +68,33 @@ public IActionResult ImportProperties([FromBody] Model.ImportPropertyModel[] mod
 
             return new JsonResult(parcels);
         }
+
+
+        /// <summary>
+        /// POST - Add an array of new properties to the datasource.
```
Fosol, line 74: Need to update the comment...

backend/api/Areas/Tools/Controllers/ImportController.cs
```diff
@@ -68,6 +68,33 @@ public IActionResult ImportProperties([FromBody] Model.ImportPropertyModel[] mod
...
+        /// <param name="stopOnError">Whether to throw an error if a failture occurs.</param>
+        /// <param name="defaults">A semi-colon separated list of key=value pairs of default values for properties if they are null or not provided.</param>
+        /// <returns>The properties added.</returns>
+        [HttpPost("projects")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(IEnumerable<Model.ProjectModel>), 200)]
+        [ProducesResponseType(typeof(Pims.Api.Models.ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "tools-import" })]
+        [HasPermission(Permissions.SystemAdmin)]
+        public IActionResult ImportProjects([FromBody] Model.ImportProjectModel[] models, bool stopOnError = true, string defaults = null)
```
Fosol, line 88: We can now import projects.

Fosol: I would like to make this more generic, but atm it is built to support SRES data.

backend/api/Startup.cs
```diff
@@ -192,6 +192,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddDbContext<PimsContext>(options =>
             {
                 options.UseSqlServer(builder.ConnectionString).UseLoggerFactory(debugLoggerFactory);
+                if (!this.Environment.IsProduction())
+                {
+                    options.EnableSensitiveDataLogging();
```
Fosol, line 197: Better error messages.

backend/api/ches.json
```diff
@@ -2,6 +2,7 @@
   "Ches": {
     "AuthUrl": "https://sso-dev.pathfinder.gov.bc.ca/auth/realms/jbd6rnxw/protocol/openid-connect/token",
     "HostUri": "https://ches-master-9f0fbe-dev.pathfinder.gov.bc.ca/api/v1",
-    "From": "Property Inventory Management System <RealPropertyDivision.Disposals@gov.bc.ca>"
+    "From": "Property Inventory Management System <RealPropertyDivision.Disposals@gov.bc.ca>",
+    "EmailEnabled": true
```
Fosol, line 6: Set to `true` so that emails are on, otherwise none will be sent.
Email will currently by default go to the user who activated them.

backend/ches/Pims.Ches.csproj
```diff
@@ -10,9 +10,9 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.4" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.4" />
-    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.6.0" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.5" />
```
Fosol, line 13: Updated dependencies in all projects.

backend/dal/Configuration/ProjectSnapshotConfiguration.cs
```diff
@@ -0,0 +1,41 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Pims.Dal.Entities;
+
+namespace Pims.Dal.Configuration
+{
+    /// <summary>
+    /// ProjectSnapshotConfiguration class, provides a way to configure project Snapshot in the database.
+    ///</summary>
+    public class ProjectSnapshotConfiguration : BaseEntityConfiguration<ProjectSnapshot>
```
Fosol, line 10: A snapshot will be used to capture their weekly/monthly reporting requirements.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -155,6 +157,38 @@ public Project Get(int id)
             return project;
         }
 
+        /// <summary>
+        /// Get the project for the specified 'id'.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="includes"></param>
+        /// <returns></returns>
+        public Project Get(int id, params Expression<Func<Project, object>>[] includes)
```
Fosol, line 166: Provides us a basic way to only return what we need.  Need to figure out how to do a `ThenInclude` like this too.

asanchezr:
Could you please see if there's a way to clean up files that seem like they shouldn't be in source control? I saw a bunch of dlls, pdbs, etc here... thanks

Fosol:
Files have been removed.  Thanks for catching that.

PR #415, opened 2020-07-14
Agency bug, tests, validation bugs

jdtoombs:


frontend/src/components/common/form/AutoCompleteText.tsx
```diff
@@ -16,6 +16,7 @@ export type IAutoCompleteProps = {
   required?: boolean;
```
jdtoombs, line 16: This felt very hacky, this is for [this story](https://pimsteam.atlassian.net/browse/PIMS-2890?atlOrigin=eyJpIjoiMGNhZGYwYTAwMzFjNDRhY2IxOTM5MGZlYzU2YjA1ZDMiLCJwIjoiaiJ9). As agency and sub agency share the same `agencyId` I struggled a little bit to come up with the best solution. I originally tried filtering the agencies from the land form but we need to have all the agencies in order to set the parent agency etc.

Fosol: The lookup has the `parentId`.  So you should always be able to filter.  If they select a sub-agency, then it should automatically set the **Agency** to the parent, and the **Sub-Agency** to the selected value.

jdtoombs: Ok should be fine then, does this all currently. Just wasn't sure if I was adding too much logic to the autocomplete component

asanchezr: It does feel like a lot of "custom logic" has been added to a "generic" auto-complete but I don't have a better solution. Other than pushing the custom logic to a new "AgencyAutoComplete" and that might be too much work. I say we leave as-is for now

PR #416, opened 2020-07-14
PIMS-1659 - ERP properties map filter, PIMS-1274 - Geocoder street ad…

goemen:
…dress


backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -135,5 +146,54 @@ public static void ThrowIfPropertyNotInProjectAgency(this Entity.Property parcel
...
+                    var statuses = context.ProjectStatus.
+                        Where(ps => propertyProjectStatuses.Select(x => x.Id).Contains(ps.Id))
+                        .Select(s => new {Id = s.Id, Code = s.Code})
+                        .ToArray();
+
+                    properties.ForEach(property =>
+                    {
+                        if (!String.IsNullOrWhiteSpace(property.ProjectNumber))
+                        {
+                            var propertyProjectStatus = propertyProjectStatuses
```
Fosol, line 177: If all we want is to get the property project status you can change this whole function.  Our business rule is a property can only be associated to a single **active** project.  Since I believe the primary purpose of this is for the ERP statuses, then just update the `Property` to include a collection association to `Projects`.

Something like the following.  This is just seudo-code.
```
this.Context.Properties(p => p.Projects.Any(pr => statusIds.Contains(pr.StatusId))).Select(p => new { Property = p, Status = p.Projects.FirstOrDefault(pr => statusIds.Contains(pr.StatusId)) })
```

backend/entities/Models/PropertyFilter.cs
```diff
@@ -52,6 +52,12 @@ public abstract class PropertyFilter : PageFilter
         /// <value></value>
         public bool? InSurplusPropertyProgram { get; set; }
 
+        /// <summary>
+        /// get/set - Flag indicating to show only properties that are in en enhanced referral process.
+        /// </summary>
+        /// <value></value>
+        public bool? InEnhancedReferralProcess { get; set; }
```
Fosol, line 59: Should rename to `InEnhancedReferralProgram`

backend/entities/Property.cs
```diff
@@ -34,6 +36,7 @@ public abstract class Property : BaseEntity
         /// <summary>
         /// get/set - The status for this property.
         /// </summary>
+        [NotMapped]
         public PropertyStatus Status { get; set; }
```
Fosol, line 40: This is awkward.  I generally only do this for transitory data.  I believe this information is only required in models returned from the API.  I see no reason to add it to the entity.

Fosol: Update the relevant API models to include this property.  Don't include it in the entity.

backend/entities/Views/Property.cs
```diff
@@ -24,6 +24,13 @@ public class Property
         /// </summary>
         public string ProjectNumber { get; set; }
 
+
+        /// <summary>
+        /// get/set - The RAEG/SPP project status
+        /// </summary>
+        [NotMapped]
+        public string ProjectStatus { get; set; }
```
Fosol, line 32: Again only include in the API models, not in the DAL.  Additionally it doesn't make sense at this level, as properties can be associated to many projects and thus have many status (although only one that is active at any given point in time).

frontend/src/components/common/form/SppButton.tsx
```diff
@@ -3,40 +3,51 @@ import './SppButton.scss';
...
-              <Col xs={2} className="marker">
-                <Image src={require('assets/images/marker-icon-2x-violet.png')} />
-              </Col>
-              <Col className="label">Surplus Properties List</Col>
-            </Row>
-          </Tooltip>
-        )}
-      </Overlay>
-    </div>
+    <ClickAwayListener onClickAway={() => setVisible(false)}>
```
devinleighsmith, line 23: This is nice

backend/tests/unit/api/Controllers/Property/SearchControllerTest.cs
```diff
@@ -211,7 +212,8 @@ public void GetProperties_Query_Success(Uri uri, bool includeParcels, bool inclu
             var service = helper.GetService<Mock<IPimsService>>();
             var mapper = helper.GetService<IMapper>();
 
-            var items = parcels.Select(p => new Entity.Views.Property(p)).Concat(buildings.Select(b => new Entity.Views.Property(b)));
+            var items = parcels.Select(p => new ProjectProperty(new Entity.Views.Property(p)))
+                .Concat(buildings.Select(b => new ProjectProperty(new Entity.Views.Property(b))));
```
devinleighsmith, line 216: Doesn't really matter, but looks like something got messed up here.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -114,11 +114,23 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
             if (filter.MaxAssessedValue.HasValue)
                 query = query.Where(p => p.Assessed <= filter.MaxAssessedValue);
 
+            if (filter.InEnhancedReferralProcess.HasValue && filter.InEnhancedReferralProcess.Value)
+            {
+                var statuses = context.Workflows.Where(w => w.Code.Equals("ERP"))
```
Fosol, line 119: Not sure if it will impact the SQL generated, but you really just want the `FirstOrDefault(w => w.Code == "ERP")`.  I also don't know if LINQ will translate `w.Code.Equals("ERP")`.  It will most likely request all data and put it into memory again.

Fosol: I also think this will result in pulling back properties that are in terminal statuses (i.e. Cancelled, Transfered, etc.).  You probably need to also filter on `!w.IsTerminal`.

PR #417, opened 2020-07-14
Updates from Hi-Fi review. Also fix spinner display within buttons.

devinleighsmith:


frontend/src/components/common/form/Button.tsx
```diff
@@ -46,7 +46,7 @@ export const Button: React.FC<ButtonProps> = ({
           size="sm"
           role="status"
           as="span"
-          style={{ marginLeft: '.5rem' }}
+          style={{ marginLeft: '.5rem', padding: '.5rem' }}
```
devinleighsmith, line 49: Part of the spinner was getting truncated and then spinning like that.

frontend/src/components/common/form/Check.tsx
```diff
@@ -109,7 +109,7 @@ export const Check: React.FC<CheckProps> = ({
               {...rest}
               value={!checked}
               placeholder={placeholder}
-              onChange={handleChange}
+              onChange={() => setFieldValue(field, !checked)}
```
devinleighsmith, line 112: prevent "false" and "true" from being saved by formik

frontend/src/components/common/form/FastFiscalYearInput.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { memo, useEffect } from 'react';
```
devinleighsmith, line 1: masked input that saves/accepts YYYY format from the DB but displays in YYYY/YYYY format.

PR #420, opened 2020-07-15
PIMS-2790 SPL workflow tests.

devinleighsmith:


backend/tests/unit/dal/Services/ProjectServiceTest.cs
```diff
@@ -2097,6 +2097,234 @@ public async void SetStatusAsync_NotInSpl_InvalidOperationException()
...
+            result.Status.Should().Be(disposed);
+            result.DisposedOn.Should().NotBeNull();
+            var property = result.Properties.First().Parcel;
+            property.AgencyId.Should().Be(2);
+            property.ClassificationId.Should().Be(2);
+            property.IsVisibleToOtherAgencies.Should().BeFalse();
+        }
+
+        [Fact]
+        public async void SetStatusAsync_Disposed_InvalidOperationException()
```
devinleighsmith, line 2302: This will need some more tests when the backend logic for disposed is finished (disposed classification, null agency).

PR #421, opened 2020-07-15
low hanging fruit:

devinleighsmith:
low hanging fruit:
    -Made the buttons slightly more disabled looking. Changed disabled cursor.
    -Removed yellow boxes on review project disposal form.
    -Fixed display of some terminal statuses.

### DIFF
--- a/frontend/src/buttons.scss
+++ b/frontend/src/buttons.scss
@@ -34,8 +34,8 @@
 }
 .btn:disabled,
 .btn:hover:disabled {
-  opacity: 0.65;
-  cursor: unset;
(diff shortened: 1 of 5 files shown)


frontend/src/buttons.scss
```diff
@@ -34,8 +34,8 @@
 }
 .btn:disabled,
 .btn:hover:disabled {
-  opacity: 0.65;
-  cursor: unset;
+  opacity: 0.45;
```
asanchezr, line 37: I rather see a different color here (like `#ccc`) but this will do for now

PR #422, opened 2020-07-15
PIMS-2621 Integrate with DataBC Geocoder - PID

asanchezr:
This PR adds frontend logic to integrate with the API geocoder PID endpoint. 

In the SubmitProperty form, when a valid address is selected (and if that address returns a non-empty `siteId`) the frontend will request the PIDs associated with that `siteId`. If no `siteId` is supplied, the PID field is cleared in the form.

**NOTE** - there is an outstanding bug that makes the PID not show until you focus on the PID input and then select another field. Some code seems to be running only `onBlur`. There's a separate issue filed to deal with that so the fix is not part of this PR. Until that bug is resolved, you will need to select the PID field and then another field to verify a value has been set from the geocoder request. 😞

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ database-clean: ## Re-creates an empty docker database - ready for seeding
 
 database-seed:
 	@echo "$(P) Seeding docker database..."
-	@cd backend/tools/import; make run
+	@cd tools/import; make run
 
 database-refresh: | server-run pause-30 database-clean database-seed ## Refreshes the docker database
 
--- a/frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
+++ b/frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
@@ -65,7 +65,7 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
   const dispatch = useDispatch();
   const history = useHistory();
   const api = useApi();
-  let formikRef = React.useRef<any>() as any;
+  const formikRef = React.useRef<FormikProps<any>>();
 
   let initialValues = getInitialValues();
   const lookupCodes = useSelector<RootState, ILookupCode[]>(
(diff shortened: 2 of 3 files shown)


frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -65,7 +65,7 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
   const dispatch = useDispatch();
   const history = useHistory();
   const api = useApi();
-  let formikRef = React.useRef<any>() as any;
+  const formikRef = React.useRef<FormikProps<any>>();
```
asanchezr, line 68: Better typing

frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -241,7 +250,7 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
       <Col>
         <Formik
           innerRef={instance => {
-            formikRef = instance;
+            formikRef.current = instance;
```
asanchezr, line 253: React ref objects are `{ current: ...}`

frontend/src/hooks/useApi.ts
```diff
@@ -6,7 +6,7 @@ import { ENVIRONMENT } from 'constants/environment';
 import * as _ from 'lodash';
 
 export interface IGeocoderResponse {
-  siteId: number;
+  siteId: string;
```
asanchezr, line 9: `siteId` is a string (a guid actually)

Makefile
```diff
@@ -85,7 +85,7 @@ database-clean: ## Re-creates an empty docker database - ready for seeding
 
 database-seed:
 	@echo "$(P) Seeding docker database..."
-	@cd backend/tools/import; make run
+	@cd tools/import; make run
```
Fosol, line 88: Thanks!

frontend/src/features/projects/spl/steps/__snapshots__/SplStep.test.tsx.snap
```diff
@@ -151,7 +151,7 @@ exports[`SPL Approval Step renders correctly 1`] = `
                       placeholder="--/--/----"
                       readOnly={false}
                       type="text"
-                      value="07/14/2020"
+                      value="07/15/2020"
```
Fosol, line 154: lol, still have this bug showing up in our tests.  Hope someone can resolve this soon.

asanchezr: the solution is to delete this snapshot test (for now) - painful to have to update this daily

PR #423, opened 2020-07-15
PIMS-2898 Return to Pre-Marketing

devinleighsmith:


backend/dal/Migrations/Initial/PostDeploy/01-Tasks.sql
```diff
@@ -143,7 +143,7 @@ INSERT INTO dbo.[Tasks] (
     , 'Bid Rigging, Collusion and Bias'
     , 0
     , 'I confirm I have reviewed the offer and to the best of my knowledge confirm there is no identifiable opportunity for bid rigging, collusion and bias'
-    , 42 --Contract in Place
```
devinleighsmith, line 146: This doesn't make sense to enforce the checkbox being ticked when going from SPL-CIP back to SPL-PM. So moving this to the Disposed status.

PR #424, opened 2020-07-15
PIMS-2898 Return to Pre-Marketing

devinleighsmith:


backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -674,7 +674,7 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
                 .Include(s => s.Tasks)
                 .FirstOrDefault(s => s.Id == project.StatusId) ?? throw new KeyNotFoundException();
 
-            if (fromStatus.StatusId != toStatus.Id)
+            if (fromStatus.StatusId != toStatus.Id && (fromStatus.Status.SortOrder <= toStatus.SortOrder || toStatus.IsTerminal))
```
devinleighsmith, line 677: I think this makes sense, but would like someone else to vet this as well.

Fosol: I'm not sure why you need this?  Isn't this section going to validate the workflow and possibly validate tasks?

devinleighsmith: There is a task against the status of contract in place. I don't want validation to occur when going from contract in place to pre-marketing. I do want it to occur when going from contract in place to disposed however.

Fosol: i wonder if we need to move the `ProjectStatus.ValidateTasks` flag to the `ProjectStatusTransition.ValidateTasks` to control when tasks should be validated.  I think this would give us better control.

devinleighsmith:
This is on hold until I have a look at Jeremy's suggestion.

backend/dal/Services/Concrete/ProjectService.cs
devinleighsmith: Applied this change above

PR #426, opened 2020-07-16
PIMS-2680 Using Classification for Dispose status

Fosol:
## Description 

Remove `ProjectStatus` as it was underused and incorrectly used throughout.
Replaced it with an updated to `ProjectClassifications`.

Project Classifications now have a `Disposed` value.  Only SRES should be able to see or set this value.

Disposed properties will no longer be returned by the Landing Page or Property List View.

- Not sure if there will be a future requirement to allow SRES to view them.

Also fixing geocoder URL so that we get content back.  Please note you'll need the developer key in your API `.env`.

```conf
Geocoder__Key={secret key}
```


backend/api/geocoder.Development.json
```diff
@@ -1,5 +1,5 @@
 {
   "Geocoder": {
-    "HostUri": "https://apps.gov.bc.ca/pub/geocoder"
+    "HostUri": "https://geocoder.api.gov.bc.ca"
```
Fosol, line 3: This URL requires a developer key.  If you want to reuse mine, just let me know and I'll pass it along.

backend/core/Comparers/ShallowPropertyCompare.cs
```diff
@@ -37,14 +38,17 @@ public int GetHashCode(object obj)
 
             var props = obj.GetType().GetCachedProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.GetProperty);
 
-            foreach (var p in props)
+            foreach (var p in props.OrderBy(p => p.Name))
```
Fosol, line 41: Order matters in hash codes.

PR #430, opened 2020-07-16
Start using Project Metadata to track non-core Project fields.

devinleighsmith:


backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -1,4 +1,5 @@
 using Mapster;
```
devinleighsmith, line 1: Overall the idea is to push the decision whether to map the metadata or not to the mapper, that way a controller can still call the project service without having to deal with the performance cost of serialization.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -57,6 +58,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.OfferAmount, src => src.OfferAmount)
                 .Map(dest => dest.Tasks, src => src.Tasks)
                 .Map(dest => dest.ProjectAgencyResponses, src => src.Responses)
+                .BeforeMapping((src, dest) => JsonConvert.PopulateObject(src.Metadata ?? "{}", src)) //use the metadata object to populate fields not stored in DB.
```
devinleighsmith, line 61: This allows the mapping to behave the same way it did before these fields were removed, all fields will be mapped directly from the entity to the model since the entity will be populated with the metadata before the mapping occurs.

Note: populateObject respects the field annotations in the Project object, so only [JsonProperty] fields will be populated.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -100,6 +102,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.OfferAmount, src => src.OfferAmount)
                 .Map(dest => dest.Tasks, src => src.Tasks)
                 .Map(dest => dest.Responses, src => src.ProjectAgencyResponses)
+                .AfterMapping((src, dest) => dest.Metadata = JsonConvert.SerializeObject(dest)) //Map all non-ignored fields into the metadata.
```
devinleighsmith, line 105: same idea, map the fields as per usual, allowing entity logic requiring fields that are stored in metadata to perform required checks. After mapping, serialize all [JsonProperty] fields to Metadata.

backend/api/Areas/Project/Mapping/Search/ProjectMap.cs
```diff
@@ -39,6 +40,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.CreatedOn, src => src.CreatedOn)
                 .Map(dest => dest.CreatedBy,
                     src => src.CreatedById != null ? src.CreatedBy.DisplayName : null)
+                .BeforeMapping((src, dest) => JsonConvert.PopulateObject(src.Metadata ?? "{}", src))
```
Fosol, line 43: The only concern I have here for the search is that we're returning an array of projects.  Ideally we won't do any extra serialization here if we don't need to.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -780,6 +781,7 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
             // Update a project
             originalProject.StatusId = toStatus.Id;
             originalProject.Status = toStatus;
+            originalProject.Metadata = project.Metadata;
```
devinleighsmith, line 784: Something to think about: From what I could tell, it shouldn't be necessary to populate originalProject with metadata values either in the update or setStatus functions as those values don't seem to be required for business logic. Correct me if I'm wrong.

backend/entities/Project.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Text.Json.Serialization;
+using Newtonsoft.Json;
 
 namespace Pims.Dal.Entities
 {
     /// <summary>
     /// Project class, provides an entity for the datamodel to manage projects.
     /// </summary>
+    [JsonObject(MemberSerialization.OptIn)]
```
devinleighsmith, line 12: Could also have marked all of the DB properties as JsonIgnore, but I think that it makes more sense to keep the property attributes grouped like this.

backend/entities/Pims.Dal.Entities.csproj
```diff
@@ -8,6 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="3.1.5" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
Fosol, line 11: We shouldn't need Newtonsoft, as we don't use it anywhere else for serialization.  Use the built in System.Text.Json.

backend/entities/Project.cs
```diff
@@ -118,6 +122,8 @@ public class Project : BaseEntity
         /// <summary>
         /// get/set - When the project was externally marketed.
         /// </summary>
+        [NotMapped]
+        [JsonProperty]
```
Fosol, line 126: I'd extract all these extra properties and create a `DisposalProject` in the `/Models` namespace within this library.

backend/tests/core/Entities/ProjectHelper.cs
```diff
@@ -62,6 +62,7 @@ public static Entity.Project CreateProject(int id, Entity.Agency agency = null,
                 PublicNote = $"publicNote-{id}",
                 PrivateNote = $"privateNote-{id}",
                 UpdatedOn = DateTime.UtcNow,
+                Metadata = "{offerAmount: 123}",
```
Fosol, line 65: Technically that isn't valid JSON, but I don't know if it'll cause any issue. `{"offerAmount": 123}`

backend/dal/Services/Concrete/ProjectService.cs
Fosol: I'd think you'd need it in `Add` and `Update`, otherwise you'll never have values there.  Since `SetStatus` also updates, you'll probably need it there too.

There are some possible business rules related to required date values before proceeding between states, but we may only enforce them currently on the frontend.  Ideally we'd enforce in the backed too.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
Fosol: Perfect!

backend/tests/core/Entities/ProjectHelper.cs
devinleighsmith: The test passes as-is.

backend/entities/Project.cs
devinleighsmith: If we go that route I think I'd need to update the Project Service to operate on the new DisposalProject instead of the Project object to keep the mappings working the way they do now. Alternatively I could map the model to the Project, and then convert the project to a DisposalProject by deserializing the metadata right after it was serialized. I'm not a fan of either of those options, but maybe I don't understand the suggestion.

backend/entities/Pims.Dal.Entities.csproj
devinleighsmith: [JsonObject(MemberSerialization.OptIn)] is only part of Newtonsoft, let me see if there is an equivalent in System.Text.Json.Serialization

backend/api/Areas/Project/Mapping/Search/ProjectMap.cs
devinleighsmith: I'll test tasking this out and see if it breaks anything.

backend/entities/Pims.Dal.Entities.csproj
Fosol: I doubt our serialization would obey a Newtonsoft specific attribute.

backend/entities/Project.cs
Fosol: You've essentially lost some of the reason to use the `metadata` property by keeping `NotMapped` properties on the entity `Project`.  It makes more sense to have a specific model for each given project type.  Thus creating a `DisposalProject` model that has these properties which can be used to work with the `Metadata` properties.

Currently yes it would require updating the `ProjectService` functions, but could theoretically be handled via generics.

devinleighsmith: maybe a partial would work here?

Fosol: A temporary option, but in truth isn't any different.

devinleighsmith: I think I'll hold off on this for now since this at least alleviates some of the pain in the database. If I have time I'll see if I can enhance this further.

frontend/src/features/projects/dispose/ProjectDisposeLayout.tsx
```diff
@@ -134,7 +134,6 @@ const ProjectDisposeLayout = ({ match, location }: { match: Match; location: Loc
           {getProjectRequest?.isFetching !== true ? (
             <Container fluid className="step-content">
               <Switch>
-                )}
```
asanchezr, line 137: Is this a bug fix for a separate issue? This frontend file doesn't seem related to the PR work...

PR #433, opened 2020-07-17
View Inventory - Group by buildings

jdtoombs:
As well as the bug so you can click the exemption check box without hitting edit, and hiding the size selector on the view inventory (going to strip it from other lists to will include in next one takes 3s).


frontend/src/components/Table/Table.tsx
```diff
@@ -405,7 +406,7 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
       {!props.hideToolbar && (
         <div className="table-toolbar">
           {props.pageSize !== -1 && <TablePagination<T> instance={instance} />}
-          {!props.lockPageSize && props.data.length > 0 && (
+          {!props.lockPageSize && props.data.length > 0 && !props.hidePageSizeSelector && (
```
devinleighsmith, line 409: is there a difference between lockPageSize and hidePageSelector?

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -103,9 +103,9 @@ const toApiProperty = (property: IProperty): IApiProperty => {
       postal: property.postal,
       provinceId: property.province,
     },
-    zoning: property.zoning,
-    zoningPotential: property.zoningPotential,
-    municipality: property.municipality,
+    zoning: property.zoning!,
```
devinleighsmith, line 106: hmm, what is the rationale for this and the above change?

frontend/src/components/Table/Table.tsx
jdtoombs: did not even notice we had that, changing now

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -64,6 +69,8 @@ const getServerQuery = (state: {
     maxLandArea: decimalOrUndefined(maxLotSize),
     page: pageIndex + 1,
     quantity: pageSize,
+    parcelId: decimalOrUndefined(parcelId!),
```
devinleighsmith, line 72: in general I'd rather we handle possibly undefined values with explicit checks/exceptions rather then the ! character.

frontend/src/features/projects/common/projectConverter.ts
jdtoombs: this was something i was going to ask about, i was getting ts errors here after my changes. would it be better to have it as `property.x ?? ''`

PR #434, opened 2020-07-17
PIMS-2781 preserve form state after keycloak token refresh

goemen:


frontend/src/utils/PrivateRoute.tsx
```diff
@@ -28,7 +29,7 @@ const PrivateRoute = (props: IPrivateRouteProps) => {
           ) {
             return (
               <Layout>
-                <Component {...props} />
+                <Component {...props} {...rest.componentProps} />
```
Fosol, line 32: I'm curious how this works?

PR #435, opened 2020-07-17
PIMS-2906 - small updates to complete project disposal.

devinleighsmith:


devinleighsmith:
Jeremy/I already completed the majority of the work for this change, so only minor updates were needed to complete the remaining requirements.

PR #437, opened 2020-07-18
PIMS-2903 Approved for Exemption Form - Required date fields

asanchezr:


frontend/src/utils/utils.ts
```diff
@@ -61,12 +61,12 @@ type FormikMemoProps = {
...
 export const formikFieldMemo = (
-  { formikProps: currentProps, field: currField }: FormikMemoProps,
-  { formikProps: prevProps, field: prevField }: FormikMemoProps,
+  { formikProps: currentProps, field: currField, disabled: currentDisabled }: FormikMemoProps,
+  { formikProps: prevProps, field: prevField, disabled: prevDisabled }: FormikMemoProps,
 ) => {
   return !(
     currField !== prevField ||
-    currentProps.disabled !== prevProps.disabled ||
+    currentDisabled !== prevDisabled ||
```
asanchezr, line 69: Fixing a bug in the memo function - formikProps doesn't have a `disabled` property

PR #438, opened 2020-07-20
PIMS-1746 Add lot size validation to map filter form

asanchezr:
Screen Shot 2020-07-19 at 8 44 29 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ bash.exe.stackdump
 .sonarqube/
 auth/keycloak/config/hsperfdata_jboss/
 auth/keycloak/config/*.log
+.DS_Store
+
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -8,3 +8,5 @@ bash.exe.stackdump
 .sonarqube/
 auth/keycloak/config/hsperfdata_jboss/
 auth/keycloak/config/*.log
+.DS_Store
```
asanchezr, line 11: Ignore `macOS` temp files

frontend/src/utils/YupSchema.ts
```diff
@@ -170,3 +170,16 @@ export const ParcelSchema = Yup.object()
     [['pin', 'pid']],
   )
   .concat(LandSchema);
+
+export const FilterBarSchema = Yup.object().shape({
+  minLotSize: Yup.number()
+    .typeError('Invalid')
+    .positive('Invalid')
+    .max(200000, 'Invalid'),
```
asanchezr, line 178: Arbitrary max value of `200,000` ha - should be high enough?

PR #444, opened 2020-07-21
PIMS-2900 ExemptionERP Tab label,

goemen:
PIMS-2864 after login redirect link


frontend/src/hooks/use-query.ts
```diff
@@ -0,0 +1,7 @@
+import { useLocation } from 'react-router';
```
devinleighsmith, line 1: Can we just make this useQuery? We don't have - in any of our other file names as far as I'm aware.

PR #446, opened 2020-07-22
PIMS-2822 OpenShift templates for DB migration job

asanchezr:
This PR sets up the ground work for running EF migrations from within the CI pipeline. 

The main idea here is to create a custom build with `dotnet-ef` bundled. Then a migration job "pod" will be created in CI builds to migrate the database.

### Running an ad-hoc migration job

```bash
oc process -f migration-job.yaml -p ENV_NAME=prod -p IMAGE_TAG=master -o yaml | oc -n jcxjin-prod create -f - 
```

image

### DIFF
--- a/backend/Dockerfile.migrations
+++ b/backend/Dockerfile.migrations
@@ -1,0 +1,18 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+
+ENV DOTNET_CLI_HOME=/tmp
+ENV PATH="$PATH:/tmp/.dotnet/tools"
+
+# Switch to root for package installs
+USER 0
+RUN dotnet tool install --global dotnet-ef
+
+WORKDIR /src
+COPY . .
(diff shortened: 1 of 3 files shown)


asanchezr:
@Fosol 
> How do you control whether the pod shutdown after it runs?

https://docs.openshift.com/container-platform/3.11/dev_guide/jobs.html

> A job, in contrast to a replication controller, runs a pod with any number of replicas to completion. 
> A job tracks the overall progress of a task and updates its status with information about active, succeeded, and failed pods.

As per the above, a job or a "bare pod" will run a task to completion. This is different than a replication controller (aka `DeploymentConfig`) which ensures pods are running at all times for things like web servers etc.

Note that `dotnet ef database update` runs and exits so the pod automatically shuts down at the end of the migration. Also, by way of setting the pod configuration `restartPolicy: Never` a failed migration will not be retried. We want that so CI builds fail on failed migrations.

[OpenShift Docs - Pod Restart Policy](https://docs.openshift.com/container-platform/3.11/architecture/core_concepts/pods_and_services.html#admin-manage-pod-restart)

PR #447, opened 2020-07-22
Form disabled bug fix / Map zoom loses pin fix

jdtoombs:
Not sure how this was working before (I am probably missing something) but I know children do not re-render if the parents props change and only if its state changes. So I tried tying it to state and using hooks but that didn't seem to work. This solution does work but not sure if it is the best one - adding a key to the evaluation and building form based off the `props.disabled` re-renders the component each time this prop changes.


frontend/src/features/properties/map/MapView.tsx
```diff
@@ -134,13 +136,17 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
...
       }
       onMarkerPopupClose={props.onMarkerPopupClosed ?? (() => dispatch(storeParcelDetail(null)))}
-      onViewportChanged={(mapFilterModel: MapViewportChangeEvent) => {
-        const apiParams = getApiParams(mapFilterModel);
-        const action = fetchParcels(apiParams);
-        _.throttle(() => {
-          dispatch(action);
-        }, 250)();
-      }}
+      onViewportChanged={
```
jdtoombs, line 139: still working on this ignore for now

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -135,11 +138,22 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
...
-          dispatch(action);
-        }, 250)();
+        if (!props.submittingProperty) {
+          const apiParams = getApiParams(mapFilterModel);
+          const action = fetchParcels(apiParams);
+          _.throttle(() => {
+            dispatch(action);
+          }, 250)();
+        } else {
+          if (!loadedProperties) {
```
jdtoombs, line 148: load the properties initially but do not filter out the pin after it is placed

PR #448, opened 2020-07-22
PIMS-2585 Close out form, tests.

devinleighsmith:


frontend/src/configureStore.spec.ts
```diff
@@ -1,22 +0,0 @@
-import { createStore, applyMiddleware } from 'redux';
```
devinleighsmith, line 1: Was getting errors from this that weren't making sense, will circle back later in the sprint.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -16,6 +17,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Name, src => src.Name)
                 .Map(dest => dest.ReportedFiscalYear, src => src.ReportedFiscalYear)
                 .Map(dest => dest.ActualFiscalYear, src => src.ActualFiscalYear)
+                .Map(dest => dest.NetBookFiscalYear, src => src.NetBookFiscalYear)
```
Fosol, line 20: I don't think we need this.  The NetBook value for the Project will be the ActualFiscalYear.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -50,14 +53,42 @@ public void Register(TypeAdapterConfig config)
...
                 .Map(dest => dest.DisposedOn, src => src.DisposedOn)
+                .Map(dest => dest.AssessedOn, src => src.AssessedOn)
+                .Map(dest => dest.AdjustedOn, src => src.AdjustedOn)
+                .Map(dest => dest.PreliminaryFormSignedOn, src => src.PreliminaryFormSignedOn)
+                .Map(dest => dest.FinalFormSignedOn, src => src.FinalFormSignedOn)
+                .Map(dest => dest.PriorYearAdjustmentOn, src => src.PriorYearAdjustmentOn)
                 .Map(dest => dest.ExemptionRequested, src => src.ExemptionRequested)
                 .Map(dest => dest.ExemptionRationale, src => src.ExemptionRationale)
                 .Map(dest => dest.NetBook, src => src.NetBook)
+                .Map(dest => dest.NetBookOverride, src => src.NetBookOverride)
```
Fosol, line 64: What is the NetBookOverride?

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -50,14 +53,42 @@ public void Register(TypeAdapterConfig config)
...
+                .Map(dest => dest.SalesProceeds, src => src.SalesProceeds)
+                .Map(dest => dest.ProgramCost, src => src.ProgramCost)
+                .Map(dest => dest.GainLoss, src => src.GainLoss)
+                .Map(dest => dest.SppCapitalization, src => src.SppCapitalization)
+                .Map(dest => dest.GainBeforeSpp, src => src.GainBeforeSpp)
+                .Map(dest => dest.GainAfterSpp, src => src.GainAfterSpp)
+                .Map(dest => dest.OcgFinancialStatement, src => src.OcgFinancialStatement)
                 .Map(dest => dest.OfferAmount, src => src.OfferAmount)
+                .Map(dest => dest.Assessed, src => src.Assessed)
+                .Map(dest => dest.AssessedOverride, src => src.AssessedOverride)
```
Fosol, line 77: What is the AssessedOverride?

backend/api/Models/Parcel/ParcelEvaluationModel.cs
```diff
@@ -14,6 +14,8 @@ public class ParcelEvaluationModel : BaseModel
         public float Value { get; set; }
 
         public string Note { get; set; }
+
+        public string Firm { get; set; }
```
Fosol, line 18: What is `Firm`?

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -1,4 +1,5 @@
 using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
```
Fosol, line 2: Don't use Newtonsoft use the built in JsonSerializer that you can get the configuration from DI.

backend/entities/NoteTypes.cs
```diff
@@ -14,6 +14,13 @@ public enum NoteTypes
         PreMarketing = 6,
         Marketing = 7,
         ContractInPlace = 8,
-        Reporting = 9
+        Reporting = 9,
+        LoanTerms = 10,
+        Adjustment = 11,
+        SppCost = 12,
+        SppGain = 13,
```
Fosol, line 21: I wonder if some of these make more sense as metadata.  Not sure though.

frontend/src/features/projects/common/components/SresManual.tsx
```diff
@@ -6,7 +6,7 @@ const SresManual: React.FunctionComponent = () => {
   const link =
     'https://intranet.gov.bc.ca/assets/intranet/mtics/real-property/sres/process_manual_for_the_surplus_properties_program_-_feb_2020_-_version_2.pdf?';
   return (
-    <div className="SresManual" onClick={() => (window.location.href = link)}>
+    <div className="SresManual" onClick={() => window.open(link, '_blank')}>
```
Fosol, line 9: Thanks

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
devinleighsmith: This is documented in the models, but if netbookoverride the api will use a static value for the netbook instead of a calculated value. Some sort of boolean flag is required here because the frontend needs a persistent way to communicate to the backend that sums should no longer be used.

backend/api/Models/Parcel/ParcelEvaluationModel.cs
devinleighsmith: This is for the appraisal firm required in the close out form

Fosol: Ah, the **Appraisal Firm**.  I'd rename to **AppraisalBy**

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
Fosol: I don't understand the need still. If I remember correctly, SRES is just going to put a number in here and that's the official `NetBook` that will be used for all reporting.  I think for now the backend will just have to not sum the `NetBook` value due to this requirement.  I think it can still sum the other values.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
devinleighsmith: Agreed, I think this was from the v1 of the design.

backend/api/Models/Parcel/ParcelEvaluationModel.cs
devinleighsmith: Does that makes sense given that this field is on the evaluation table, and technically could be used by assessments as well?

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
devinleighsmith: If this is what is desired, then I will also have to remove/rename the "Sum Netbook" column from the approval requests and project list tables as well. Not sure if it makes sense to continue to include that column at all if SRES is the only one entering a value.

devinleighsmith: Would this apply for assessedOverride as well? that boolean is being used for essentially the same purpose, just for assessed instead of netbook. I'm actually not sure if the summed value for that is being used in the frontend.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
devinleighsmith: That serializer doesn't support populateObject which is critical to my use case. I can probably find some reflection code on stack overflow that does the same thing, but that seems worse then using a published library.

https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#populate-existing-objects

Fosol: Is this to support `metadata`?

backend/api/Models/Parcel/ParcelEvaluationModel.cs
Fosol: Good point, but `Firm` really doesn't explain itself.  We essentially want it to be self descriptive `ValueFrom`  `SourceOfValue` `Note` would even be better.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
Fosol: We don't have a place to enter the value at that point.  The value is coming from the owning agency updating their properties.  For some reason when they sell they are only receiving a final number.  I think it's an issue with their process, but for now the only single `NetBook` value we care about at the point of SPL disposal is the one they manually enter once.

devinleighsmith: Before I make this change, please confirm the following:
1) I should remove all evaluation/fiscal summation logic from the backend. (netbook, assessed, estimated)
2) I should remove all frontend logic that depends on # 1, such as the project list view.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
devinleighsmith: yes.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
Fosol: Based on our last conversation with Mikaela, don't remove any summation logic.  Add a new separate `Project.CloseOutNetBook` property to capture the value they enter on the Close Out Form.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
Fosol: Ugh, this isn't ideal on any level other than to quickly implement today.  Using the DAL layer (entities) for serialization is a big mistake that bites later on.  Sure it helps with RAD, but boy does it cause problems.  It's why we have models in the API for all serialization.  Due to the timing, we may as well let this slide, but we'll have to add it to tech-debt.

PR #449, opened 2020-07-22
PIMS-2860 Surplus Property Program - Tier 1

goemen:


backend/dal/Migrations/Initial/PostDeploy/01-Tasks.sql
```diff
@@ -22,7 +22,7 @@ INSERT INTO dbo.[Tasks] (
     2
     , 'Triple Bottom Line completed and sent'
     , 0
-    , 'Triple Bottom Line document emailed to SRES.'
+    , 'Triple Bottom Line document emailed to SRES OR Project is in Tier 1.'
     , 4
     , 0
```
Fosol, line 27: I think you need to change this to `IsOptional=1` so that it becomes optional.  Then in the app you'll need to add validation logic based on the `Tier`.  Additionally in the api you should also include the logic, but I don't think it's critical at this point there.

goemen: I talked to @devinleighsmith and he says that changing the description of the task will force the user to check the box

Fosol: Sounds good

PR #453, opened 2020-07-23
PIMS-2489 Allowing building ownership

Fosol:
## Description

Provide a way to allow agencies to view and edit buildings they own when they don't own the parcel.

Adding `Geocoder` Secret to Deployment Configuration.

Fixing possible bug in production when adding projects.

Updating Keycloak configuration for production.


frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -71,6 +74,10 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
     state => (state.lookupCode as ILookupCodeState).lookupCodes,
   );
 
+  let allowEdit =
```
goemen, line 77: change to `const`

PR #454, opened 2020-07-24
PIMS-1836 Financial - if amount has not been entered by user do not s…

goemen:
…how any value


frontend/src/features/properties/components/forms/subforms/EvaluationForm.tsx
```diff
@@ -254,6 +254,8 @@ const EvaluationForm = <T extends any>(props: EvaluationProps & FormikProps<T>)
                                 formikProps={props}
                                 disabled={props.disabled}
                                 field={withNameSpace('value', type, year)}
+                                placeholder={props.disabled ? 'n/a' : ' '}
```
Fosol, line 257: Will this result in the rows with the value `' '` to not be saved to the database?

goemen: It does not, its just an input placeholder not the value

PR #457, opened 2020-07-24
PIMS-2822 Run migrations in CI build

asanchezr:
New bash script created `oc-migrate.sh` that will run a pod that has `dotnet ef` installed - in a similar fashion as you would do `docker run ...`

The `oc` command returns the exit code from the migration process:
- If successful the CI pipeline continues building. 
- If the migration cannot be applied, an error code is returned and the CI pipeline aborts

Once this is merged we will need to test success and failure scenarios to get ready to launch in production.

### DIFF
--- a/docs/DEVOPS.md
+++ b/docs/DEVOPS.md
@@ -1,4 +1,4 @@
-# DevOps Continuous Integration / Continous Deployment
+# DevOps Continuous Integration / Continuous Deployment
 
 First it was [Waterfall](https://en.wikipedia.org/wiki/Waterfall_model), next it was [Agile](https://en.wikipedia.org/wiki/Agile_software_development), and now it's [DevOps](https://aws.amazon.com/devops/what-is-devops/). This is how modern developers approach building great products. With the rise of DevOps has come the new methods of Continuous Integration, Continuous Delivery, (CI/CD) and Continuous Deployment. Conventional software development and delivery methods are rapidly becoming obsolete. Historically, in the agile age, most companies would deploy/ship software in monthly, quarterly, bi-annual, or even annual releases. Now we build and deploy multiple times a day.
 
@@ -8,7 +8,7 @@ First it was [Waterfall](https://en.wikipedia.org/wiki/Waterfall_model), next it
 
 The PIMS project currently uses [GitHub Actions](https://github.com/features/actions), [OpenShift](https://www.openshift.com/) and [Jenkins](https://www.jenkins.io/) to support CI/CD.
 
-The general highlevel workflow is as follows;
+The general high-level workflow is as follows;
 
 1. Fork Repo
 1. Clone Repo
@@ -65,7 +65,7 @@ As changes are merged the CI/CD pipeline build and deploy to the [DEV](https://p
 
 The CI/CD process determines if changes were made. If they were not, it will ask if you want to force a rebuild. If the build, test and scan process completed successfully it will deploy the new build and spin up new containers.
 
-This makes the **DEV** branch unstable, as it can change mulitple times per day.
+This makes the **DEV** branch unstable, as it can change multiple times per day.
 
 Initial QA testing is performed in the **DEV** environment.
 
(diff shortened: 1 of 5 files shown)


docs/DEVOPS.md
```diff
@@ -1,4 +1,4 @@
-# DevOps Continuous Integration / Continous Deployment
+# DevOps Continuous Integration / Continuous Deployment
```
asanchezr, line 1: Fixing typos

openshift/pipelines/Jenkinsfile.test-to-prod
```diff
@@ -49,7 +49,7 @@ pipeline {
     ENABLE_VERSION_PROMPT = true // ignored - version prompt always enabled in the PROD pipeline
     OC_JOB_NAME = sh(script: 'echo "${OC_JOB_NAME:-master}"', returnStdout: true).trim()
     DESTINATION = sh(script: 'echo "${DESTINATION:-prod}"', returnStdout: true).trim()
-    VANITY_URL = sh(script: 'echo "${VANITY_URL:-https://pims-prod.pathfinder.gov.bc.ca/}"', returnStdout: true).trim()
+    VANITY_URL = sh(script: 'echo "${VANITY_URL:-https://pims.gov.bc.ca/}"', returnStdout: true).trim()
```
asanchezr, line 52: We now have a prod URL

openshift/pipelines/Jenkinsfile.test-to-prod
```diff
@@ -127,10 +127,15 @@ pipeline {
     }
 
     stage("Database migration") {
-      options { timeout(time: 5, unit: "MINUTES") }
+      options { timeout(time: 15, unit: "MINUTES") }
+      environment {
+        DB_PASSWORD = sh(script: "oc -n jcxjin-prod get secret pims-db-prod-secret -o jsonpath='{.data.DB_PASSWORD}' | base64 -d", returnStdout: true).trim()
+        DB_CONNECTION_STRING = sh(script: "oc -n jcxjin-prod get cm pims-api-prod-db-client -o jsonpath='{.data.CONNECTION_STRINGS_PIMS}'", returnStdout: true).trim()
+      }
```
asanchezr, line 134: This "extracts" OpenShift secrets into environment variables required by the migrations script below

openshift/scripts/oc-migrate.sh
```diff
@@ -0,0 +1,75 @@
...
+MIGRATION_IMAGE=${MIGRATION_IMAGE:-pims-migrations}
+MIGRATION_TAG=${MIGRATION_TAG:-dev}
+COMMAND=${COMMAND:-}
+
+FULL_IMAGE="docker-registry.default.svc:5000/${PROJ_TOOLS}/${MIGRATION_IMAGE}:${MIGRATION_TAG}"
+
+# Rebuild migrations image - e.g. 'pims-migrations-dev' OR 'pims-migrations-master'
+#
+BUILD_NAME="${BUILD_NAME:-${MIGRATION_IMAGE}-${MIGRATION_TAG}}"
+start_build "${BUILD_NAME}"
```
asanchezr, line 48: We make sure to rebuild the `pims-migrations` image so we have the latest code and migrations in the docker image

openshift/scripts/oc-migrate.sh
```diff
@@ -0,0 +1,75 @@
...
+  --image-pull-policy=Always \
+  --attach \
+  --rm \
+  --restart=Never \
+  --env='ASPNETCORE_ENVIRONMENT=${ASPNETCORE_ENVIRONMENT}' \
+  --env='ConnectionStrings__PIMS=${DB_CONNECTION_STRING}' \
+  --env='DB_PASSWORD=${DB_PASSWORD}' \
+  --limits='cpu=1000m,memory=1Gi' \
+  --requests='cpu=500m,memory=512Mi'"
+[ "${COMMAND}" ] && OC_MIGRATE="${OC_MIGRATE} --command -- ${COMMAND}"
```
asanchezr, line 64: This command creates and runs a pod to completion (or error), passing DB password and connection string as env vars.
With `--rm`, the pod will be deleted the pod once finished.
With `--restart=Never`, if the pod crashes or errors out it will not be retried and an error code will be returned to the calling script.

openshift/scripts/oc-migrate.sh
Fosol: This may cause a duplicate build, as `dotnet ef database update` will cause a build too.

PR #459, opened 2020-07-27
PIMS-2974 empty cityId field after clearing input text value

goemen:


frontend/src/components/common/form/AutoCompleteText.tsx
```diff
@@ -75,6 +75,8 @@ export const AutoCompleteText: React.FC<IAutoCompleteProps> = ({
       } else {
         dynamicSuggestions = options.filter(v => regex.test(v.label));
       }
+    } else {
+      setFieldValue(field, null);
```
devinleighsmith, line 79: Does formik throw an error in the console for this? If so you may need to set this to '' instead.

goemen: It did nt throw an error, but I think your suggestion is safer

PR #460, opened 2020-07-27
UI for grouping buildings

jdtoombs:
Still ironing out some bugs in my last few tasks - putting this up in the meantime to split things up


goemen:
Hey, is this the ticket to hide the properties table if there are properties?

PR #462, opened 2020-07-27
Concurrency Bug

jdtoombs:
Little bit of unfamiliar territory but this seems to fix the bug - let me know if there is a better alternative!


backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -493,7 +493,28 @@ public async System.Threading.Tasks.Task<Project> UpdateAsync(Project project)
             {
                 originalProject.Note = project.Note;
                 originalProject.PublicNote = project.PublicNote;
-                this.Context.SaveChanges();
+                try
```
Fosol, line 496: regrettably this is incorrect.  A concurrency error was correctly being thrown by the API.  The frontend however wasn't correctly handling an update and then subsequent updates were using an old `RowVersion`.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -540,7 +561,29 @@ public async System.Threading.Tasks.Task<Project> UpdateAsync(Project project)
             originalProject.Merge(project, this.Context);
             originalProject.Metadata = project.Metadata;
 
-            this.Context.SaveChanges();
+            try
```
Fosol, line 564: Regrettably incorrect.  Same as above comment.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -799,7 +841,28 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
             originalProject.Status = toStatus;
             originalProject.Metadata = project.Metadata;
             project.CopyRowVersionTo(originalProject);
-            this.Context.CommitTransaction();
+            try
```
Fosol, line 844: Regrettably incorrect. Same as above comment.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -799,7 +841,28 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
...
+                    {
+                        var proposedValue = proposedValues[property];
+                        var databaseValue = databaseValues[property];
+
+                        proposedValues[property] = proposedValue;
+                    }
+
+                    entry.OriginalValues.SetValues(databaseValues);
+                }
+            }
```
goemen, line 865: This code block is a duplicate of the other 2 above, it be nice the create a method and call it

backend/dal/Services/Concrete/ProjectService.cs
jdtoombs: Gotchya, that makes sense. I originally thought it was a frontend error but read the docs the error message pointed to and then changed my mind and went with that approach. Adjusting now!

frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
```diff
@@ -98,6 +98,7 @@ const ReviewApproveStep = ({ formikRef }: IStepProps) => {
         innerRef={formikRef}
         onSubmit={(values: IProject) => {
           const workflowCode = getNextWorkflowCode(submitStatusCode, values);
+          if (values.rowVersion !== project.rowVersion) values.rowVersion = project.rowVersion;
```
devinleighsmith, line 101: Should be able to fix this by setting enablereinitialize to true on the <Formik> component within the ReviewApproveStep.tsx

jdtoombs: Ooh, thank you - works!

PR #467, opened 2020-07-28
Correct inability to save review form multiple times.

devinleighsmith:


frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
```diff
@@ -94,6 +94,7 @@ const ReviewApproveStep = ({ formikRef }: IStepProps) => {
   return (
     <Container fluid className="ReviewApproveStep">
       <Formik
+        enableReinitialize={true}
```
devinleighsmith, line 97: This is how we currently get the updated rowverison from the API.

PR #469, opened 2020-07-28
PIMS-2489 Fix editing properties

Fosol:
## Description

Fix - Property Detail View was disabled when adding new properties.
Fix - Submitting properties with 0 Rentable Area
Fix - Transfer within GRE


frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -76,6 +76,7 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
 
   let allowEdit =
     keycloak.hasClaim(Claims.ADMIN_PROPERTIES) ||
+    !props.parcelDetail ||
```
Fosol, line 79: new properties should allow editing.

PR #470, opened 2020-07-28
PIMS-2981 Fix building financials.

devinleighsmith:


backend/dal/Migrations/Initial/PostDeploy/01-ViewProperties.sql
```diff
@@ -188,35 +188,35 @@ OUTER APPLY (
     SELECT TOP 1
         [Value]
         , [Date]
-    FROM dbo.[ParcelEvaluations]
-    WHERE [ParcelId] = p.[Id]
+    FROM dbo.[BuildingEvaluations]
```
devinleighsmith, line 191: This was probably just copy/pasta

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -73,7 +73,7 @@ export const toFlatProject = (project?: IApiProject) => {
     return undefined;
   }
   const flatProperties = project.properties.map(pp => {
-    const apiProperty: IApiProperty = (pp.parcel ?? pp.building) as IApiProperty;
+    const apiProperty: IApiProperty = (pp.building ?? pp.parcel) as IApiProperty;
```
devinleighsmith, line 76: Sorry, the order here doesn't actually matter.

backend/dal/Migrations/Initial/PostDeploy/01-ViewProperties.sql
Fosol: Thanks for resolving this!

PR #474, opened 2020-07-28
Ensure that API applied dates are saved in metadata correctly.

devinleighsmith:


backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -725,31 +727,31 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
                     // This must be done first because it requires its own transaction.
                     var projectNumber = this.Context.GenerateProjectNumber(_options.Project.NumberFormat);
                     this.Context.UpdateProjectNumber(originalProject, projectNumber);
-                    originalProject.SubmittedOn = DateTime.UtcNow;
+                    project.SubmittedOn = DateTime.UtcNow;
```
Fosol: The `originalProject.Merge(project, this.Context);` occurs before this.  Which means that these may not be correctly merged.  Or are all of these `Metadata` and will be handled by the last serialize you've added?

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -734,10 +736,10 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
                     var now = DateTime.UtcNow;
                     originalProject.ApprovedOn = now;
                     // Default notification dates.
-                    originalProject.InitialNotificationSentOn = now;
-                    originalProject.ThirtyDayNotificationSentOn = now.AddDays(30);
-                    originalProject.SixtyDayNotificationSentOn = now.AddDays(60);
-                    originalProject.NinetyDayNotificationSentOn = now.AddDays(90);
+                    project.InitialNotificationSentOn = now;
```
Fosol, line 739: Ah yes only set the `Metadata` properties this way.

PR #475, opened 2020-07-28
PIMS-2983

devinleighsmith:


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -408,6 +408,10 @@ public static bool IsProjectInDraft(this Entity.Project project, ProjectOptions
         /// <returns></returns>
         public static bool IsProjectClosed(this Entity.Project project)
         {
+            if(project?.Status == null)
```
devinleighsmith, line 411: This was causing failures for me when submitting parcels - it shouldn't.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -239,7 +239,7 @@ public Parcel Update(Parcel parcel)
 
             // Users who don't own the parcel, but only own a building cannot update the parcel.
 
-            if (!allowEdit)
+            if (allowEdit)
```
devinleighsmith, line 242: if allowEdit is true, update the database with the value from the API.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
Fosol: This would be the result of it not being loaded correctly by the caller.  It doesn't really make sense to return false if the `project == null`.  It should throw an `ArgumentException` in that case.

You can also simplify to `return project.Status?.IsTerminal ?? false`.  Although I think it probably makes more sense to throw an exception as the `Status` hasn't been loaded correctly for this function.

devinleighsmith: I think I'm missing something here, why is this even being called during parcel submission? this is on a vanilla DB with no projects. If I throw an exception here the submission will fail. However, I get that returning false here may not make sense. I'll need to dig deeper on this I guess.

devinleighsmith: nevermind, just repro'd this again. This occurs when trying to update a parcel that is in a project. Digging deeper.

devinleighsmith: @Fosol this was introduced by this change:

```
// update only an active project with any financial value changes.
if (!String.IsNullOrWhiteSpace(parcel.ProjectNumber) && (!this.Context.Projects.FirstOrDefault(p => p.ProjectNumber == parcel.ProjectNumber)?.IsProjectClosed() ?? false))
{
    this.Context.UpdateProjectFinancials(parcel.ProjectNumber);
}
```

correct me if I'm wrong, but won't this always fail since this.Context.Projects is not eager loading the Project.Status?

devinleighsmith: Well, going ahead with adding the include here for now since it allows me to remove the above if statement.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -387,7 +387,8 @@ public Parcel Update(Parcel parcel)
                 }
 
                 // update only an active project with any financial value changes.
-                if (!String.IsNullOrWhiteSpace(parcel.ProjectNumber) && (!this.Context.Projects.FirstOrDefault(p => p.ProjectNumber == parcel.ProjectNumber)?.IsProjectClosed() ?? false))
+                if (!String.IsNullOrWhiteSpace(parcel.ProjectNumber)
```
devinleighsmith, line 390: isProjectClosed requires that the status be loaded.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
Fosol: We could load the Project if it is required at this point

PR #481, opened 2020-07-29
Include buildings in property list

jdtoombs:
Include buildings in property list in addition to them being grouped with the parcel - but do not show a folder for buildings

### DIFF
--- a/frontend/src/components/Table/Table.tsx
+++ b/frontend/src/components/Table/Table.tsx
@@ -96,6 +96,8 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   detailsPanel?: DetailsOptions<T>;
   footer?: boolean;
   hideToolbar?: boolean;
+  // Limit where you would like an expansion button to appear based off this props criteria
+  canRowExpand?: (val: any) => boolean;
 }
 
 const IndeterminateCheckbox = React.forwardRef(({ indeterminate, ...rest }: any, ref) => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/Table/Table.tsx
```diff
@@ -96,6 +96,7 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   detailsPanel?: DetailsOptions<T>;
   footer?: boolean;
   hideToolbar?: boolean;
+  hideBuildingFolder?: boolean;
```
goemen, line 99: This table can be used for other list, not just for properties. Can you use a more generic name for this property

Fosol: Good suggestion.  Expanding rows should be a reusable generic implementation.

jdtoombs: Making a generic prop that lets you pass a function to determine specific case where to allow expansion

asanchezr: yeah, I agree this should be generic if it applies to all tables or create a new component just for this use case

frontend/src/components/Table/Table.tsx
```diff
@@ -96,6 +96,8 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   detailsPanel?: DetailsOptions<T>;
   footer?: boolean;
   hideToolbar?: boolean;
+  // Restrict where you would like an expansion button to appear
+  restrictExpansion?: (val: any) => boolean;
```
asanchezr, line 100: I think I'd prefer if this was named in the positive. Something like `canRowExpand?: (row: any) => boolean`
What do you think? 
It's usually harder the other way: "is it not restrictExpansion that allows expansion?" _ugh_

asanchezr: Also I think your logic below might not reflect the "restrict" naming...

jdtoombs: I agree, was having trouble naming. Started with showFolderWhere -> allowExpansionWhere -> restrictExpansion lol. Changing now

PR #484, opened 2020-07-29
PIMS-2160 add name and description to parcel/building detail and popups.

devinleighsmith:
Note that this does not include: PIMS-2371 and PIMS-2373 as those stories probably require some requirements or UI changes in order to fit the new data into the existing tables.


Fosol:
I've pulled this PR branch and re-published from my account to rebase and merge.

PR #486, opened 2020-07-30
PIMS-2610 Search by PID or PIN

Fosol:
## Description

You can now search by PID or PIN on the Landing Page, Property List View and the Project Select Properties View.

> Thanks Jonny!

image

### DIFF
--- a/backend/dal/Helpers/Extensions/PropertyExtensions.cs
+++ b/backend/dal/Helpers/Extensions/PropertyExtensions.cs
@@ -77,6 +77,12 @@ namespace Pims.Dal.Helpers.Extensions
             if (!String.IsNullOrWhiteSpace(filter.Description))
                 query = query.Where(p => EF.Functions.Like(p.Description, $"%{filter.Description}%"));
 
+            if (!String.IsNullOrWhiteSpace(filter.PID))
+            {
+                var pidValue = filter.PID.Replace("-", "").Trim();
+                if (Int32.TryParse(pidValue, out int pid))
+                    query = query.Where(p => p.PID == pid || p.PIN == pid);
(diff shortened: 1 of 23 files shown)


backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -77,6 +77,12 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
             if (!String.IsNullOrWhiteSpace(filter.Description))
                 query = query.Where(p => EF.Functions.Like(p.Description, $"%{filter.Description}%"));
 
+            if (!String.IsNullOrWhiteSpace(filter.PID))
+            {
+                var pidValue = filter.PID.Replace("-", "").Trim();
+                if (Int32.TryParse(pidValue, out int pid))
+                    query = query.Where(p => p.PID == pid || p.PIN == pid);
```
Fosol, line 84: Search for properties with either PID or PIN value that matches.

frontend/src/actionCreators/parcelsActionCreator.spec.ts
```diff
@@ -10,7 +10,7 @@ import axios from 'axios';
 import { fetchParcelDetail, fetchParcels } from 'actionCreators/parcelsActionCreator';
 import * as genericActions from 'actions/genericActions';
 import * as API from 'constants/API';
-import { IParcelListParams, IParcelDetailParams } from 'constants/API';
+import { IPropertySearchParams, IParcelDetailParams } from 'constants/API';
```
Fosol, line 13: Renamed to be more accurate.

PR #490, opened 2020-07-30
PIMS-1839 Landing Page - create a legend for the map

goemen:


frontend/src/components/maps/leaflet/Legend/LegendControl.scss
```diff
@@ -0,0 +1,34 @@
...
+
+.legend-button {
+  background-color: #ffffff !important;
+  color: #121212 !important;
+  width: 40px !important;
+  height: 40px !important;
+  font-size: 25px !important;
+  display: flex !important;
+  align-items: center !important;
+}
```
asanchezr, line 11: Why the need for `!important` here? 
Isn't it a better way to override the styles? i.e. 
```css
.LegendControl .legend-button {
  ...
}
```

frontend/package.json
```diff
@@ -28,6 +28,7 @@
     "react-error-boundary": "1.2.5",
     "react-icons": "3.9.0",
     "react-leaflet": "2.7.0",
+    "react-leaflet-control": "^2.1.2",
```
asanchezr, line 31: By convention we use exact versions in `package.json` to avoid dependencies being auto-updated and breaking things. 
Please remove the `^` - thanks!

frontend/src/components/maps/leaflet/Legend/LegendControl.scss
Fosol: Ideally we wouldn't use !important.

PR #505, opened 2020-09-23
PA-739 Avoid redirect to MapView after saving in SubmitProperty

goemen:


jdtoombs (approved):

frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -270,26 +270,22 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
...
+            try {
+              if (!values.id) {
+                const data = await createParcel(apiValues)(dispatch);
                 dispatch(clear(actionTypes.ADD_PARCEL));
-                dispatch(clear(actionTypes.UPDATE_PARCEL));
-              });
+                history.replace(`/submitProperty/${data.id}`);
+              } else {
+                await updateParcel(apiValues)(dispatch);
+                window.location.reload();
```
devinleighsmith, line 281: Not a big fan of this. This will force everything in the system to reload, redux, keycloak, all of the code values, etc. If we are going to do some kind of networking action, we should always use react router since this is a SPA.

I think your intention here is to forcibly get the page to show whatever values were saved to the DB? For that you could use history.go(0) to re-render the current page.

frontend/src/actionCreators/parcelsActionCreator.ts
```diff
@@ -81,38 +81,39 @@ export const fetchPropertyDetail = (
...
+  } catch (axiosError) {
+    dispatch(error(actionTypes.ADD_PARCEL, axiosError?.response?.status, axiosError));
+    dispatch(hideLoading());
+    throw Error(axiosError.response?.data.details);
+  }
 };
 
-export const updateParcel = (parcel: IParcel) => (dispatch: Function) => {
+export const updateParcel = (parcel: IParcel) => async (dispatch: Function) => {
   dispatch(request(actionTypes.UPDATE_PARCEL));
```
Fosol, line 101: It would be great to abstract a lot of this out to common functions, we're seeing repetition far too much.  Makes it hard to maintain and slows development.  Don't make changes in this PR though

frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
Fosol: I'd agree, we already have too many page refreshes for a SPA

Fosol (changes requested):

frontend/src/actionCreators/parcelsActionCreator.ts
devinleighsmith: one solution is handleAxiosResponse in utils.ts

frontend/src/actionCreators/parcelsActionCreator.ts
```diff
@@ -81,38 +81,39 @@ export const fetchPropertyDetail = (
...
-      dispatch(fetchParcelDetail(response.data));
-      dispatch(hideLoading());
-    })
-    .catch((axiosError: AxiosError) => {
-      dispatch(error(actionTypes.ADD_PARCEL, axiosError?.response?.status, axiosError));
-      throw Error(axiosError.response?.data.details);
-    })
-    .finally(() => dispatch(hideLoading()));
+  try {
+    const { data, status } = await CustomAxios().post(ENVIRONMENT.apiUrl + API.PARCEL_ROOT, parcel);
```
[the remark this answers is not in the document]
devinleighsmith: yeah, it would be nice to use the queryparams object that axios supports. Something like: await axios.get('https://httpbin.org/get', { params: { answer: 42 } });

Fosol (commented):
Have you been able to update to reduce reloads?

Fosol (approved):

PR #506, opened 2020-09-23
Codecov temp fix

jdtoombs:
Just commented out the old method in case we can use it again (actively looking for better solution). After some research this was the only solution that appeared.


jdtoombs:
Going to see what's going on with the frontend failure

PR #517, opened 2020-09-30
PA-750 Display Toast Feedback when saving user, unit tests

devinleighsmith:


frontend/src/App.tsx
```diff
@@ -36,6 +37,17 @@ const App = () => {
           <>
             <AppRouter />
             <OnLoadActions />
+            <ToastContainer
```
devinleighsmith, line 40: This is the root toast container in the DOM. All toasts are displayed within this container. Also, global toast defaults are defined here. These defaults can be overriden when creating a toast as well via:

toast('message' { toast_specific_options})

Fosol: I think having a consistent location for all toasts makes sense.

frontend/src/actionCreators/usersActionCreator.ts
```diff
@@ -66,20 +68,30 @@ export const fetchUserDetail = (id: API.IUserDetailParams) => (dispatch: Functio
...
       dispatch(adminActions.updateUser(response.data));
-      dispatch(success(reducerTypes.PUT_USER_DETAIL));
-      dispatch(hideLoading());
-    })
-    .catch(() => dispatch(error(reducerTypes.PUT_USER_DETAIL)))
-    .finally(() => dispatch(hideLoading()));
+      return Promise.resolve(response);
+    });
+
+  return handleAxiosResponseWithToasts(
```
Fosol, line 87: I this this needs to be abstracted one more layer.  In this current implementation you are essentially wrapping CustomAxios to give it the toast feature.  But this means any place we want toasts we need to manually wrap.

frontend/src/constants/toasts.ts
```diff
@@ -0,0 +1,16 @@
+import { toast } from 'react-toastify';
+
+/**
+ * The purpose of this file is to centralize the toasts in use in the application in one location. This should allow us to minimize duplication of toast messages using toastIds. https://fkhadra.github.io/react-toastify/prevent-duplicate
+ * In the future we may need to split this file out (at least into /features).
+ */
+
+/** These toasts are used by the update user api */
+export const USER_UPDATING_TOAST_ID = 'UPDATING_USER';
```
Fosol, line 9: I like centralization of string constants, but I think you're missing out on an opportunity to namespace this.  Presently this would result in a massive constant list of variables for all different types of toasts.  Find a way to namespace it would result in better maintainability and discoverability with intelisense.

Fosol (approved):

frontend/src/constants/toasts.ts
devinleighsmith: The way I was thinking we would namespace this is 
import * as toasts from toasts;
this allows you to treat this entire file as an object, meaning that intellisense works
ie toasts.USER_UPDATING_TOAST_ID

Fosol: Wouldn't it be cleaner to have something like `toasts.user.UPDATING_TOAST_ID`?  Then each file or object would contain their own toasts.

frontend/src/actionCreators/usersActionCreator.ts
devinleighsmith: I think that is what we want isn't it? I could force all network requests to display toasts, but for that to work we would have to use extremely generic toast messages, and also we would end up displaying toasts for all async actions in the system, which would either result in a huge number of toasts, or toasts would end up frequently cancelling each other if we force unique ids on them.

Possibly I'm not understanding what you are recommending here. Not that handleAxiosResponseWithToasts is just an altered version of handleAxiosResponse. The original purpose of handleAxiosResponse was to provide all of the redux network actions against an axios response to prevent that code from having to be written for all async actions. The new method is similar, but shouldn't be applied to all network actions because we don't need toasts for every api call in the system.

Fosol: Couldn't you provide an abstraction of `CustomAxios` that includes toasts?  You could then pass in the toast messages you want to use.

goemen (approved):

frontend/src/actionCreators/usersActionCreator.ts
devinleighsmith: alright, I added the toasts to CustomAxios instead.

frontend/src/constants/toasts.ts
devinleighsmith: sure, I can do that.

frontend/src/actionCreators/usersActionCreator.ts
Fosol: Awesome, now it should be much easier to add the feature anywhere else we want it.

PR #518, opened 2020-09-30
PA - 749 

devinleighsmith:
Disable unused roles from administration via DB migration. update frontend code using roles to use lookups hook. update lookups hook to filter out disabled items by default.

Add tests to check that enabled lookups are displayed and disabled lookups are not displayed on effected pages.


devinleighsmith:
I noticed the version naming convention on the other migration in the repo, but I was under the impression we were going to not create releases for specific migrations and instead group all migrations in a sprint under one migration and release that way. Maybe I don't fully understand how that looks.

Fosol:
Migration naming within a sprint isn't as critical as the version that is released at the end of the sprint.  Generally best practice is to either name them after the user story, or similar to true versioning standards MAJOR.MINOR.PATCH.  Since we don't want to have to support multiple migrations per release, the naming we choose during the sprint isn't as important.  I'm open to any practice we want to do.

backend/dal/Migrations/20200930210531_DisableRoles.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class DisableRoles : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("UPDATE dbo.Roles SET IsDisabled = 1 WHERE Name in ('Assistant Deputy Minister', 'Assistant Deputy Minister Assistant', 'Executive Director')");
```
Fosol, line 9: Move all scripts to the seed folder to be maintained as separate SQL.  You can see how this is accomplished by the `Initial` migration.  This requires you to use the `SeedMigration` instead.

backend/dal/Migrations/20200930210531_DisableRoles.cs
```diff
@@ -0,0 +1,17 @@
...
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("UPDATE dbo.Roles SET IsDisabled = 0 WHERE Name in ('Assistant Deputy Minister', 'Assistant Deputy Minister Assistant', 'Executive Director')");
```
Fosol, line 14: You should update the `SeedMigration` to be able to easily run rollback scripts the same way you can run deployment scripts.

frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -38,7 +50,9 @@ function useCodeLookups() {
     getOptionsByType: getOptionsByType,
```
Fosol, line 50: Why not be consistent in syntax?

frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -13,7 +14,18 @@ function useCodeLookups() {
     return lookupCodes.filter(code => code.type === type && code.id === id)?.find(x => x)?.code;
   };
 
-  const getByType = (type: string) => lookupCodes.filter(code => code.type === type);
+  const getByType = useCallback(
```
Fosol, line 17: I think this is a good use of `useCallback`.  

One thing I don't fully understand related to this though is whether the results are also cached in memory, or is it just the function?  If it's just the function, I wonder how valuable `useCallback` is.

https://dmitripavlutin.com/dont-overuse-react-usecallback/

Fosol (changes requested):

frontend/src/hooks/useLookupCodes.tsx
devinleighsmith: I believe that we are in this situation here (from your article):

**However, there are cases when you need to keep one instance of a function:

A component wrapped inside React.memo() (or shouldComponentUpdate) accepts a callback prop**

ManageUsersPage has a useMemo which accepts getByType as a callback prop.

I don't think ManageUsersPage really requires the memoized agencies/roles. However, that was how that code was written prior to my change, and I was fine leaving it as-is

frontend/src/hooks/useLookupCodes.tsx
devinleighsmith: Not sure what you mean here.

Fosol: Maybe I'm missing something, but it just caught my eyes when scanning the code.  It appeared inconsistent.
```
{
    getOptionsByType: getOptionsByType,
    getCodeById: getCodeById,
    getByType,
    getPublicByType,
    filterByParent,
    lookupCodes,
  }
```
Why not
```
{
    getOptionsByType,
    getCodeById,
    getByType,
    getPublicByType,
    filterByParent,
    lookupCodes,
}
```
or
```
{
    getOptionsByType: getOptionsByType,
    getCodeById: getCodeById,
    getByType: getByType,
    getPublicByType: getPublicByType,
    filterByParent: filterByParent,
    lookupCodes: lookupCodes,
  }
```

Fosol (approved):

frontend/src/hooks/useLookupCodes.tsx
devinleighsmith: Oh I see. Probably just multiple people touching the code. I'll update for consistency.

jdtoombs (approved):

PR #519, opened 2020-10-01
Zoom out to show more results and filter by city

goemen:


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -89,6 +92,17 @@ const Map: React.FC<MapProps> = ({
   });
   const [baseLayers, setBaseLayers] = useState<BaseLayer[]>([]);
   const [activeBasemap, setActiveBasemap] = useState<BaseLayer | null>(null);
+  const smallScreen = useMediaQuery({ maxWidth: 1800 });
+
+  useEffect(() => {
```
goemen, line 97: @devinleighsmith this is the code that trigger the fetch if there are no results in the current bounds

Fosol (approved):

frontend/src/components/maps/leaflet/Map.tsx
devinleighsmith: Cool thanks, I get it now.

devinleighsmith (approved):

jdtoombs (approved):

PR #521, opened 2020-10-02
Pa 857

devinleighsmith:


frontend/src/customAxios.ts
```diff
@@ -69,13 +66,8 @@ const CustomAxios = ({
         lifecycleToasts.errorToast();
       }
 
-      const status = error.response ? error.response.status : null;
       //const errorMessage =
       //  errorToastMessage || (error.response && error.response.data.message) || String.ERROR;
-
-      if (status === UNAUTHORIZED || status === MAINTENANCE) {
-        window.location.reload(false);
```
devinleighsmith, line 77: I think this is the most likely cause of your issue @Fosol . This logic could easily get into a loop where it would try and load a url, get a 401, reload, and then get a 401 again indefinitely. This inspired me to find and remove the other non-react router navigations throughout the app.

frontend/src/features/admin/users/constants/index.tsx
```diff
@@ -4,14 +4,15 @@ import { CellProps } from 'react-table';
...
+import { Link } from 'react-router-dom';
 
 export const columnDefinitions: ColumnWithProps<IUserRecord>[] = [
   {
     Header: 'IDIR/BCeID',
     accessor: 'username',
     align: 'left',
     Cell: (props: CellProps<IUserRecord>) => {
-      return <a href={`/admin/user/${props.row.original.id}`}>{props.row.original.username}</a>;
+      return <Link to={`/admin/user/${props.row.original.id}`}>{props.row.original.username}</Link>;
```
devinleighsmith, line 15: React router links work the same as a tags for navigation but the won't force the page to refresh.

frontend/src/features/admin/access/ManageAccessRequests.test.tsx
```diff
@@ -81,8 +84,6 @@ describe('Manage access requests', () => {
   it('Snapshot matches', () => {
     const component = componentRender(successStore);
     expect(component.toJSON()).toMatchSnapshot();
-    const instance = component.getInstance();
```
devinleighsmith, line 84: this is tested in the next test.

Fosol (approved):
Very happy to see this!

goemen (approved):

PR #523, opened 2020-10-02
Popup for deleting building 

jdtoombs:
Ended up removing my modal test as it was causing havoc. Original test still works for testing of removal functionality.

### DIFF
--- a/frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.test.tsx
+++ b/frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.test.tsx
@@ -15,10 +15,13 @@ import { ILookupCode } from 'actions/lookupActions';
 import _ from 'lodash';
 import MockAdapter from 'axios-mock-adapter';
 import axios from 'axios';
+import { createMemoryHistory } from 'history';
+import { Router } from 'react-router-dom';
 
 const mockAxios = new MockAdapter(axios);
 mockAxios.onAny().reply(200, {});
 
+const history = createMemoryHistory();
 const mockBuilding: IFormBuilding = {
   address: {
     cityId: 1,
@@ -28,6 +31,8 @@ const mockBuilding: IFormBuilding = {
     city: 'city',
     province: 'province',
   },
+  name: 'name',
+  classification: 'class',
   agencyId: 3,
   buildingConstructionTypeId: 4,
   buildingFloorCount: 5,
@@ -97,9 +102,11 @@ describe('PagedBuildingForms functionality', () => {
   const getPagedBuildingForms = (initialValues: IFormParcel, onSubmit: any) => {
     return (
       <Provider store={store}>
-        <Formik initialValues={initialValues} onSubmit={onSubmit}>
-          <PagedBuildingForms allowEdit={true}></PagedBuildingForms>
-        </Formik>
+        <Router history={history}>
+          <Formik initialValues={initialValues} onSubmit={onSubmit}>
+            <PagedBuildingForms allowEdit={true}></PagedBuildingForms>
+          </Formik>
+        </Router>
       </Provider>
     );
   };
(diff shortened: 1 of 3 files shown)


frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.test.tsx
```diff
@@ -15,10 +15,13 @@ import { ILookupCode } from 'actions/lookupActions';
 import _ from 'lodash';
```
jdtoombs, line 15: I'll come back to this and get the removed modal test working

goemen (approved):

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
```diff
@@ -58,6 +73,22 @@ const PagedBuildingForms: React.FC<PagedBuildingFormsProps> = (props: PagedBuild
         name="buildings"
         render={arrayHelpers => (
           <div>
+            {showDelete && (
```
devinleighsmith, line 76: you may not need this conditional render if you are using the display prop.

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
```diff
@@ -33,6 +34,18 @@ const getPageErrors = (errors: any, nameSpace: any) => {
   );
 };
 
+const deletePagedBuilding = (
+  buildings: IFormBuilding[],
+  currentPage: number,
+  setCurrentPage: Function,
+  arrayHelpers: any,
+) => {
+  if (currentPage === buildings.length - 1) {
```
devinleighsmith, line 43: Does this work ok with a single building?

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
```diff
@@ -91,10 +122,9 @@ const PagedBuildingForms: React.FC<PagedBuildingFormsProps> = (props: PagedBuild
                     variant="link"
                     title="Remove Building"
                     onClick={() => {
-                      if (currentPage === buildings.length - 1) {
-                        setCurrentPage(currentPage - 1);
-                      }
-                      arrayHelpers.remove(currentPage);
+                      formikProps.dirty
```
devinleighsmith, line 125: Does this work as intended? To me this means:
If the entire form has been modified from the initial state, show the delete dialog. Otherwise just delete the building.

I think it would make more sense if it:
Checked to see if the building being deleted has any non-default content, if so, show the delete dialog when delete is clicked
- or -
Check to see if the building being deleted has an ID (meaning it exists in the DB) if it does, show the delete dialog.

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
```diff
@@ -58,6 +73,22 @@ const PagedBuildingForms: React.FC<PagedBuildingFormsProps> = (props: PagedBuild
         name="buildings"
         render={arrayHelpers => (
           <div>
+            {showDelete && (
+              <GenericModal
+                display={showDelete}
+                cancelButtonText="Close"
+                okButtonText="Delete"
+                handleOk={() => {
+                  deletePagedBuilding(buildings, currentPage, setCurrentPage, arrayHelpers);
```
devinleighsmith, line 82: This may be outside the scope of this story, but it would be nice if the modal tried to submit the form whenever a building is deleted. The problem with this is that you would have to check for other validation errors in the form and display an error to prevent situations where an auto-submitted deleted building fails.

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
jdtoombs: Hmm yes you're correct, going to implement a different solution as you suggested

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
jdtoombs: yup, seems to be working. need to implement changes for the conditional dirty render we talked about below though

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
jdtoombs: ahh yes that's a good idea. right now it might surprise a few users when they don't hit submit after removing... the modal 'feels' like a submit. maybe for now I should add a message saying finalize by submitting the form in the meantime...

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
jdtoombs: So I removed this and it worked functionally fine, but I was getting a bunch of failures on the unit tests...

frontend/src/features/properties/components/forms/subforms/PagedBuildingForms.tsx
```diff
@@ -33,6 +34,18 @@ const getPageErrors = (errors: any, nameSpace: any) => {
   );
 };
 
+const deletePagedBuilding = (
+  buildings: IFormBuilding[],
+  currentPage: number,
+  setCurrentPage: Function,
```
Fosol, line 40: It looks like `setCurrentPage(currentPage - 1)` requires a number parameter.  I would be better to provide that detail here so that it doesn't accept a generic function.

Fosol (approved):

PR #527, opened 2020-10-06
Integrate with geo coder to zoom to city on city filter

goemen:


jdtoombs (approved):

Fosol (approved):
Does this now only perform the zoom if the city is part of the filter?

PR #530, opened 2020-10-06
PIMS-937

jdtoombs:
image


backend/api/Areas/Property/Controllers/ParcelController.cs
```diff
@@ -78,6 +78,22 @@ public IActionResult IsPidAvailable(int parcelId, int pid)
...
+        /// <summary>
+        /// Check if PIN is available
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("check/pin-available")]
+        [HasPermission(Permissions.PropertyView)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.CheckPidAvailabilityResponseModel), 200)]
+        [SwaggerOperation(Tags = new[] { "parcel" })]
+        public IActionResult IsPinAvailable(int parcelId, int pin)
```
Fosol, line 90: If you're trying to determine if a PIN is available why do you ask for the `parcelId`?

frontend/src/hooks/useApi.ts
```diff
@@ -68,6 +72,18 @@ export const useApi = (): PimsAPI => {
     return data;
   };
 
+  axios.isPinAvailable = async (
```
Fosol, line 75: I'm wondering if this is the best approach.  This requires making a separate request to determine if the PIN is available.  But you essentially get this information when you submit the property, as the error returned I believe states this.

Fosol:
I have mixed opinion on the solution.  I kinda like it, as it gives them feedback immediately when they enter the PIN.  But I kinda don't love it as now it's a separate AJAX request to find out.  I just don't understand why you're including the PID in the request to find out if the PIN is available?

backend/api/Areas/Property/Controllers/ParcelController.cs
jdtoombs: This is the `id `not the actual `pid`. It is the same way we check for the duplicate `pid` as well. It sends the `parcelId` and `pid`.  This also ensures that if it is the same parcel it finds the duplicate on it will not cause the error

jdtoombs: Or am I confusing myself

frontend/src/hooks/useApi.ts
jdtoombs: This is true, I guess the tradeoff would be the instant feedback. This is also the same way we check for the PID duplicate. Open to changing this approach though

backend/api/Areas/Property/Controllers/ParcelController.cs
Fosol: Ah I see, misread that.  Thanks!

Fosol: I may need to double check the constraints in the DB on this.  I'm wondering if it's possible to have duplicate PIN if the PID is different.

Fosol (approved):

frontend/src/hooks/useApi.ts
devinleighsmith: Yeah, I think the bigger question here is how we want to handle problems like this in the future. For every instance of a form validation error that can only be determined by the API, we can
1) Create a new api request to handle that validation
OR
2) We can update the api to return some json that the frontend can use to set errors in formik.

There are some potential issues with #2 that we should consider. In general manually setting errors in formik isn't great, as formik will overwrite those errors the next time validation runs. We would probably have to write some more custom formik code to handle this scenario.

Also, even if we tell the user they have a duplicate on the correct field we aren't able to tell them what values are allowed which may result in the user having to try to submit random values until the error goes away.

devinleighsmith: Personally I'd lean towards #1 if we think there will only be a few instances of this problem across the application. If this is a common issue that will occur on every form, lets consider #2.

Fosol: Generally speaking they shouldn't be entering random numbers.  These numbers are supposed to be generated by another tool, which gives them unique values.  In PIMS they are simply transcribing them.

frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -201,10 +201,18 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
       pidDuplicated = !(await isPidAvailable(values));
     }
 
+    let pinDuplicated = false;
```
devinleighsmith, line 204: I missed this on Goeme's review but I think this is important enough that it shouldn't be ignored.

Since we are performing expensive, field specific validation I think we should be doing that validation at the level of the problematic field and not of the whole form:

https://formik.org/docs/guides/validation#field-level-validation

devinleighsmith: That way we aren't performing this validation every time the form changes (or at least, checking to see if we need to perform this validation every time the form changes).

devinleighsmith: Some more context on this since it is relevant and not obvious.

We rely heavily on memoized form components in most of our forms for performance reasons. The consequence of this is that a change in field A will not trigger a re-render in field B. This means that any component that needs to be triggered by another component needs to be handled at the global form level (ie evaluation sums).

In this case I think the validation logic is specific to each individual field, and so I don't think the above will be an issue for field based validation.

jdtoombs: Ah ok this makes sense, I'll try to switch it over to field level validation!

jdtoombs: Would it be bad practice to just add the following: 
` if (values.pin && initialValues.pin !== values.pin) {
      pinDuplicated = !(await isPinAvailable(values));
  }`? 
I noticed the PID was not firing every time to check for duplicate but mine was as I did not check it against its initial value

frontend/src/components/layout/AppNavBar/AppNavBar.test.tsx
```diff
@@ -225,33 +225,12 @@ describe('AppNavBar', () => {
       });
     });
 
-    it('AppNavBar include View Projects Link', () => {
```
jdtoombs, line 228: duplicate test

frontend/src/utils/YupSchema.ts
```diff
@@ -172,7 +172,7 @@ export const ParcelSchema = Yup.object()
         otherwise: Yup.string()
           .min(1)
           .required('PID or PIN Required')
-          .max(15, 'Please enter a valid PIN'),
+          .max(9, 'Please enter a valid PIN no longer than 9 digits.'),
```
jdtoombs, line 175: anything larger than 9 will cause a validation error with api calls

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -164,7 +164,7 @@ const toApiProperty = (property: IProperty): IApiProperty => {
     parcelId: property.propertyTypeId === 0 ? property.id : undefined,
     buildingId: property.propertyTypeId === 1 ? property.id : undefined,
     pid: property.pid,
-    pin: property.pin,
+    pin: Number(property.pin),
```
jdtoombs, line 167: Was causing the bad request on project creation, was doing this in test too. I guess we just never really submitted properties with PINs that often

frontend/src/features/properties/components/forms/ParcelDetailForm.tsx
```diff
@@ -201,10 +202,18 @@ const ParcelDetailForm = (props: ParcelPropertyProps) => {
...
+    if (values.pin && initialValues.pin !== values.pin && values.pin.toString().length < 10) {
```
jdtoombs, line 206: Just thought of the scenario where the person changes the PIN for whatever reason. I guess this would still fire at form level as is.

devinleighsmith (approved):

PR #536, opened 2020-10-08
React Table filter paging fix, reset to page 0 when filter changes.

devinleighsmith:


jdtoombs (approved):

frontend/src/components/Table/TablePagination.tsx
```diff
@@ -54,6 +54,7 @@ const TablePagination = <T extends object>(props: TablePaginationProps<T>): Reac
       marginPagesDisplayed={2}
       pageRangeDisplayed={5}
       onPageChange={handleChangePage}
+      forcePage={pageIndex}
```
devinleighsmith, line 57: This ensures that the pagination control stays in sync with the internal state within react table.

frontend/src/features/admin/users/ManageUsersPage.tsx
```diff
@@ -62,6 +62,13 @@ export const ManageUsersPage = () => {
     state => (state.network as any)[actionTypes.GET_USERS] as IGenericNetworkAction,
   );
 
+  const onRequestData = useCallback(
```
devinleighsmith, line 65: callbacks are required for functions that control react-table behaviour (ie. pageIndex, pageSize, etc). Otherwise these functions run on every render of the table.

frontend/src/features/admin/users/ManageUsersPage.tsx
```diff
@@ -98,20 +105,22 @@ export const ManageUsersPage = () => {
         value={filter}
         agencyLookups={agencies}
         rolesLookups={roles}
-        onChange={value => dispatch(getUsersFilterAction(value))}
+        onChange={value => {
```
devinleighsmith, line 108: whenever the filter changes, update the filter in redux and reset the page index to 0.

frontend/src/components/Table/Table.tsx
```diff
@@ -209,7 +213,15 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
 
   // Listen for changes in pagination and use the state to fetch our new data
   useEffect(() => {
-    onRequestData?.({ pageIndex, pageSize });
+    pageIndexProp !== undefined && pageIndexProp >= 0 && instance.gotoPage(pageIndexProp);
+  }, [pageIndexProp, instance]);
+
+  useEffect(() => {
```
devinleighsmith, line 219: update the internal state of the table based on the external pageSize prop.

frontend/src/components/Table/Table.tsx
```diff
@@ -209,7 +213,15 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
 
   // Listen for changes in pagination and use the state to fetch our new data
   useEffect(() => {
-    onRequestData?.({ pageIndex, pageSize });
+    pageIndexProp !== undefined && pageIndexProp >= 0 && instance.gotoPage(pageIndexProp);
```
devinleighsmith, line 216: Update the internal state of the table based on the external pageIndex prop.

frontend/src/components/Table/Table.tsx
```diff
@@ -151,7 +153,9 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
       columns,
       data,
       defaultColumn,
-      initialState: pageSizeProp ? { pageIndex: 0, pageSize: pageSizeProp } : { pageIndex: 0 },
+      initialState: pageSizeProp
```
devinleighsmith, line 156: use external props to initialize table settings when provided.

Fosol (approved):
I couldn't tell by reviewing this PR, and it may be irrelevant, however there is an existing bug where if a filter is applied after paging through results, the paging doesn't work correctly.  Essentially it gets stuck on a page that doesn't exist in the new filtered results.  Does this new Table component handle this correctly?

devinleighsmith:
The changes fix two distinct issues that I observed:
1) Now whenever the filter is changed or the filter search button is clicked, the page that is requested/saved is **always** reset to 0 (1).
2) The pageSize prop no longer resets back to the initial value after it has been changed. This fixes an issue where setting the page size from 10 to 5 and then navigating to page 2 would request page=2&pageSize=10 from the api. This was resulting in the API returning a black page when there were between 5 and 10 results and the user wanted to view the second page.

devinleighsmith:
Of course, if the user has a filter saved and they try and navigate between pages, they are still able to do this.

PR #537, opened 2020-10-08
Sensitive property changes

devinleighsmith:
swap checkbox with radio button.
add a linke w/tooltip


frontend/src/features/properties/components/forms/subforms/LandForm.tsx
```diff
@@ -193,13 +195,25 @@ const LandForm = <T extends any>(props: LandProps & FormikProps<T>) => {
           </Form.Row>
 
           <Form.Row>
-            <Form.Label column md={2}>
-              Sensitive Land
+            <Form.Label column md={12}>
+              Harmful if Released?{' '}
+              <TooltipWrapper toolTipId="sensitive-harmful" toolTip={senstiveTooltip}>
```
Fosol, line 200: Probably no great solution, but separating the tooltip text from the link implementation requires two places to be edited if anything changes.  I'm not a fan of the dangerous HTML in text for this, but wonder if there is another way?

Fosol (approved):

frontend/src/features/properties/components/forms/subforms/LandForm.tsx
devinleighsmith: I'm not sure if "two places to be edited if anything changes" is the concern here. If the link needs to change, then change the link. If the tooltip needs to change, then change the tooltip. I don't think these two things are tied together in such a way that a change to one would or should require a change in the other.

However, if we re-use this component in multiple places then a future dev may need to update the link href in multiple places if the link changes since the link is hardcoded here. I'll move the link to strings.js as well to mitigate that.

jdtoombs (approved):

goemen (approved):

PR #539, opened 2020-10-09
PA-982 Config API to send logs to Elasticsearch

goemen:


backend/api/Program.cs
```diff
@@ -48,8 +53,41 @@ public static IWebHostBuilder CreateWebHostBuilder(string[] args)
                     config.AddEnvironmentVariables();
                     config.AddCommandLine(args);
                 })
+                .UseSerilog()
                 .UseUrls(config.GetValue<string>("ASPNETCORE_URLS"))
                 .UseStartup<Startup>();
         }
+
+        private static void ConfigureLogging()
```
Fosol, line 61: I find it odd we need to configure logging here.  I'm surprised we wouldn't configure it in the `Startup.ConfigureServices(...)` function.  As that is where we setup DI.

backend/api/appsettings.Development.json
```diff
@@ -1,13 +1,19 @@
...
+      "Override": {
+        "Microsoft": "Warning",
+        "System": "Warning",
+        "Microsoft.Hosting.Lifetime": "Information",
+        "System.Net.Http.HttpClient": "Error",
+        "HealthChecks.UI": "Error"
+      }
     }
   },
+  "ElasticConfiguration": {
```
Fosol, line 14: Shouldn't need this if the value is the same in the `appsettings.json`

backend/api/appsettings.Test.json
```diff
@@ -1,11 +1,16 @@
...
       "Default": "Information",
-      "Microsoft": "Warning",
-      "Microsoft.Hosting.Lifetime": "Information"
+      "Override": {
+        "Microsoft": "Warning",
+        "Microsoft.Hosting.Lifetime": "Information"
+      }
     }
   },
+  "ElasticConfiguration": {
```
Fosol, line 11: Shouldn't need this if the value is the same in the `appsettings.json`

Fosol (approved):

PR #543, opened 2020-10-11
Reorganizing migration script implementation

Fosol:
## Description

Discovered issue with how I merged migration.  Had to manually run scripts in TEST and PROD.  This will fix the issue and keep migrations consistent moving forward.


backend/dal/Helpers/Migrations/SeedMigration.cs
```diff
@@ -54,39 +54,39 @@ public SeedMigration()
 
         #region Methods
```
Fosol, line 55: By default it will look for scripts in `/Migrations/Up` and `/Migrations/Down` and run them during up or down migrations.  Extended this by providing `PreUp`, `PostUp`, `PreDown`, `PostDown`.  Each of these deal with common use-case for running scripts before and after the up and down migrations.

PR #547, opened 2020-10-14
Correct time format in PIMS by converting times from utc to the user'…

devinleighsmith:
…s local time.


frontend/src/utils/utils.ts
```diff
@@ -158,6 +158,15 @@ export const formatDateTime = (date: string | undefined) => {
   return !!date ? moment(date).format('YYYY-MM-DD hh:mm a') : '';
 };
 
+export const formatApiDateTime = (date: string | undefined) => {
```
Fosol, line 161: Add some jsdoc and it'll be even clearer what the function does with intellisense 👍

Fosol (approved):

jdtoombs (approved):

PR #550, opened 2020-10-14
Show appropriate buildings when 2+ folders open

jdtoombs:


frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -201,8 +201,14 @@ const PropertyListView: React.FC = () => {
     if (expandedRows.length > 0) {
       await Promise.all(
         expandedRows.map(async property => {
-          property.propertyTypeId === 0 &&
-            setExpandData((await service.loadBuildings(property.id)).items);
+          if (property.propertyTypeId === 0) {
+            if (expandData[property.id] === undefined) {
```
jdtoombs, line 205: So it doesn't reload data every time

devinleighsmith (approved):

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -95,8 +95,8 @@ const PropertyListView: React.FC = () => {
 
   // We'll start our table without any data
   const [data, setData] = useState<IProperty[]>([]);
-  // For getting the buildings on parcel click
-  const [expandData, setExpandData] = useState<IProperty[]>([]);
+  // For getting the buildings on parcel folder click
+  const [expandData, setExpandData] = useState<any>({});
```
Fosol, line 99: Is there a reason to use `any`?

Fosol (approved):

frontend/src/features/properties/list/PropertyListView.tsx
jdtoombs: We could make our own interface but without the type `any` typescript will throw an error as you can't use a number to index type `{}`

PR #552, opened 2020-10-16
PA-194

devinleighsmith:
Also includes fixes to listview queries and mapview throttling.

### DIFF
--- a/frontend/src/hooks/useRouterFilter.ts
+++ b/frontend/src/hooks/useRouterFilter.ts
@@ -1,0 +1,67 @@
+import { useDispatch, useSelector } from 'react-redux';
+import { useHistory } from 'react-router-dom';
+import React, { useState } from 'react';
+import { saveFilter } from 'reducers/filterSlice';
+import { RootState } from 'reducers/rootReducer';
+import _ from 'lodash';
+import { BasePropertyFilter } from 'components/common/interfaces';
+
+const defaultFilter: BasePropertyFilter = {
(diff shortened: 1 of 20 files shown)


jdtoombs (approved):

frontend/src/hooks/useRouterFilter.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+  return obj;
+};
+
+/**
+ * Control the state of the passed filter using url search params.
+ * NOTE: URLSearchParams not supported by IE.
+ */
+export const useRouterFilter = (filter: any, setFilter: (val: any) => void, key: string) => {
```
Fosol, line 38: Does this mean that on IE the page won't work?  Or just that they can't get to a specific filter by copying a URL?

Fosol (approved):

frontend/src/hooks/useRouterFilter.ts
devinleighsmith: Haven't tested, but my guess is that if IE calls that function it will throw an exception. So year, this will likely break the map and listview for IE.

devinleighsmith: Could add a try catch here to essentially do no filter saving for IE.

Fosol: I think we should spend a very little time to see if it doesn't barf on IE.  Although I don't know if IE can even login at this point.

PR #553, opened 2020-10-16
PA-980, 993

jdtoombs:
Quick PR!

### DIFF
--- a/frontend/src/features/properties/components/forms/ParcelDetailForm.test.tsx
+++ b/frontend/src/features/properties/components/forms/ParcelDetailForm.test.tsx
@@ -265,7 +265,8 @@ describe('ParcelDetailForm', () => {
       await fillInput(container, 'isSensitive', true, 'radio');
       // await fillInput(container, 'financials.1.date', exampleData.evaluations[1].date);
       // TODO: add a function capable of filling this type of field
-      await fillInput(container, 'financials.0.netbook.value', exampleData.fiscals[1].value);
+      // Net Book no longer required
+      // await fillInput(container, 'financials.0.netbook.value', exampleData.fiscals[1].value);
       (updateParcel as any).mockImplementation((parcel: IParcel) => {
         expect(parcel.isSensitive).toBeTruthy();
         expect(parcel.pin).toEqual(exampleData.pin);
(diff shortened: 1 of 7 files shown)


frontend/src/features/properties/components/forms/subforms/EvaluationForm.tsx
```diff
@@ -112,12 +112,13 @@ export const validateFinancials = (
   financialYears.forEach((financialYear, index) => {
```
devinleighsmith, line 112: I think we may be at the point where literally none of this validation is required anymore.

devinleighsmith: Confirm this with Mikaela, but we don't show appraisal anymore so that validation should not be required, and then I think netbook was the only field left that had any validation on it. If we aren't validating netbook anymore I don't think we are validating anything.

devinleighsmith (approved):

frontend/src/features/properties/components/forms/ParcelDetailForm.test.tsx
```diff
@@ -265,7 +265,8 @@ describe('ParcelDetailForm', () => {
       await fillInput(container, 'isSensitive', true, 'radio');
       // await fillInput(container, 'financials.1.date', exampleData.evaluations[1].date);
       // TODO: add a function capable of filling this type of field
-      await fillInput(container, 'financials.0.netbook.value', exampleData.fiscals[1].value);
+      // Net Book no longer required
+      // await fillInput(container, 'financials.0.netbook.value', exampleData.fiscals[1].value);
```
Fosol, line 269: I think at this point we shouldn't just comment it out.

frontend/src/features/properties/components/forms/subforms/EvaluationForm.tsx
Fosol: Essentially we don't want to block them from saving the property.   There is no validation now on financial values (other than perhaps they are valid positive numbers 0-infinity)

Fosol (approved):

frontend/src/features/properties/components/forms/ParcelDetailForm.test.tsx
jdtoombs: Ah ok, I will update it!

frontend/src/features/properties/components/forms/subforms/EvaluationForm.tsx
jdtoombs: Ok that's what I was thinking originally was just wondering if I should have left it incase the requirements change again. But it's easy enough to add back in so removal is probably the best option.

goemen (approved):

PR #554, opened 2020-10-16
Timestamp on parcel detail view

devinleighsmith:


backend/api/Mapping/BaseMap.cs
```diff
@@ -11,6 +11,8 @@ public void Register(TypeAdapterConfig config)
             config.NewConfig<Entity.BaseEntity, Models.BaseModel>()
                 .Map(dest => dest.CreatedOn, src => src.CreatedOn)
                 .Map(dest => dest.UpdatedOn, src => src.UpdatedOn)
+                .Map(dest => dest.UpdatedByName, src => src.UpdatedBy != null ? src.UpdatedBy.DisplayName : src.CreatedBy != null ? src.CreatedBy.DisplayName : "unknown")
```
devinleighsmith, line 14: Yeah, so I'm not entirely sure this is the best way forward, maybe we should be returning an actual user object here instead of these specific fields?

jdtoombs (approved):

goemen (approved):

backend/api/Models/BaseModel.cs
```diff
@@ -9,6 +9,10 @@ public abstract class BaseModel
 
         public DateTime? UpdatedOn { get; set; }
 
+        public string? UpdatedByName { get; set; }
```
Fosol, line 12: Not sure why we need to use a nullable string here?

backend/api/Models/BaseModel.cs
```diff
@@ -9,6 +9,10 @@ public abstract class BaseModel
...
+
+        public string? UpdatedByEmail { get; set; }
```
Fosol, line 14: Not sure why we need to use a nullable string here?

Fosol (approved):

PR #557, opened 2020-10-19
pa-745 & pa-635 - Adding v01.01.00 migration

Fosol:
## Description

Creating new database migration **v01.01.00** which contains updates to the **Project** and **ProjectSnapshot** tables.

Updating makefile with a few more shortcuts to get things done quicker;
- `make db-add n={migration name}` = create a new migration and auto complete the code changes for scripts
- `make db-drop` = drop the whole database
- `make npm-refresh` = If npm packages change, get the frontend rebuilt and working again
- `make db-remove` = remote the prior migration and files from code and the db
- `make db-migrations` = list all the migrations
- `make db-rollback n={migration name}` = rollback the specified db migration
` `make build n={service name}` = build the specific service
- `make up n={service name}` = spin up the specific service
- `make stop n={service name}` = stop the specific service

### DIFF
--- a/backend/entities/Project.cs
+++ b/backend/entities/Project.cs
@@ -263,7 +263,7 @@ namespace Pims.Dal.Entities
         [NotMapped]
         [DataMemberAttribute]
         public DateTime? PriorYearAdjustmentOn { get; set; } // TODO: Move to metadata property.
-        
+
         /// <summary>
         /// get/set - When the project was completed.
         /// </summary>
@@ -289,7 +289,7 @@ namespace Pims.Dal.Entities
 
         #region Financials
         /// <summary>
-        /// get/set - The netbook value which is the sum of the properties.
+        /// get/set - The netbook value.
         /// </summary>
         public decimal NetBook { get; set; }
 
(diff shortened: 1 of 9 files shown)


backend/entities/Project.cs
```diff
@@ -289,7 +289,7 @@ public class Project : BaseEntity
 
         #region Financials
         /// <summary>
-        /// get/set - The netbook value which is the sum of the properties.
+        /// get/set - The netbook value.
```
Fosol, line 292: These are no longer sums, they will be manually entered by user.

scripts/db-migration.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
Fosol, line 1: This script provides a way to automatically update the migration auto-generated file so that it is converted to a seed migration.  Which means it can run scripts in the appropriate folder structure.

devinleighsmith:
I'm confused. I thought I was assigned to the DB story on PIMS-635, so I ended up creating the same migration. Any ideas how we can communicate better to avoid duplicated work in the future?

devinleighsmith (approved):

devinleighsmith:
@Fosol I also made the API changes for this story, so hopefully you aren't working on that in parallel as well.

Fosol:
@devinleighsmith I mentioned it this morning in Scrum.  But if you have yours done, feel free to submit the PR.  I was working on the db changes to support pa-745 and felt they were closely related and it wouldn't take much to do it all.

devinleighsmith:
If you wouldn't mind, assign yourself the story in Jira next time. That would greatly assist me.

scripts/db-migration.sh
devinleighsmith: This is nice.

jdtoombs (approved):

backend/entities/ProjectSnapshot.cs
```diff
@@ -23,22 +24,59 @@ public class ProjectSnapshot : BaseEntity
...
+        /// get/set - Whether this snapshot represents a final report (i.e. a report presented to the ADM).
+        /// </summary>
+        /// <value></value>
+        public bool IsFinal { get; set; }
+
+        /// <summary>
+        /// get/set - The date the snapshot was taken.
+        /// </summary>
+        /// <value></value>
+        public DateTime SnapshotOn { get; set; }
```
goemen, line 43: Can we rename this to something like `SnapshotTimestamp`, `SnapshotOn` feels like its a boolean value

goemen (approved):

PR #561, opened 2020-10-20
Pa 635

devinleighsmith:
recreated this PR because it is so different from the old one.


backend/api/Areas/Project/Controllers/ReportController.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// Get all of the project snapshots for the specified report id.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("snapshots/{reportId:int}")]
+        [HasPermission(Permissions.AdminProjects)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(ProjectSnapshotModel[]), 200)]
+        [ProducesResponseType(typeof(ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public IActionResult GetProjectReportSnapshots(int reportId)
```
devinleighsmith, line 92: To me it made sense to group some of this snapshot apis with the report apis since all of the snapshot data being returned is driven by the reports table.

Fosol: I agree, although you still may want a separate controller in the Reports area.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+            return Get(originalReport.Id);
+        }
+
+        /// <summary>
+        /// Generate an updated list of snapshots based on current project data.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <exception cref="NotAuthorizedException">User does not have permission to get refreshed snapshots.</exception>
+        /// <returns></returns>
+        public IEnumerable<ProjectSnapshot> Refresh(int reportId)
```
Fosol, line 189: I think this ideally would be returning a new `ProjectReport` as the `To` date will change.  However because the relationship is loosely defined with a date that wouldn't provide the appropriate information.  Not sure anything needs to change here.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+            this.Context.CommitTransaction();
+        }
+
+
+        /// <summary>
+        /// Generate snapshots for all SPL projects using the passed 'to' date as the snapshot date.
+        /// </summary>
+        /// <param name="to"></param>
+        /// <returns></returns>
+        private IEnumerable<ProjectSnapshot> GenerateSnapshots(DateTime to)
```
Fosol, line 236: You don't need to pass `to` here.  We're simply generating a new collection of snapshots for the most current moment in time.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+                .Include(p => p.Status)
+                .Where(p => p.Workflow.Code == "SPL") ?? throw new KeyNotFoundException();
+
+            IEnumerable<ProjectSnapshot> projectSnapshots = new ProjectSnapshot[0];
+            foreach (Project project in splProjects)
+            {
+                this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+
+                ProjectSnapshot snapshot = new ProjectSnapshot(project);
+                snapshot.SnapshotOn = to;
```
Fosol, line 249: All the snapshots should have the same `to` value, but it doesn't need to be passed into the function.  It can be generated when the function is called.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (originalReport.To == null) throw new ArgumentNullException();
+            if (originalReport.IsFinal && report.IsFinal) throw new InvalidOperationException($"Unable to update FINAL project reports.");
+            if (originalReport.From == originalReport.To) throw new InvalidOperationException($"Project report start and end dates cannot be the same.");
+
+            this.Context.Entry(originalReport).CurrentValues.SetValues(report);
+
+            //If the report 'To' date has changed regenerate all the snapshots based on that date.
+            if(!originalReport.To.Equals(report.To))
+            {
+                this.Context.ProjectSnapshots.AddRange(GenerateSnapshots((DateTime)report.To));
```
Fosol, line 174: If you are generating new snapshots for this moment in time, then you also need to recalculate the variance for each snapshot.  Seems like you could just call the `Refresh(...)` function at this point.  It should be noted that this action will likely result in different report than what the user is viewing presently.  This means that you will need to ensure the returned report contains the latest snapshot and the UI is updated.  This is particular important before they export the results.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+                report.To = DateTime.UtcNow;
+            }
+            var mostRecentReport = this.Context.ProjectReports
+                .Where(r => r.ReportTypeId == ReportTypes.SPL)
+                .OrderByDescending(r => r.To)
+                .FirstOrDefault();
+            report.From = mostRecentReport?.To;
+            this.Context.Add(report);
+
+            this.Context.ProjectSnapshots.AddRange(GenerateSnapshots((DateTime)report.To));
```
Fosol, line 141: It should be noted that this action will likely result in different report than what the user is viewing presently. This means that you will need to ensure the returned report contains the latest snapshot and the UI is updated. This is particular important before they export the results.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            if (report.To == null)
+            {
+                report.To = DateTime.UtcNow;
+            }
+            var mostRecentReport = this.Context.ProjectReports
+                .Where(r => r.ReportTypeId == ReportTypes.SPL)
+                .OrderByDescending(r => r.To)
+                .FirstOrDefault();
+            report.From = mostRecentReport?.To;
```
Fosol, line 138: Incorrect.  The `From` will be whatever the user has set (which could be `null`).  The `To` will be whatever is set in the `GenerateSnapshots(...)` function

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// </summary>
+        /// <param name="project"></param>
+        /// <exception cref="ArgumentNullException">Argument 'report' is required.</exception>
+        /// <exception cref="NotAuthorizedException">User does not have permission to add report.</exception>
+        /// <returns></returns>
+        public ProjectReport Add(ProjectReport report)
+        {
+            if (report.To == null)
+            {
+                report.To = DateTime.UtcNow;
```
Fosol, line 132: Based on requested changes to the `GenerateSnapshots(...)` function this would be removed and you would determine the value from the results of the function.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: You will also need to calculate variance.

backend/entities/ProjectReport.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <summary>
+        /// get/set - The oldest date of comparison snapshots used in the report.
+        /// </summary>
+        public DateTime? From { get; set; }
+
+        /// <summary>
+        /// get/set - The most recent date of comparison snapshots used in the report.
+        /// </summary>
+        public DateTime? To { get; set; }
```
Fosol, line 35: Shouldn't be nullable

backend/entities/ProjectReport.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        public ReportTypes ReportTypeId { get; set; }
+
+        #endregion
+
+        #region Constructors
+        /// <summary>
+        /// Create a new instance of a ProjectReport class.
+        /// </summary>
+        public ProjectReport() { }
```
Fosol, line 48: Isn't generally required unless you have other overloads for constructors.  For example if you want to create a constructor that helps with know which fields are required you could create an other constructor `ProjectReport(DateTime to, bool isFinal = false)`

backend/entities/ProjectSnapshot.cs
```diff
@@ -49,6 +60,12 @@ public class ProjectSnapshot : BaseEntity
         /// </summary>
         public decimal? NetProceeds { get; set; } // TODO: Move to metadata property.
 
+        /// <summary>
+        /// get/set - The net proceed baseline integrity
+        /// </summary>
+        [NotMapped]
+        public decimal? BaselineIntegrity { get; set; } // TODO: Move to metadata property.
```
Fosol, line 67: This isn't useful as far as I can determine as a `NotMapped` property.  If you are thinking we should store the `NetProceedsVariance` rather than recalculate each time, then it would make more sense to just add it to the table with in the DB.  It may actually make a lot of sense to store the value, but we'll need to think it through.  It would mean that a prior report could fetch the `To` snapshots and would know the variance for that moment in time without having go and fetch all the `From` snapshots and recalculating.  I kinda like adding it to the DB.

backend/entities/ReportTypes.cs
```diff
@@ -0,0 +1,7 @@
+namespace Pims.Dal.Entities
+{
+    public enum ReportTypes
+    {
+        SPL = 0,
```
Fosol, line 5: Add some documentation for intellisense.

Fosol (changes requested):
A number of changes and suggestions.

backend/api/Areas/Project/Controllers/ReportController.cs
```diff
@@ -0,0 +1,187 @@
...
+        #endregion
+
+        #region Endpoints
+        /// <summary>
+        /// Get the project report for the specified 'id'.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        [HttpGet("{id:int}")]
+        [HasPermission(Permissions.AdminProjects)]
```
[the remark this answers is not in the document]
devinleighsmith: FYI, I don't find it helpful to put a comment on every instance of the same issue. You can do this if it helps you to keep track of things in the review of course.

backend/api/Areas/Project/Controllers/ReportController.cs
```diff
@@ -0,0 +1,187 @@
...
+using Pims.Api.Areas.Project.Models.Report;
+using Pims.Api.Areas.Project.Models.Snapshot;
+using Pims.Api.Models;
+using Pims.Api.Policies;
+using Pims.Dal;
+using Pims.Dal.Security;
+using Swashbuckle.AspNetCore.Annotations;
+using Entity = Pims.Dal.Entities;
+
+namespace Pims.Api.Areas.Project.Controllers
```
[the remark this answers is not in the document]
devinleighsmith: I don't think the two have any overlap. the /Areas/Reports is all about generated excel/csv files in my mind.

backend/api/Areas/Project/Controllers/ReportController.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// Get all of the project snapshots for the specified report id.
+        /// </summary>
+        /// <returns></returns>
+        [HttpPost("snapshots/{reportId:int}")]
+        [HasPermission(Permissions.AdminProjects)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(ProjectSnapshotModel[]), 200)]
+        [ProducesResponseType(typeof(ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public IActionResult GetProjectReportSnapshots(int reportId, ProjectReportModel model)
```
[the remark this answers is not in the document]
devinleighsmith: this one is get by passed model, which allows you to pass an model that may or may not be saved in the DB. Used to show the user data based on changes to the report From dates that haven't actually been saved yet.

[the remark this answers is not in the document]
devinleighsmith: The other one is just get the snapshots based on a report id in the DB.

backend/api/Areas/Project/Models/Snapshot/ProjectReportModel.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// get/set - The oldest date of comparison snapshots used in the report.
+        /// </summary>
+        public DateTime? From { get; set; }
+
+        /// <summary>
+        /// get/set - The most recent date of comparison snapshots used in the report.
+        /// </summary>
+        public DateTime? To { get; set; }
```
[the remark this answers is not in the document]
devinleighsmith: I ended up doing this because I wanted the backend to generate the To value initially, but can go the other way if you like.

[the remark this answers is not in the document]
Fosol: Ah yes I see, good point.  It can be `null` for the model then.  One of those scenarios where the model would be different depending on whether it's a response or request.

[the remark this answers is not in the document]
devinleighsmith: The other issue with non-nullable dates is that if you don't pass in a date, .net core substitutes in a default time:
1/1/0001 12:00:00 AM
Explicitly checking for this is a lot messier than checking for null in my opinion.

backend/api/Areas/Project/Models/Snapshot/ProjectSnapshotModel.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// get/set - The project corresponding to this snapshot.
+        /// </summary>
+        public ProjectModel Project { get; set; }
+
+        /// <summary>
+        /// get/set - The date when this snapshot's data was taken from the project.
+        /// </summary>
+        public DateTime? SnapshotOn { get; set; }
```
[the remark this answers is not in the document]
devinleighsmith: See above, I don't like how .net core provides 1/1/0001 12:00:00 AM when a date isn't provided.

backend/api/Areas/Project/Controllers/ReportController.cs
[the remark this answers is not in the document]
Fosol: I thought about this, but felt that if I was specific it may either imply only this one was incorrect.  I probably should have just made a comment at the top saying change all.  But at the time I was reviewing each function independently.  I'll go back and summarize next time to keep it cleaner.  It does have the benefit of tracking each change though.

backend/dal/Migrations/20200925194028_v01.00.01.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Microsoft.EntityFrameworkCore.Migrations;
+using Microsoft.EntityFrameworkCore.Migrations;
```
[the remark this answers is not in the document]
devinleighsmith: yeah, something weird happened when I merged these, I'll see if I can clean this up.

backend/api/Areas/Project/Models/Snapshot/ProjectSnapshotModel.cs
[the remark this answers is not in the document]
Fosol: I agree with you.  The model due to the way we use it should be nullable.`

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+            {
+                var previousSnapshot = snapshots.Where(s => s.ProjectId == snapshot.ProjectId && s.Id != snapshot.Id).FirstOrDefault();
+                snapshot.BaselineIntegrity = (snapshot?.NetProceeds ?? 0) - (previousSnapshot?.NetProceeds ?? 0);
+            }
+
+            return snapshots;
+        }
+
+        /// <summary>
+        /// Get all project snapshots matching either the 'To' or 'From' dates of the passed report.
```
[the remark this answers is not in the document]
devinleighsmith: Is it? this calls function the above function.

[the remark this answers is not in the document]
Fosol: I think it should be
Get all project snapshots for the specified `reportId`.  Depending on whether you introduce saving the `Variance` to the database, it would only actually need to return the snapshots that match the `To` date.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: Not necessarily incorrect, but it does sound like how I have this working is different than what you had in mind. In the current version of the frontend this add function gets called the instant the user clicks the button. As a result there is no option to fill in a from date here.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <exception cref="ArgumentNullException">Argument 'report' is required.</exception>
+        /// <exception cref="NotAuthorizedException">User does not have permission to add report.</exception>
+        /// <returns></returns>
+        public ProjectReport Add(ProjectReport report)
+        {
+            if (report.To == null)
+            {
+                report.To = DateTime.UtcNow;
+            }
+            var mostRecentReport = this.Context.ProjectReports
```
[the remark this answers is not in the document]
devinleighsmith: Because of how the frontend is implemented, the from will never be populated during the add workflow. I can change this if necessary.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: variance is calculated on all of the get snapshot apis, so I don't think I need to add any additional variance calculations.

devinleighsmith: And yes, since the report and the snapshot data isn't tightly coupled I re-request the snapshot data whenever the report changes.

backend/dal/Migrations/20201020221329_v01.01.00.cs
```diff
@@ -0,0 +1,42 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<decimal>(
+                name: "Appraised",
+                table: "ProjectSnapshots",
+                type: "MONEY",
+                nullable: false,
+                defaultValue: 0m);
+
+            migrationBuilder.AlterColumn<DateTime>(
```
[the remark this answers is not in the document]
devinleighsmith: Yeah, this is weird, I'll figure out what happened here.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <exception cref="NotAuthorizedException">User does not have permission to edit report.</exception>
+        /// <returns></returns>
+        public ProjectReport Update(ProjectReport report)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+
+            var originalReport = this.Context.ProjectReports
+                .SingleOrDefault(p => p.Id == report.Id) ?? throw new KeyNotFoundException();
+
+            if (originalReport.To == null) throw new ArgumentNullException();
```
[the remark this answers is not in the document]
devinleighsmith: I'm going to resolve this as I think this was covered by other discussion.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
Fosol: Ideally the initial report generated would use the `From` from the most recent `IsFinal` report (or the most recent report if there is no final one).  You could determine that before clicking the `Add` button, as you would have a list of reports in memory.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: See comment about about getting a `From` before submitting.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: Hmm... maybe the way I've done things is confusing.

So first of all I don't persist the variance calculations (I'm assuming you are referring to Baseline Integrity) in the DB at all. Baseline integrity is calculated based on the snapshot values when they are fetched from the backend.

As a result, there should never be a time that I need to update Baseline Integrity when updating a ProjectReport.

This does mean that I need to re-fetch the snapshots when the report changes, but since they are decoupled I think I would need to do that in either case.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
Fosol: I don't think we want the entity `To` to be nullable.  Just the models used by the API correct?

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// </summary>
+        /// <param name="project"></param>
+        /// <exception cref="NotAuthorizedException">User does not have permission to get refreshed snapshots.</exception>
+        /// <returns></returns>
+        public IEnumerable<ProjectSnapshot> Refresh(int reportId)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+            var report = Get(reportId);
+            var currentSnapshots = this.Context.ProjectSnapshots
+                .Include(s => s.Project).ThenInclude(p => p.Agency)
```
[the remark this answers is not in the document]
devinleighsmith: This gives me a human readable agency and status in the frontend. Other option would be to use the agency and status lookups on the frontend to get the human readable values from the agencyId and the statusId.

Went with this as the code is cleaner, but it does add processing overhead to the db. Although at least the joins here are small.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// Generate an updated list of snapshots based on current project data.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <exception cref="NotAuthorizedException">User does not have permission to get refreshed snapshots.</exception>
+        /// <returns></returns>
+        public IEnumerable<ProjectSnapshot> Refresh(int reportId)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+            var report = Get(reportId);
+            var currentSnapshots = this.Context.ProjectSnapshots
```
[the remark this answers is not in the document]
devinleighsmith: Agreed, ran into this yesterday.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: Variance = Baseline Integrity.  Which is the difference between the Net Proceeds from one snapshot to the next.
I'm wondering based on other comments I've added if it would be wise to store the variance in the DB.
This function will add a bunch of new snapshots to the database for this moment in time.

Not to confuse the subject, but I wonder if we're polluting the database with snapshots for projects when those values haven't changed.  Maybe another conversation for another time.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <param name="to"></param>
+        /// <returns></returns>
+        private IEnumerable<ProjectSnapshot> GenerateSnapshots(DateTime to)
+        {
+            var splProjects = this.Context.Projects
+                .Include(p => p.Agency)
+                .Include(p => p.Status)
+                .Where(p => p.Workflow.Code == "SPL") ?? throw new KeyNotFoundException();
+
+            IEnumerable<ProjectSnapshot> projectSnapshots = new ProjectSnapshot[0];
```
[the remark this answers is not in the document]
devinleighsmith: It does work, however I agree that this looks confusing. I'll change it.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+
+        /// <summary>
+        /// Generate snapshots for all SPL projects using the passed 'to' date as the snapshot date.
+        /// </summary>
+        /// <param name="to"></param>
+        /// <returns></returns>
+        private IEnumerable<ProjectSnapshot> GenerateSnapshots(DateTime to)
+        {
+            var splProjects = this.Context.Projects
+                .Include(p => p.Agency)
```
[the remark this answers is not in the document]
devinleighsmith: Partially, if this function is called as part of a report update, this isn't required. However, it is required if called as part of the refresh api.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+            var report = Get(reportId);
+            var currentSnapshots = this.Context.ProjectSnapshots
+                .Include(s => s.Project).ThenInclude(p => p.Agency)
+                .Include(s => s.Project).ThenInclude(p => p.Status)
+                .Where(s => s.SnapshotOn == report.To)
+                .ToArray();
+
+            var snapshots = GenerateSnapshots(DateTime.UtcNow).Concat(currentSnapshots);
```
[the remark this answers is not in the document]
devinleighsmith: I assumed that we wanted Baseline Integrity values for all of the snapshots not just the the To: values. I guess this makes sense as otherwise the From: dates would just show the same value negated.

backend/entities/ProjectSnapshot.cs
devinleighsmith: I made this a NotMapped property so that I can calculate and return the NetProceedsVariance/BaselineIntegrity at the DAL level. I decided not to store it as I was going to need to run calculations on snapshots that aren't stored in 2 different scenarios:
1) Refreshes
2) Getting temporary snapshot data when the report **From** changes.
 I could add this at the API level instead but I don't really like doing calculations at the model level. Could also do this calculation on the frontend which I think has similar +/-

backend/entities/ReportTypes.cs
devinleighsmith: Something I've been wondering, is there value in adding comments like this?

/// PropertyTypes enum, provides the valid property types that can be used.
or this:
/// TaskModel class, provides a model to represent a task.

To me these comments are redundant, and functionally contain the same information as the actual variable name and type.

So instead of writing something like this:
// ReportTypes enum, provides the valid report types that can be used.
I guess I would rather provide something like this:
// ReportTypes enum, provides the valid project report types that can be used in the ProjectReportService.
but now I've introduced a comment that is tied to the enums current use, if someone else starts using it or the ProjectReportService stops using it the comment is inaccurate and at high risk of being missed for updates.

I don't know if I actually have a point here, maybe this can be something to discuss in knowledge sharing.

Fosol: It's mixed bagged.  Generally a comment should provide a little more information than just repeat the name.  However where the value is more relevant are auto documentation tools (i.e. Swagger).  They will include the information in their output. Without it, it is rather ugly.

backend/entities/ProjectSnapshot.cs
Fosol: I think your reasoning is sound.  Generally it isn't good to populate entity classes with NotMapped properties, unless those properties actually do the logic (i.e. `get { return a+b; }`).
However I'm wondering if we should make it a DB property.  If we store the variance, it doesn't need to be calculated every time we load the report.  Technically it should also be the same value every time that report is requested anyways.  Your thoughts?

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: Ok, this may be something else that I misinterpreted from your design. Based on this:
https://preview.uxpin.com/bcfcebd6d46b70cc4030fc7ec0b363b72d9d2d99#/pages/133385268/simulate/sitemap
I assumed that I always need to return the To: and From: snapshots for every matching project. This results in two snapshots per project, each on a separate row, as in the linked design.

I can easily change this to only return the To: snapshots. I just thought some of the other differences might be useful as well. IE, if the netbook changed between snapshots.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: My intention was to actually have the backend do what you are recommending in your comment. Does that work?

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
Fosol: The UI only needs the `To` snapshots + variances.
However depending on whether you recalculate each time.  If you recalculate you would need to load the `From` snapshots before returning the results and doing the variance math.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: maybe something like this:
```
var mostRecentReport = this.Context.ProjectReports
                .Where(r => r.ReportTypeId == ReportTypes.SPL)
                .OrderByDescending(r => r.IsFinal)
                .ThenByDescending(r => r.To)
                .FirstOrDefault();
```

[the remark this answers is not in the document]
Fosol: I think it does make sense for the UI to be able to submit the `From` value.  As it would increase the flexibility.  But allow the backend to figure it out if no `From` is provided.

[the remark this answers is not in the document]
Fosol: That LINQ looks correct.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: if only the entity is null, you end up with the same problem as you do when the model is null, which is that you won't get an error from a null/not defined To: value, instead you will get a Date value of: 1/1/0001 12:00:00 AM.

In my mind, the entire point of making a DateTime not nullable is that you should get an error if the API caller or automapper doesn't provide a value.

I actually think the alternative is much worse, because before I was using a nullable datetime I ended up with a whole bunch of 1/1/0001 12:00:00 AM saved to the DB.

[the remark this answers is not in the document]
Fosol: There is a relationship between the model and the entity, but in the DB the `To` value should never be `null`.  Additionally nullable values in a DB are not performant, nor are they considered good practice.  Especially when the value shouldn't be nullable in any case.

The only time I can see the `To` value being null is when a user makes an AJAX request and they are specifying the `To` because the backend is going to.  But the reality is the backend will always set the `To` to the current moment in time, never null.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: Thinking about this more, I actually think adding the variance to the DB is not a good idea unless we create a join table of some kind.

This is my personal definition of a Project Snapshot: **A snapshot of project financial values at a given time.**

Based on the above, putting baseline integrity directly on a snapshot doesn't make sense, because a snapshot intuitively should not directly depend on another snapshot. Even more so in this case, as the only way to figure out where the baseline integrity came from is to look up the matching snapshot in the ProjectReports table which doesn't even have a foreign key linkage.

Something that could make sense is to create another table, such as SnapshotDiff. This could just be join table between two snapshots, or we could actually store the results of calculations in it. I would probably lean towards the first option.

We could then get associate a list of snapshotDiffs to a report which would greatly simplify the frontend logic.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// <returns></returns>
+        public IEnumerable<ProjectSnapshot> Refresh(int reportId)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
+            var report = Get(reportId);
+            var currentSnapshots = this.Context.ProjectSnapshots
+                .Include(s => s.Project).ThenInclude(p => p.Agency)
+                .Include(s => s.Project).ThenInclude(p => p.Status)
+                .Where(s => s.SnapshotOn == report.To)
+                .ToArray();
```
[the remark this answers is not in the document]
devinleighsmith: I assumed looping through a returned list of values would be faster then a DB lookup, but I'll do this since the requested change is small.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: Yes it can go either way.  I originally designed it to be as you said, a snapshot for a moment in time.  However when I saw you added the `NotMapped` property it made me think we might have another option.

There are some pros to making it in the DB.
- performance: we don't need to recalculate each time we look at the report
- consistency: the variance should never change for any given report when fetching data
- network: less data transferred on the wire each time
- because we don't have a hard link (foreign key relationship) with the snapshots it means those snapshots could be deleted and then a report would become invalid.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            var splProjects = this.Context.Projects
+                .Include(p => p.Agency)
+                .Include(p => p.Status)
+                .Where(p => p.Workflow.Code == "SPL") ?? throw new KeyNotFoundException();
+
+            IEnumerable<ProjectSnapshot> projectSnapshots = new ProjectSnapshot[0];
+            foreach (Project project in splProjects)
+            {
+                this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);
```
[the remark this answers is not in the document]
devinleighsmith: oof

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: Does that mean that you would be fine if I left the entity as-is if I added this to the DB configuration?

```
builder.Property(m => m.From).HasColumnType("DATETIME2").IsRequired();
builder.Property(m => m.To).HasColumnType("DATETIME2").IsRequired();
```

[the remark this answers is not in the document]
Fosol: Interesting solution.  Does it correctly handle the table creation?  I suppose if it does it would satisfy the majority of the issues.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: A few comments:

**performance: we don't need to recalculate each time we look at the report.**
Not exactly, we'll still need to calculate this each time we look at unsaved report data (when refreshing data or when changing the from date). However, I agree that making the calculation occur during save/update would result in fewer calculations overall.
**consistency: the variance should never change for any given report when fetching data**
I'm not seeing any increase in consistency. The call to get snapshots occurs completely independently of this call as far as the backend is concerned. Either way the frontend needs to call the api's update, wait for a success, and then call the getsnapshots method.
**network: less data transferred on the wire each time**
I think the opposite. The frontend will send/receive the same amount of data to the API with this change. However, the DB will be required to send/receive the result of the variance calculation to the backend which is net new.
**because we don't have a hard link (foreign key relationship) with the snapshots it means those snapshots could be deleted and then a report would become invalid. Having the variances in the DB removes this issue.**
This is true, although I don't think we have a way for users to delete snapshots currently so more of an issue for future changes in the app.

devinleighsmith: Also the delete method orphans all of its snapshots as well. Could update to selectively delete if no other snapshot is using its To: snapshots.

devinleighsmith: There is also no hard enforcement on the From: value actually pointing to anything here. Theoretically the frontend could just send in a random date that doesn't correspond to any other reports To: date. Should we be enforcing that as well?

backend/entities/ProjectSnapshot.cs
devinleighsmith: We continued this conversation in the ProjectReportService update call. Closing this out as a duplicate.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: So here was my reasoning for this:
Some of the callers of this function are actually saving a report to the DB. In order for the snapshots generated for that report to actually be connected, the To value needs to be identical (Note that this function is a private, internal helper function, not a service function).

I think every time you call DateTime.UtcNow you will get a new, slightly progressed value. This would therefore be a problem if the report was saved with an earlier To: value using DateTime.UtcNow, and then updated with a slightly progressed To: value obtained by calling DateTime.UtcNow again.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: I did think about that and forgot to add it in the requirements.  An invalid `From` should result in an error returned.  Our `From` dates must match an existing snapshot.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: This answer: https://github.com/bcgov/PIMS/pull/561/files/7986bf90439eb03cb11569b09887dcf430130f9b#r509670817
is also relevant for this question.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: Deleting a report doesn't technically orphan the snapshots, as they aren't hard linked.  Which is a good thing too, as we would like to keep historical snapshots of projects.  However it does highlight a design issue where a user wouldn't know there was a snapshot for a given date if there isn't a report to inform them.  Probably a problem we can look into later.

Fosol: performance - if it's saved in the DB we only calculate it once for the report.  The user would have to regenerate it to recalc.
consistency - more to do with the fact that there is no hard link from one snapshot to the next, which means historical data may be purged and now the snapshot can't calculate the original variance.
network - the network is within OpenShift too.  Every byte is cpu/power.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: You wouldn't call `DateTime.UtcNow` multiple times.  Just once at the beginning of the function.  Having it as a parameter implies the caller decides.  Which would mean that the data should reflect that specified `to` date and time.  This function doesn't do that.  We grab the most recent project information and generate new snapshots.

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: I think my only actual qualm with storing the variance is intended use. This isn't really a valid concern since to my knowledge we aren't actually using Snapshots anywhere. We can always rename ProjectSnapshot to ProjectReportSnapshot if it becomes obvious that these snapshots will only be used for reporting.

devinleighsmith: Or we create a new table with variance calculations that is linked to the snapshot table.

Fosol: If we create a separate table are we essentially create a single record for each snapshot?  While that would be normalized, it would be terrible for performance.  So I'd either not store it in the DB, or I'd just use the snapshots.  Your call, pros/cons either way.

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
devinleighsmith: image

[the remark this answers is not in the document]
devinleighsmith: image

backend/dal/Services/Concrete/ProjectReportService.cs
devinleighsmith: I think what you are asking for is for GenerateSnapshots to be turned into this:

```
private IEnumerable<ProjectSnapshot> GenerateSnapshots()
        {
            var to = DateTime.UtcNow;
            var splProjects = this.Context.Projects
                .Include(p => p.Agency)
                .Include(p => p.Status)
                .Where(p => p.Workflow.Code == "SPL") ?? throw new KeyNotFoundException();

            IEnumerable<ProjectSnapshot> projectSnapshots = new ProjectSnapshot[0];
            foreach (Project project in splProjects)
            {
                this.User.ThrowIfNotAuthorized(Permissions.AdminProjects);

                ProjectSnapshot snapshot = new ProjectSnapshot(project);
                snapshot.SnapshotOn = to;

                projectSnapshots = projectSnapshots.Append(snapshot);
            }
            return projectSnapshots;
        }
```
If the above is what you were thinking, I debugged this and it does not work.
Here is an example:
- The API calls the Add function. The add function saves the current DateTime.UtcNow to a local variable. (debugged value: 23:27:42.5129541)
- The add function creates a new report using the above DateTime.UtcNow() value.
- The add function calls the above GenerateSnapshots function. 
- The GenerateSnapshots saves another, slightly progressed value of DateTime.UtcNow (debugged value: 23:27:44.6235520)
- The GenerateSnapshots function generates all of the snapshots using the same DateTime.UtcNow value from the previous step.
- Now we have a report with a To value of 23:27:42.5129541 and a whole bunch of snapshots with a snapshotOn value of 23:27:44.6235520. Since the To value doesn't match the snapshotOn value, no data is returned if this report is requested.

Now interestingly, when I tested this without debugging I had this pass a few times. Apparently this is because the UtcNow function is only accurate to a few ms. As a result its possible that two calls happening in quick succession could return the same ms value. However, I wouldn't want to depend on that.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,258 @@
...
+
+
+        /// <summary>
+        /// Generate snapshots for all SPL projects using the passed 'to' date as the snapshot date.
+        /// </summary>
+        /// <param name="to"></param>
+        /// <returns></returns>
+        private IEnumerable<ProjectSnapshot> GenerateSnapshots(DateTime to)
+        {
+            var splProjects = this.Context.Projects
```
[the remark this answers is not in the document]
devinleighsmith: I'd argue the above is not needed here. This is a utility function not a service call. Services should be responsible for enforcing permissions. 

If I were to enforce this check here all of the service functions currently calling GenerateSnapshots would end up making the same check twice.

devinleighsmith:
FYI I regenerated the migration which seemed to fix the issues we were seeing.

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: The date used needs to be identical everywhere.  If the `GenerateSnapshots()` is setting the datetime then it needs to be used by the calling code.  This is where it's a little ugly and I don't like my simple solution - The calling code could retrieve the datetime from the first snapshot returned.  This highlights something not great about the implementation I'm suggesting though.  However if you pass the date, then in theory it should be used as part of the query (although technically we can do whatever we want).

backend/dal/Services/Concrete/ProjectReportService.cs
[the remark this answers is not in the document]
Fosol: I agree, didn't notice it was private.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -0,0 +1,285 @@
...
+                    .Where(s => s.SnapshotOn == from)
+                    .ToDictionary(s => (int?)s.ProjectId, s => s);
+                if (fromSnapshots.Count() == 0) throw new KeyNotFoundException("If specified report FROM value must return valid snapshots");
+            }
+
+            IEnumerable<ProjectSnapshot> projectSnapshots = new HashSet<ProjectSnapshot>();
+            
+            foreach (Project project in splProjects)
+            {
+                JsonConvert.PopulateObject(project.Metadata ?? "{}", project);
```
devinleighsmith, line 273: Not ideal, but until the backend is refactored to not require newtonsoft I still need to do this.

goemen (approved):

backend/api/Areas/Project/Controllers/ReportController.cs
```diff
@@ -0,0 +1,186 @@
...
+        [ProducesResponseType(typeof(ErrorResponseModel), 400)]
+        [SwaggerOperation(Tags = new[] { "project" })]
+        public IActionResult GetProjectReports()
+        {
+            var report = _pimsService.ProjectReport.GetAll();
+            return new JsonResult(_mapper.Map<ProjectReportModel[]>(report));
+        }
+
+        /// <summary>
+        /// Get all of the project snapshots for the specified report id.
```
Fosol, line 82: Document the difference between this and the GET below.  Swagger users will not be able to tell the difference.

backend/dal/Configuration/ProjectReportConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+            builder.HasKey(m => m.Id);
+            builder.Property(m => m.Id).IsRequired();
+            builder.Property(m => m.Id).ValueGeneratedOnAdd();
+
+            builder.Property(m => m.Name).HasMaxLength(250).IsNullable();
+
+            builder.Property(m => m.From).HasColumnType("DATETIME2");
+            builder.Property(m => m.To).HasColumnType("DATETIME2").IsRequired();
+
+            builder.HasIndex(m => new { m.Id, m.To, m.From, m.IsFinal });
```
Fosol, line 27: Don't need to change, but you'll get better performance if the property order matches the most common query parameters.  If the majority of queries state `where IsFinal = 1` then it should be second property after the primary key.

backend/dal/Configuration/ProjectSnapshotConfiguration.cs
```diff
@@ -20,19 +21,23 @@ public override void Configure(EntityTypeBuilder<ProjectSnapshot> builder)
...
 
+            builder.Property(m => m.SnapshotOn).HasColumnType("DATETIME2");
+
             builder.Property(m => m.NetBook).HasColumnType("MONEY");
             builder.Property(m => m.Estimated).HasColumnType("MONEY");
             builder.Property(m => m.Assessed).HasColumnType("MONEY");
+            builder.Property(m => m.Appraised).HasColumnType("MONEY");
             builder.Property(m => m.SalesCost).HasColumnType("MONEY");
             builder.Property(m => m.NetProceeds).HasColumnType("MONEY");
+            builder.Property(m => m.BaselineIntegrity).HasColumnType("MONEY");
```
Fosol, line 32: You decided to put the variance in the db.

backend/dal/Security/Permissions.cs
```diff
@@ -64,6 +64,12 @@ public enum Permissions
         ProjectEdit = 32,
 
         [Display(GroupName = "project", Name = "project-delete", Description = "Can delete projects.")]
-        ProjectDelete = 33
+        ProjectDelete = 33,
+
+        [Display(GroupName = "project", Name = "reports-view", Description = "Can view project reports.")]
```
Fosol, line 69: You will need to create a new role **SRES Financial Manager**, add these claims to it.  This will require updating the database.  We should be able to run the keycloak-sync tool to get them into keycloak afterwards.

backend/dal/Migrations/20201021213937_v01.01.00.cs
```diff
@@ -0,0 +1,116 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010100 : Migration
```
Fosol, line 6: You will need to inherit from `SeedMigration` so that you can update the db with the new role and claims.

Fosol (approved):

PR #568, opened 2020-10-23
Remove and correct duplicate agencies.

devinleighsmith:


devinleighsmith:
Due to what I can only describe as severe incompetence, I included some correctAgencies artifacts in a previous PR by mistake. We regret the error.

Fosol (approved):

PR #581, opened 2020-10-28
PA-635 add frontend tests for SPL Reports page and related functionality

devinleighsmith:


frontend/src/features/splReports/components/SplReportLayout.scss
```diff
@@ -61,7 +61,6 @@
 
   .tbody {
     max-height: calc(100vh - #{$footer-height} - #{$header-height} - #{$navbar-height} - 110px);
-    background-color: red;
```
devinleighsmith, line 64: How this got in here I'll never understand.

frontend/src/features/splReports/components/ReportControls.tsx
```diff
@@ -147,7 +146,12 @@ const ReportControls: React.FunctionComponent<IReportControlsProps> = ({
                 <Button
                   className="h-75 mr-auto"
                   type="submit"
-                  disabled={(!dirty && originalReport?.to === currentReport?.to) || !values.id}
+                  disabled={
+                    (!dirty &&
+                      originalReport?.to === currentReport?.to &&
+                      originalReport?.from === currentReport?.from) ||
```
devinleighsmith, line 152: from has the same issue as to.

frontend/src/components/common/form/Check.tsx
```diff
@@ -85,7 +85,6 @@ export const Check: React.FC<CheckProps> = ({
             required={required}
             disabled={disabled}
             custom={custom}
-            defaultChecked={!!checked}
```
devinleighsmith, line 88: Having both checked and defaultchecked throws a warning, controlled components should be using the checked prop so I've switched to that.

frontend/src/components/common/form/Input.tsx
```diff
@@ -94,30 +94,29 @@ export const Input: React.FC<InputProps> = ({
           {!!tooltip && <TooltipIcon toolTipId={`${field}-tooltip`} toolTip={tooltip} />}
         </Form.Label>
       )}
-      <span className={className}>
```
devinleighsmith, line 97: As far as I can see there is no reason for this span to be here, I've had a look through the pages affected by removing this and there doesn't appear to be a difference. Removing this:
simplifies the dom
makes findByLabelText work

frontend/src/features/splReports/containers/SplReportContainer.tsx
```diff
@@ -101,19 +95,6 @@ const SplReportContainer: React.FunctionComponent<ISplReportContainerProps> = ()
     setCurrentReport(report);
     setSnapshots(data);
   };
-  const exportReport = (report: IReport, accept: 'csv' | 'excel') => {
```
devinleighsmith, line 104: This makes more sense to be in the api hook.

frontend/src/features/splReports/hooks/useRouterReport.ts
```diff
@@ -34,17 +34,16 @@ export const useRouterReport = ({
...
 
   //When this hook loads, override the value of the filter with the search params. Should run once as originalSearch should never change.
   useDeepCompareEffect(() => {
-    if (reports?.length) {
+    if (reports?.length && currentReport?.id === undefined) {
       const filterFromParams = paramsToObject(originalSearch);
       if (filterFromParams.reportId) {
         const report = _.find(reports, { id: +filterFromParams.reportId });
         setCurrentReport(report ?? reports[0]);
-        setOriginalSearch('');
```
devinleighsmith, line 47: setting original search like this was causing an unnecessary call of this useEffect. The change has the same effect of only setting the report if unset, but without this extra evaluation.

devinleighsmith:
You may want to skip that majority of the snapshots on the code review.

jdtoombs (approved):

frontend/package.json
```diff
@@ -69,6 +69,7 @@
     "@types/lodash": "4.14.149",
     "@types/mui-datatables": "2.13.7",
     "@types/node": "12.12.17",
+    "@types/pretty": "^2.0.0",
```
Fosol, line 72: Should this be a specific version

Fosol (approved):
That should improve our coverage

frontend/package.json
devinleighsmith: It should be, thanks.

devinleighsmith:
snyk died for some reason. I'll force push at some point to force it to run again.

PR #587, opened 2020-11-03
pa-1170 Updating Parcel and Building Relationship

Fosol:
## Description

Changing the relationship of the **Parcel** and **Building** to support more complex relationships (i.e. many-to-many).  This will allow with future enhancements the ability to add buildings without parcels to inventory.  It will also allow for owning buildings without owning parcels.

[Story](https://pimsteam.atlassian.net/browse/PA-1170)

### Additional Enhancements

- **City** and **Municipality** have been merged into **Administrative Area** (or Location).  This requires additional effort in this sprint, but for now essentially city has been replaced by location.
- SQL is now using GIS data types and we have enhanced functionality for performing queries related to GIS.  Read more [here](https://nettopologysuite.github.io/NetTopologySuite/api/NetTopologySuite.html).  This means `Property.Latitude` and `Property.Longitude` have been replaced with `Property.Location`.
- NuGet packages have been updated across the board

### Known Issue

- The map search appears to not return all properties.  It looks like the bounding area is not correctly calculated in the query and it requires the user to move the map to see additional properties.  Will need to debug.
- The Latitude and Longitude values do not have as many decimal places for some reason with the new GIS features.  I am not sure if this is an issue.

### To Do

- **Administrative Areas** need to be populated from another source (Data BC Warehouse).  Also need a way to keep them up-to-date
- Need to test the whole application to ensure it is still working
- The frontend is still essentially bound to the `Building.Parcel` relationship.  A future sprint will be required to change this.


backend/api/Areas/Property/Controllers/BuildingController.cs
```diff
@@ -5,9 +5,9 @@
 using Pims.Dal;
 using Pims.Dal.Security;
 using Swashbuckle.AspNetCore.Annotations;
-using Model = Pims.Api.Models.Building;
+using Model = Pims.Api.Areas.Property.Models.Building;
```
Fosol, line 8: Moved models and maps to the correct location.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -234,84 +232,6 @@ private Entity.Agency GetOrCreateAgency(Model.ImportPropertyModel property)
...
-        /// <summary>
-        /// Get or create a new city for the specified code.
-        /// This is a recursive function so that it can generate a new city with a unique code.
-        /// It has a limitation of which the code must be 4 characters and therefore it can only handle 9 variations (i.e. BRe1, BRe2...)
-        /// </summary>
-        /// <param name="code"></param>
-        /// <param name="name"></param>
-        /// <param name="attempts"></param>
-        /// <returns></returns>
-        private Entity.City GetOrCreateCity(string code, string name, int attempts = 1)
```
Fosol, line 246: We don't want to create cities anymore.  We want administrative areas to come from BC Data Warehouse.  This functionality will need to be added.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -358,21 +279,22 @@ private Entity.Parcel AddUpdateParcel(Model.ImportPropertyModel property, int pi
                 p_e.ClassificationId = propClassification.Id;
                 p_e.Classification = propClassification;
 
-                var city = GetOrCreateCity(property.City);
+                // TODO: Handle this issue more gracefully.
+                var city = _pimsAdminService.AdministrativeArea.Get(property.City.ConvertToUTF8()) ?? throw new InvalidOperationException($"Administrative area '{property.City}' does not exist in the datasource.");
```
Fosol, line 283: This will cause developer issues when they use the `make db-seed` command because their data will have cities that don't exist in the `Administrative Areas`.  Just temporarily update your seed data to use only cities we have.

backend/api/Pims.Api.csproj
```diff
@@ -15,39 +15,39 @@
     <None Remove="healthchecksdb" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="AspNetCore.HealthChecks.UI.Client" Version="3.1.1" />
+    <PackageReference Include="AspNetCore.HealthChecks.UI.Client" Version="3.1.2" />
```
Fosol, line 18: Updated NuGet packages

backend/core/Comparers/DeepPropertyCompare.cs
```diff
@@ -24,10 +24,11 @@ public new bool Equals([AllowNull] object x, [AllowNull] object y)
         {
             // Only root objects cannot be null.  Child objects that are null will be equal.
             if (x == null || y == null || GetHashCode(x) != GetHashCode(y) || x.GetType() != y.GetType()) return false;
+            if (x == y) return true;
 
             var type = x.GetType();
             var children = type.GetCachedProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.GetProperty).Where(p => !p.PropertyType.IsValueType && p.PropertyType != typeof(string) && !p.PropertyType.IsEnumerable());
-            foreach (var p in children)
+            foreach (var p in children.Where(p => !p.GetIndexParameters().Any()))
```
Fosol, line 31: The new GIS objects were causing issues when doing equal comparison tests.  This isn't an ideal solution, but appears to work for our use-case.

backend/core/Extensions/StringExtensions.cs
```diff
@@ -78,5 +78,19 @@ public static HttpMethod GetHttpMethod(this string method)
...
         }
+
+        /// <summary>
+        /// Convert the specified 'value' from default to UTF8 encoding.
+        /// Replace linebreaks with spaces.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <param name="replaceLineBreaks"></param>
+        /// <returns></returns>
+        public static string ConvertToUTF8(this string value, bool replaceLineBreaks = true)
```
Fosol, line 89: Without this the db was getting populated with \n newline characters.  Which causes issues when searching.

backend/dal/Configuration/AddressConfiguration.cs
```diff
@@ -26,14 +26,13 @@ public override void Configure(EntityTypeBuilder<Address> builder)
             builder.Property(m => m.ProvinceId).HasMaxLength(2);
             builder.Property(m => m.ProvinceId).IsRequired();
 
-            builder.HasOne(m => m.City).WithMany().HasForeignKey(m => m.CityId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.Province).WithMany().HasForeignKey(m => m.ProvinceId).OnDelete(DeleteBehavior.ClientSetNull);
 
-            builder.HasIndex(m => new { m.Id, m.CityId }).IncludeProperties(m => new { m.Address1, m.Address2 });
+            builder.HasIndex(m => new { m.Id, m.AdministrativeArea }).IncludeProperties(m => new { m.Address1, m.Address2 });
```
Fosol, line 31: AdministrativeArea is just a string.  Theoretically the address can have any value, but we will control the value by validating it against a list.

backend/dal/Configuration/AdministrativeAreaConfiguration.cs
```diff
@@ -5,14 +5,14 @@
...
     ///</summary>
-    public class CityConfiguration : CodeEntityConfiguration<City, int>
+    public class AdministrativeAreaConfiguration : LookupEntityConfiguration<AdministrativeArea, int>
     {
         #region Methods
-        public override void Configure(EntityTypeBuilder<City> builder)
+        public override void Configure(EntityTypeBuilder<AdministrativeArea> builder)
         {
-            builder.ToTable("Cities");
+            builder.ToTable("AdministrativeAreas");
```
Fosol, line 15: Replaced Cities.  This may be a temporary table just to keep us going until we have a better solution for this.  Although perhaps we'll keep it and it'll be our cached list to validate against.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -14,18 +15,18 @@ namespace Pims.Dal.Helpers.Extensions
...
+        /// - No two buildings should have the same name on a parcel.
         /// </summary>
-        /// <param name="buildings"></param>
+        /// <param name="context"></param>
+        /// <param name="parcel"></param>
         /// <param name="building"></param>
-        /// <exception type="DbUpdateException">The PID and PIN must be unique.</exception>
-        public static void ThrowIfNotUnique(this DbSet<Entity.Building> buildings, Entity.Building building)
+        /// <exception type="DbUpdateException">The name within a parcel should be unique.</exception>
+        public static void ThrowIfNotUnique(this PimsContext context, Entity.Parcel parcel, Entity.Building building)
```
Fosol, line 25: This could possibly lead to confusion due to the fact buildings can now exist on multiple parcels.  A unique name may not be required anymore, but it generally doesn't make sense to have multiple buildings with the same name.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -57,13 +57,20 @@ public static IQueryable<Entity.Parcel> GenerateQuery(this PimsContext context,
...
-                query = query.Where(p =>
-                    p.Latitude != 0 &&
-                    p.Longitude != 0 &&
-                    p.Latitude <= filter.NELatitude &&
-                    p.Latitude >= filter.SWLatitude &&
-                    p.Longitude <= filter.NELongitude &&
-                    p.Longitude >= filter.SWLongitude);
+            {
+                var pfactory = new NetTopologySuite.Geometries.GeometryFactory();
+                var ring = new NetTopologySuite.Geometries.LinearRing(
```
Fosol, line 62: Need to make a helper extension function to do this.  Essentially it provides a way to query the db for all properties that fall within the polygon.  However I think this isn't working correctly, as all properties are not being returned when they should be.

backend/dal/Migrations/v01.02.00/Up/PostUp/01-ViewProperties.sql
```diff
@@ -0,0 +1,219 @@
...
+OUTER APPLY (
+    SELECT TOP 1
+        [Value]
+        , [FiscalYear]
+    FROM dbo.[ParcelFiscals]
+    WHERE [ParcelId] = p.[Id]
+        AND [Key] = 0 -- NetBook
+    ORDER BY [FiscalYear] DESC
+) AS fn
+UNION ALL
```
Fosol, line 110: Had to switch to a `UNION ALL` because spatial data types aren't allowed in an normal `UNION`.  I don't know if this makes this perform poorly or not.

backend/dal/Migrations/v01.02.00/Up/PostUp/03-SpatialIndex.sql
```diff
@@ -0,0 +1,4 @@
+PRINT 'Create Spatial Indexes'
+
+CREATE SPATIAL INDEX IX_Parcels_SpatialLocation ON Parcels(Location) using geography_auto_grid
```
Fosol, line 3: Spatial indexes can't be created by EF, so we have to manually create them

backend/api/Areas/Admin/Mapping/Parcel/BuildingMap.cs
```diff
@@ -42,12 +43,10 @@ public void Register(TypeAdapterConfig config)
             config.NewConfig<Model.BuildingModel, Entity.Building>()
                 .IgnoreNonMapped(true)
                 .Map(dest => dest.Id, src => src.Id)
-                .Map(dest => dest.LocalId, src => src.LocalId)
                 .Map(dest => dest.AgencyId, src => src.AgencyId)
                 .Map(dest => dest.Name, src => src.Name)
                 .Map(dest => dest.Description, src => src.Description)
-                .Map(dest => dest.Latitude, src => src.Latitude)
-                .Map(dest => dest.Longitude, src => src.Longitude)
+                .Map(dest => dest.Location, src => new NetTopologySuite.Geometries.Point(src.Longitude, src.Latitude) { SRID = 4326 })
```
devinleighsmith, line 49: what is this magic number?

backend/api/Areas/Project/Mapping/Dispose/BuildingMap.cs
```diff
@@ -11,48 +13,45 @@ public void Register(TypeAdapterConfig config)
...
-                .Map(dest => dest.LocalId, src => src.LocalId)
+                .Map(dest => dest.Parcels, src => new [] { new Entity.ParcelBuilding() { ParcelId = src.ParcelId, BuildingId = src.Id } })
                 .Map(dest => dest.ProjectNumber, src => src.ProjectNumber)
                 .Map(dest => dest.ClassificationId, src => src.ClassificationId)
                 .Map(dest => dest.Classification, src => src.Classification)
                 .Map(dest => dest.AgencyId, src => src.AgencyId)
                 .Map(dest => dest.Description, src => src.Description)
-                .Map(dest => dest.Latitude, src => src.Latitude)
-                .Map(dest => dest.Longitude, src => src.Longitude)
+                .Map(dest => dest.Location, src => new NetTopologySuite.Geometries.Point(src.Longitude, src.Latitude) { SRID = 4326 })
```
goemen, line 54: Not a deal breaker, but it be nice to create Geometry Extension that returns a Point object with SRID=4326 for reusability

goemen (approved):

goemen:
@Fosol can we run the Administration Areas sql script that I created as part of this story migration to populate the database?

backend/api/Pims.Api.csproj
```diff
@@ -15,39 +15,39 @@
...
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Version="2.0.3" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="3.1.5" />
-    <PackageReference Include="Microsoft.Extensions.Identity.Core" Version="3.1.5" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="3.1.9" />
+    <PackageReference Include="Microsoft.Extensions.Identity.Core" Version="3.1.9" />
     <PackageReference Include="Microsoft.Extensions.PlatformAbstractions" Version="1.1.0" />
-    <PackageReference Include="prometheus-net.AspNetCore" Version="3.6.0" />
+    <PackageReference Include="prometheus-net.AspNetCore" Version="4.0.0" />
```
devinleighsmith, line 44: This is a major version update, are you confident this won't brick anything?

backend/core/Comparers/DeepPropertyCompare.cs
devinleighsmith: I think I would prefer to explicitly choose to ignore problematic fields rather then silently ignore everything that is not indexed. I worry this approach will lead to false positives, as anyone who isn't familiar with the above may not realize that it is ignoring some fields.

backend/core/Extensions/StringExtensions.cs
devinleighsmith: So this extension gets called without us having to call it explicitly?

backend/dal/Configuration/BuildingConfiguration.cs
```diff
@@ -17,22 +17,18 @@ public override void Configure(EntityTypeBuilder<Building> builder)
...
             builder.Property(m => m.TransferLeaseOnSale).HasDefaultValue(false);
             builder.Property(m => m.OccupantName).HasMaxLength(100);
 
             builder.HasOne(m => m.Agency).WithMany(m => m.Buildings).HasForeignKey(m => m.AgencyId).OnDelete(DeleteBehavior.ClientSetNull);
-            builder.HasOne(m => m.Parcel).WithMany(m => m.Buildings).HasForeignKey(m => m.ParcelId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.BuildingConstructionType).WithMany().HasForeignKey(m => m.BuildingConstructionTypeId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.BuildingPredominateUse).WithMany().HasForeignKey(m => m.BuildingPredominateUseId).OnDelete(DeleteBehavior.ClientSetNull);
             builder.HasOne(m => m.BuildingOccupantType).WithMany().HasForeignKey(m => m.BuildingOccupantTypeId).OnDelete(DeleteBehavior.ClientSetNull);
 
-            builder.HasIndex(m => new { m.Id, m.Latitude, m.Longitude, m.IsSensitive, m.AgencyId, m.ParcelId, m.ClassificationId, m.AddressId, m.ProjectNumber, m.BuildingConstructionTypeId, m.BuildingPredominateUseId, m.BuildingOccupantTypeId, m.BuildingFloorCount, m.BuildingTenancy });
```
devinleighsmith, line 34: huh, weird that we used to have two indexes.

backend/dal/Configuration/PropertyConfiguration`.cs
```diff
@@ -14,10 +14,10 @@ public abstract class PropertyConfiguration<TBase> : BaseEntityConfiguration<TBa
         public override void Configure(EntityTypeBuilder<TBase> builder)
         {
             builder.Property(m => m.ProjectNumber).HasMaxLength(50);
-            builder.Property(m => m.Name).HasMaxLength(150);
+            builder.Property(m => m.Name).HasMaxLength(250);
             builder.Property(m => m.Description).HasMaxLength(2000);
-            builder.Property(m => m.Latitude).IsRequired();
-            builder.Property(m => m.Longitude).IsRequired();
+            builder.Property(m => m.Location).HasColumnType("GEOGRAPHY").IsRequired();
```
devinleighsmith, line 19: Should this be an enum/constant?

backend/api/Areas/Admin/Mapping/Parcel/BuildingMap.cs
Fosol: A spatial reference identifier (SRID) is a unique identifier associated with a specific coordinate system, tolerance, and resolution.  4326 is a number used for geographical coordinates.  I don't know more than this.  I'm going to add a commit to centralize the creation of points.

backend/api/Areas/Project/Mapping/Dispose/BuildingMap.cs
Fosol: I'll add a commit to clean this up.

backend/api/Pims.Api.csproj
Fosol: Testing so far has been promising.

backend/dal/Configuration/PropertyConfiguration`.cs
Fosol: No it's a string for a specific data type within MS SQL.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
devinleighsmith: Will have to consider whether or not to enforce this on the frontend as well.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -57,13 +58,20 @@ public static IQueryable<Entity.Building> GenerateQuery(this PimsContext context
...
-                    b.Longitude >= filter.SWLongitude);
+            {
+                var pfactory = new NetTopologySuite.Geometries.GeometryFactory();
+                var ring = new NetTopologySuite.Geometries.LinearRing(
+                    new[] {
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.NELatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.SWLongitude.Value, filter.NELatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.SWLongitude.Value, filter.SWLatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.SWLatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.NELatitude.Value)
```
devinleighsmith, line 69: isn't this coordinate the same as the first one?

devinleighsmith: is the last one to complete the shape?

backend/dal/Services/Admin/Concrete/BuildingService.cs
```diff
@@ -68,89 +68,86 @@ public Building Get(int id)
                 .Include(p => p.BuildingConstructionType)
                 .Include(p => p.BuildingPredominateUse)
                 .Include(p => p.BuildingOccupantType)
-                .Include(p => p.Address)
-                .Include(p => p.Address.City)
-                .Include(p => p.Address.Province)
-                .Include(p => p.Agency)
-                .Include(p => p.Agency.Parent)
+                .Include(p => p.Address).ThenInclude(a => a.Province)
```
devinleighsmith, line 71: So we don't return the admin area by default?

backend/dal/Services/Admin/Concrete/BuildingService.cs
```diff
@@ -68,89 +68,86 @@ public Building Get(int id)
...
-                .Include(p => p.Address)
-                .Include(p => p.Address.City)
-                .Include(p => p.Address.Province)
-                .Include(p => p.Agency)
-                .Include(p => p.Agency.Parent)
-                .AsNoTracking().SingleOrDefault(u => u.LocalId == localId) ?? throw new KeyNotFoundException();
+                .Include(p => p.Address).ThenInclude(a => a.Province)
+                .Include(p => p.Agency).ThenInclude(a => a.Parent)
+                .AsNoTracking()
+                .Where(b => b.Parcels.Any(pb => pb.Parcel.PID == pid) && (name == null || EF.Functions.Like($"{name}%", b.Name)));
```
devinleighsmith, line 94: do we want a startsWith or includes here?

backend/dal/Services/Admin/Concrete/ParcelService.cs
```diff
@@ -51,13 +51,20 @@ public Paged<Parcel> Get(ParcelFilter filter)
             var query = this.Context.Parcels.AsNoTracking();
```
devinleighsmith, line 51: I like that you are adding better variable names but keep in mind that it makes it much harder to review what actually changed.

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -187,7 +193,8 @@ public Building Update(Building building)
             // Only administrators can dispose a property.
             if (building.ClassificationId == 4 && !isAdmin) throw new NotAuthorizedException("Building classification cannot be changed to disposed.");
 
-            this.Context.Buildings.ThrowIfNotUnique(building);
+            // A building should have a unique name within the parcel it is located on.
+            existingBuilding.Parcels.ForEach(pb => this.Context.ThrowIfNotUnique(pb.Parcel, building));
```
devinleighsmith, line 197: Shouldn't this be thrown if the new building names aren't unique, not the existing building names?

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -354,42 +359,58 @@ public Parcel Update(Parcel parcel)
...
+                    var matchingParcelBuilding = parcel.Buildings.FirstOrDefault(pb => pb.BuildingId == parcelBuilding.BuildingId);
+                    if (matchingParcelBuilding == null)
                     {
-                        this.ThrowIfNotAllowedToUpdate(building, _options.Project);
-                        this.Context.Buildings.Remove(building);
+                        // TODO: This will delete the building from the datasource.  This isn't entirely ideal as it may be related to multiple parcels.
+                        this.ThrowIfNotAllowedToUpdate(matchingParcelBuilding.Building, _options.Project);
+
+                        var parcelBuildings = this.Context.ParcelBuildings.Where(pb => pb.BuildingId == parcelBuilding.BuildingId).ToArray();
+                        parcel.Buildings.Remove(matchingParcelBuilding);
```
devinleighsmith, line 371: Not crazy about the naming of the join table here, but there are bigger fish.

backend/entities/Address.cs
```diff
@@ -68,14 +63,14 @@ public class Address : BaseEntity
...
+        /// <param name="administrativeArea"></param>
         /// <param name="provinceId"></param>
         /// <param name="postal"></param>
-        public Address(string address1, string address2, int cityId, string provinceId, string postal)
+        public Address(string address1, string address2, string administrativeArea, string provinceId, string postal)
         {
             this.Address1 = address1;
             this.Address2 = address2;
-            this.CityId = cityId;
+            this.AdministrativeArea = administrativeArea ?? throw new ArgumentNullException(nameof(administrativeArea));
```
devinleighsmith, line 73: it seems weird to do the enforcement at this level no?

backend/entities/Parcel.cs
```diff
@@ -60,19 +55,19 @@ public class Parcel : Property
         /// get/set - A collection of buildings on this parcel.
         /// </summary>
         /// <typeparam name="Building"></typeparam>
-        public ICollection<Building> Buildings { get; private set; } = new List<Building>();
+        public ICollection<ParcelBuilding> Buildings { get; } = new List<ParcelBuilding>();
```
devinleighsmith, line 58: maybe ParcelBuildings?

backend/dal/Services/Concrete/BuildingService.cs
Fosol: I think that is what it is doing.

backend/tests/core/Entities/AddressHelper.cs
```diff
@@ -15,13 +15,13 @@ public static partial class EntityHelper
         /// <param name="id"></param>
         /// <param name="address1"></param>
         /// <param name="address2"></param>
-        /// <param name="city"></param>
+        /// <param name="adminsitrativeArea"></param>
```
devinleighsmith, line 18: minor typo

backend/dal/Services/Concrete/ParcelService.cs
Fosol: This is just a standard many-to-many naming conventions, Parcel -> ParcelBuilding -> Building

backend/entities/Address.cs
Fosol: A standard practice in constructors to provide immediate feedback on required arguments.  I should actually enforce more of them as it is weird it is the only one.

backend/entities/Parcel.cs
Fosol: Some would recommend that because of the object, but it would not be the normal naming convention for a relationship.  Due to .net core entity framework not automatically handling many-to-many relationships it requires us to manually perform them.  I still prefer the mental relationship naming convention which has been the practice for many years now with ORMs.  However I do understand it is odd since the object doesn't match 100%.

frontend/src/features/properties/components/forms/subforms/AddressForm.tsx
```diff
@@ -7,11 +7,12 @@ import { ILookupCode } from 'actions/lookupActions';
 import { ILookupCodeState } from 'reducers/lookupCodeReducer';
```
devinleighsmith, line 7: Thankfully this all looks simple enough to integrate with my changes.

devinleighsmith (approved):

frontend/src/features/properties/components/forms/subforms/AddressForm.tsx
Fosol: I attempted to essentially make one breaking change, which was the `AdministrativeArea` instead of `City` and `Municipality`. The frontend should continue working as it has been.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -57,13 +57,20 @@ public static IQueryable<Entity.Parcel> GenerateQuery(this PimsContext context,
...
+                    new[] {
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.NELatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.SWLongitude.Value, filter.NELatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.SWLongitude.Value, filter.SWLatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.SWLatitude.Value),
+                        new NetTopologySuite.Geometries.Coordinate(filter.NELongitude.Value, filter.NELatitude.Value)
+                    });
+                var poly = pfactory.CreatePolygon(ring);
+                poly.SRID = 4326;
+                query = query.Where(p => poly.Contains(p.Location));
```
goemen, line 72: I like this

backend/dal/Services/Concrete/ParcelService.cs
devinleighsmith: I meant the Parcel.Buildings.Buildings

PR #588, opened 2020-11-04
PA-106, 1203

jdtoombs:
Edit/Add/View Agencies 

Also quick bug fix for the view inventory page. Working on the other issues, but want to get this in for testing as it is functional.

Known issues:

- Need to change the way I assign parents to the corresponding agencies

- add typeahead component instead of custom autocomplete after grouping work completed

### DIFF
--- a/backend/dal/Services/Admin/Concrete/AgencyService.cs
+++ b/backend/dal/Services/Admin/Concrete/AgencyService.cs
@@ -5,6 +5,9 @@ using Pims.Dal.Entities;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using Pims.Dal.Entities.Models;
+using Pims.Dal.Helpers.Extensions;
+using Pims.Dal.Security;
 
 namespace Pims.Dal.Services.Admin
 {
@@ -28,6 +31,52 @@ namespace Pims.Dal.Services.Admin
         #endregion
 
         #region Methods
+        
+        /// <summary>
+        /// Get a page of agencies from the datasource.
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        /// <returns></returns>
+        public Paged<Agency> Get(int page, int quantity)
+        {
+            return Get(new AgencyFilter(page, quantity));
+        }
+
+        /// <summary>
+        /// Get a page of agencies from the datasource.
+        /// The filter will allow queries to search for the following property values; Name, Description, ParentId.
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public Paged<Agency> Get(AgencyFilter filter = null)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin);
+
+            var query = this.Context.Agencies.AsNoTracking();
+
+            if (filter != null)
+            {
+                if (filter.Page < 1) filter.Page = 1;
+                if (filter.Quantity < 1) filter.Quantity = 1;
+                if (filter.Quantity > 50) filter.Quantity = 50;
+                if (filter.Sort == null) filter.Sort = new string[] { };
+
+                if (!string.IsNullOrWhiteSpace(filter.Name))
+                    query = query.Where(a => EF.Functions.Like(a.Name, $"%{filter.Name}%"));
+                if (filter.IsDisabled != null)
+                    query = query.Where(a => a.IsDisabled == filter.IsDisabled);
+                if (filter.Id > 0)
+                    query = query.Where(a => a.Id == filter.Id);
+
+                if (filter.Sort.Any())
+                    query = query.OrderByProperty(filter.Sort);
+            }
+            var agencies = query.Skip((filter.Page - 1) * filter.Quantity).Take(filter.Quantity);
+            return new Paged<Agency>(agencies.ToArray(), filter.Page, filter.Quantity, query.Count());
+        }
+
+
         /// <summary>
         /// Get a page of agencies from the datasource.
         /// </summary>
--- a/backend/entities/Models/AgencyFilter.cs
+++ b/backend/entities/Models/AgencyFilter.cs
@@ -1,0 +1,112 @@
+using Pims.Core.Extensions;
+using System;
+using System.Collections.Generic;
+
+namespace Pims.Dal.Entities.Models
+{
+    /// <summary>
+    /// AgencyFilter class, provides a model for filtering agencies.
+    /// </summary>
+    public class AgencyFilter
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - The page number.
+        /// </summary>
+        /// <value></value>
+        public int Page { get; set; } = 1;
+
+        /// <summary>
+        /// get/set - The quantity to return in each page.
+        /// </summary>
+        /// <value></value>
+        public int Quantity { get; set; } = 10;
+
+        /// <summary>
+        /// get/set - The name of the agency.
+        /// </summary>
+        /// <value></value>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// get/set - The parent id of given agency.
+        /// </summary>
+        /// <value></value>
+        public int ParentId { get; set; }
+
+        /// <summary>
+        /// get/set - account status
+        /// </summary>
+        /// <value></value>
+        public bool? IsDisabled { get; set; }
+
+        /// <summary>
+        /// get/set - The agency ID
+        /// </summary>
+        /// <value></value>
+        public int? Id { get; set; }
+
+        /// <summary>
+        /// get/set - An array of sorting conditions (i.e. FirstName desc, LastName asc)
+        /// </summary>
+        /// <value></value>
+        public string[] Sort { get; set; }
+        #endregion
+
+        #region Constructors
+        /// <summary>
+        /// Creates a new instance of a AgencyFilter class.
+        /// </summary>
+        public AgencyFilter() { }
+
+        /// <summary>
+        /// Creates a new instance of a AgencyFilter class, initializes it with the specified arguments.
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        public AgencyFilter(int page, int quantity)
+        {
+            this.Page = page;
+            this.Quantity = quantity;
+        }
+
+        /// <summary>
+        /// Creates a new instance of a AgencyFilter class, initializes it with the specified arguments.
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        /// <param name="name"></param>
+        /// <param name="description"></param>
+        /// <param name="parentId"></param>
+        /// <param name="isDisabled"></param>
+        /// <param name="sort"></param>
+        /// <returns></returns>
+        public AgencyFilter(int page, int quantity, int id, String name, string description, int parentId, bool? isDisabled, string[] sort) : this(page, quantity)
+        {
+            this.Name = name;
+            this.ParentId = parentId;
+            this.IsDisabled = isDisabled;
+            this.Sort = sort;
+            this.Id = id;
+        }
+
+        /// <summary>
+        /// Creates a new instance of a AgencyFilter class, initializes it with the specified arguments.
+        /// Extracts the properties from the query string to generate the filter.
+        /// </summary>
+        /// <param name="query"></param>
+        public AgencyFilter(Dictionary<string, Microsoft.Extensions.Primitives.StringValues> query)
+        {
+            // We want case-insensitive query parameter properties.
+            var filter = new Dictionary<string, Microsoft.Extensions.Primitives.StringValues>(query, StringComparer.OrdinalIgnoreCase);
+            this.Page = filter.GetIntValue(nameof(this.Page), 1);
+            this.Quantity = filter.GetIntValue(nameof(this.Quantity), 10);
+            this.Name = filter.GetStringValue(nameof(this.Name));
+            this.ParentId = filter.GetIntValue(nameof(this.ParentId));
+            this.IsDisabled = filter.GetValue<bool?>(nameof(this.IsDisabled));
+            this.Id = filter.GetIntValue(nameof(this.Id));
+            this.Sort = filter.GetStringArrayValue(nameof(this.Sort));
+        }
+        #endregion
+    }
+}
--- a/frontend/src/actionCreators/agencyActionCreator.ts
+++ b/frontend/src/actionCreators/agencyActionCreator.ts
@@ -1,0 +1,97 @@
+import { showLoading, hideLoading } from 'react-redux-loading-bar';
+import { request, success, error } from 'actions/genericActions';
+import * as actionTypes from 'constants/actionTypes';
+import * as adminActions from 'actions/adminActions';
+import * as API from 'constants/API';
+import { ENVIRONMENT } from 'constants/environment';
+import CustomAxios, { LifecycleToasts } from 'customAxios';
+import { AxiosResponse, AxiosError } from 'axios';
+import { IAddAgency, IAgency, IAgencyDetail } from 'interfaces';
+import { handleAxiosResponse } from 'utils';
+import * as pimsToasts from 'constants/toasts';
+
+export const getAgenciesAction = (params: API.IPaginateParams) => (dispatch: Function) => {
+  dispatch(request(actionTypes.GET_AGENCIES));
+  dispatch(showLoading());
+  return CustomAxios()
+    .post(ENVIRONMENT.apiUrl + API.POST_AGENCIES(), params)
+    .then((response: AxiosResponse) => {
+      dispatch(success(actionTypes.GET_AGENCIES, response.status));
+      dispatch(adminActions.storeAgencies(response.data));
+      dispatch(hideLoading());
+    })
+    .catch((axiosError: AxiosError) =>
+      dispatch(error(actionTypes.GET_AGENCIES, axiosError?.response?.status, axiosError)),
+    )
+    .finally(() => dispatch(hideLoading()));
+};
+
+export const fetchAgencyDetail = (id: API.IAgencyDetailParams) => (dispatch: Function) => {
+  dispatch(request(actionTypes.GET_AGENCY_DETAILS));
+  dispatch(showLoading());
+  return CustomAxios()
+    .get(ENVIRONMENT.apiUrl + API.AGENCY_DETAIL(id))
+    .then((response: AxiosResponse) => {
+      dispatch(success(actionTypes.GET_AGENCY_DETAILS));
+      dispatch(adminActions.storeAgencyDetails(response.data));
+      dispatch(hideLoading());
+    })
+    .catch(() => dispatch(error(actionTypes.GET_AGENCY_DETAILS)))
+    .finally(() => dispatch(hideLoading()));
+};
+
+const agencyToasts: LifecycleToasts = {
+  loadingToast: pimsToasts.agency.AGENCY_UPDATING,
+  successToast: pimsToasts.agency.AGENCY_UPDATED,
+  errorToast: pimsToasts.agency.AGENCY_ERROR,
+};
+
+export const getUpdateAgencyAction = (
+  id: API.IAgencyDetailParams,
+  updatedAgency: IAgencyDetail,
+) => (dispatch: Function) => {
+  const axiosPromise = CustomAxios({ lifecycleToasts: agencyToasts })
+    .put(ENVIRONMENT.apiUrl + API.AGENCY_DETAIL(id), updatedAgency)
+    .then((response: AxiosResponse) => {
+      dispatch(adminActions.updateAgency(response.data));
+      return Promise.resolve(response);
+    });
+  return handleAxiosResponse(
+    dispatch,
+    actionTypes.PUT_AGENCY_DETAILS,
+    axiosPromise,
+  ).catch(() => {});
+};
+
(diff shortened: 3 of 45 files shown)


backend/entities/Models/AgencyFilter.cs
```diff
@@ -0,0 +1,120 @@
+using Pims.Core.Extensions;
+using System;
+using System.Collections.Generic;
+
+namespace Pims.Dal.Entities.Models
+{
+    /// <summary>
+    /// AgencyFilter class, provides a model for filtering agencies.
```
jdtoombs, line 8: Not all these fields are used, should I just remove them? Thought they would be good to  have in case we want further filtering in the future.

frontend/src/features/admin/agencies/ManageAgencies.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { generateSortCriteria } from 'utils';
+import { AgencyFilterBar } from './AgencyFilterBar';
+import useCodeLookups from 'hooks/useLookupCodes';
+import { useHistory } from 'react-router-dom';
+
+const ManageAgencies: React.FC<any> = ({ filterable, title, mode }) => {
+  const columns = useMemo(() => columnDefinitions, []);
+  const dispatch = useDispatch();
+  const [filter, setFilter] = useState<IAgencyFilter>({});
+  const lookupCodes = useCodeLookups();
```
jdtoombs, line 24: This kind of defeats the purpose of the server side paging... but I was not sure how to get the string `label `for the parent without it as I can't match up the `parentId `with the `id `without it. I cannot compare it against the paged agencies as they do not contain all the potential parents, should I work on implementing something on the backend? Or is this ok.

backend/entities/Models/AgencyFilter.cs
Fosol: I don't think we should filter by anything we won't put an index on `Description`.  You also won't ever filter on `Id`.

frontend/src/components/common/form/PlusButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
Fosol, line 1: I suspect we'll want to reorganize where we keep buttons.  Maybe they belong in a `/components/common/button` folder

frontend/src/constants/actionTypes.ts
```diff
@@ -21,6 +21,21 @@ export const DELETE_PARCEL = 'DELETE_PARCEL';
 export const STORE_LOOKUP_CODE_RESULTS = 'STORE_LOOKUP_CODE_RESULTS';
```
Fosol, line 21: I really hate these files lol.  Just pointless boiler-plate that don't do anything for us.  It's why I'm leaning heavily into the new redux toolkit.

frontend/src/features/admin/agencies/AgencyFilterBar.tsx
```diff
@@ -0,0 +1,46 @@
...
+      initialValues={value}
+      onChange={onChange}
+      searchClassName="bg-primary"
+      plusButton={true}
+      filterBarHeading="Agencies"
+      handleAdd={handleAdd}
+    >
+      <Col className="d-item">
+        {/* TODO: Swap out with new auto complete after switchover */}
+        <AutoCompleteText
```
Fosol, line 36: Doesn't Goeme have the new autocomplete implemented already?

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+import { clear } from 'actions/genericActions';
+import * as actionTypes from 'constants/actionTypes';
+import GenericModal from 'components/common/GenericModal';
+
+interface IEditAgencyPageProps {
+  id: number;
+  match?: any;
+}
+
+const EditAgencyPage = (props: IEditAgencyPageProps) => {
```
Fosol, line 31: Would love to see more jsdoc in our frontend.  We don't document things well here.

frontend/src/features/admin/agencies/ManageAgencies.tsx
Fosol: You could do this a few different ways.  You could update the models when fetching agencies to include some minor details about their parent, or vise-versa.  The way you are doing will only work until they add or rename an agency, as the lookup will become invalid.

Fosol: Which highlights an issue that will appear to be a bug for some once someone does add/edit/delete an agency.

frontend/src/router.tsx
```diff
@@ -149,6 +151,30 @@ const AppRouter: React.FC = () => {
...
+        protected
+        path="/admin/agencies"
+        component={ManageAgencies}
+        layout={AuthLayout}
+        claim={Claims.ADMIN_USERS}
+        title={getTitle('Agency Management')}
+      />
+      <AppRoute
+        protected
+        path="/admin/agency/:id"
```
Fosol, line 164: I like that you kept nice SEO URL paths.  However you may want to use the same path conventions we use in the API.  I don't have a preference though.

Fosol (approved):

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Will add some more!

frontend/src/features/admin/agencies/AgencyFilterBar.tsx
jdtoombs: Yup, I just wanted to get this up for review as I've been keeping it to myself recently... swapping it out before merge. Just want to get my grouping working first

frontend/src/features/admin/agencies/ManageAgencies.tsx
jdtoombs: Gotchya, thank you!

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -28,6 +31,54 @@ public class AgencyService : BaseService<Agency>, IAgencyService
...
+
+        /// <summary>
+        /// Get a page of agencies from the datasource.
+        /// The filter will allow queries to search for the following property values; Name, Description, ParentId.
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public Paged<Agency> Get(AgencyFilter filter = null)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin);
```
devinleighsmith, line 54: I think this makes sense for now, because this list is only used on the agency form.

frontend/src/actionCreators/agencyActionCreator.ts
```diff
@@ -0,0 +1,90 @@
...
+    actionTypes.PUT_AGENCY_DETAILS,
+    axiosPromise,
+  ).catch(() => {});
+};
+
+export const createAgency = (agency: IAddAgency) => async (dispatch: Function) => {
+  dispatch(request(actionTypes.ADD_AGENCY));
+  dispatch(showLoading());
+  try {
+    const { data, status } = await CustomAxios().post(
```
devinleighsmith, line 63: add some lifecycle toasts to your CustomAxios instance to give feedback to user driven actions (so for creates/updates/deletes, not for gets)

frontend/src/components/common/form/PlusButton.tsx
devinleighsmith: If this component is going in the top level components/common/form area it should try and be generic. This component is quite specific to the add agency page, so in its current form it should be located under something like /features/adminAgencies/components.

If you want to leave this here, strip out all the feature specific text/logic and make them settable via params (in this case just the tooltip)

frontend/src/features/admin/agencies/AgencyFilterBar.tsx
```diff
@@ -0,0 +1,46 @@
...
+  handleAdd: (value: any) => void;
+}
+
+export const AgencyFilterBar: React.FC<IProps> = ({ value, onChange, handleAdd }) => {
+  const mapLookupCode = (code: ILookupCode): SelectOption => ({
+    label: code.name,
+    value: code.id.toString(),
+  });
+
+  const lookupCodes = useCodeLookups();
```
devinleighsmith, line 21: Looks like you might be able to use getOptionsByType here instead.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+    selected: !!agency.parentId,
+  });
+  //
+  const selectAgencies = agencies.map(c => mapLookupCode(c));
+
+  const checkAgencies = (
+    <Select
+      label="Parent Agency"
+      field="parentId"
+      data-testid="agency"
```
devinleighsmith, line 61: I think I've fixed the selects so that *byLabel works properly now, so data-testid may not be required.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+      options={selectAgencies}
+      placeholder={agency?.parentId ? undefined : 'Please select if applicable'}
+    />
+  );
+
+  const goBack = () => {
+    history.push('/admin/agencies');
+  };
+
+  const initialValues: IAgencyDetail = {
```
devinleighsmith, line 71: I recommend doing something like this:
initialValues = {...newValues, ...agency}
Reason is that formik will complain if any of your inputs aren't set to '', so unless you are explicitly guaranteeing this at some point after getting the agency from the api you may run into issues.

also, it looks like you should just be able to use ...agency in general rather then explicitly setting all the fields like this.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                dispatch(clear(actionTypes.ADD_PARCEL));
+                history.replace(`/admin/agency/${data.id}`);
+              }
+              setSubmitting(false);
+            }}
+          >
+            {props => (
+              <Form className="agencyInfo">
+                <Label>Agency</Label>
+                <Input data-testid="agency" field="name" value={props.values.name} type="text" />
```
devinleighsmith, line 144: Could use the label prop here instead potentially.
If not (possibly due to styling issues) would prefer you use label for instead of data-testid

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                      isDisabled: values.isDisabled,
+                      sendEmail: values.sendEmail,
+                      parentId: Number(values.parentId),
+                      description: values.description,
+                      rowVersion: values.rowVersion,
+                    },
+                  ),
+                );
+              } else {
+                const data = await createAgency({
```
devinleighsmith, line 126: consider if you need a try/catch here to handle a promise rejection.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+        </Navbar.Brand>
+      </Navbar>
+      <Container fluid={true}>
+        <Row className="agency-edit-form-container">
+          <Formik
+            enableReinitialize
+            initialValues={newAgency ? newValues : initialValues}
+            onSubmit={async (values, { setSubmitting }) => {
+              if (!newAgency) {
+                dispatch(
```
devinleighsmith, line 109: you may need a .error here to deal with promise rejection.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                const data = await createAgency({
+                  name: values.name,
+                  code: values.code,
+                  email: values.email,
+                  isDisabled: values.isDisabled,
+                  sendEmail: values.sendEmail,
+                  parentId: Number(values.parentId),
+                  description: values.description,
+                })(dispatch);
+                dispatch(clear(actionTypes.ADD_PARCEL));
```
devinleighsmith, line 135: is this necessary?

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                <Input data-testid="shortName" field="code" value={props.values.code} type="text" />
+                {checkAgencies}
+                <Label>Description</Label>
+                <Input field="description" type="text" />
+                <Label>Agency e-mail address</Label>
+                <Input data-testid="email" field="email" type="text" />
+
+                <Form.Group className={'is-disabled'}>
+                  <Form.Label>
+                    Is Disabled?{' '}
```
devinleighsmith, line 155: Slight preference for getting rid of these spaces after the label text in favor of css for cross browser consistency.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                </Form.Group>
+                <Form.Group className="send-email">
+                  <Form.Label>
+                    Send e-mail{' '}
+                    <TooltipIcon
+                      toolTipId="email-tooltip"
+                      toolTip={'Toggle to change send email and click save.'}
+                    />{' '}
+                  </Form.Label>
+                  <Field type="checkbox" name="sendEmail" />
```
devinleighsmith, line 171: There is a checkbox (or maybe check) component that you may be able to reuse here.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                </Form.Group>
+
+                <div className="horizontal-line"></div>
+                <Row className="buttons">
+                  <ButtonToolbar className="cancelSave">
+                    <Button
+                      className="bg-danger mr-5"
+                      type="button"
+                      onClick={async () => setShowDelete(true)}
```
devinleighsmith, line 180: is this async necessary?

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                    Send e-mail{' '}
+                    <TooltipIcon
+                      toolTipId="email-tooltip"
+                      toolTip={'Toggle to change send email and click save.'}
+                    />{' '}
+                  </Form.Label>
+                  <Field type="checkbox" name="sendEmail" />
+                </Form.Group>
+
+                <div className="horizontal-line"></div>
```
devinleighsmith, line 174: instead of an hr?

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+                <Row className="buttons">
+                  <ButtonToolbar className="cancelSave">
+                    <Button
+                      className="bg-danger mr-5"
+                      type="button"
+                      onClick={async () => setShowDelete(true)}
+                    >
+                      Delete Agency
+                    </Button>
+                    <Button className="mr-5" type="submit">
```
devinleighsmith, line 184: this may be silently throwing some warnings in your client. If it is you'll need to e.preventDefault on the html submit event and call formiks submit method manually.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -0,0 +1,215 @@
...
+          </Formik>
+        </Row>
+      </Container>
+    </div>
+  );
+};
+
+export default EditAgencyPage;
+
+const DeleteModal = ({ showDelete, setShowDelete, history, dispatch, agency }: any) => (
```
devinleighsmith, line 200: Yes, this is nicer than just putting the modal in directly in the main component tsx

frontend/src/features/admin/agencies/ManageAgencies.scss
```diff
@@ -0,0 +1,37 @@
+$header-height: 72px;
```
devinleighsmith, line 1: Is there a reason you can't import these from _variables.css?

frontend/src/features/admin/agencies/ManageAgencies.scss
```diff
@@ -0,0 +1,37 @@
...
+  background-color: white;
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+  .agency-toolbar {
+    padding: 0;
+    background-color: var(--light);
+    .search-bar {
+      .filterBarHeading {
+        margin-left: 20%;
```
devinleighsmith, line 16: % based margins can be troublesome for responsive designs, how does this page perform when it is resized?

frontend/src/features/admin/agencies/ManageAgencies.tsx
devinleighsmith: Not to throw a wrench into this, but do we need server side paging at all here? this seems like a relatively small (kb) dataset that shouldn't grow much over time.

frontend/src/features/admin/agencies/ManageAgencies.tsx
```diff
@@ -0,0 +1,112 @@
...
+          onChange={value => {
+            setFilter({ ...value });
+          }}
+          handleAdd={() => {
+            history.push('/admin/agency/new');
+          }}
+        />
+      </Container>
+      <div className="table-section">
+        <Table<any>
```
devinleighsmith, line 95: You can probably strongly type this now right?

devinleighsmith:
I think this work is a strong candidate to be placed into /features as it is self-contained functionality.

frontend/src/features/admin/agencies/ManageAgencies.tsx
jdtoombs: Yeah I talked about this with Jeremy briefly after I had realized it may be unnecessary.. we decided on it won't hurt anything but in the future for small data sets always do client side

frontend/src/components/common/form/PlusButton.tsx
jdtoombs: Sounds good!

frontend/src/actionCreators/agencyActionCreator.ts
jdtoombs: Oh right good call, I can use your user one's for reference.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Wooo, good news! I'll clean them up

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Forgot to remove that.. woops

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Will do

frontend/src/features/admin/agencies/ManageAgencies.scss
jdtoombs: Terrible at the moment, all the styling is just a preliminary thing right now. Wanted to get the functionality code looked at then was going to clean this up afterwards.

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Learned this practice from you 😛

frontend/src/features/admin/agencies/EditAgencyPage.tsx
jdtoombs: Lol woops, that is a duh moment.

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -207,10 +207,7 @@ const PropertyListView: React.FC = () => {
   };
 
   const checkExpanded = (row: IProperty, property: IProperty) => {
```
jdtoombs, line 209: Not sure if I had rationale behind this or I was just having a long day....

frontend/src/components/SearchBar/FilterBar.scss
```diff
@@ -9,4 +8,11 @@
       margin: 0;
     }
   }
+  .vl {
```
goemen, line 11: What is `vl`?

jdtoombs: vertical line to separate the buttons of different functionality! based off of the UI

frontend/src/components/common/form/PlusButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+  toolTipText: string;
+  /** set the id of the tool tip use for on hover of the plus buttons */
+  toolTipId: string;
+}
+
+/**
+ * PlusButton displaying a plus button, used to add new items.
+ * @param param0
+ */
+const PlusButton: React.FC<IPlusButtonProps> = ({ ...props }) => {
```
goemen, line 17: No need to use the spread function

frontend/src/components/common/form/SearchButton.tsx
```diff
@@ -10,7 +10,12 @@ import { Button, ButtonProps } from '.';
 const SearchButton: React.FC<ButtonProps> = ({ ...props }) => {
```
goemen, line 10: No need to use the spread functions

goemen (approved):

PR #589, opened 2020-11-05
PA-633 Replacing Local ID

Fosol:
## Description

Replacing `Building.LocalId` with `Building.Name` property.

Updated the following forms by removing **Local ID** and cleaning up the formatting.

- **Map View** - **Parcel Popup**
- **Map View** - **Building Popup**
- **Property Detail View** - **Building Information Section**

The following formatting has changed on the **Parcel Popup** and **Building Popup**

- PID at the top centered for parcels
- Name bolded at top
- Name label removed
- Description displayed if different than the name
- Description label removed
- Address label removed
- View and Update buttons applied the same to both building and parcel

### DIFF
--- a/frontend/src/components/maps/BuildingPopupView.scss
+++ b/frontend/src/components/maps/BuildingPopupView.scss
@@ -7,6 +7,17 @@
   text-align: left;
   padding: 10px;
 
+  .row .col .list-group {
+    margin-top: 6px;
+
+    .list-group-item {
+      padding: 0px 0px 5px 0px;
+      border: none;
+      font-family: 'BCSans', Fallback, sans-serif;
+      font-size: 12px;
+    }
+  }
+
   .row .col :first-child {
     margin-top: 0px;
   }
@@ -19,15 +30,6 @@
     font-weight: bold;
   }
 
-  .list-group {
-    margin-top: 6px;
-  }
-  .list-group-item {
-    padding: 0px 0px 5px 0px !important;
-    border: none;
-    font-family: 'BCSans', Fallback, sans-serif;
-    font-size: 12px;
-  }
   .menu {
     text-align: center;
   }
--- a/frontend/src/components/maps/BuildingPopupView.scss
+++ b/frontend/src/components/maps/BuildingPopupView.scss
@@ -5,7 +5,7 @@
   max-width: 250px;
   box-shadow: 1px 2px 2px;
   text-align: left;
-  padding: 12px;
+  padding: 10px;
 
   .row .col :first-child {
     margin-top: 0px;
@@ -13,16 +13,23 @@
 
   .label {
     font-weight: 700;
+    padding-right: 5px;
+  }
+  .name {
+    font-weight: bold;
   }
 
   .list-group {
     margin-top: 6px;
   }
   .list-group-item {
-    padding: 0;
+    padding: 0px 0px 5px 0px !important;
     border: none;
     font-family: 'BCSans', Fallback, sans-serif;
     font-size: 12px;
+  }
+  .menu {
+    text-align: center;
   }
   p {
     margin: 0;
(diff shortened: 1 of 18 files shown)


frontend/src/components/maps/BuildingPopupView.scss
```diff
@@ -5,25 +5,32 @@
...
+  .name {
+    font-weight: bold;
   }
 
   .list-group {
     margin-top: 6px;
   }
   .list-group-item {
-    padding: 0;
+    padding: 0px 0px 5px 0px !important;
```
Fosol, line 26: ugh, another `!important`... styling is a mess

frontend/src/components/maps/BuildingPopupView.tsx
```diff
@@ -24,63 +24,72 @@ export const BuildingPopupView: React.FC<IBuildingDetailProps> = (props: IBuildi
...
+                  <ListGroup.Item>
+                    <Label>SPP:</Label>
+                    {buildingDetail?.projectNumber}
+                  </ListGroup.Item>
+                )}
+              </ListGroup>
+            </Col>
+          </Row>
+
+          {buildingDetail?.parcelId && !props?.disabled && (
```
Fosol, line 76: Now building and parcel have the same logic

frontend/src/components/maps/ParcelPopupView.scss
```diff
@@ -4,27 +4,37 @@
...
     font-weight: 700;
+    padding-right: 5px;
   }
 
   .list-group {
     margin-top: 6px;
   }
   .list-group-item {
-    padding: 0;
+    padding: 0px 0px 5px 0px !important;
```
Fosol, line 30: Uggh, another `important!`.  Styling is a mess...

frontend/src/components/maps/ParcelPopupView.tsx
```diff
@@ -23,77 +23,74 @@ export const ParcelPopupView = (props: IParcelDetailProps | null) => {
...
+                </ListGroup.Item>
+                <ListGroup.Item>
+                  <Label>Classification: </Label>
+                  {parcelDetail?.classification}
+                </ListGroup.Item>
+              </ListGroup>
+            </Col>
+          </Row>
 
           {parcelDetail?.id && !props?.disabled && (
```
Fosol, line 80: Now building and parcel have the same logic

jdtoombs (approved):
Thank you!!

goemen (approved):

frontend/src/components/maps/BuildingPopupView.scss
devinleighsmith: I don't think this is necessary. Map.scss is causing this class not to work. That was a change that Goeme introduced for the layers popup. Ideally the styling would be consistent, if not we need a specific class for that popup so that we can differentiate it.

frontend/src/components/maps/ParcelPopupView.scss
devinleighsmith: Same comment as for the building popups.

frontend/src/components/maps/BuildingPopupView.scss
devinleighsmith: Just as a note, in your browser you can see what css is overriding your css, which should help you determine if the override is from conflicting _style_ _bootstrap_ or our css. Apologies if you were already aware of this, but we should avoid falling back on important where possible, such as instances like this.

Fosol: image
This is the cause of the issue and why I needed to use `!important`.  How do I fix this?

goemen: We can fix by doing this: `
.parcelPopup {
  .row {
    .col {
      .list-group {
        margin-top: 6px;
        .list-group-item {
         padding: 0px 0px 5px 0px
        }
      }
    }
  }
}`

goemen: Or for a building: 
`.buildingPopup { .row { .col { .list-group { margin-top: 6px; .list-group-item { padding: 0px 0px 5px 0px } } } } }`

PR #597, opened 2020-11-07
Group autocomplete for filters

jdtoombs:
Grouped auto complete filter now on map landing page and property list view. Further customization may be needed but this is a good start to get some feedback. 

Bug fixes included in this PR are as follows:

- styling changes so the agency page now scales and doesn't look terrible

- email validation on edit user page

- popup display for when agency has associated buildings while trying to delete

- only select parent agencies where appropriate

- name and code validation when updating or adding an agency 

- agencies now appear in appropriate lists when they are newly added or edited

- parent agency field is locked when it is a parent agency

- leaving the parent agency on the placeholder (blank) will lock that field and change the text to no parent on save

**New since last edit**

- grouped agency filter now on the page (search by parent, child, or code)

- fixed paging + filter bug

### DIFF
--- a/frontend/src/components/SearchBar/ParentGroupedFilter.tsx
+++ b/frontend/src/components/SearchBar/ParentGroupedFilter.tsx
@@ -1,0 +1,66 @@
+import { SelectOption, SelectOptions } from 'components/common/form';
+import { TypeaheadField } from 'components/common/form/Typeahead';
+import { groupBy, sortBy } from 'lodash';
+import React, { Fragment } from 'react';
+import { Highlighter, Menu, MenuItem } from 'react-bootstrap-typeahead';
+
+interface IParentGroupedFilterProps {
+  name: string;
+  options: SelectOptions;
+  className: string;
+  inputSize?: 'small' | 'large';
+  filterBy?: string[];
+  placeholder?: string;
+}
+
+export const ParentGroupedFilter: React.FC<IParentGroupedFilterProps> = ({
+  name,
+  options,
+  className,
+  inputSize,
+  filterBy,
+  placeholder,
+}) => {
+  return (
+    <TypeaheadField
+      name={name}
+      options={options}
+      inputProps={{ className: className }}
+      bsSize={inputSize}
+      filterBy={filterBy}
+      placeholder={placeholder}
+      filter={true}
+      renderMenu={(results, menuProps) => {
+        const parents = groupBy(
+          results.map((x: SelectOption) => {
+            return {
+              ...x,
+              parentId: x.parentId || x.value,
+            };
+          }),
+          x => x.parentId,
+        );
+        const items = Object.keys(parents)
+          .sort()
+          .map(parent => (
+            <Fragment key={parent}>
+              {!!results.find((x: SelectOption) => x.value === parent) && (
+                <Menu.Header>
+                  <b>{results.find(x => x.value === parent)?.label}</b>
+                </Menu.Header>
+              )}
+              {sortBy(parents[parent], (x: SelectOption) => x.value).map((i, index) => {
+                return (
+                  <MenuItem key={index + 1} option={i} position={index + 1}>
+                    <Highlighter>{i.label}</Highlighter>
+                  </MenuItem>
+                );
+              })}
+            </Fragment>
+          ));
+
+        return <Menu {...menuProps}>{items}</Menu>;
+      }}
+    />
+  );
+};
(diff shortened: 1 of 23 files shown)


frontend/src/features/properties/list/FilterBar.test.tsx
```diff
@@ -40,7 +40,7 @@ describe('FilterBar', () => {
     );
```
jdtoombs, line 40: Will come back to this file on the weekend, need to adjust for the auto complete component which is a bit more tricky. Have some higher priority bugs I want to do today first.

frontend/src/hooks/useRouterFilter.ts
```diff
@@ -51,8 +51,8 @@ export const useRouterFilter = <T extends BasePropertyFilter>(
   React.useEffect(() => {
```
jdtoombs, line 51: Goeme and I found this. The route was not being changed after the filter was applied on the landing page. Seems to be working now

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -8,6 +8,8 @@ interface ITypeaheadFieldProps<T extends TypeaheadModel> extends TypeaheadProps<
   name: string;
```
jdtoombs, line 8: Thank you @goemen for the peer programming with this!

frontend/src/hooks/useRouterFilter.ts
```diff
@@ -51,8 +51,8 @@ export const useRouterFilter = <T extends BasePropertyFilter>(
...
     const filterParams = new URLSearchParams(filter as any);
     const allParams = {
-      ...queryString.parse(filterParams.toString()),
       ...queryString.parse(history.location.search),
+      ...queryString.parse(filterParams.toString()),
```
Fosol, line 55: Is there a reason for the order change?

Fosol (approved):

frontend/src/hooks/useRouterFilter.ts
jdtoombs: it started applying them values to the url properly afterward... not exactly sure the reason why but it works now. (ex) in dev right now if you apply a filter the URL will not change

frontend/src/components/SearchBar/ParentGroupedFilter.tsx
```diff
@@ -0,0 +1,64 @@
+import { SelectOption, SelectOptions } from 'components/common/form';
+import { TypeaheadField } from 'components/common/form/Typeahead';
+import { groupBy, sortBy } from 'lodash';
+import React, { Fragment } from 'react';
+import { Highlighter, Menu, MenuItem } from 'react-bootstrap-typeahead';
+
+interface IParentGroupedFilterProps {
```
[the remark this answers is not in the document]
jdtoombs: I think I am making some of the optional type ahead props required here

frontend/src/hooks/useRouterFilter.ts
devinleighsmith: the old order would overwrite the new params with any existing values, I must have messed this up at some point.

frontend/src/features/admin/agencies/AgencyFilterBar.tsx
```diff
@@ -26,16 +26,13 @@ export const AgencyFilterBar: React.FC<IProps> = ({ value, onChange, handleAdd }
       toolTipAddId="agency-filter-add"
       toolTipAddText="Add a new Agency"
     >
-      <Col className="d-item">
```
jdtoombs, line 29: Unnecessary Col I added, was messing with DOM and styling

frontend/src/features/admin/agencies/EditAgencyPage.tsx
```diff
@@ -169,7 +175,18 @@ const EditAgencyPage = (props: IEditAgencyPageProps) => {
                       <Button
                         className="bg-danger mr-5"
                         type="button"
-                        onClick={() => setShowDelete(true)}
+                        onClick={async () => {
+                          const data = await service.getPropertyList({
```
jdtoombs, line 179: Only able to delete agencies when no buildings associated to it

jdtoombs: Need to add a check for users

frontend/src/components/maps/MapFilterBar.tsx
```diff
@@ -1,17 +1,19 @@
 import './MapFilterBar.scss';
 
-import React from 'react';
+import React, { useMemo } from 'react';
 import { Col } from 'react-bootstrap';
 import { Formik, useFormikContext } from 'formik';
 import { ILookupCode } from 'actions/lookupActions';
-import { Form, Select, InputGroup, Input, SelectOption, AutoCompleteText } from '../common/form';
```
jdtoombs, line 7: Almost time for depreciation :(

PR #598, opened 2020-11-07
Fix disposal project bug

Fosol:
This should resolve the issue currently blocking the ability to add properties to projects.


backend/api/Areas/Project/Mapping/Dispose/BuildingMap.cs
```diff
@@ -67,10 +67,13 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Fiscals, src => src.Fiscals)
                 .AfterMapping((src, dest) =>
                 {
-                    var parcel = dest.Parcels.FirstOrDefault().Parcel;
-                    parcel.LandArea = src.LandArea;
-                    parcel.Zoning = src.Zoning;
-                    parcel.ZoningPotential = src.ZoningPotential;
+                    var parcel = dest.Parcels.FirstOrDefault()?.Parcel;
```
Fosol, line 70: This was causing the issue.  Oddly enough it wasn't causing an issue locally.

PR #608, opened 2020-11-12
PA-1667 using formik persist to local storage

goemen:
The setting `persistable` to enable local storage persistence, but if set to `true` then the `persistProps` will be required and an error is thrown when not set.


jdtoombs (approved):
Wooo

Fosol (approved):

PR #609, opened 2020-11-13
PA 645 - ChesExceptions test cases

LawrenceLau2020:
Added 5 tests for the ChesException constructors.

### DIFF
--- a/backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
+++ b/backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
@@ -1,0 +1,146 @@
+using FluentAssertions;
+using Pims.Core.Exceptions;
+using Pims.Core.Http;
+using Pims.Core.Test;
+using Pims.Ches.Models;
+using Moq;
+using System.Diagnostics.CodeAnalysis;
+using System.Net;
+using Xunit;
+using Pims.Core.Http.Models;
+using Microsoft.Extensions.Options;
+using Pims.Ches.Configuration;
+using Pims.Ches;
+using System.Linq;
+using System.Net.Http;
+
+namespace Pims.Dal.Test.Libraries.Ches
+{
+    [Trait("category", "unit")]
+    [Trait("category", "ches")]
+    [Trait("group", "notification")]
+    [ExcludeFromCodeCoverage]
+    public class ChesExceptionTest
+    {
+
+        #region Tests
+        #region ChesException Constructors
+        [Fact]
+        public void ChesExceptionClientModel_Success()
+        {
+            // Arrange
+            var message = "test2";
+            var status = HttpStatusCode.OK;
+            var helper = new TestHelper();
+
+            var user = PrincipalHelper.CreateForPermission();
+
+            var options = Options.Create(new ChesOptions()
+            {
+                AuthUrl = "https:/test.com",
+                EmailEnabled = true,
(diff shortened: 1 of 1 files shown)


backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
```diff
@@ -0,0 +1,149 @@
...
+            var exception = new HttpClientRequestException(message, status);
+            var client = helper.GetService<Mock<IHttpRequestClient>>().Object;
+            var model = new ErrorResponseModel();
+
+            // Act
+            var chesException = new ChesException(exception, client, model);
+
+            // Assert
+            Assert.NotNull(exception);
+            Assert.NotNull(client);
```
Fosol, line 61: No need to assert a variable you created.

backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
```diff
@@ -0,0 +1,149 @@
...
+            Assert.NotNull(model);
```
Fosol, line 62: No need to assert a variable you created.

backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
```diff
@@ -0,0 +1,149 @@
...
+            chesException.Message.Should().Be(message + System.Environment.NewLine);
```
Fosol, line 63: Probably should find out why it has an extra `NewLine` at the end of the string.  It's probably not ideal.

LawrenceLau2020: There is a `NewLine` in the constructor:

`public ChesException(HttpClientRequestException exception, IHttpRequestClient client, ErrorResponseModel model)
            : this($"{exception.Message}{Environment.NewLine}", exception, exception.StatusCode)`

where it concatenates the new line right after the exception.Message.

backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
LawrenceLau2020: Should I remove those 3 assertions, or can we assert the chesException properties as below?

`Assert.NotNull(chesException.Message);
 Assert.NotNull(chesException.Client);
 Assert.NotNull(chesException.Detail);`

backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
Fosol: Ah leave it for now.  Looks a little messy, but it isn't critical.

backend/tests/unit/dal/Libraries/Ches/ChesExceptionTest.cs
Fosol: Remove the two assertions that are related to the `Act`.  You will always know `client` and `model` are not null because you created them yourself manually.

jdtoombs (approved):

Fosol (approved):
Thanks

goemen (approved):

devinleighsmith (approved):

LawrenceLau2020:
Merging chesException test cases into BCGov Pims dev branch

PR #610, opened 2020-11-13
PA-1445: Update tooltip design

agalbrai:
Updated tooltip design as per Bobbi's design


jdtoombs (approved):

Fosol (approved):
Nice

frontend/src/components/common/TooltipIcon.scss
```diff
@@ -1,12 +1,25 @@
...
 .tooltip {
   white-space: pre-wrap;
   .tooltip-inner {
     text-align: left;
     max-width: 25rem;
+    background-color: $secondary-variant-color;
+    color: #ffffff;
+  }
+  .arrow:before {
+    border-top-color: $secondary-variant-color;
```
devinleighsmith, line 19: Can you confirm that this works if the tooltip is drawn to the left, right or bottom of the icon?

devinleighsmith:
This will fix the tooltip icon's tooltips, but won't affect the tooltip wrappers tooltips. Not sure if that was on purpose or you haven't come across the tooltip wrapper yet.

For the tooltip wrapper you'll either need to override some of the global styling in App.scss or create a new scss file for TooltipWrapper.tsx

devinleighsmith:
I'm fine if you want to merge this and then fix the other tooltips in another PR @agalbrai

frontend/src/components/common/TooltipIcon.scss
agalbrai: it seems it doesn't, I'll fix that

agalbrai: it doesn't, but from what I've seen all of our tooltips are from the top (the only one on the right is being removed)

agalbrai:
@devinleighsmith from what I've seen all of the tooltipWrappers tooltips also changed with this

frontend/src/components/common/TooltipIcon.scss
devinleighsmith: @agalbrai

PR #615, opened 2020-11-18
Pa-1668: property detail view tooltips

agalbrai:
Created some tool tips/text that Jonny can add to the new add building/land wizard. Also fixed other small spelling/wording issues

### DIFF
--- a/frontend/src/components/common/TooltipIcon.scss
+++ b/frontend/src/components/common/TooltipIcon.scss
@@ -4,6 +4,8 @@
   min-width: 14px;
   min-height: 14px;
   color: $secondary-variant-color;
+  margin-left: -15px;
+  margin-right: 1px;
 }
 
 .tooltip {
--- a/frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
+++ b/frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
@@ -10,9 +10,9 @@ import {
   Check,
 } from 'components/common/form';
 import { useFormikContext } from 'formik';
-import TooltipWrapper from 'components/common/TooltipWrapper';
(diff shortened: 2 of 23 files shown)


frontend/src/components/common/TooltipIcon.scss
```diff
@@ -4,6 +4,7 @@
   min-width: 14px;
   min-height: 14px;
   color: $secondary-variant-color;
+  margin-right: 1px;
```
devinleighsmith, line 7: I just had a look in dev and noticed that all of the fields with tooltips don't have their inputs aligned. you can correct this by adding a negative margin equal to the width of the tooltip icon (you'll need to take into account the margin you've added above as well.

frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
```diff
@@ -48,14 +48,17 @@ const InformationForm: FunctionComponent<InformationFormProps> = (props: Informa
...
         <Form.Label className="required">Classification</Form.Label>
-        <FastSelect
-          formikProps={formikProps}
-          disabled={props.disabled}
-          type="number"
-          placeholder="Must Select One"
-          field={withNameSpace('classificationId')}
-          options={props.classifications}
-        />
+        <Form.Group>
```
devinleighsmith, line 51: FastSelect contains a Form.Group already, so this will result in nested form groups on a single field. Would recommend removing this unless you absolutely need it.

agalbrai: working on it

devinleighsmith (approved):

Fosol (approved):

PR #618, opened 2020-11-19
PA-1380 SIS Disposal Projects

Fosol:
Enhancements and bug fixes to support the latest SIS Disposal Projects.

** DRAFT PR ** - For early feedback

- New database migration v01.03.00
- Enhanced Converter Tool
- Updated project import endpoint
- Fix Importer Tool bug
- Fix Disposal Project UI validation
- Updated Keycloak Sync Tool configuration for SRES Financial Manager role
- Updated relevant dependencies
- Remove Newtonsoft
- Added `Project.Metadata` property
- Added `ProjectSnapShot.Metadata` property
- Renamed `Estimated` to `Market` to reflect actual usage

## To Do

- Fix backend unit-tests
- Fix notifications when `to` address is not provided
- Fix notifications when shared notes are updated
- Review import data manually to ensure imported data is valid


backend/api/Areas/Project/Models/Dispose/ProjectModel.cs
```diff
@@ -265,14 +275,9 @@ public class ProjectModel : Api.Models.BaseModel
...
-        /// get/set - A static netbook value used on the close out form.
+        /// get/set - The market value.
         /// </summary>
-        public decimal? CloseOutNetbook { get; set; }
-
-        /// <summary>
-        /// get/set - The estimated value.
-        /// </summary>
-        public decimal Estimated { get; set; }
+        public decimal Market { get; set; }
```
Fosol, line 280: All `Estimate` properties have been renamed to `Market`.  This is because it better reflects the purpose and reduced confusion.  There were over three different labels which all were the same field (Estimated Market Value, Current Market Value, Sales Proceeds).

backend/api/Areas/Reports/Models/Project/SurplusPropertyListModel.cs
```diff
@@ -49,11 +49,11 @@ public class SurplusPropertyListModel
 
         [DisplayName("Sales_Cost")]
         [CsvHelper.Configuration.Attributes.Name("Sales_Cost")]
-        public decimal SalesCost { get; set; }
+        public decimal? SalesCost { get; set; }
```
Fosol, line 52: These should NOT default to $0

backend/api/Helpers/Mapping/ServiceCollectionExtensions.cs
```diff
@@ -30,5 +32,37 @@ public static IServiceCollection AddMapster(this IServiceCollection services, Ac
...
+        /// <param name="serializerOptions"></param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddMapster(this IServiceCollection services, JsonSerializerOptions serializerOptions, Action<TypeAdapterConfig> options = null)
+        {
+            var config = TypeAdapterConfig.GlobalSettings;
+
+            var optionsSerializer = Microsoft.Extensions.Options.Options.Create(serializerOptions);
+            var assemblies = new[] { Assembly.GetAssembly(typeof(Startup)) };
+            var registers = assemblies.Select(assembly => assembly.GetTypes()
```
Fosol, line 50: Fixing unit-tests to allow for DI serialization.

backend/api/Startup.cs
```diff
@@ -86,7 +86,14 @@ public Startup(IConfiguration configuration, IWebHostEnvironment env)
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSerilogging(this.Configuration);
-            services.AddMapster(options =>
+            var jsonSerializerOptions = new JsonSerializerOptions()
```
Fosol, line 89: Cleaning up code to reduce repetition.

backend/dal/Helpers/Extensions/DbContextExtensions.cs
```diff
@@ -44,6 +44,11 @@ public static void UpdateProjectNumber(this PimsContext context, Project project
 
             project.Properties.ForEach(p =>
             {
+                if (p.PropertyType == PropertyTypes.Land && p.Parcel == null)
```
Fosol, line 47: Load the properties if they are not in context yet.

backend/dal/Migrations/Initial/Up/PostUp/01-Agencies.sql
```diff
@@ -232,6 +232,13 @@ INSERT INTO dbo.[Agencies] (
     , 'BC Assessment'
```
Fosol, line 232: Add missing agencies

backend/dal/Migrations/v01.03.00/Down/PostDown/01-NotificationTemplates.sql
```diff
@@ -0,0 +1,10 @@
+PRINT 'Updating NotificationTemplates'
+
+UPDATE dbo.[NotificationTemplates]
+SET [Body] = '
+@using System.Linq
+@using Pims.Dal.Entities
+<html><head><title>@Model.Environment.Title</title></head>
+<body><p>Dear @Model.Project.Manager,</p><p>Your project @Model.Project.ProjectNumber has been updated with the following note;</p><p>@Model.Project.PublicNote</p></body></html>'
```
Fosol, line 8: Fixing template now that the note is part of a collection instead of on the project.

backend/dal/PIMSContext.cs
```diff
@@ -195,6 +203,28 @@ public int CommitTransaction()
...
+
+        /// <summary>
+        /// Deserialize the specified 'json' to the specified type of 'T'.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="json"></param>
+        /// <returns></returns>
+        public T Deserialize<T>(string json)
+        {
+            return JsonSerializer.Deserialize<T>(json, _serializerOptions);
```
Fosol, line 215: Provides an easy way to serialize/deserialize objects with DI

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -104,8 +100,11 @@ public IEnumerable<ProjectSnapshot> GetSnapshots(ProjectReport report)
 
             foreach (ProjectSnapshot snapshot in toSnapshots)
             {
-                var previousSnapshot = fromSnapshots.FirstOrDefault(s => s.Key == snapshot.ProjectId);
-                snapshot.BaselineIntegrity = (snapshot?.NetProceeds ?? 0) - (previousSnapshot.Value?.NetProceeds ?? 0);
+                var metadata = !String.IsNullOrWhiteSpace(snapshot.Metadata) ? this.Context.Deserialize<DisposalProjectSnapshotMetadata>(snapshot.Metadata) : new DisposalProjectSnapshotMetadata();
```
Fosol, line 103: A snapshot now gets a full copy of the `metadata`

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -526,10 +521,11 @@ public async System.Threading.Tasks.Task<Project> UpdateAsync(Project project)
             var responses = project.GetResponseChanges(project);
 
             // If the note was changed generate a notification for it.
-            var noteChanged = originalProject.PublicNote != project.PublicNote;
+            var noteChanged = !String.IsNullOrWhiteSpace(project.GetNoteText(NoteTypes.Public)) && originalProject.GetNoteText(NoteTypes.Public) != project.GetNoteText(NoteTypes.Public);
```
Fosol, line 524: There is a bug with this presently.

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -198,12 +198,6 @@ public Building Update(Building building)
 
             this.Context.Entry(existingBuilding).CurrentValues.SetValues(building);
 
-            // update only an active project with any financial value changes.
-            if (!String.IsNullOrWhiteSpace(existingBuilding.ProjectNumber) && (!this.Context.Projects.Include(p => p.Status).FirstOrDefault(p => p.ProjectNumber == existingBuilding.ProjectNumber)?.IsProjectClosed() ?? false))
```
Fosol, line 202: We no longer use the properties for their financial values within a project.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -318,12 +318,6 @@ public Parcel Update(Parcel parcel)
                             this.Context.Entry(existingBuildingFiscal).CurrentValues.SetValues(buildingFiscal);
                         }
                     }
-
-                    // update only an active project with any financial value changes.
-                    if (updateProject && !String.IsNullOrWhiteSpace(existingBuilding.ProjectNumber) && (!this.Context.Projects.Include(p => p.Status).FirstOrDefault(p => p.ProjectNumber == existingBuilding.ProjectNumber)?.IsProjectClosed() ?? false))
```
Fosol, line 323: We no longer use the properties for their financial values within a project.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -411,13 +405,6 @@ public Parcel Update(Parcel parcel)
                         this.Context.ParcelFiscals.Remove(parcelFiscals);
                     }
                 }
-
-                // update only an active project with any financial value changes.
-                if (!String.IsNullOrWhiteSpace(parcel.ProjectNumber)
```
Fosol, line 416: We no longer use the properties for their financial values within a project.

frontend/src/features/projects/common/components/columns.tsx
```diff
@@ -175,19 +175,14 @@ const useProjectFinancialValues = (project: IProject, editable?: boolean) => {
 export const defaultPropertyColumns: any[] = [
   {
     Header: 'Agency',
-    accessor: 'agencyCode', // accessor is the "key" in the data
+    accessor: (row: IProperty) =>
+      row.subAgency ? `${row.subAgency} (${row.agencyCode})` : row.agencyCode, // accessor is the "key" in the data
```
Fosol, line 179: Improving the table layout by combining agency information into a single column

frontend/src/features/projects/common/components/columns.tsx
```diff
@@ -175,19 +175,14 @@ const useProjectFinancialValues = (project: IProject, editable?: boolean) => {
...
-    Header: 'Sub Agency',
-    accessor: 'subAgency',
-    align: 'left',
-    clickable: true,
-  },
-  {
-    Header: 'Property Name',
-    accessor: 'description',
+    Header: 'Name',
+    accessor: (row: IProperty) => row.name ?? row.pid,
```
Fosol, line 185: Ensures a name is displayed

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -309,7 +309,7 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
               </Button>
             </TooltipWrapper>
           )}
-          {keycloak.hasClaim(Claims.REPORTS_VIEW) && (
+          {keycloak.hasClaim(Claims.PROJECT_VIEW) && (
```
Fosol, line 312: Those who can use this page should be able to export to Excel

frontend/src/features/projects/list/columns.tsx
```diff
@@ -65,16 +65,7 @@ export const columns = (onDelete?: (id: string) => void): ColumnWithProps<IProje
...
-      align: 'left',
-      responsive: true,
-      clickable: true,
-      width: spacing.medium,
-      minWidth: 80,
-    },
-    {
-      Header: 'Sub Agency',
-      accessor: 'subAgency',
+      accessor: item => (item.subAgency ? `${item.subAgency} (${item.agency})` : item.agency),
```
Fosol, line 68: Improving table layout by combining two columns

frontend/src/features/projects/spl/forms/CloseOutFinancialSummaryForm.tsx
```diff
@@ -28,7 +28,7 @@ const CloseOutFinancialSummaryForm = (props: CloseOutFinancialSummaryFormProps)
               formikProps={formikProps}
               disabled={props.isReadOnly}
               outerClassName="col-md-8"
-              field="salesProceeds"
+              field="market"
```
Fosol, line 31: Sales Proceeds is the same as Current Market Value, as that is how much was paid.

frontend/src/features/projects/spl/forms/CloseOutFinancialSummaryForm.tsx
```diff
@@ -50,7 +50,7 @@ const CloseOutFinancialSummaryForm = (props: CloseOutFinancialSummaryFormProps)
               formikProps={formikProps}
               disabled={props.isReadOnly}
               outerClassName="col-md-8"
-              field="closeOutNetbook"
+              field="netBook"
```
Fosol, line 53: Removed `CloseOutNetBook` as it was a duplication.

tools/keycloak/sync/realm.json
```diff
@@ -84,6 +84,16 @@
         "Description": "Ability to view projects.",
         "Composite": false
       },
+      {
+        "Name": "reports-view",
```
Fosol, line 88: Missing role/claim values for keycloak sync

tools/keycloak/sync/realm.json
```diff
@@ -199,6 +209,10 @@
           "project-delete",
           "dispose-approve"
         ]
+      },
+      {
+        "Name": "SRES Financial Manager",
```
Fosol, line 214: Missing group/role for keycloak sync

jdtoombs (approved):

backend/api/Areas/Project/Mapping/Report/ProjectSnapshotMap.cs
```diff
@@ -1,11 +1,29 @@
...
 {
     public class ProjectSnapshotMap : IRegister
     {
+        #region Variables
+        private readonly JsonSerializerOptions _serializerOptions;
+        #endregion
+
+        #region Constructors
+        /// <summary>
+        /// Creates a new instance of a ProjectMap, initializes with specified arguments.
```
devinleighsmith, line 17: ProjectMap -> ProjectSnapshotMap

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -108,45 +140,117 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
 
                 _logger.LogDebug($"Importing Project '{model.ProjectNumber}', agency:'{model.Agency}', workflow:'{model.Workflow}', status:'{model.Status}'");
 
+                var status = GetStatus(model.Status);
+
+                // Default to the last workflow this status is associated with.
+                // This isn't ideal and could end up with projects assocaited with an incorrect workflow, however
```
devinleighsmith, line 146: however?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -108,45 +140,117 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+                var workflow = !String.IsNullOrWhiteSpace(model.Workflow) ? GetWorkflow(model.Workflow) : model.Activity switch
```
devinleighsmith, line 147: this is some funky syntax, so if model.workflow is null switch on the value of model.Activity?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -108,45 +140,117 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
-                if (model.PriorNetProceeds.HasValue)
-                {
-                    project.Snapshots.Add(new Entity.ProjectSnapshot(project) { NetProceeds = model.PriorNetProceeds.Value });
-                }
+                project.Market = model.Market;
+                project.Appraised = model.Appraised;
 
-                project.MarketedOn = model.MarketedOn;
+                project.SubmittedOn = new DateTime(model.ReportedFiscalYear - 1, 1, 1); // Don't have a source for this information.
+                project.ApprovedOn = project.SubmittedOn; // Don't have a source for this information.
```
devinleighsmith, line 208: double checking that this is correct.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -108,45 +140,117 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+                {
+                    InitialNotificationSentOn = null, // Don't have a source for this information.
+                    ThirtyDayNotificationSentOn = null, // Don't have a source for this information.
+                    SixtyDayNotificationSentOn = null, // Don't have a source for this information.
+                    NinetyDayNotificationSentOn = null, // Don't have a source for this information.
+                    OnHoldNotificationSentOn = null, // Don't have a source for this information.
+                    InterestedReceivedOn = null, // Don't have a source for this information.
+                    TransferredWithinGreOn = null, // Don't have a source for this information.
+                    ClearanceNotificationSentOn = null, // Don't have a source for this information.
+                    RequestForSplReceivedOn = model.RequestForSplReceivedOn, // Don't have a source for this information.
```
devinleighsmith, line 221: you say you don't have a source here, but it seems like you do?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -108,45 +140,117 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+                    OcgFinancialStatement = model.OcgFinancialStatement,
+                    AppraisedBy = model.AppraisedBy,
+                    ProgramCost = model.ProgramCost,
+                    SalesCost = model.SalesCost,
+                    InterestComponent = model.InterestComponent,
+                    NetProceeds = model.NetProceeds,
+                    GainLoss = model.GainLoss,
+                    SaleWithLeaseInPlace = model.SaleWithLeaseInPlace,
+                    DisposedOn = model.DisposedOn ?? (project.Status.Code == "DIS" ? model.CompletedOn : null),
+                    IsContractConditional = model.IsContractConditional.HasValue ? model.IsContractConditional : false, // Don't have a source for this information.
```
devinleighsmith, line 234: where is model.iscontractConfitional coming from then?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -187,7 +283,8 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
                     project.AddTask(tasks.DistinctBy(t => t.Id).ToArray());
                     project.Tasks.ForEach(t =>
                     {
-                        t.IsCompleted = !t.Task.IsOptional;
+                        t.IsCompleted = !t.Task.IsOptional; // Complete all required steps.
+                        t.Task = null;
```
devinleighsmith, line 287: curios why this is necessary.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -203,14 +300,100 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+        /// <param name="noteType"></param>
+        /// <param name="note"></param>
+        private void AddNote(Entity.Project project, string noteType, string note)
+        {
+            if (!String.IsNullOrWhiteSpace(note))
+            {
+                if (!Enum.TryParse(noteType, true, out Entity.NoteTypes type))
+                {
+                    type = noteType switch
+                    {
```
devinleighsmith, line 317: Should these checks ignore caps?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -203,14 +300,100 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+        /// Assumption is made that if the parcel is added, then also add all the buildings on the parcel.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <param name="pid"></param>
+        /// <returns></returns>
+        private Entity.Project AddProperty(Entity.Project project, string pid)
+        {
+            try
+            {
+                var pidValue = Int32.Parse(pid.Replace("-", ""));
```
devinleighsmith, line 362: .net used to throw a warning unless you specified the desired base here, I assume it doesn't?

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -203,14 +300,100 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
...
+                var parcel = _adminService.Parcel.GetByPid(pidValue);
+                project.AddProperty(parcel).ForEach(p => {
```
devinleighsmith, line 364: I'm finding the two functions both called AddProperty to be confusing.

backend/api/Areas/Tools/Models/Import/ImportProjectModel.cs
```diff
@@ -1,38 +1,192 @@
...
+        /// </summary>
+        public decimal NetBook { get; set; }
+
+        /// <summary>
+        /// get/set - The assessed value.
+        /// </summary>
+        public decimal? Assessed { get; set; }
+
+        /// <summary>
+        /// get/set - The net book value.
```
devinleighsmith, line 136: typo

backend/dal/Helpers/Extensions/NoteExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        /// <summary>
+        /// Add or update the note for the specified 'type' on the 'project'.
+        /// Presently only one note of each type is supported.
+        /// </summary>
+        /// <param name="project"></param>
+        /// <param name="type"></param>
+        /// <param name="text"></param>
+        /// <returns></returns>
+        public static ProjectNote AddOrUpdateNote(this Project project, NoteTypes type, string text)
```
devinleighsmith, line 19: AddOrReplaceNote perhaps?

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -57,7 +56,7 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
             if (!String.IsNullOrWhiteSpace(filter.ProjectNumber))
                 query = query.Where(p => EF.Functions.Like(p.ProjectNumber, $"%{filter.ProjectNumber}%"));
             if (!String.IsNullOrWhiteSpace(filter.Name))
-                query = query.Where(p => EF.Functions.Like(p.Name, $"%{filter.Name}%"));
+                query = query.Where(p => EF.Functions.Like(p.Name, $"%{filter.Name}%") || EF.Functions.Like(p.ProjectNumber, $"%{filter.Name}%"));
```
devinleighsmith, line 59: is this safe? dumb example but if I name my project "2" do I expect to get back all of the projects with 2 in the project number?

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -419,11 +379,14 @@ public static void Merge(this Entity.Project originalProject, Entity.Project upd
         {
             // Update a project
             var agency = originalProject.Agency;
-            var metadata = originalProject.Metadata;
+            var originalMetadata = context.Deserialize<DisposalProjectMetadata>(originalProject.Metadata); // TODO: Need to test whether automatically overwriting the metadata is correct.
```
devinleighsmith, line 382: I'm curious about this as well.

backend/dal/Services/Admin/Concrete/ProjectService.cs
```diff
@@ -136,23 +168,12 @@ public void Add(IEnumerable<Project> projects)
...
-                {
-                    r.Agency = null;
-                });
-
-                project.Tasks.ForEach(t =>
+                if (String.IsNullOrWhiteSpace(project.ProjectNumber))
                 {
-                    t.Task = null;
-                });
+                    project.ProjectNumber = $"TEMP-{DateTime.UtcNow.Ticks:00000}";
```
devinleighsmith, line 175: Are temp projects viewable on the FE? If so does it display ok?

backend/dal/Services/Admin/Concrete/WorkflowService.cs
```diff
@@ -51,6 +51,23 @@ public Workflow Get(int id)
...
 
+        /// <summary>
+        /// Get an array of workflows that are associated with the specified project 'code'.
+        /// </summary>
+        /// <param name="code"></param>
+        /// <returns></returns>
+        public IEnumerable<Workflow> GetForStatus(string code)
+        {
+            return (
+                from s in this.Context.ProjectStatus
```
devinleighsmith, line 62: reason for the different sql query format?

backend/dal/Services/Concrete/ProjectService.cs
devinleighsmith, line 524: so deleting a note should not generate a notification?

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -595,7 +591,10 @@ public async System.Threading.Tasks.Task RemoveAsync(Project project)
 
             await this.CancelNotificationsAsync(project.Id);
 
+            originalProject.Notifications.ForEach(n => this.Context.NotificationQueue.Remove(n));
             originalProject.Notifications.Clear(); // TODO: Need to test this to determine if it'll let us delete a project with existing notifications.
+            originalProject.Notes.ForEach(n => this.Context.ProjectNotes.Remove(n));
+            originalProject.Notes.Clear();
```
devinleighsmith, line 597: so cascades don't work here I guess?

tools/converters/excel/Converters/StringConverters.cs
```diff
@@ -30,26 +31,37 @@ public static class StringConverters
...
-        /// 
+        ///
         /// </summary>
         /// <param name="value"></param>
         /// <param name="nullable"></param>
         /// <returns></returns>
         public static DateTime? ConvertToDate(this string value, bool nullable = false)
         {
-            if (Regex.Match(value, "[0-9]{2}/[0-9]{2}").Success)
+            if (Regex.Match(value, "^[0-9]{2,4}/[0-9]{2,4}$").Success)
```
devinleighsmith, line 41: I wonder if this could be converted to use tryparse instead:
https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tryparse?view=net-5.0
not important for now, maybe tech debt.

devinleighsmith (approved):

backend/dal/Services/Concrete/ProjectService.cs
Fosol: Correct.  The current template includes the new note.  Which would be empty and cause more confusion.  I didn't want to increase scope to update the template to give a different message.  Additionally if they use this as a tool to communicate with the agency in some fashion, it doesn't really make sense to send an email when there is no note.

backend/dal/Services/Admin/Concrete/WorkflowService.cs
Fosol: fluent annotation just makes certain queries easier to write and read.

backend/dal/Services/Admin/Concrete/ProjectService.cs
Fosol: They are viewable, as they are just projects that haven't been submitted.

backend/dal/Helpers/Extensions/ProjectExtensions.cs
Fosol: I'm not entirely certain what's best here. We don't current provide a way to find a project by its `ProjectNumber` in the UI.  Which is one of the most common ways to find a project.  So I did this to essentially bypass the need to add yet another filter on the project list view page.   Essentially a workaround to get us closer to user friendly.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: The `AddProperty(...)` extension function just does a handful of required steps for adding either `parcel` or `building` to a `project`.  In this case because we're dealing with unattached entities if we reference them it results in attempts to add duplicate values to the database.  Which either results in actually inserting duplicate records into the database, or an error because it already exists in the database.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: No warnings are thrown here, it is strongly typed at this point.  The only possible issue is that the `pid` is null, or results in not a number.  But I'm hesitant to add the logic at this point in time due to the current use-case for this helper.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: Probably, but similar to other implementation decisions, I've taken a few short-cuts.  This helper is currently only to support imports.  Which at this point in time are only for development and one time run in production.  I do try to make them as generic and future proof as I can within the timeframe.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: This question came up today with Maggie too.  It's one of those scenarios where I don't think there is a good solution.  If I let them remain false it will almost certainly block future saves because we validate whether all required tasks are completed.  Which means I'd have to allow for overriding that behaviour...  This gets rather complicated.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: I could add it to the JSON sent to the import, but SIS data doesn't have the data.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: The source is the JSON file sent with the import.  But SIS doesn't have anything that matches.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: `SubmittedOn` is when the project was submitted for assessment.  `ApprovedOn` is when the project was approved by SRES to be added to SPP.  SIS data doesn't have either of these dates in their data regrettably.  So I've simply assumed when it was reported last will be the date.  I'm wondering if I should leave them null.

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
Fosol: If they didn't provide the exact workflow in the JSON, then determine it based on the `Activity` value.

PR #619, opened 2020-11-19
PA-1533 highlight parcel on map when marker clicked, or when map clicked

devinleighsmith:
Also fix an issue with our version of formik.


frontend/package.json
```diff
@@ -14,7 +14,7 @@
     "crypto-js": "4.0.0",
     "dequal": "1.0.0",
     "epsg": "0.5.0",
-    "formik": "2.1.4",
+    "formik": "2.2.5",
```
devinleighsmith, line 17: 2.1.4 has known issues with initialValues not working correctly:
https://github.com/formium/formik/releases/tag/v2.1.5

frontend/src/components/maps/leaflet/Spiderfier.ts
```diff
@@ -141,6 +141,7 @@ export class Spiderfier {
   }
 
   private handleChildMarkerClick(e: LeafletMouseEvent) {
+    e.originalEvent.stopPropagation();
```
devinleighsmith, line 144: This prevents the marker click from being registered on the map.

Fosol (approved):

jdtoombs (approved):

PR #620, opened 2020-11-20
PA-1660 component to edit/add/delete polygons for properties on a map

goemen:
This component will allow us to edit/add/delete polygons for properties on a map. We can wrap this component with a formik field to add validation to the boundaries, etc.


frontend/public/index.html
```diff
@@ -25,13 +25,18 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
 
-    <title>Property Inventory Management System</title>
-  </head>
+  <title>Property Inventory Management System</title>
+  <style type="text/css">
```
Fosol, line 29: Is this idea?

Fosol (approved):

frontend/src/components/maps/leaflet/LeafletDraw/LeafletDraw.tsx
```diff
@@ -0,0 +1,124 @@
...
+          remove: !readonly,
+        },
+      });
+      map.addControl(drawControl);
+    }
+  }, [map, readonly, canAdd]);
+
+  useEffect(() => {
+    if (!!map) {
+      map.removeEventListener((L as any).Draw.Event.CREATED);
```
devinleighsmith, line 81: what's going on with these event listeners?

frontend/src/components/maps/leaflet/LeafletDraw/LeafletDraw.tsx
```diff
@@ -0,0 +1,124 @@
...
+          color,
+          weight: 3,
+          opacity: 1,
+        },
+        filter: feature => feature.properties.editable,
+      }).eachLayer(layer => {
+        featureGroup.addLayer(layer);
+      });
+
+      map.addLayer(featureGroup);
```
devinleighsmith, line 96: does this layer need to be cleaned up at some point?

frontend/src/components/maps/leaflet/LeafletDraw/LeafletDraw.tsx
devinleighsmith: might make more sense to put these in the useEffects cleanup then like this, because presumably this isn't required initially, but is required after the component is cleaned up.

devinleighsmith (approved):

frontend/src/components/maps/leaflet/LeafletDraw/LeafletDraw.tsx
goemen: Thats a great idea

PR #622, opened 2020-11-20
Provide zoom to marker functionality on the mapview.

devinleighsmith:
Adds a zoom button that takes you to a level where you should be able to differentiate points (as long as they aren't on the exact same lat/lng).


Fosol (approved):
Love the hook implementation

PR #625, opened 2020-11-23
Ensure app name env var is used.

devinleighsmith:
Fix bug with the alpha pipeline where the app name environment variable was being ignored.

### DIFF
--- a/openshift/pipelines/Jenkinsfile.cicd
+++ b/openshift/pipelines/Jenkinsfile.cicd
@@ -8,7 +8,7 @@ def notify
 
 def FE_COV_STASH = 'fe-test-coverage'
 def API_COV_STASH = 'api-test-coverage'
-def RELEASE_VERSION = ""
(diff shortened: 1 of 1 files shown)


jdtoombs (approved):

openshift/pipelines/Jenkinsfile.cicd
```diff
@@ -17,25 +17,25 @@ pipeline {
...
-    buildDiscarder(logRotator(numToKeepStr: "5")) // keep 5 builds only
+    buildDiscarder(logRotator(numToKeepStr: '5')) // keep 5 builds only
   }
   environment {
-    CI = "true"
+    CI = 'true'
 
     // The name of your application
-    APP_NAME = "pims"
+    APP_NAME = 'pims-alpha'
```
Fosol, line 26: Isn't this the same file used by the other pipelines?

Fosol (approved):

PR #632, opened 2020-11-25
Dev alpha

devinleighsmith:
This is a merge of dev into dev-alpha

71428c6 merge commit

9cd58f1 corrections


openshift/pipelines/Jenkinsfile.cicd
```diff
@@ -22,9 +22,6 @@ pipeline {
   environment {
     CI = 'true'
 
-    // The name of your application
```
Fosol, line 25: I thought you just added this value, but it looks like you're removing it now

frontend/src/features/properties/components/forms/subforms/PidPinForm.tsx
```diff
@@ -50,11 +50,10 @@ const PidPinForm: FunctionComponent<PidPinProps> = (props: PidPinProps) => {
   return (
     <>
       <Form.Row className="d-inline-flex flex-nowrap">
-        <Form.Label className="required">PID</Form.Label>
+        <Form.Label className="req">*</Form.Label>PID
```
Fosol, line 53: Seems odd to have the label outside like this?

Fosol (approved):

frontend/src/features/properties/components/forms/subforms/PidPinForm.tsx
devinleighsmith: Agreed, but this is how it is in current dev-alpha

openshift/pipelines/Jenkinsfile.cicd
devinleighsmith: I added the ability to have this value come from the env vars, not just hardcoded.

jdtoombs (approved):

PR #636, opened 2020-11-26
Fix Agency Filter

Fosol:
The agency filter was incorrectly displaying the `Agency.Id` rather than the `Agency.Name`.  This PR should fix this.

Can someone confirm this doesn't break something else, as I'm not sure who broke this in the first place?

### DIFF
--- a/frontend/src/components/common/form/Typeahead.tsx
+++ b/frontend/src/components/common/form/Typeahead.tsx
@@ -61,7 +61,7 @@ export function TypeaheadField<T extends TypeaheadModel>({
         isValid={!filter && isValid}
         selected={getOptionByValue(getIn(values, name))}
         onChange={(newValues: T[]) => {
-          setFieldValue(name, getIn(newValues[0], 'value') ?? newValues[0]);
+          setFieldValue(name, getIn(newValues[0], 'name') ?? newValues[0]);
         }}
         onBlur={() => setFieldTouched(name, true)}
         id={`${name}-field`}


frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -61,7 +61,7 @@ export function TypeaheadField<T extends TypeaheadModel>({
         isValid={!filter && isValid}
         selected={getOptionByValue(getIn(values, name))}
         onChange={(newValues: T[]) => {
-          setFieldValue(name, getIn(newValues[0], 'value') ?? newValues[0]);
+          setFieldValue(name, getIn(newValues[0], 'name') ?? newValues[0]);
```
devinleighsmith, line 64: does this work? I think the api expects the agency id not the agency name to be sent over the wire.

Fosol: The basic tests seem to work.  I was able to filter.  But I'm worried this isn't right, as I really don't feel comfortable with my knowledge in this area.

devinleighsmith: Probably the safest way then would be to create a prop that takes onChange as a function, and then you can override that for the specific case that isn't working:

onChange={
          onChange
            ? onChange
            : (newValues: T[]) => {
                setFieldValue(name, getIn(newValues[0], 'value') ?? newValues[0]);
              }
        }

Fosol:
Jonny has submitted one to fix this issue.

PR #638, opened 2020-11-26
Agency filter fix

jdtoombs:
Fix agency filter populating with id on change.

Based off of Jeremy's solution and Devin's comment!

### DIFF
--- a/frontend/src/components/SearchBar/ParentGroupedFilter.tsx
+++ b/frontend/src/components/SearchBar/ParentGroupedFilter.tsx
@@ -1,5 +1,6 @@
 import { SelectOption, SelectOptions } from 'components/common/form';
 import { TypeaheadField } from 'components/common/form/Typeahead';
+import { getIn, useFormikContext } from 'formik';
 import { groupBy, sortBy } from 'lodash';
 import React, { Fragment } from 'react';
 import { Highlighter, Menu, MenuItem } from 'react-bootstrap-typeahead';
@@ -21,6 +22,7 @@ export const ParentGroupedFilter: React.FC<IParentGroupedFilterProps> = ({
   filterBy,
   placeholder,
 }) => {
+  const { setFieldValue } = useFormikContext();
   return (
     <TypeaheadField
       name={name}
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -9,12 +9,14 @@ interface ITypeaheadFieldProps<T extends TypeaheadModel> extends TypeaheadProps<
   name: string;
   label?: string;
   required?: boolean;
-  /** whether or not this component is being used to filter so we can ignore the validation checkmark */
-  filter?: boolean;
+  /** whether or not this component should show the validation checkmark */
```
jdtoombs, line 12: changed on alpha first coming through on here now, less confusing name

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -58,11 +61,15 @@ export function TypeaheadField<T extends TypeaheadModel>({
         {...rest}
         inputProps={{ ...rest.inputProps, name: name, id: `${name}-field` }}
         isInvalid={hasError as any}
-        isValid={!filter && isValid}
+        isValid={!hideValidation && isValid}
         selected={getOptionByValue(getIn(values, name))}
-        onChange={(newValues: T[]) => {
-          setFieldValue(name, getIn(newValues[0], 'value') ?? newValues[0]);
-        }}
+        onChange={
```
Fosol, line 66: The default uses the value, but it can now be overridden?

frontend/src/components/SearchBar/ParentGroupedFilter.tsx
```diff
@@ -29,7 +31,10 @@ export const ParentGroupedFilter: React.FC<IParentGroupedFilterProps> = ({
       bsSize={inputSize}
       filterBy={filterBy}
       placeholder={placeholder}
-      filter={true}
+      onChange={vals => {
```
Fosol, line 34: This overrides to display the name?  Does it still use `id` value for querying?

frontend/src/components/common/form/Typeahead.tsx
jdtoombs: Correct, has its' initial implementation still so location still behaves as expected but for our agency case we override it in order to function as expected

frontend/src/components/SearchBar/ParentGroupedFilter.tsx
jdtoombs: Yes still uses the id for querying, it alters the filter value onFilterChange in MapFilterBar.tsx

Fosol (approved):

frontend/src/components/SearchBar/ParentGroupedFilter.tsx
devinleighsmith: Can you confirm this is working on the GRETransformForm? this appears to be the only other place the typeahead is used for agencies.

devinleighsmith (approved):

PR #639, opened 2020-11-26
Pa:1776: Remove asterix and darken label field

agalbrai:
Ran my changes past Bobbi, she approves.

### DIFF
--- a/frontend/src/App.scss
+++ b/frontend/src/App.scss
@@ -1,6 +1,6 @@
 @import './colors.scss';
 .error {
-  color: red;
+  color: $danger-color;
 }
 
 ul.pagination li {
@@ -52,4 +52,21 @@ div.bootstrap-table {
 //React table global styling
 .table .tr .td {
   word-break: break-word;
+}
+
+//by default highlight required fields in blue.
+.form-group.required input {
+  border: 1px solid $secondary-variant-color;
+}
+
+.form-group.required select {
+  border: 1px solid $secondary-variant-color;
+}
+
+.input-group.required input {
+  border: 1px solid $secondary-variant-color;
+}
+
+.form-check.required input {
+  border: 1px solid $secondary-variant-color;
 }
--- a/frontend/src/components/common/form/Check.tsx
+++ b/frontend/src/components/common/form/Check.tsx
@@ -22,7 +22,7 @@ type OptionalAttributes = {
   placeholder?: string;
   /** Adds a custom class to the input element of the <Input> component */
   className?: string;
-  /** Whether the field is required. Shows an asterisk after the label. */
+  /** Whether the field is required. Makes the field border blue. */
   required?: boolean;
   /** Specifies that the HTML element should be disabled */
   disabled?: boolean;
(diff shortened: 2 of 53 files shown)


agalbrai:
Required fields

frontend/src/App.scss
```diff
@@ -53,3 +53,20 @@ div.bootstrap-table {
 .table .tr .td {
   word-break: break-word;
 }
+
+//by default highlight required fields in blue.
+.form-group.required input {
```
devinleighsmith, line 58: could maybe use :first-child here instead?

devinleighsmith: something like this:
.form-group.required > *:first-child

frontend/src/components/common/form/Check.tsx
```diff
@@ -74,13 +74,12 @@ export const Check: React.FC<CheckProps> = ({
   return (
     <Form.Group
       controlId={`input-${field}`}
-      className={classNames(outerClassName, !!required ? 'required' : '')}
```
devinleighsmith, line 77: the order doesn't matter does it?

frontend/src/components/common/form/FastCurrencyInput.tsx
```diff
@@ -81,7 +84,7 @@ const CurrencyInput = ({
   const isInvalid = error && touch ? 'is-invalid ' : '';
   const isValid = !error && touch && value && !disabled ? 'is-valid ' : '';
   return (
-    <Form.Group className={classNames(outerClassName, 'fast-currency-input')}>
+    <Form.Group className={classNames(!!required ? 'required' : '', outerClassName)}>
```
devinleighsmith, line 87: last 'fast-currency-input' class with this change, not sure if that is used anywhere. Please confirm.

devinleighsmith (approved):

frontend/src/features/projects/common/forms/TasksForm.tsx
```diff
@@ -17,6 +17,7 @@ interface ITaskFormParams {
 const TasksForm = ({ tasks, isReadOnly, className }: ITaskFormParams) => {
   const { values } = useFormikContext<IProject>();
   return (
+    //<Form.Group
```
Fosol, line 20: delete line if not required.

Fosol (approved):

frontend/src/components/common/form/Check.tsx
agalbrai: I thought it did, but no it doesn't

frontend/src/components/common/form/FastCurrencyInput.tsx
agalbrai: I'm not sure what you mean

frontend/src/App.scss
agalbrai: Just tried this, didn't work because a lot of the groups have the label as the first child, so the label gets the border rather than the input field

frontend/src/components/common/form/FastCurrencyInput.tsx
agalbrai: @devinleighsmith

PR #645, opened 2020-12-01
New Agency Component

jdtoombs:
Please refer to GIF below to see UI functionality. Note the field is only editable if you are an admin. Also - have updated the filter with this PR as well on the `MapFilterBar`. Will be phasing out my old `ParentGroupedFilter` and replacing with `ParentSelect`. This specific PR just addresses the new agency component to be used for the Building and Land Wizard.

**Updated** with some changes to the building form as well. Sticky submit bar so it stays visible, classification component adjustments, bug fixes etc. Will need to ensure this integrates with Devin's prior PR.

newagency


frontend/src/features/mapSideBar/SidebarContents/subforms/IdentificationForm.tsx
```diff
@@ -45,7 +55,7 @@ export const IdentificationForm: React.FC<IIdentificationProps> = ({
       <Row>
         <Col>
           <InformationForm
-            isAdmin
+            isAdmin={isAdmin ?? false}
```
goemen, line 58: You can do `!!isAdmin`

goemen (approved):

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -0,0 +1,132 @@
...
+  label?: string;
+  /** flag used to control whehter to display red astrix indicating required */
+  required?: boolean;
+  /** determine whether this component is disabled or not */
+  disabled?: boolean;
+  /** if this component is being used as a filter do not pre populate the value */
+  isFilter?: boolean;
+}
+
+export const ParentSelect: React.FC<IParentSelect> = ({
```
Fosol, line 32: Documentation should be on the component so that intelisense works

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -0,0 +1,132 @@
...
+  /** filterBy is optional - if used can determine what you can filter with */
+  filterBy?: string[];
+  /** provide the opitional label text*/
+  label?: string;
+  /** flag used to control whehter to display red astrix indicating required */
+  required?: boolean;
+  /** determine whether this component is disabled or not */
+  disabled?: boolean;
+  /** if this component is being used as a filter do not pre populate the value */
+  isFilter?: boolean;
```
Fosol, line 29: Not an ideal implementation.  I think you should use the more standard approach which would be to set the default value instead.

Fosol (approved):

Fosol:
Looks like you could add the component to the `dev` branch if you split this up.

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -0,0 +1,132 @@
...
+  /** give the component the desired options to group by parent with */
+  options: SelectOptions;
+  /** if you would like placeholder text within the input */
+  placeholder?: string;
+  /** filterBy is optional - if used can determine what you can filter with */
+  filterBy?: string[];
+  /** provide the opitional label text*/
+  label?: string;
+  /** flag used to control whehter to display red astrix indicating required */
+  required?: boolean;
```
devinleighsmith, line 25: Since you are in dev you should be aware that we aren't using the red asterix to denote required inputs anymore, we are now highlighting the field in light blue. see the property detail form in pims-dev as an example.

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -46,6 +53,12 @@ export function TypeaheadField<T extends TypeaheadModel>({
   if (!getOptionByValue) {
     getOptionByValue = (value: T) => (!!value ? ([value] as T[]) : ([] as T[]));
   }
+  const ref = useRef<any>();
+  React.useEffect(() => {
+    if (clearMenu && ref.current) {
```
devinleighsmith, line 58: might as well check for ref.current?.blur

devinleighsmith (approved):

PR #647, opened 2020-12-02
PA-1833 Adding basic GeoJSON endpoints

Fosol:
Incremental update to switch the map over to using GeoJSON endpoints.  This will open the door to a number of GIS related enhancements moving forward.

Created `InventoryLayer` component and replaced current functionality for displaying PIMS Inventory.

This new layer can be turned on and off in the layers menu.

## To Do

- Regrettably I ran out of time to discover how to use a `TileLayer`.  The documented examples don't work with an `accessToken` regrettably.  Which means we'll have to build a custom tile layer.
- Improve caching.  Presently I have hardcoded it to 5 minutes.
- The GeoJSON endpoints are not true WFS endpoints.  They only provide custom filters related to PIMS.


jdtoombs (approved):

goemen (approved):

backend/core/Extensions/DictionaryExtensions.cs
```diff
@@ -217,6 +217,22 @@ public static T GetEnumValue<T>(this IDictionary<string, Microsoft.Extensions.Pr
...
+        /// <param name="key"></param>
+        /// <param name="defaultValue"></param>
+        /// <returns></returns>
+        public static NetTopologySuite.Geometries.Envelope GetEnvelopNullValue(this IDictionary<string, Microsoft.Extensions.Primitives.StringValues> dict, string key, NetTopologySuite.Geometries.Envelope defaultValue = null)
+        {
+            if (!dict.TryGetValue(key, out Microsoft.Extensions.Primitives.StringValues value))
+                return defaultValue;
+
+            var values = value.ToString().Split(',');
+            return NetTopologySuite.Geometries.Envelope.Parse($"Env[{values[0]}:{values[1]},{values[2]}:{values[3]}]");
```
Fosol, line 233: Provides a way to deserialize a boundary box (aka envelope).

backend/core/Extensions/GeometryExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// GeometryExtensions static class, provides extension methods for geometry objects.
+    /// </summary>
+    public static class GeometryExtensions
+    {
+        /// <summary>
+        /// Convert an Envelope to a Polygon.
+        /// </summary>
+        /// <param name="envelope"></param>
+        /// <returns></returns>
+        public static Polygon ToPolygon(this Envelope envelope)
```
Fosol, line 15: Centralize this function so that all the queries that use it can be updated in one place.

backend/core/Extensions/GeometryExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Convert an array of Coordinate to a Polygon.
+        /// If the coordinate array is only a length of two it will be treated like a boundary.
+        /// If the coordinate array does not self close, it will by default close with the inital coordinates.
+        /// The coordinate array should be expressed counter-clockwise.
+        /// </summary>
+        /// <param name="coords"></param>
+        /// <returns></returns>
+        public static Polygon ToPolygon(this Coordinate[] coords)
```
Fosol, line 44: Centralize this function so that all the queries that use it can be updated in one place.

backend/core/Extensions/ObjectExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Copies the public properties from the 'source' object to the 'destination' object.
+        /// It will only copy public properties that have a 'set' and are of the following types, primitive, enum, string.
+        /// Or an array of those same types.
+        /// </summary>
+        /// <param name="source"></param>
+        /// <param name="destination"></param>
+        /// <typeparam name="ST"></typeparam>
+        /// <typeparam name="DT"></typeparam>
+        public static DT CopyValues<ST, DT>(this ST source, DT destination)
```
Fosol, line 21: Quick and dirty clone one object into another.   I believe the mapper can do this too, but will look into it later.

backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -78,6 +78,8 @@ where projectNumbers.Contains(p.Project.ProjectNumber)
                 }
             }
 
+            // TODO: Add optional paging ability to query.
```
Fosol, line 81: WFS has paging, but this endpoint doesn't currently support it.

backend/entities/Models/PageFilter.cs
```diff
@@ -71,6 +72,19 @@ public virtual bool IsValid()
...
                 && this.Quantity > 0;
         }
+
+        /// <summary>
+        /// Convert a start index to a page for paging. // TODO: Should switch to using start index instead of paging because it has a little more control.
+        /// </summary>
+        /// <param name="startIndex"></param>
+        /// <param name="quantity"></param>
+        /// <returns></returns>
+        private int CalcPage(int startIndex, int quantity)
```
Fosol, line 82: WFS uses `startIndex` for paging.  Will need to consider switching over to using it in the future.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    bounds.getNorthEast().lng,
+    bounds.getNorthEast().lat,
+  ] as BBox;
+};
+
+/**
+ * Displays the search results onto a layer with clustering.
+ * This component makes a request to the PIMS API properties search WFS endpoint.
+ */
+export const InventoryLayer: React.FC<InventoryLayerProps> = ({
```
Fosol, line 44: Instead of the `Map` and `MapView` components doing all the work for making requests for PIMS inventory, it has been all moved into this `InventoryLayer` component.  At some point if I have time I will look into creating a `InventoryTileLayer` so that the requests to the DB will be more performant and cacheable.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -0,0 +1,119 @@
...
+    classificationId: filter?.classificationId,
+    minLandArea: filter?.minLandArea,
+    maxLandArea: filter?.maxLandArea,
+    inSurplusPropertyProgram: filter?.inSurplusPropertyProgram,
+    inEnhancedReferralProcess: filter?.inEnhancedReferralProcess,
+  } as IGeoSearchParams)}`;
+
+  // Fetch the geoJSON collection of properties.
+  useDeepCompareEffect(() => {
+    fetch(url, {
```
Fosol, line 90: I didn't want to use `redux` to fetch this content.  It isn't currently shared by anything.  I however need to figure out how to hook it up to error handling.  If anyone has suggestions it would great.

Fosol: I should also switch to `axios`.  Need to look into how to do that.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -431,6 +418,17 @@ const Map: React.FC<MapProps> = ({
                   )}
                   <LegendControl />
                   <LayersControl position="topright">
```
Fosol, line 420: Users can now turn PIMS Inventory off if they want to

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -431,6 +418,17 @@ const Map: React.FC<MapProps> = ({
...
+                    <LayersControl.Overlay checked name="PIMS Inventory">
+                      <InventoryLayer
+                        zoom={zoom}
+                        onMarkerClick={onSingleMarkerClick}
+                        filter={{
```
Fosol, line 425: Just pass the filter params down to update the map.

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -28,22 +27,6 @@ const defaultLatLng = {
   lng: -124.67285156250001,
 };
 
-const parcelBounds: IPropertySearchParams = {
```
Fosol, line 31: Moved to the `API`

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -82,63 +65,17 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
   const lotSizes = fetchLotSizes();
   const dispatch = useDispatch();
 
-  const getApiParams = (e: MapViewportChangeEvent): IPropertySearchParams | null => {
```
Fosol, line 85: All logic has been moved to the `InventoryLayer`

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -82,63 +65,17 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
...
-    };
-    return apiParams;
-  };
   const saveLatLng = (e: LeafletMouseEvent) => {
     if (!props.disabled) {
       dispatch(saveLeafletMouseEvent(e));
     }
   };
 
-  const throttledFetch = useRef(
```
Fosol, line 128: `InventoryLayer` now handles the request for properties.  Not sure if I need to apply throttling to that layer as testing so far appears to work well without it.

backend/api/Areas/Property/Controllers/SearchController.cs
```diff
@@ -58,6 +60,7 @@ public SearchController(IPimsService pimsService, IMapper mapper)
         [Produces("application/json")]
         [ProducesResponseType(typeof(IEnumerable<PropertyModel>), 200)]
         [SwaggerOperation(Tags = new[] { "property" })]
+        [ResponseCache(Duration = 300, Location = ResponseCacheLocation.Any)]
```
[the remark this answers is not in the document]
devinleighsmith: I have a requirement to focus the map on a newly created property. I could probably temporarily add the new property to the map right after creation to force it to display, but the pin would then disappear if the map was moved/scrolled if I'm understanding this correctly.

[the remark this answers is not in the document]
devinleighsmith: So if we want to keep this we may need to add a disclaimer? Newly added properties may not show up on the map immediately or something?

backend/core/Extensions/DictionaryExtensions.cs
```diff
@@ -217,6 +217,22 @@ public static T GetEnumValue<T>(this IDictionary<string, Microsoft.Extensions.Pr
...
         }
 
+        /// <summary>
+        /// Get the value from the dictionary for the specified 'key' and return it as an double.
+        /// </summary>
+        /// <param name="dict"></param>
+        /// <param name="key"></param>
+        /// <param name="defaultValue"></param>
+        /// <returns></returns>
+        public static NetTopologySuite.Geometries.Envelope GetEnvelopNullValue(this IDictionary<string, Microsoft.Extensions.Primitives.StringValues> dict, string key, NetTopologySuite.Geometries.Envelope defaultValue = null)
```
devinleighsmith, line 227: Envelop

backend/core/Converters/GeometryJsonConverter.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <returns></returns>
+        private Geometry Create(Type type, string value)
+        {
+            if (String.IsNullOrWhiteSpace(value)) return null;
+
+            switch (type.Name)
+            {
+                case nameof(Point):
+                    {
+                        var values = value.Split(',').Select(v => double.Parse(v)).ToArray();
```
devinleighsmith, line 91: null check on value?

backend/core/Extensions/DictionaryExtensions.cs
```diff
@@ -217,6 +217,22 @@ public static T GetEnumValue<T>(this IDictionary<string, Microsoft.Extensions.Pr
             return dict.TryGetValue(key, out Microsoft.Extensions.Primitives.StringValues dValue) && Enum.TryParse(typeof(T), dValue, ignoreCase, out object value) ? (T)value : defaultValue;
         }
 
+        /// <summary>
+        /// Get the value from the dictionary for the specified 'key' and return it as an double.
```
devinleighsmith, line 221: Does this really return a double?

backend/core/Extensions/DictionaryExtensions.cs
devinleighsmith: do we want some sanity checking around the number of values so we don't end up with an indexoutofbounds?

backend/core/Extensions/GeometryExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+                    bounds = new[]
+                    {
+                        new Coordinate(coords[0].X, coords[0].Y),
+                        new Coordinate(coords[1].X, coords[0].Y),
+                        new Coordinate(coords[1].X, coords[1].Y),
+                        new Coordinate(coords[0].X, coords[1].Y),
+                        new Coordinate(coords[0].X, coords[0].Y),
+                    };
+                }
+                else if (coords[0].Equals(coords[^1]))
```
devinleighsmith, line 61: should this be != ? 
This is saying "if the first coord is equal to the last coord, make sure the last coord is equal to the first coord"
but I think you want to be doing this correction only if it is not equal.

backend/core/Extensions/ObjectExtensions.cs
devinleighsmith: yeah, I think the mapper does this by default if IgnoreNonMapped is not specified.

frontend/src/constants/API.ts
```diff
@@ -51,6 +51,22 @@ export interface IPropertySearchParams {
...
   `/properties/search?${params ? queryString.stringify(params) : ''}`; // get filtered properties or all if not specified.
 
+export interface IGeoSearchParams {
+  bbox?: string | undefined;
+  address?: string;
+  administrativeArea?: string;
+  pid?: string;
+  projectNumber?: string;
+  agencies?: string; // TODO: Switch to number[]
+  classificationId?: string; // TODO: Switch to number
```
devinleighsmith, line 61: these used to be in this format:

classificationId: number | null;
  minLandArea: number | null;
  maxLandArea: number | null;

I can see why you might want to delay fixing agencies, but shouldn't you be able to just keep these all as number?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -0,0 +1,119 @@
...
+        const data = (await response.json()) as Feature[];
+        const points = data.map(f => {
+          return {
+            ...f,
+          } as PointFeature;
+        });
+        setFeatures(points);
+      })
+      .catch(error => {
+        console.log(error);
```
devinleighsmith, line 105: At least make this a console.error
Probably need a todo to enhance this with either a toast or to reuse the existing redux error modal.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
devinleighsmith: The draft marker hook uses the list of properties. If the properties aren't in redux that functionality will not work.

devinleighsmith: Maybe not an issue, I've been told that we don't want draft marker functionality.

backend/api/Areas/Property/Controllers/SearchController.cs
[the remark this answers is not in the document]
Fosol: We can reduce the cache to 2 minutes.  Generally it's unlikely a user will create a property in such a short time.  The caching simply allows for better performance when viewing the map map in some scenarios.
As for redux, I'm hesitant to continue using it this way.  The current use-case of the map is we pull the whole database down by default.  Pulling that into redux just for the few properties that will be added from time to time doesn't make a lot of sense.  Not only that, but the current implementation of redux with the map properties is that we're constantly changing it as we move around.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
Fosol: It's now hooked up with the global axios error handling, along with a friendly toast message.

backend/api/Areas/Property/Controllers/SearchController.cs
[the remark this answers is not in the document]
Fosol: Additionally the geojson data is a very different format to what we were using previously.  So the parcel you would be adding would need to be converted as well.

PR #648, opened 2020-12-02
PA-1722 Added fields to Financial Information

Fosol:
Added Estimated Sales Costs
Added Estimated Program Recovery Fees

image

### DIFF
--- a/backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
+++ b/backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
@@ -141,7 +141,8 @@ namespace Pims.Api.Areas.Project.Mapping.Dispose
                 .Map(dest => dest.Tasks, src => src.Tasks)
                 .Map(dest => dest.Responses, src => src.ProjectAgencyResponses)
                 .Map(dest => dest.Notes, src => src.Notes)
-                .AfterMapping((src, dest) => {
+                .AfterMapping((src, dest) =>
+                {
                     var metadata = new Entity.Models.DisposalProjectMetadata()
                     {
                         Purchaser = src.Purchaser,
--- a/frontend/src/features/projects/common/ApprovalTransitionPage.tsx
+++ b/frontend/src/features/projects/common/ApprovalTransitionPage.tsx
@@ -43,15 +43,17 @@ const ApprovalTransitionPage: React.FunctionComponent<IApprovalTransitionPagePro
       return;
     }
     if (project !== undefined && !isTransitioned) {
+      // Look for a possible transition within the same workflow.
+      const next = toStatus?.filter(s => s.workflowCode === project.workflowCode);
       if (
         project.statusCode === ReviewWorkflowStatus.ApprovedForExemption ||
         project.statusCode === ReviewWorkflowStatus.NotInSpl
       ) {
         history.replace(`erp?projectNumber=${project.projectNumber}`);
-      } else if (toStatus?.length !== 1) {
+      } else if (next?.length !== 1) {
         history.replace(`${project.status?.route}?projectNumber=${project.projectNumber}`);
       } else {
-        const toStatusCode = toStatus[0].code;
+        const toStatusCode = next[0].code;
         setIsTransitioned(true);
         transitionFunction(
           dispatch(updateWorkflowStatus(project, toStatusCode, project.workflowCode)),
--- a/frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
+++ b/frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
@@ -6,20 +6,20 @@ import { useFormikContext } from 'formik';
 export interface IProjectFinacialTableProps {
   /** Whether form fields are disabled. */
   disabled: boolean;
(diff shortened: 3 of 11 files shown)


frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
```diff
@@ -33,15 +48,29 @@ const ProjectFinancialTable = ({ disabled }: { disabled: boolean }) => {
...
             <Form.Label column md={4}>
               Appraised Value
             </Form.Label>
             <FastCurrencyInput field="appraised" formikProps={context} disabled={disabled} />
-          </Row>
+          </Form.Row>
+        </Col>
+        <Col md={4}>
+          <Form.Row>
+            <Form.Label column md={4}>
```
jdtoombs, line 61: This would have been my error from when I added the original 4 fields, I can include it in a later PR - but wondering if we should add `margin-bottom: 5px` to the labels so they align with the inputs a bit better

jdtoombs (approved):

goemen (approved):

frontend/src/features/projects/common/ApprovalTransitionPage.tsx
```diff
@@ -43,15 +43,17 @@ const ApprovalTransitionPage: React.FunctionComponent<IApprovalTransitionPagePro
       return;
     }
     if (project !== undefined && !isTransitioned) {
+      // Look for a possible transition within the same workflow.
+      const next = toStatus?.filter(s => s.workflowCode === project.workflowCode);
```
Fosol, line 47: This resolves the issue when there are multiple possible transitions.  The assumption is this component will only auto transition to a status within the same workflow.

frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
Fosol: What's the best way to do this?  I don't like adding a style manually to each.

backend/api/Areas/Project/Mapping/Dispose/ProjectMap.cs
```diff
@@ -141,7 +141,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Tasks, src => src.Tasks)
                 .Map(dest => dest.Responses, src => src.ProjectAgencyResponses)
                 .Map(dest => dest.Notes, src => src.Notes)
-                .AfterMapping((src, dest) => {
+                .AfterMapping((src, dest) =>
```
devinleighsmith, line 144: I would love to get automated formatting running on the backend as well.

frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
devinleighsmith: probably create a styled component wrapping Form.Label with the margin-bottom.

frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
```diff
@@ -1,29 +1,44 @@
...
+import { Col, Form } from 'react-bootstrap';
 import { FastCurrencyInput } from 'components/common/form';
 import { useFormikContext } from 'formik';
 
-const ProjectFinancialTable = ({ disabled }: { disabled: boolean }) => {
+export interface IProjectFinacialTableProps {
+  /** Whether form fields are disabled. */
+  disabled: boolean;
+  /** An optional label to display. */
+  label?: string;
```
devinleighsmith, line 10: I think label is misleading, this is a header or title.

frontend/src/features/projects/common/tabs/ProjectInformationTab.tsx
```diff
@@ -19,6 +19,7 @@ const ProjectInformationTab: React.FunctionComponent<IProjectInformationTabProps
 
       <UpdateInfoForm
         isReadOnly={isReadOnly}
+        showRisk={true}
```
devinleighsmith, line 22: hmm.. what is this?

devinleighsmith (approved):

frontend/src/features/projects/common/tabs/ProjectInformationTab.tsx
Fosol: Whether the risk dropdown will be displayed.  As this component is used during submit we don't want the owning agency setting the risk value.  Only SRES should be setting this value.

frontend/src/features/projects/common/components/ProjectFinancialTable.tsx
devinleighsmith: @Fosol not seeing the requested change here, or a TODO. My concern is that this'll get forgotten if we don't at least have a tech-debt item for it.

Fosol: Spoke with jonny and he mentioned we can just correct the shared css instead.  That was part of the PR as well.

devinleighsmith: thanks,sounds good.

PR #651, opened 2020-12-02
PA-747: Direct IE user to use Edge, Chrome or Firefox

agalbrai:
image
Bobbi and Mikaela approve of the page look

### DIFF
--- a/frontend/src/features/account/IENotSupportedPage.tsx
+++ b/frontend/src/features/account/IENotSupportedPage.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { Col, Container, Row } from 'react-bootstrap';
+
+/**
+ * The IENotSupportedPage prevents users from entering PIMS
+ * if they are using any version of Internet Explorer, as most
+ * libraries no longer support IE.
+ */
+export const IENotSupportedPage: React.FC = () => {
+  return (
+    <Container className="unauth" fluid={true}>
+      <h1>PIMS does not support Internet Explorer</h1>
+      <Row className="sign-in">
+        <Col xs={16} md={6} className="block">
+          <h6>Please use a supported internet browser such as Chrome, Firefox or Edge.</h6>
+        </Col>
+      </Row>
+    </Container>
+  );
+};
--- a/frontend/src/features/account/Login.tsx
+++ b/frontend/src/features/account/Login.tsx
@@ -11,11 +11,22 @@ import * as actionTypes from 'constants/actionTypes';
 import { useQuery } from 'hooks/use-query';
 import { FaExternalLinkAlt } from 'react-icons/fa';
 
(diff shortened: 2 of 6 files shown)


frontend/src/features/account/Login.tsx
```diff
@@ -28,55 +40,66 @@ const Login = () => {
...
-                <p>
-                  If you're not yet registered, <br></br>2.{' '}
-                  <a href="https://www.bceid.ca/os/?7169" target="_blank" rel="noopener noreferrer">
-                    Register for your Business BCeID
-                  </a>{' '}
-                  <FaExternalLinkAlt />
-                </p>
-              </Jumbotron>
-            )}
+      {isIE ? (
```
Fosol, line 45: Might be cleaner if instead of showing a different component you redirect the user to the `IENotSupported` page

frontend/src/features/account/LoginErrorPage.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Col, Row } from 'react-bootstrap';
+
+export const LoginErrorPage: React.FC<any> = () => {
```
Fosol, line 4: This component name is generic, but the implementation is specific to IE not supported.  Rename, or implement generically.

Fosol (commented):
Looks nice

frontend/src/features/account/LoginErrorPage.tsx
goemen, line 4: No need for `<any>`

goemen (approved):

frontend/src/features/account/IENotSupportedPage.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Col, Container, Row } from 'react-bootstrap';
+
+export const IENotSupportedPage: React.FC = () => {
```
devinleighsmith, line 4: We're trying to have top level component jsdoc for all new components, so please add some here.

frontend/src/features/account/IENotSupportedPage.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <Container className="unauth" fluid={true}>
+      <h1>PIMS does not support Internet Explorer</h1>
```
devinleighsmith, line 7: Technically it just doesn't support older versions of IE right? is this text correct?

devinleighsmith (approved):

frontend/src/features/account/IENotSupportedPage.tsx
agalbrai: @devinleighsmith I thought it wasn't supporting IE at all

Fosol (approved):

PR #652, opened 2020-12-03
Swap snapshot style to a format supported by jest-styled-components

devinleighsmith:
The old syntax was taken from the react testing library website, but isn't (directly) supported by styled components.
The new syntax is taken from jest-styled-components library for react testing library and appears to work.


goemen (approved):

frontend/src/components/common/form/AutoCompleteText.test.tsx
```diff
@@ -28,6 +28,6 @@ describe('Auto Complete Text', () => {
         )}
       </Formik>,
     );
-    expect(pretty(container.innerHTML)).toMatchSnapshot();
+    expect(container.firstChild).toMatchSnapshot();
```
Fosol, line 31: Wouldn't a `firstChild` be different than `innerHTML`?  As the `innerHTML` could have many children.

Fosol (approved):

frontend/src/components/common/form/AutoCompleteText.test.tsx
devinleighsmith: I thought the same thing, but if you look at the generated snapshots the contents match (minus some formatting changes).

PR #657, opened 2020-12-05
PA-1678 Land Classification form updates, includes new usage form.

devinleighsmith:


frontend/src/components/common/Label.tsx
```diff
@@ -1,16 +1,19 @@
 import * as React from 'react';
 import './Label.scss';
 interface ILabelProps {
-  children?: string;
+  content?: string;
```
devinleighsmith, line 4: using children as a string is confusing given the react naming convention for nested child components.

frontend/src/components/common/Label.tsx
```diff
@@ -1,16 +1,19 @@
...
   required?: boolean;
 }
 
-/** generic inline label element */
-export const Label = (props: ILabelProps | null | undefined) => {
+/** Generic inline label element */
+export const Label: React.FunctionComponent<ILabelProps & React.HTMLAttributes<HTMLDivElement>> = ({
```
devinleighsmith, line 9: Labels should support all of the standard html attribute tags.

frontend/src/components/maps/__snapshots__/BuildingPopupView.test.tsx.snap
```diff
@@ -42,7 +42,7 @@ exports[`building popup view renders correctly 1`] = `
           onClick={[Function]}
         >
           <p
-            className="label"
+            className="label undefined"
```
devinleighsmith, line 45: You can safely ignore this as it has been corrected in a future revision.

frontend/src/components/common/Label.tsx
jdtoombs: Ooh good call

jdtoombs (approved):

PR #658, opened 2020-12-05
PA-1977: Message for users who don't have permission to submit disposal projects

agalbrai:


Fosol (approved):

frontend/src/features/admin/access-request/AccessRequestPage.tsx
```diff
@@ -101,7 +101,6 @@ const AccessRequestPage = () => {
         toolTip="To select multiple roles, hold Ctrl and select options."
       >
         <Select
-          label="Role"
```
agalbrai, line 104: There was already a label so "Roles" and "Role" were both appearing

goemen (approved):

PR #665, opened 2020-12-08
Refactoring the Supercluster rendering for easy testing

goemen:
The supercluster render in the PointCluster component was easy to test, interacting directly with leaflet in the react tests is not any playground. I refactored the component to use react component in the render. 

I used this refactored code to write unit tests, showcasing how we can write more tests in the leaflet component

### DIFF
--- a/frontend/src/components/maps/leaflet/InventoryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/InventoryLayer.tsx
@@ -1,14 +1,14 @@
-import { GEO_PROPERTIES, IGeoSearchParams } from 'constants/API';
-import { ENVIRONMENT } from 'constants/environment';
+import { IPropertyDetail } from 'actions/parcelsActions';
+import { IGeoSearchParams } from 'constants/API';
 import { BBox, Feature } from 'geojson';
 import useDeepCompareEffect from 'hooks/useDeepCompareEffect';
-import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 import { LatLngBounds } from 'leaflet';
 import React from 'react';
 import { useLeaflet } from 'react-leaflet';
 import { toast } from 'react-toastify';
 import { PointFeature } from '../types';
 import PointClusterer from './PointClusterer';
+import { useApi } from 'hooks/useApi';
 
 export type InventoryLayerProps = {
   /** Latitude and Longitude boundary of the layer. */
@@ -23,6 +23,8 @@ export type InventoryLayerProps = {
   filter?: IGeoSearchParams;
   /** What to do when the point feature is clicked. */
   onMarkerClick?: (point: PointFeature, position?: [number, number]) => void;
+
+  selected?: IPropertyDetail | null;
 };
 
 /**
@@ -49,10 +51,11 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
   maxZoom,
   filter,
   onMarkerClick,
+  selected,
 }) => {
-  const keycloak = useKeycloakWrapper();
   const [features, setFeatures] = React.useState<Array<PointFeature>>([]);
   const { map } = useLeaflet();
+  const { loadProperties } = useApi();
 
   if (!map) {
     throw new Error('<InventoryLayer /> must be used under a <Map> leaflet component');
@@ -67,12 +70,16 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
       setBbox(getBbox(bounds));
     };
     map.on('moveend', rebound);
+
+    return () => {
+      map.off('moveend');
+    };
   }, [map]);
 
   minZoom = minZoom ?? 0;
   maxZoom = maxZoom ?? 18;
 
-  const url = `${ENVIRONMENT.apiUrl}${GEO_PROPERTIES({
+  const params = {
     bbox: (bounds ?? map.getBounds()).toBBoxString(),
     address: filter?.address,
     administrativeArea: filter?.administrativeArea,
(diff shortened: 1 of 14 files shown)


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -67,12 +70,16 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
       setBbox(getBbox(bounds));
     };
     map.on('moveend', rebound);
+
+    return () => {
+      map.off('moveend');
```
Fosol, line 75: Removes listener?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -84,34 +91,32 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
+            } as PointFeature;
+          });
         setFeatures(points);
       })
       .catch(error => {
         toast.error((error as Error).message, { autoClose: 7000 });
         console.error(error);
       });
-  }, [url, bbox]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 118: Is this because the linter doesn't understanding `useDeepCompareEffect(...)`?

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -100,6 +100,29 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
   return createSingleMarker(feature, latlng);
 };
 
+export const getMarkerIcon = (feature: ICluster) => {
```
Fosol, line 103: Documentation

frontend/src/components/maps/leaflet/Spiderfier.ts
```diff
@@ -69,12 +66,12 @@ export class Spiderfier {
   }
 
   // expand a cluster (spiderfy)
-  spiderfy(cluster: ICluster) {
+  spiderfy(cluster: ICluster): { lines?: any[]; markers?: any[] } {
```
Fosol, line 69: I don't understand this change?

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -89,47 +83,145 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+  );
 
   return (
-    <GeoJsonLayer
-      ref={ref}
-      data={clusters as any}
-      pointToLayer={pointToLayer}
-      onclick={onLayerClick}
-    ></GeoJsonLayer>
+    <>
```
Fosol, line 113: Why was this change necessary?

Fosol (approved):

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -158,96 +162,73 @@ describe('MapProperties View', () => {
         </Router>
       </Provider>,
     );
-    await act(() => (promise as unknown) as Promise<void>);
+    wait(() => expect(mapRef.current).toBeDefined(), { timeout: 500 });
```
agalbrai, line 165: what's the difference between wait() and await act()?

agalbrai (commented):
I'm a little fuzzy on all the details so I won't approve it but I assume it works well :)

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -223,7 +222,6 @@ const Map: React.FC<MapProps> = ({
   const closeMarkerPopup = (e: any) => {
     if (e.target._animateToZoom === mapRef.current?.leafletElement.getZoom()) {
       setLayerPopup(undefined);
-      dispatch(storeParcelDetail(null));
```
devinleighsmith, line 226: Can you confirm that this doesn't adversely affect the 'zoom' functionality please?

frontend/src/components/maps/leaflet/MapProperties.test.tsx
devinleighsmith: wait is used to allow an expectation to finish before continuing. Behind the scenes the wait runs in a loop, and just ignores any failures or errors until the timeout is reached. 

act is used to wrap an action that will cause a change in a rendered component. For example, clicking a submit button. Sometimes the act will also need to check for something to be true before exiting.

Generally use wait if you are waiting for something that isn't a direct result of one of your actions. Use act otherwise.

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -158,96 +162,73 @@ describe('MapProperties View', () => {
...
-            zoom={14}
-            properties={mockParcels}
-            selectedProperty={mockDetails}
-            agencies={[]}
-            propertyClassifications={[]}
-            lotSizes={[]}
-            onMarkerClick={jest.fn()}
-            mapRef={mapRef}
-          />
+          <div style={{ width: 500, height: 500 }}>
```
devinleighsmith, line 178: Ah, so this ended up being required?

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -158,96 +162,73 @@ describe('MapProperties View', () => {
...
     const { getByText } = render(
       <Provider store={store}>
         <Router history={history}>
           <ParcelPopupView parcel={emptyDetails} />
         </Router>
       </Provider>,
     );
-    await act(() => (promise as unknown) as Promise<void>);
+
     const alert = getByText('Failed to load parcel details.');
```
devinleighsmith, line 219: How is this test working with the mock being removed? does it throw a warning?

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -89,47 +83,145 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+      {clusters.map((cluster, index) => {
+        // every cluster point has coordinates
+        const [longitude, latitude] = cluster.geometry.coordinates;
+        // the point may be either a cluster or a crime point
```
devinleighsmith, line 117: what is a crime point?

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: This is a pretty big render block with a lot of different categories of functionality. Any way we could split this up? Even just creating functions for these for readability would be fine.

frontend/src/features/properties/map/MapView.tsx
```diff
@@ -91,7 +91,17 @@ const MapView: React.FC<MapViewProps> = (props: MapViewProps) => {
           selectedDraftProperty?.parcelDetail?.longitude ??
           defaultLatLng.lng
         }
-        properties={properties}
+        properties={
```
devinleighsmith, line 94: Hmm, what's going on here.

frontend/src/hooks/useApi.ts
```diff
@@ -113,5 +116,12 @@ export const useApi = (): PimsAPI => {
     return data;
   };
 
+  axios.loadProperties = async (params?: IGeoSearchParams): Promise<any[]> => {
+    const { data } = await axios.get<any[]>(
```
devinleighsmith, line 120: Please at least some basic error handling logic since this is a core application request that fails periodically.

devinleighsmith: Also I think this will not display the loading bar anymore right? Refactoring is great but lets not regress functionality.

devinleighsmith:
@goemen I think this needs to be demoed to @Fosol or myself as this PR contains a lot of map changes that aren't covered by existing tests.

devinleighsmith (changes requested):
Please setup a demo for this before merging.

devinleighsmith (approved):

PR #667, opened 2020-12-08
Removing Server Side Cache

Fosol:
Originally wanted to test the ability to quickly add server side caching for some endpoints, but I think it's causing memory issues in the docker container.

### DIFF
--- a/backend/api/Areas/Property/Controllers/SearchController.cs
+++ b/backend/api/Areas/Property/Controllers/SearchController.cs
@@ -2,7 +2,6 @@ using MapsterMapper;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
-using NetTopologySuite.IO;
 using Pims.Api.Areas.Property.Models.Search;
 using Pims.Api.Helpers.Exceptions;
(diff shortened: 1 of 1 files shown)


backend/api/Areas/Property/Controllers/SearchController.cs
```diff
@@ -2,7 +2,6 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
-using NetTopologySuite.IO;
```
devinleighsmith, line 5: this is unrelated right? I wonder if we could get basic static analysis going for .net at some point.

devinleighsmith (approved):

jdtoombs (approved):

agalbrai (approved):
Did some reading to understand what was happening 👍

goemen:
Looks good to me

goemen (approved):

backend/api/Areas/Property/Controllers/SearchController.cs
Fosol: Unrelated, but an unused `using`.  It would be helpful to have more linting on the backend, but I don't have time to address.

PR #671, opened 2020-12-09
v01.04.00

Fosol:
Enhancing Inventory to support buildings without land or on multiple parcels.  Correcting SPL Workflow and Importing SIS Disposal Projects.  Updating SSO domain name across environments.    Numerous enhancements to Disposal Projects and Excel export reports.  Adding GeoJSON endpoints for inventory.  Enhancing Map layer selection pane.  General UX/UI improvements.  Transition Planning.


devinleighsmith:
@Fosol what are we looking for in a review like this? Just that all of the included commits make sense? or are you concerned that the merge may have introduced an issue and need us to review all of the commits individually?

Fosol:
No review is necessary.  I just use the same process to update the `master` branch for record keeping.  It's composed of PRs that have already been reviewed, so we should be okay.

PR #672, opened 2020-12-09
PA-2141 Imported Projects Assign to Agency

Fosol:
- Code cleanup
- Additional unit-tests

### DIFF
--- a/backend/dal/Services/Admin/Concrete/ProjectService.cs
+++ b/backend/dal/Services/Admin/Concrete/ProjectService.cs
@@ -178,19 +178,22 @@ namespace Pims.Dal.Services.Admin
             projects.ThrowIfNull(nameof(projects));
             this.User.ThrowIfNotAuthorized(Permissions.SystemAdmin, Permissions.AgencyAdmin);
 
-            projects.ForEach((project) =>
+            if (projects.Any())
             {
-                if (project == null) return;
-
-                if (String.IsNullOrWhiteSpace(project.ProjectNumber))
+                projects.ForEach((project) =>
                 {
-                    project.ProjectNumber = $"TEMP-{DateTime.UtcNow.Ticks:00000}";
-                }
+                    if (project == null) return;
 
-                this.Context.Projects.Add(project);
-            });
+                    if (String.IsNullOrWhiteSpace(project.ProjectNumber))
+                    {
+                        project.ProjectNumber = $"TEMP-{DateTime.UtcNow.Ticks:00000}";
+                    }
 
-            this.Context.CommitTransaction();
+                    this.Context.Projects.Add(project);
+                });
+
+                this.Context.CommitTransaction();
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


backend/dal/Services/Admin/Concrete/ProjectService.cs
```diff
@@ -178,19 +178,22 @@ public void Add(IEnumerable<Project> projects)
...
                 {
-                    project.ProjectNumber = $"TEMP-{DateTime.UtcNow.Ticks:00000}";
-                }
+                    if (project == null) return;
 
-                this.Context.Projects.Add(project);
-            });
+                    if (String.IsNullOrWhiteSpace(project.ProjectNumber))
+                    {
+                        project.ProjectNumber = $"TEMP-{DateTime.UtcNow.Ticks:00000}";
```
jdtoombs, line 189: I know this isn't new but out of curiosity, is it pretty common to use `DateTime` to generate unique identifiers?

jdtoombs (approved):

goemen (commented):
Looks good to me

goemen (approved):

backend/dal/Services/Admin/Concrete/ProjectService.cs
devinleighsmith: I haven't really seen this before but that is just me. I've seen random GUIDS which is kinda similar.

backend/tests/core/ServiceHelper.cs
```diff
@@ -142,6 +142,46 @@ public static class ServiceHelper
...
+        /// Creates an instance of an object of the specified 'T' type and initializes it with the specified arguments.
+        /// Will use any 'args' passed in instead of generating defaults.
+        /// This will add the new instance to the ServiceCollection, so that you can continue to configure your ServiceCollection before building it.
+        /// </summary>
+        /// <typeparam name="TService"></typeparam>
+        /// <typeparam name="TImplmentation"></typeparam>
+        /// <param name="helper"></param>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        public static TService Add<TService, TImplmentation>(this TestHelper helper, params object[] args)
```
devinleighsmith, line 175: Implmentation

backend/tests/core/TestHelperExtensions.cs
```diff
@@ -22,17 +25,26 @@ public static class TestHelperExtensions
...
 
+            var tempProvider = helper.Services.BuildServiceProvider();
+
+            var types = new Dictionary<Type, object>();
             var gmock = typeof(Mock<>);
             var cargs = ci.GetParameters();
             foreach (var carg in cargs)
             {
-                if (helper.Services.Any(s => s.ServiceType == carg.ParameterType)) continue;
+                var exists = tempProvider.GetService(carg.ParameterType);// helper.Services.FirstOrDefault(sd => sd.ServiceType == carg.ParameterType);
```
devinleighsmith, line 42: comment intentional?

backend/tests/core/TestHelperExtensions.cs
```diff
@@ -22,17 +25,26 @@ public static class TestHelperExtensions
         /// <param name="helper"></param>
         /// <param name="args"></param>
         /// <typeparam name="T"></typeparam>
-        public static void MockConstructorArguments<T>(this TestHelper helper, params object[] args)
+        /// <returns></returns>
```
devinleighsmith, line 28: returns a Dictionary?

backend/tests/unit/api/Controllers/Tools/ImportControllerTest.cs
```diff
@@ -759,6 +761,85 @@ public void ImportProjects_WithNoNote_Success()
...
+                ProjectNumber = "TEST-00001",
+                Description = "Model 1",
+                Agency = "TEST",
+                Activity = activity,
+                Status = "TEST",
+                Risk = "GREEN",
+                ReportedFiscalYear = 2020,
+            };
+
+            var adminServiceMock = helper.GetService<Mock<IPimsAdminService>>();
```
devinleighsmith, line 813: not convinced the mock needs to be part of the test (could be in a before block), but since there is only the one test probably not a big deal.

devinleighsmith (approved):

backend/dal/Services/Admin/Concrete/ProjectService.cs
Fosol: This isn't a perfect solution, as it is theoretically possible for more than one project to receive the same `tick`.  Although it's unlikely, especially in relation to our architecture and use-case.   Having an incrementing number is generally friendlier than a GUID.  It also sorts far better.

backend/tests/core/TestHelperExtensions.cs
Fosol: Good catch!  Just realized while this will work great for most use-cases, it is possible to have multiple arguments of the same type.  I do like dictionaries because they're easy to grab the correct key/value.  I'll switch to an array of `KeyValuePair<,>`

backend/tests/unit/api/Controllers/Tools/ImportControllerTest.cs
Fosol: The mock is part of the `Arrange`.  It's necessary to setup everything for the test.  Sometimes it makes sense to setup things for a number of tests, but that can also be problematic as there is no guarantee on thread-safe and order the tests are run.  Generally best practice is to keep each test independent.

PR #673, opened 2020-12-09
Minor improvements to resolve memory issues

Fosol:
The massive request for all properties is killing the DB and API.  Constant `Out of Memory` issues keep showing up.  Which results in a `Gateway Timeout` on the app.

This change reimplements `_.debounce(...)` to reduce the number of requests.  In addition a temporary DAL method that does not incur an additional request for project status information has been created to reduce the load.  In reality we shouldn't be making the project status request in this endpoint anyway, but I don't have time to refactor other features presently.

### DIFF
--- a/backend/dal/Services/Concrete/PropertyService.cs
+++ b/backend/dal/Services/Concrete/PropertyService.cs
@@ -84,6 +84,39 @@ namespace Pims.Dal.Services
         }
 
         /// <summary>
+        /// Get an array of properties within the specified filters.
+        /// Will not return sensitive properties unless the user has the `sensitive-view` claim and belongs to the owning agency.
+        /// Note that the 'parcelFilter' will control the 'page' and 'quantity'.
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public IEnumerable<ProjectProperty> Search(AllPropertyFilter filter)
+        {
+            this.User.ThrowIfNotAuthorized(Permissions.PropertyView);
+            filter.ThrowIfNull(nameof(filter));
+            if (!filter.IsValid()) throw new ArgumentException("Argument must have a valid filter", nameof(filter));
+
+            var parcelFilter = (ParcelFilter)filter;
+            var buildingFilter = (BuildingFilter)filter;
+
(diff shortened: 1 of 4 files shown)


backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -83,6 +83,39 @@ where projectNumbers.Contains(p.Project.ProjectNumber)
...
         }
 
+        /// <summary>
+        /// Get an array of properties within the specified filters.
+        /// Will not return sensitive properties unless the user has the `sensitive-view` claim and belongs to the owning agency.
+        /// Note that the 'parcelFilter' will control the 'page' and 'quantity'.
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public IEnumerable<ProjectProperty> Search(AllPropertyFilter filter)
```
Fosol, line 93: New DAL function that won't make an additional request for project status

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -93,28 +94,39 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
-    loadProperties(params)
-      .then(async (data: Feature[]) => {
-        const points = data
-          .filter(feature => {
-            return !(
-              feature.properties!.propertyTypeId === selected?.propertyTypeId &&
-              feature.properties!.id === selected?.parcelDetail?.id
-            );
+  const search = React.useCallback(
+    _.debounce(
```
Fosol, line 98: Debounce to stop hammering the API with requests.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -9,6 +9,7 @@ import { toast } from 'react-toastify';
 import { PointFeature } from '../types';
 import PointClusterer from './PointClusterer';
 import { useApi } from 'hooks/useApi';
+import _ from 'lodash';
```
goemen, line 12: You can do `import debounce from 'lodash/debounce';` since we only care about debounce

goemen (approved):

PR #680, opened 2020-12-11
PA-2114 Zoom to property when filtering on Map

goemen:
When using the filter using the filter bar the map should zoom to the results. The map will resume the default behavior, zooming and moving. I wrapped the inventory results with a featureGroup so that I can zoom to the bounds of the layers/markers on the group. 

I also remove some dead code in the map component, @devinleighsmith ping me if you have an questions

### DIFF
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -1,15 +1,8 @@
 import './Map.scss';
 
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect } from 'react';
 import axios from 'axios';
-import {
-  LatLngBounds,
-  LeafletMouseEvent,
-  LeafletEvent,
-  LatLng,
-  Map as LeafletMap,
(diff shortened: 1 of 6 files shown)


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -317,6 +247,13 @@ const Map: React.FC<MapProps> = ({
     }
   };
 
+  const handleBounds = (e: any) => {
+    const boundsData: LatLngBounds = e.sourceTarget.getBounds();
+    if (!isEqual(boundsData.getNorthEast(), boundsData.getSouthWest())) {
```
jdtoombs, line 252: I think I gotta start reading into more of the lodash capabilities, seem to have a lot of useful methods rather than doing vanilla JS operations

jdtoombs (approved):
Can't comment too much on the code removal as I have not done much of the work in these areas but from what I've read it looks good.

Fosol (approved):
I like the `FilterProvider` context implementation

PR #681, opened 2020-12-11
PA-2236 remove enablereinitialize from spl form.

devinleighsmith:


frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -153,14 +153,14 @@ const SplStep = ({ formikRef }: IStepProps) => {
       <Formik
         initialValues={initialValues}
         validateOnMount={true}
-        enableReinitialize={true}
```
devinleighsmith, line 156: So, we've had a few errors caused by this over the course of this project and so I'm starting to think we need to avoid using this in the future.

Basically what this does is reinitialize the entire formik form with the passed initialValues whenever anything above the formik form in the render tree changes. This causes all unsaved changes on the form to be wiped, which is what is happening in this case (all unsaved changes are being lost right before the form is submitted, or whenever the tab changes).

Generally there should be nothing above the formik form changing in the render tree so this is usually safe, but this assumption has proven to be brittle in this and other instances so I think we should try and avoid using it from now on to avoid situations like this and in the future.

The alternative is to manually update the form values with formik's setValues method. This ensures that you only overwrite the form with an explicit call, which is nice. It does mean that we need to be careful to remember to call setValues, or the form will end up out of sync with the DB causing concurrency exceptions. Overall I think those end up being easier to debug, so I think I prefer this approach.

frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -153,14 +153,14 @@ const SplStep = ({ formikRef }: IStepProps) => {
...
-        onSubmit={(values: IProject) => {
+        onSubmit={(values: IProject, actions) => {
           return onSubmitReview(
             values,
             formikRef,
             submitStatusCode,
             getStatusTransitionWorkflow(submitStatusCode),
           ).then((project: IProject) => {
+            actions.setValues(project);
```
devinleighsmith, line 163: This is required to sync the db project with the form (which will update the rowVersion at the lease).

frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -212,7 +212,9 @@ const SplStep = ({ formikRef }: IStepProps) => {
                       formikRef,
                       submitStatusCode,
                       getStatusTransitionWorkflow(submitStatusCode),
-                    );
+                    ).then((project: IProject) => {
+                      setValues(project);
```
devinleighsmith, line 216: This is an example of the gotcha with this approach. If you have more then one way of updating the object used by the form (besides the formik submit) you won't get those changes automagically anymore.

frontend/src/router.tsx
```diff
@@ -25,6 +25,10 @@ import EditAgencyPage from 'features/admin/agencies/EditAgencyPage';
 import { IENotSupportedPage } from 'features/account/IENotSupportedPage';
 
 const AppRouter: React.FC = () => {
+  const location = useLocation();
```
devinleighsmith, line 28: Snuck this in here: this ensures that changing the route loads the top of the newly loaded page.

frontend/src/components/common/form/Check.tsx
```diff
@@ -98,7 +98,10 @@ export const Check: React.FC<CheckProps> = ({
             value={!!checked}
             placeholder={placeholder}
             checked={!!checked}
-            onChange={() => setFieldValue(field, !checked)}
+            onChange={() => {
+              setFieldTouched(field);
```
devinleighsmith, line 102: while debugging noticed that touching the check fields wasn't being recorded in formik.

frontend/src/components/common/form/Check.scss
```diff
@@ -7,3 +8,10 @@
     padding-right: 5px;
   }
 }
+.required.form-group {
```
devinleighsmith, line 11: Sorry, another distraction. When we removed all of the required asterixes we forgot to add something to the checkboxes to indicate required.

I played around a bit and came up with this (note that styling the checkbox box itself in css is actually extremely annoying):

image

frontend/src/features/projects/spl/forms/SurplusPropertyListForm.scss
```diff
@@ -14,7 +14,7 @@
   .form-check-label {
     font-size: 12px;
     font-family: 'BCSans', Fallback, sans-serif;
-    padding-right: 1rem;
```
devinleighsmith, line 17: In most cases margin should be used to provide gaps between adjacent fields not padding.

frontend/src/components/common/form/Check.scss
jdtoombs: Looks good to me, seems to fit the rest of our required styling as well

jdtoombs (approved):

frontend/src/features/projects/spl/steps/SplStep.tsx
Fosol: I would have taken days to debug this...  Thanks!

frontend/src/features/projects/spl/steps/SplStep.tsx
Fosol: We need to remember this in all our forms, as we need to implement optimistic concurrency.

Fosol (approved):
Much appreciated!

PR #682, opened 2020-12-11
PA-2234 Fixing map filter

Fosol:
Due to rushing in last minute PRs we broke the map filter bar.  This should resolve those issues.  However I have discovered some other issues along the way that will need to be reviewed.  Regrettably there appears to be inconsistent behaviour which results in the following bugs showing up from time-to-time;

- agencies filter input `undefined` and not functioning
- An infinite loop that results in a React error `Maximum update depth exceeded`
- Pins remaining on the map that shouldn't be there after a filter has been entered

### DIFF
--- a/frontend/src/components/maps/MapFilterBar.tsx
+++ b/frontend/src/components/maps/MapFilterBar.tsx
@@ -35,7 +35,7 @@ const SearchBar: React.FC = () => {
   const {
     values: { searchBy },
     setFieldValue,
-  } = useFormikContext<MapFilterChangeEvent>();
+  } = useFormikContext<IMapFilter>();
   const desc = state.placeholders[searchBy] || '';
 
   const reset = () => {
@@ -56,7 +56,7 @@ const SearchBar: React.FC = () => {
   );
 };
 
-export interface MapFilterChangeEvent extends BasePropertyFilter {
+export interface IMapFilter extends BasePropertyFilter {
   searchBy: string;
   pid: string;
   address: string;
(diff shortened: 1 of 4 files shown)


frontend/src/components/maps/MapFilterBar.tsx
```diff
@@ -35,7 +35,7 @@ const SearchBar: React.FC = () => {
   const {
     values: { searchBy },
     setFieldValue,
-  } = useFormikContext<MapFilterChangeEvent>();
+  } = useFormikContext<IMapFilter>();
```
Fosol, line 38: It isn't an event, it's an object with filter parameters.  It is also an interface, so renamed accordingly.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -86,8 +86,8 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
 
   const search = React.useCallback(
     _.debounce(
-      () => {
-        loadProperties(params)
+      (filter: IGeoSearchParams) => {
```
Fosol, line 89: filter params were not being passed to the query

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -106,17 +121,8 @@ const Map: React.FC<MapProps> = ({
...
-    searchBy: 'address',
-    address: '',
-    administrativeArea: '',
-    projectNumber: '',
-    agencies: '',
-    classificationId: '',
-    minLotSize: '',
-    maxLotSize: '',
-  });
+  const [mapFilter, setMapFilter] = useState<IMapFilter>(defaultMapFilter);
```
Fosol, line 124: If I find time I'll get this down to management of one single state.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -125,7 +131,6 @@ const Map: React.FC<MapProps> = ({
   useRouterFilter(mapFilter, setMapFilter, 'mapFilter');
   const municipalitiesService = useLayerQuery(MUNICIPALITY_LAYER_URL);
   const parcelsService = useLayerQuery(PARCELS_LAYER_URL);
-  const filterState = useFilterContext();
```
Fosol, line 128: Had to remove because of infinite loop

frontend/src/components/maps/MapFilterBar.tsx
jdtoombs: Ooh, reminds me I should do the same thing with `SelectOption` and `SelectOptions`

jdtoombs (approved):

agalbrai (commented):
I hope this fixes things, I'm not much help here :/

PR #684, opened 2020-12-11
PA-2232 Export all land and buildings in one spreadsheet

agalbrai:


frontend/src/features/properties/list/FilterBar.tsx
```diff
@@ -140,13 +140,11 @@ const FilterBar: React.FC<FilterBarProps> = ({
               <SearchBar />
             </Col>
             <Col className="bar-item">
-              <ParentGroupedFilter
-                name="agencies"
+              <ParentSelect
```
agalbrai, line 143: Fix Agency filter to be new dropdown

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -40,12 +40,20 @@ const initialQuery: IPropertyFilter = {
...
-  pageIndex: number;
-  pageSize: number;
-  filter: IFilterBarState;
-  agencyIds: number[];
-}) => {
+/**
+ * Get the server query
+ * @param ignorePropType - Whether to ignore the PropertyTypes (for doing excel export with all values)
+ * @param state - Table state
+ */
```
agalbrai, line 47: Did I do this right?

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -283,12 +291,12 @@ const PropertyListView: React.FC = () => {
           </div>
           <TooltipWrapper toolTipId="export-to-excel" toolTip="Export to Excel">
             <FileIcon>
-              <FaFileExcel size={36} onClick={() => fetch('excel')} />
+              <FaFileExcel size={36} title="Export to Excel" onClick={() => fetch('excel')} />
```
agalbrai, line 294: so it can be found in tests

frontend/src/features/properties/list/FilterBar.tsx
jdtoombs: Thanks!

agalbrai: Np!

frontend/src/features/properties/list/PropertyListView.tsx
jdtoombs: This will actually cause some text to appear when you hover over it. If this is desired that's fine but I also see we have a tooltip wrapper here which makes it a little redundant. Another option is to add a `data-testid` attribute. After that you can query by it. See more [here](https://testing-library.com/docs/dom-testing-library/api-queries/#bytestid). 

I see that we do this on the project page though, so maybe this is something we could clean up there as well. See pic below for reference.

image

Fosol: We don't want both, and generally the tooltip is preferred over the `title` approach.  However the `title` approach may work better inside of input dropdowns.

agalbrai: I did notice that as well, I'll look into that Jonny

frontend/src/features/properties/list/PropertyListView.tsx
Fosol: You can test by hovering over code that calls `getServerQuery`.  It should show your description.

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -283,12 +291,12 @@ const PropertyListView: React.FC = () => {
...
             </FileIcon>
           </TooltipWrapper>
           <TooltipWrapper toolTipId="export-to-excel" toolTip="Export to CSV">
             <FileIcon>
-              <FaFileAlt size={36} onClick={() => fetch('csv')} />
+              <FaFileAlt size={36} title="Export to CSV" onClick={() => fetch('csv')} />
```
Fosol, line 299: I'd refrain from having the title and the tooltip.  I prefer the tooltip because you have more control.

agalbrai: Fixing this right now!

goemen (approved):
looks good to me

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -55,12 +55,20 @@ const defaultFilterValues: IPropertyFilter = {
...
-  filter: IPropertyFilter;
-  agencyIds: number[];
-}) => {
+/**
+ * Get the server query
+ * @param ignorePropType - Whether to ignore the PropertyTypes (for doing excel export with all values)
+ * @param state - Table state
+ */
+const getServerQuery = (
+  ignorePropertyType: boolean,
```
goemen, line 64: I would make this the second parameter.

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -320,7 +320,7 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
                 <FileIcon>
                   <FaFileExcel
                     size={36}
-                    title="Export to Excel"
+                    data-testid="excel-icon"
```
Fosol, line 323: Interesting, is this something that gets stripped by the transpiler for the production build, or is it just a simple way to use `queryByTestId(...)`?

Fosol (approved):

frontend/src/features/projects/list/ProjectListView.tsx
agalbrai: I have no idea about the first part, but I added it for ease of testing

PR #687, opened 2020-12-13
PA-2236 Cleaned Property Filter code

Fosol:
Organizing the code to be clearer and generic.

I'm not certain ignoring the exhaustive dependencies is ideal, but in this example we would only ever want this to execute the effect on Mount.

### DIFF
--- a/frontend/src/constants/trueFalse.ts
+++ b/frontend/src/constants/trueFalse.ts
@@ -1,0 +1,8 @@
+/**
+ * TrueFalse enum type to ensure strict compile time checking.
+ * Can't be named 'Boolean' because typescript has a type with that name.
+ */
+export enum TrueFalse {
+  True = 'true',
+  False = 'false',
+}
--- a/frontend/src/features/properties/filter/IPropertyFilter.ts
+++ b/frontend/src/features/properties/filter/IPropertyFilter.ts
@@ -1,3 +1,6 @@
+import { TrueFalse } from 'constants/trueFalse';
+import { PropertyTypes } from 'constants/propertyTypes';
+
 /**
  * Property filter options used by Formik.
  */
@@ -25,9 +28,9 @@ export interface IPropertyFilter {
   /** The maxium lot size of the property. */
   maxLotSize: string;
   /** Whether the property is in SPP. */
-  inSurplusPropertyProgram?: string;
+  inSurplusPropertyProgram?: TrueFalse;
   /** Whether the property is in ERP. */
-  inEnhancedReferralProcess?: string;
+  inEnhancedReferralProcess?: TrueFalse;
   /** Select on of the property types [Land, Building]. */
-  propertyType?: string;
+  propertyType?: PropertyTypes;
 }
--- a/frontend/src/features/properties/filter/PropertyFilter.tsx
+++ b/frontend/src/features/properties/filter/PropertyFilter.tsx
@@ -17,38 +17,49 @@ import { PropertyFilterOptions } from './';
 import { useRouterFilter } from 'hooks/useRouterFilter';
 import { IPropertyFilter } from './IPropertyFilter';
 import { TableSort } from 'components/Table/TableSort';
+import { TrueFalse } from 'constants/trueFalse';
 
-export type PropertyFilterProps = {
+/**
+ * PropertyFilter component properties.
+ */
+export interface IPropertyFilterProps {
+  /** The default filter to apply if a different one hasn't been set in the URL or stored in redux. */
   defaultFilter: IPropertyFilter;
+  /** An arry of agency lookup codes. */
   agencyLookupCodes: ILookupCode[];
+  /** An array of classification codes. */
   propertyClassifications: ILookupCode[];
+  /** Callback event when the filter is changed during Mount. */
   onChange: (filter: IPropertyFilter) => void;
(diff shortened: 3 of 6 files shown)


frontend/src/hooks/useMount.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
jdtoombs, line 1: Oh interesting! I've been wondering what the best way of doing the `componentDidMount` equivalent is since `useEffect` seemed to combine `componentDidMount`, `componentDidUpdate`, and `componentWillUnmount`

jdtoombs: I'm guessing just passing an empty dependency array to `useEffect` wasn't desired behaviour because you also want it to run on unmount?

jdtoombs (approved):

frontend/src/hooks/useMount.ts
Fosol: Just using `useEffect(..., [])` would be the same thing, but it'll complain about missing dependencies.  This way it's obvious what we're doing and reduces the extra lines of code.

jdtoombs: Ah gothcya! Makes sense

frontend/src/hooks/useMount.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Accepts a function that contains imperative, possibly effectful code.
+ * This function will only be run on Mount and Unmount.
+ * @param effect Imperative function that can return a cleanup function.
+ */
+export const useMount = (effect: React.EffectCallback) => {
```
jdtoombs, line 8: Ah just saw the callback and the empty dependency array, I take back my previous question lol

jdtoombs: This is cleaner though!

frontend/src/constants/trueFalse.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * TrueFalse enum type to ensure strict compile time checking.
+ * Can't be named 'Boolean' because typescript has a type with that name.
+ */
+export enum TrueFalse {
+  True = 'true',
```
Fosol, line 6: Use this instead of magic string Boolean values.

frontend/src/features/properties/filter/IPropertyFilter.ts
```diff
@@ -25,9 +28,9 @@ export interface IPropertyFilter {
   /** The maxium lot size of the property. */
   maxLotSize: string;
   /** Whether the property is in SPP. */
-  inSurplusPropertyProgram?: string;
+  inSurplusPropertyProgram?: TrueFalse;
```
Fosol, line 31: Compile time safety.

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -16,32 +16,41 @@ import { ParentSelect } from 'components/common/form/ParentSelect';
 import { PropertyFilterOptions } from './';
 import { useRouterFilter } from 'hooks/useRouterFilter';
 import { IPropertyFilter } from './IPropertyFilter';
+import { TrueFalse } from 'constants/trueFalse';
 
-export type PropertyFilterProps = {
+/**
+ * PropertyFilter component properties.
+ */
+export interface IPropertyFilterProps {
```
Fosol, line 24: Generally it makes far more sense to use an `interface` than a `type`.

frontend/src/constants/trueFalse.ts
agalbrai: why does it need to be a string rather than an actual boolean?

Fosol: Good question and one I'd like to know too!  It appears all the Formik forms require the use of string values.  I haven't had opportunity to look into why we don't have a wrapper at that level rather than having to deal with it higher up in the food chain.

agalbrai: huh, weird

goemen: @Fosol can we chat about this enum and its usage. Something does not seem right

frontend/src/hooks/useRouterFilter.ts
```diff
@@ -5,10 +5,13 @@ import { saveFilter } from 'reducers/filterSlice';
 import { RootState } from 'reducers/rootReducer';
 import _ from 'lodash';
 import queryString from 'query-string';
+import { useMount } from './useMount';
 
 /**
  * Extract the specified properties from the source object.
  * Does not extract 'undefined' property values.
+ * This provides a consistent deconscrutor implementation.
```
devinleighsmith, line 13: deconscrutor

frontend/src/constants/trueFalse.ts
devinleighsmith: formik doesn't require all of its values to be strings, we could easily represent boolean flags as booleans as is done elsewhere in the application (such as the edit field in the edit user page). the one caveat is that all booleans require a non-null non-undefined initial value, so either true or false. If that doesn't work, you would need a boolean | '' type with an initial value of ''.

essentially the field itself is responsible for determining what gets saved to formik. In this case it is saving string representations of true and false, which is likely incorrect.

PR #688, opened 2020-12-14
Vertically stack search fields, seperate search from form logic on la…

devinleighsmith:
…nd and associated land forms.

clean up some of the parcel/layer fetch logic.

### DIFF
--- a/frontend/src/features/mapSideBar/SidebarContents/AssociatedLandForm.tsx
+++ b/frontend/src/features/mapSideBar/SidebarContents/AssociatedLandForm.tsx
@@ -31,6 +31,7 @@ import { useBuildingApi } from '../hooks/useBuildingApi';
 import _ from 'lodash';
 import { IFormParcel } from '../containers/MapSideBarContainer';
 import { useState } from 'react';
+import { Prompt } from 'react-router-dom';
 import { defaultBuildingValues } from './BuildingForm';
 
 const Container = styled.div`
@@ -138,6 +139,8 @@ export const valuesToApiFormat = (
     const parcelApiValues = landValuesToApiFormat({ data: p } as any);
     if (!!agencyId && p.agencyId === '') {
       parcelApiValues.agencyId = agencyId;
+    } else {
+      parcelApiValues.agencyId = +(parcelApiValues.agencyId as any).value;
     }
     return parcelApiValues;
   });
@@ -178,7 +181,11 @@ const Form: React.FC<IAssociatedLandForm> = ({
   const agencies = getOptionsByType(API.AGENCY_CODE_SET_NAME);
   const classifications = getOptionsByType(API.PROPERTY_CLASSIFICATION_CODE_SET_NAME);
   const currentParcelNameSpace = `data.parcels.${stepper.currentTab}`;
-  useParcelLayerData({ formikRef, nameSpace: currentParcelNameSpace });
+  useParcelLayerData({
+    formikRef,
+    nameSpace: currentParcelNameSpace,
+    agencyId: +formikProps.values.data.agencyId,
+  });
 
   const render = (): React.ReactNode => {
     switch (stepper.current) {
@@ -283,13 +290,19 @@ interface IAssociatedLandForm {
   isAdmin: boolean;
 }
 
+interface IAssociatedLandParentForm extends IAssociatedLandForm {
+  /** signal the parent that the associated land process has been completed. */
+  setAssociatedLandComplete: (show: boolean) => void;
+}
+
 /**
  * A component used for land associated to a building.
  * This form will appear after a user enters a new building after navigating to Manage Property > Submit Property in PIMS
  * @component
  */
-
-const AssociatedLandForm: React.FC<IAssociatedLandForm> = (props: IAssociatedLandForm) => {
+const AssociatedLandForm: React.FC<IAssociatedLandParentForm> = (
+  props: IAssociatedLandParentForm,
+) => {
   const keycloak = useKeycloakWrapper();
   const { createBuilding, updateBuilding } = useBuildingApi();
   const dispatch = useDispatch();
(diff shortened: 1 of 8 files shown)


frontend/src/features/mapSideBar/SidebarContents/AssociatedLandForm.tsx
```diff
@@ -138,6 +139,8 @@ export const valuesToApiFormat = (
     const parcelApiValues = landValuesToApiFormat({ data: p } as any);
     if (!!agencyId && p.agencyId === '') {
       parcelApiValues.agencyId = agencyId;
+    } else {
+      parcelApiValues.agencyId = +(parcelApiValues.agencyId as any).value;
```
devinleighsmith, line 143: the agency Id is currently an object as set by the new agency component. Already discussed this with Johnny, one of us will take a longer look at fixing this time permitting.

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.tsx
```diff
@@ -90,50 +87,73 @@ export const ParcelIdentificationForm: React.FC<IIdentificationProps> = ({
...
-                handlePidChange(pid, nameSpace);
-              } else if (pin?.length > 0) {
-                handlePinChange(pin, nameSpace);
-              } else if (geocoderResponse) {
-                handleGeocoderChanges(geocoderResponse, nameSpace);
-              }
-            }}
-          />
-          <Form.Row className="justify-content-end">
+          <Form.Row>
```
devinleighsmith, line 90: the search fields are now separate from the actual form data used by the PIMS backend. That means that the search fields no longer need to be required fields, and that the actual data is displayed below the search fields.

frontend/src/features/mapSideBar/containers/MapSideBarContainer.tsx
```diff
@@ -104,37 +104,51 @@ const MapSideBarContainer: React.FunctionComponent<IMapSideBarContainerProps> =
...
-  /** query pims for the given pid, set data within the form if match found. Fallback to querying the parcel data layer. */
-  const handlePidChange = (pid: string, nameSpace?: string) => {
-    fetchParcelsDetail({ pid } as any)(dispatch).then(resp => {
+  /**
+   * Attempt to fetch the parcel within PIMS matching the passed pid or pin value. If that request fails, make another request to the parcel layer with the same data.
+   * @param pidOrPin
+   * @param parcelLayerSearchCallback
+   * @param nameSpace
+   */
+  const fetchPimsOrLayerParcel = (
```
devinleighsmith, line 113: this method is now used in all areas where a pid/pin search occurs, which ensures consistency.

frontend/src/features/mapSideBar/containers/MapSideBarContainer.tsx
```diff
@@ -287,6 +301,9 @@ const MapSideBarContainer: React.FunctionComponent<IMapSideBarContainerProps> =
       case SidebarContextType.ADD_ASSOCIATED_LAND:
         if (propertyType !== 'land') {
           setPropertyType('land');
+        } else if (!buildingToAssociateLand?.id) {
```
devinleighsmith, line 304: if the user somehow loads the associated land form without an attached building, kick them out to the selector since there is no way to "fix" the underlying building object within the associated land form.

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.tsx
```diff
@@ -90,50 +87,73 @@ export const ParcelIdentificationForm: React.FC<IIdentificationProps> = ({
...
+              onClick={(e: any) => {
+                e.preventDefault();
+                handlePidChange(getIn(formikProps.values, withNameSpace('searchPid')), nameSpace);
+              }}
+            />
+          </Form.Row>
+          <Form.Row>
+            <Label>PIN</Label>
+            <FastInput
+              formikProps={formikProps}
```
Fosol, line 115: I wonder if we could use a `Context` for our Formik controls.  We're always passing down the props anyway.

Fosol (approved):

agalbrai (commented):
should look good :)

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.tsx
devinleighsmith: formik already puts all of its data into a context provider that is internal to formik. You can get access to that via useFormikContext().

This is required because I'm using the old style of memo checking provided in FastInput. Could log tech debt to remove this prop, but I'm not convinced we would see any value other then one less prop on the Fast* variants of fields.

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.scss
```diff
@@ -60,7 +74,11 @@
           justify-content: center;
         }
       }
+      .form-group {
+        margin-bottom: 0px;
+      }
       .form-row {
+        margin-bottom: 1rem;
```
jdtoombs, line 81: Silly question, but I've always been confused with `rem` -  all my understanding of it is that it is relative to the font size. Just wondering what scenarios `rem `is most often used over `px ` etc

jdtoombs (approved):

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.scss
devinleighsmith: We have some inconsistency in this area for sure. REM is indeed relative to the font size, so any area where spacing relative to the font size makes sense is where rems should be used (many areas).

PR #695, opened 2020-12-15
PA-2202 Added Exemption Approved On

Fosol:
When a Disposal Project is submitted with a request to be exempted from ERP the assessor must now include the date when the ADM approved the exemption.

image


backend/dal/Migrations/20201215161602_v01.05.01.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using Pims.Dal.Helpers.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010501 : SeedMigration
```
Fosol, line 6: Need a minor update migration to run the script to update the task description.  This will get merged down to a major release `v01.05.00` at the end of the sprint.

PR #696, opened 2020-12-15
ensure that useApi uses the current bearer token stored in memory.

devinleighsmith:


Fosol (approved):
Thanks tons!

frontend/src/hooks/useApi.ts
```diff
@@ -46,6 +49,7 @@ export const useApi = (): PimsAPI => {
```
devinleighsmith, line 49: The issue is that the axios instance is created once when the page is loaded and then persists until this hook is recreated (so probably only when a parent is re-rendered). 

Since the current token is only assigned when the custom axios instance is created at the top of this hook, the axios instance will not receive the updated token when the refresh fires in the background.

Other use* hooks in the application don't have this issue because they follow a pattern like this:
CustomAxios().get('blah').then(() => //do something);
meaning that the axios instance is created the instant before the request is sent, ensuring that the token is up to date.

devinleighsmith: I've tested locally to ensure that this change ensures the most up to date token is being sent out.

devinleighsmith: @goemen @jdtoombs @Fosol let me know if this makes sense as the above is critical infrastructure on the frontend.

Fosol: If I'm reading this correctly when axios makes a request you are intercepting it and updating the header to use the latest redux stored token.  It looks correct to me.

Fosol (approved):

frontend/src/hooks/useApi.ts
jdtoombs: Gotchya, good find.

jdtoombs (approved):

PR #703, opened 2020-12-16
Fixing map crashing and zoom to results after search

goemen:
Now subscribing to the map `ondragend` to monitor the map bounds and make API query to load more data. `onzoomend` was too spammy and causing the map to make unnecessary api calls, and infinite loops. 

Also reenabled zoom results after searching


frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -170,7 +166,7 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
             position={[latitude, longitude]}
             icon={getMarkerIcon(cluster)}
           >
-            <Popup>
+            <Popup autoPan={false}>
```
Fosol, line 169: What does this option do?

Fosol (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
goemen: Stops the map from refocussing to an open marker popup

devinleighsmith (approved):

PR #706, opened 2020-12-17
PA-1963 Populate details link only when submit inventory form open

agalbrai:


frontend/src/components/maps/leaflet/LayerPopup/LayerPopUpContent.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+          config={mockLayer.config}
+          onAddToParcel={mockLayer.onAddToParcel}
+        />
+      </Router>,
+    );
+    const link = queryByText(/Populate property details/i);
+    expect(link).toBeNull();
+  });
+
+  xit('Populate details link appears when sideBar open', () => {
```
agalbrai, line 68: The link wasn't appearing for this test, and I can't figure out why

frontend/src/components/maps/leaflet/LayerPopup/LayerPopUpContent.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { render, cleanup } from '@testing-library/react';
+import { IPopupContentProps, LayerPopupContent } from './LayerPopupContent';
+import { Router } from 'react-router-dom';
+import queryString from 'query-string';
+
+const history = createMemoryHistory();
+jest.mock('hooks/useApi');
+
+const mockLayer: IPopupContentProps = {
+  config: {},
```
agalbrai, line 13: I hope leaving this empty isn't causing issues?

Fosol: Entirely depends on if your test or the code requires properties to be defined within it.  Otherwise it's entirely acceptable.

frontend/src/components/maps/leaflet/LayerPopup/LayerPopUpContent.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  data: {
+    feature_area_sqm: '10000',
```
Fosol, line 15: Everything being a string highlights the issue I was mentioning in a prior PR.  I mentioned Formik appears to require this, but Devin mentioned it doesn't.  Not asking you to change anything, simply highlighting the ugliness of how it's currently implemented.

frontend/src/components/maps/leaflet/LayerPopup/LayerPopUpContent.test.tsx
Fosol: The link should only be displayed when the Property Detail View is present.  I suspect this test isn't mimicking the scenario correctly.

Fosol (approved):
Ideally you'd fix the unit-test

frontend/src/components/maps/leaflet/LayerPopup/LayerPopUpContent.test.tsx
agalbrai: it won't let me assign it anything but a string

Fosol: Yeah it won't because it has been designed to only allow strings.  You don't need to change anything here.  I'm just highlighting one of the issues we have in our implementation due to either Formik, javascript or something else.

devinleighsmith (approved):

PR #709, opened 2020-12-18
PA-2238 Replace ParentGroupedFilter with ParentSelect

agalbrai:


frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -57,13 +57,15 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
     if (!loaded) {
-      if (value?.value) {
-        setFieldValue(field, value);
-      } else if (value && !value.value) {
-        setFieldValue(field, option);
-      } else {
-        setFieldValue(field, option);
-        setLoaded(true);
+      if (value !== undefined) {
+        if (value?.value) {
```
agalbrai, line 61: A test kept failing unless this was added, I can't remember what the reasoning was but @jdtoombs figured it out

frontend/src/features/admin/users/ManageUsers.test.tsx
```diff
@@ -103,12 +106,7 @@ describe('Manage Users Component', () => {
         },
       });
     });
-    expect(getAllByRole('option')[0]).toHaveTextContent('agencyVal');
-  });
-
-  it('Does not display disabled agencies', () => {
-    const { queryByText } = testRender(getStore());
-    expect(queryByText('disabledAgency')).toBeNull();
```
agalbrai, line 111: For whatever reason the test can't actually see the dropdown options with ParentSelect, so this test wasn't actually testing anything

frontend/src/components/common/form/ParentSelect.tsx
devinleighsmith: I don't really see why the if statement on 60 is required, unless the else statement shouldn't execute when value is undefined?

frontend/src/features/admin/users/ManageUsers.test.tsx
devinleighsmith: something to try in situations like this is dumping the entire dom with prettydom (something like console.log(prettyDOM(container, 99999))

alternatively sometimes you need to wait for your condition to be true:
```
wait(() => {
//wait will run and expectations in a loop until it times out or the expectation stops failing.
  expect(true).toBeTruthy();
});
```

devinleighsmith (approved):

frontend/src/components/common/form/ParentSelect.tsx
jdtoombs: I think it does throw an error threw the test if I remember correctly, I feel like this logic could be rearranged a bit better (from what I did originally). But I have a ticket for cleaning this component up/fixing warnings so I could do it then!

jdtoombs (approved):

frontend/src/components/common/form/ParentSelect.tsx
Fosol: `if (value?.value)` will return false if `value === undefined`, so I'm not sure why the need for `if (value !== undefined)`

Fosol (approved):

PR #718, opened 2020-12-21
Implementing properties column filter

goemen:
Implementing table column filter. The table is configurable using column settings. The table header is wrapped with formik and render the filterable columns as a form. 
Hnet-image (9)

### DIFF
--- a/frontend/src/components/Table/ColumnFilter.tsx
+++ b/frontend/src/components/Table/ColumnFilter.tsx
@@ -1,0 +1,88 @@
+import * as React from 'react';
+import { ColumnInstanceWithProps } from '.';
+import styled from 'styled-components';
+import ClickAwayListener from 'react-click-away-listener';
+import { getIn, useFormikContext } from 'formik';
+import clsx from 'classnames';
+import { FiFilter } from 'react-icons/fi';
+import { FaFilter } from 'react-icons/fa';
+
+interface IColumnFilterProps {
+  /** Column instance from react table */
+  column: ColumnInstanceWithProps<any>;
+  /** Handle column filter change */
+  onFilter: (values: any) => void;
+}
+
+const Wrapper = styled('div')`
+  display: flex;
+  align-items: center;
+  justify-items: center;
+  position: relative;
+  font-size: 12px;
+`;
+
+const InputContainer = styled('div')`
+  position: absolute;
+  bottom: -55px;
+  left: 0;
+  z-index: 1;
+
+  .form-control {
+    border-radius: 0;
+    border-color: #606060;
+  }
+`;
+
+/**
+ * Component to display column filter component, used internally by the table
+ */
+const ColumnFilter: React.FC<IColumnFilterProps> = ({ column, onFilter, children }) => {
+  const [open, setOpen] = React.useState(false);
+  const context = useFormikContext();
+
+  const handleClick = () => {
+    if (open) {
+      onFilter(context.values);
+      setOpen(false);
+    } else {
+      setOpen(true);
+    }
+  };
+
+  if (!column.filterable) {
+    return <>{children}</>;
+  }
+
+  if (!column.filter) {
+    throw new Error('Column filter settings are required when the column is filterable.');
+  }
+
+  const hasValue = !!getIn(context.values, (column.filter.props || {}).name);
+  const Control = column.filter!.component as any;
+
+  return (
+    <Wrapper className={clsx('filter-wrapper', { active: hasValue })}>
+      {hasValue ? (
+        <FaFilter onClick={handleClick} style={{ fontSize: 10, marginRight: 5 }} />
+      ) : (
+        <FiFilter onClick={handleClick} style={{ fontSize: 10, marginRight: 5 }} />
+      )}
+
+      <span onClick={handleClick}>{children}</span>
+      {open && (
+        <ClickAwayListener onClickAway={handleClick}>
+          <InputContainer>
+            {(column.filter?.props as any)?.injectFormik ? (
+              <Control {...column.filter?.props} formikProps={context} />
+            ) : (
+              <Control {...column.filter?.props} />
+            )}
+          </InputContainer>
+        </ClickAwayListener>
+      )}
+    </Wrapper>
+  );
+};
+
+export default ColumnFilter;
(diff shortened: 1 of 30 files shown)


frontend/src/hooks/useRouterFilter.ts
```diff
@@ -63,27 +63,30 @@ export const useRouterFilter = <T extends object>({
...
+        const sort = resolveSortCriteriaFromUrl(
+          typeof params.sort === 'string' ? [params.sort] : params.sort,
+        );
+        if (!_.isEmpty(sort)) {
+          setSorting(sort as any);
+        }
       }
     }
+
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
Fosol, line 90: This line shouldn't be required here as it is using `useMount(...)`

frontend/src/features/properties/list/columns.tsx
```diff
@@ -4,9 +4,13 @@ import { ReactComponent as LandSvg } from 'assets/images/icon-lot.svg';
 import React from 'react';
 import { CellProps } from 'react-table';
 import { Link } from 'react-router-dom';
-import { formatMoney, formatNumber } from 'utils';
+import { formatMoney, formatNumber, mapLookupCode } from 'utils';
 import { IProperty } from '.';
 import { ColumnWithProps } from 'components/Table';
+import { ParentGroupedFilter } from 'components/SearchBar/ParentGroupedFilter';
```
agalbrai, line 10: Why not use ParentSelect instead?

agalbrai (commented):
It feels a little weird to have both the filter bar and the filter options in the table, but maybe that's just me

frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  align-items: center;
+  justify-items: center;
+  position: relative;
+  font-size: 12px;
+`;
+
+const InputContainer = styled('div')`
+  position: absolute;
+  bottom: -55px;
```
devinleighsmith, line 26: Does this work ok if the columns get resized on smaller screens?

frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const handleClose = React.useCallback(() => {
+    onFilter(context.values);
+    setOpen(false);
+  }, [onFilter, context.values, setOpen]);
+
+  if (!column.filterable) {
+    return <>{children}</>;
+  }
+
+  const hasValue = !!getIn(context.values, column.filter?.props!.name);
```
devinleighsmith, line 48: the combination of ? and ! has me a bit worried here. So we don't know if filter.props will always be non-null/undefined, but props.name will always be specified?

frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const Control = (column.filter?.component || Form.Control) as any;
+
+  return (
+    <Wrapper
+      className={clsx('filter-wrapper', { active: hasValue })}
+      onClick={e => {
+        if ((e.target as any)?.className?.indexOf('filter-wrapper') !== -1) {
```
devinleighsmith, line 55: won't this onClick event always have a target element with classname of filter-wrapper? the clsx statement above should create two classnames, 'filter-wrapper' and then conditionally 'active' if hasValue is true.

frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -0,0 +1,83 @@
...
+          if (!open) {
+            setOpen(true);
+          } else {
+            handleClose();
+          }
+        }
+      }}
+    >
+      {hasValue && <FaFilter onClick={handleClose} style={{ fontSize: 10, marginRight: 5 }} />}
```
devinleighsmith, line 64: ternary here might make more sense.

frontend/src/components/Table/Table.tsx
```diff
@@ -390,7 +414,7 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
       );
     return (
       props.detailsPanel && (
-        <TooltipWrapper toolTipId="expand-all-rows" toolTip={open ? 'Collapse Row' : 'Expand Row'}>
+        <TooltipWrapper toolTipId="filter-rows" toolTip={open ? 'Collapse Row' : 'Expand Row'}>
```
devinleighsmith, line 417: the tooltip text should probably be update for this as well right?

frontend/src/components/Table/Table.tsx
```diff
@@ -439,24 +454,64 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
...
-                    'th',
-                    column.isSorted ? (column.isSortedDesc ? 'sort-desc' : 'sort-asc') : '',
-                  )}
-                >
-                  {renderHeaderCell(column)}
-                </div>
-              ))}
+              <Formik
+                initialValues={props.filter || {}}
+                enableReinitialize
```
devinleighsmith, line 459: I'm starting to feel like I want us to try and stop using this in new code because of all the trouble its caused us in previous bugs. Essentially my issue with this is if a component above this one re-renders then all of the data on this form will be wiped. If we can switch this to using a ref to update the contents that would be my preference, but open for discussion.

frontend/src/features/properties/list/columns.tsx
goemen: The parent group filter seems to work better for me, and the ParentSelector breaks the code

frontend/src/components/Table/ColumnFilter.tsx
goemen: yes it works

frontend/src/components/Table/ColumnFilter.tsx
goemen: Thanks

frontend/src/components/Table/Table.tsx
goemen: Ooops, this did not have to change

devinleighsmith:
> It feels a little weird to have both the filter bar and the filter options in the table, but maybe that's just me

Yeah, we talked about that, but that is what they want apparently.

Fosol:
The request was for excel like filtering and sorting.  I don't think much thought was put into the related details on the page.  They can come back with a new request to change the page filter if they want

devinleighsmith (approved):

agalbrai (approved):

Fosol (approved):

PR #719, opened 2020-12-21
Correct map not using bounds to filter api response.

devinleighsmith:
Critical bugfix ensures that the map boundary is actually used when passed to the API.

### DIFF
--- a/frontend/src/components/maps/leaflet/InventoryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/InventoryLayer.tsx
@@ -42,6 +42,19 @@ const getBbox = (bounds: LatLngBounds): BBox => {
 };
 
 /**
+ * Get a new instance of a BBox from the specified 'bounds'.
+ * @param bounds The latitude longitude boundary.
+ */
+const getApiBbox = (bounds: LatLngBounds): BBox => {
+  return [
+    bounds.getSouthWest().lng,
+    bounds.getNorthEast().lng,
+    bounds.getSouthWest().lat,
+    bounds.getNorthEast().lat,
+  ] as BBox;
+};
+
+/**
  * Displays the search results onto a layer with clustering.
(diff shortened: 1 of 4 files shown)


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -41,6 +41,19 @@ const getBbox = (bounds: LatLngBounds): BBox => {
   ] as BBox;
 };
 
+/**
+ * Get a new instance of a BBox from the specified 'bounds'.
+ * @param bounds The latitude longitude boundary.
+ */
+const getApiBbox = (bounds: LatLngBounds): BBox => {
+  return [
```
devinleighsmith, line 49: expected api envelope format is x,x,y,y

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -113,7 +114,10 @@ export const getMarkerIcon = (feature: ICluster) => {
     ].includes(projectStatus)
   ) {
     return erpIcon;
-  } else if (projectNumber !== undefined) {
+  } else if (
```
devinleighsmith, line 117: checking to see if the project number is populated as a way of choosing the purple pin doesn't match up with how our filters work.

Fosol: We don't generally return the project status of a property with the geoJSON endpoints.  Neither do we want to moving forward, as it will be another hit on performance.  We may need to rethink how the map displays coloured pins.

Fosol (approved):

frontend/src/components/maps/leaflet/mapUtils.tsx
devinleighsmith: ok... but is checking the classification like this ok? the projectStatus stuff predates my change.

goemen (approved):
Looks good to me

PR #723, opened 2020-12-24
Add support for new encumbrance reason field.

devinleighsmith:
Update geocoder logic to populate building form.
Update geococer to populate parcel forms when no layer or pims data available.
Add new fiscal field effective date, update related forms.
Remove dead code in geocoder.


frontend/src/features/properties/components/forms/subforms/columns.tsx
```diff
@@ -22,15 +22,28 @@ const getEditableMoneyCell = (disabled: boolean | undefined, namespace: string,
...
-  return (cellInfo: any) => {
-    return formatFiscalYear(cellInfo.row.values[field]);
-  };
+/**
+ * Create a formik date picker using the passed cellinfo to get the associated data.
+ * This information is only editable if this cell belongs to a parcel row.
+ * @param cellInfo provided by react table
+ * @param minDate restrict the minimum date that can be selected
+ * @param oldDateWarning warn if the user selects an old date
+ */
```
agalbrai, line 31: are the minDate and oldDateWarning params supposed to be there? I don't see them

frontend/src/features/properties/hooks/useGeocoder.tsx
```diff
@@ -98,48 +98,42 @@ const useGeocoder = ({ formikRef, fetchPimsOrLayerParcel }: IUseGeocoderProps) =
...
-          }
+        if (data.latitude && data.longitude) {
+          parcelsService
+            .findOneWhereContains({
+              lat: data.latitude,
+              lng: data.longitude,
+            } as LatLng)
+            .then(response => {
+              const pid = getIn(response, 'features.0.properties.PID');
+              //it is possible the the geocoder will fail to get the pid but the parcel layer service request will succeed. In that case, double check that the pid doesn't exist within pims.
```
agalbrai, line 118: the the

agalbrai (approved):

backend/entities/Property.cs
```diff
@@ -39,6 +39,11 @@ public abstract class Property : BaseEntity
         /// </summary>
         public PropertyClassification Classification { get; set; }
 
+        /// <summary>
+        /// get/set - The encumbrance reason for this property.
+        /// </summary>
+        public string EncumbranceReason { get; set; }
```
Fosol, line 45: You should update the `PropertyConfiguration` to set the max length of this field.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -24,11 +24,11 @@ public static class BuildingExtensions
         /// <exception type="DbUpdateException">The name within a parcel should be unique.</exception>
         public static void ThrowIfNotUnique(this PimsContext context, Entity.Parcel parcel, Entity.Building building)
         {
-            if(parcel == null)
+            if(parcel == null || String.IsNullOrEmpty(building.Name))
```
Fosol, line 27: Are you making the `name` a required field now?  Not sure the frontend has the same expectation.  Additionally this will most likely cause the import to fail as many buildings don't have names in the SIS data.

backend/dal/Migrations/20201223235542_01.06.00.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Pims.Dal.Migrations
+{
+    public partial class _010600 : SeedMigration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AddColumn<string>(
+                name: "EncumbranceReason",
```
Fosol, line 13: You should edit the `ParcelConfiguration` to set the correct size of this field.

backend/dal/Migrations/20201223235542_01.06.00.cs
```diff
@@ -0,0 +1,58 @@
...
+                nullable: true);
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "EffectiveDate",
+                table: "ParcelFiscals",
+                type: "DATE",
+                nullable: true);
+
+            migrationBuilder.AddColumn<string>(
+                name: "EncumbranceReason",
```
Fosol, line 24: You should edit the `BuildingConfiguration` to set the correct size of this field.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
devinleighsmith: I think this may have the opposite effect that you are thinking it does. If the building name is null or empty all of this duplicate name checking will be skipped.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -24,11 +24,11 @@ public static class BuildingExtensions
...
             {
                 return;
             }
-            var parcelBuildings = context.Parcels.Where(p => p.Id == parcel.Id).SelectMany(p => p.Buildings.Select(b => b.Building.Name)).Distinct().ToArray();
+            var parcelBuildings = context.Parcels.Where(p => p.Id == parcel.Id).SelectMany(p => p.Buildings).Where(b => b.BuildingId != building.Id).Select(b => b.Building.Name).Distinct().ToArray();
```
devinleighsmith, line 31: previously updating a building without changing its name would trigger a duplicate name exception, this ensures that duplicates are checked for all of the OTHER buildings on the parcel during an update, ignoring the building being updated.

frontend/src/components/common/TooltipWrapper.tsx
```diff
@@ -30,16 +30,19 @@ interface ITooltipWrapperProps extends Partial<OverlayTriggerProps> {
 const TooltipWrapper: React.FunctionComponent<ITooltipWrapperProps> = props => {
   return (
     <>
-      {props.toolTip?.length ? (
```
devinleighsmith, line 33: This conditional render was causing input fields to lose focus when they shifted from an in error to not in error state.

frontend/src/features/mapSideBar/SidebarContents/BuildingForm.tsx
```diff
@@ -186,12 +236,45 @@ interface IBuildingForm {
   setMovingPinNameSpace: (nameSpace: string) => void;
   /** to help determine the namespace of the field (eg. address.line1) */
   nameSpace: string;
-  /** Notify the parent that the building has been saved, potentially starting a new workflow. */
-  setBuildingToAssociateLand: (building: IBuilding) => void;
+  /** Go to the associated land form directly */
+  goToAssociatedLand: (building: IBuilding) => void;
```
devinleighsmith, line 240: this is for the modify associated land button, which doesn't need to go to the generate associated land page.

frontend/src/features/mapSideBar/SidebarContents/BuildingForm.tsx
```diff
@@ -239,28 +326,86 @@ const BuidingForm: React.FC<IBuildingForm> = ({
   let initialValues = {
```
goemen, line 326: change this to `const`

frontend/src/features/mapSideBar/SidebarContents/LandForm.tsx
```diff
@@ -299,15 +298,13 @@ const LandForm: React.FC<IParentLandForm> = (props: IParentLandForm) => {
   const keycloak = useKeycloakWrapper();
   const dispatch = useDispatch();
   const api = useApi();
-  let initialValueProps = {
+  let initialValues = {
```
goemen, line 301: change to `const`

goemen (approved):
Looks good to me

Fosol (commented):
Still missing `BuildingConfiguration` for the encumbrance reason.

PR #728, opened 2020-12-29
ERP/SPP checkboxes for filter

jdtoombs:
Adding checkboxes for ERP and SPP properties.  Alex's #726 will adjust the padding on the other sections
image


frontend/src/features/properties/filter/IPropertyFilter.ts
```diff
@@ -27,9 +25,9 @@ export interface IPropertyFilter {
   /** The maxium lot size of the property. */
   maxLotSize: string;
   /** Whether the property is in SPP. */
-  inSurplusPropertyProgram?: TrueFalse;
```
jdtoombs, line 30: Only switching because these fields are using a checkbox now

agalbrai (approved):

Fosol (approved):

PR #731, opened 2020-12-29
PA-2340 Change "Net Rentable Area" to "Net Usable Area" (dev)

agalbrai:
Same as the other pull request, this one just adds the changes in dev rather than alpha

### DIFF
--- a/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
+++ b/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
@@ -112,7 +112,7 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
             label="Construction type"
           />
           <VerticalLine />
-          <NumberInput label="Net rentable area" field="rentableArea" />
+          <NumberInput label="Net usable area" field="rentableArea" />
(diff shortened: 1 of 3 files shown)


frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
```diff
@@ -112,7 +112,7 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
             label="Construction type"
           />
           <VerticalLine />
-          <NumberInput label="Net rentable area" field="rentableArea" />
+          <NumberInput label="Net usable area" field="rentableArea" />
```
Fosol, line 115: The other labels use caps.  Any reason to keep this one small?

Fosol (approved):

jdtoombs (approved):

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
agalbrai: Seems that's how Bobbi wants it, I just checked the UX pin

PR #733, opened 2020-12-30
PA-2255 Enhanced Referral Process Agency Responses

Fosol:
User can now select specific agencies and add their responses to projects within ERP.

image

### DIFF
--- a/frontend/src/features/projects/common/components/FormikTable.tsx
+++ b/frontend/src/features/projects/common/components/FormikTable.tsx
@@ -21,7 +21,7 @@ const FormikTable: React.FC<IFormikTableProps> = ({
   disabled,
 }: IFormikTableProps) => {
   const { values } = useFormikContext();
-  const data = getIn(values, field);
+  const data = getIn(values, field) ?? [];
 
   return (
     <Container fluid className="PropertyListView">
@@ -37,6 +37,7 @@ const FormikTable: React.FC<IFormikTableProps> = ({
               footer
             />
           ),
+          // TODO: This is almost guaranteed to be an invalid list of dependencies for this Memo.
           // eslint-disable-next-line react-hooks/exhaustive-deps
           [data.length, name],
         )}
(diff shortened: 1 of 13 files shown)


jdtoombs (approved):
Looks good!

frontend/src/features/projects/erp/forms/AgencyInterest.tsx
```diff
@@ -0,0 +1,123 @@
...
+  );
+  const agencies = _.filter(lookupCodes, (lookupCode: ILookupCode) => {
+    return lookupCode.type === API.AGENCY_CODE_SET_NAME;
+  });
+  const agencyOptions = (agencies ?? []).map(c => mapLookupCodeWithParentString(c, agencies));
+
+  const onAddAgency = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>): void => {
+    event.preventDefault();
+    const selectedAgency = getIn(values, 'addAgencyResponse');
+    const agency = agencies.find(a => a.id == selectedAgency?.value);
```
jdtoombs, line 39: `==` -> `===`

Fosol: Actually this was intentional, but looks like our linting won't allow it.  Essentially `a.id` in typescript is a string, but during runtime it is a number.  Which means `===` will never be `true`...  I'll modify to handle this and make the code way harder to read...

frontend/src/features/projects/common/components/FormikTable.tsx
```diff
@@ -21,7 +21,7 @@ const FormikTable: React.FC<IFormikTableProps> = ({
   disabled,
 }: IFormikTableProps) => {
   const { values } = useFormikContext();
-  const data = getIn(values, field);
+  const data = getIn(values, field) ?? [];
```
Fosol, line 24: Fixes unit-tests where `data = null | undefined`

frontend/src/features/projects/erp/steps/ErpStep.tsx
```diff
@@ -50,8 +49,7 @@ const ErpStep = ({ formikRef }: IStepProps) => {
   const currentTab = useSelector<RootState, string | null>(state => state.erpTab) ?? defaultTab;
   const dispatch = useDispatch();
   const history = useHistory();
-  const { projectAgencyResponses } = useAgencyResponseTable();
-  const initialValues: IProject = { ...project, projectAgencyResponses: projectAgencyResponses };
```
Fosol, line 54: Don't populate the agency responses this way

agalbrai (commented):
That'll be way better, no more scrolling

PR #736, opened 2020-12-30
PA-2246 Merging DB Migrations

Fosol:
Merging Agency 'Crown Land Opportunities'
Updating Converter configuration

Since I've merged the migrations, DEV and TEST will need to be reset so that they run the latest scripts.


tools/converters/excel/appsettings.json
```diff
@@ -59,6 +59,10 @@
               {
                 "Search": "Insurance Coporation of BC",
                 "Value": "Insurance Corporation of BC"
+              },
+              {
+                "Search": "LNG, Crown Land Opportunities and Restoration",
+                "Value": "Crown Land Opportunities"
```
Fosol, line 65: Needed to update the Converter Tool so that future generated files will contain the correct agency name.

PR #739, opened 2020-12-31
Classification - remove disposed from non-sres

jdtoombs:
Originally had this checking the label not the value and when I switched it over to use the value and enum I forgot to cast it to a number

### DIFF
--- a/frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
+++ b/frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
@@ -47,7 +47,7 @@ const InformationForm: FunctionComponent<InformationFormProps> = (props: Informa
   /** only SRES can change to Disposed  */
   const classifications = keycloak.hasClaim(Claims.ADMIN_PROPERTIES)
     ? props.classifications
-    : props.classifications.filter(x => x.value !== Classifications.Disposed);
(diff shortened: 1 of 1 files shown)


frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
```diff
@@ -47,7 +47,7 @@ const InformationForm: FunctionComponent<InformationFormProps> = (props: Informa
   /** only SRES can change to Disposed  */
   const classifications = keycloak.hasClaim(Claims.ADMIN_PROPERTIES)
     ? props.classifications
-    : props.classifications.filter(x => x.value !== Classifications.Disposed);
+    : props.classifications.filter(x => Number(x.value) !== Classifications.Disposed);
```
Fosol, line 50: Ha this kind of stuff isn't supposed to occur if we use typescript appropriately apparently, but looks like the `any` value is getting us

Fosol (approved):

frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
jdtoombs: I think we have the `SelectOption` value set up as `string | number` and happens to be string in this case haha

jdtoombs: But `any `has definitely got me before

PR #752, opened 2021-01-06
PA-2365 prevent display of empty popup when clicking on clusters.

devinleighsmith:


frontend/src/components/maps/leaflet/Spiderfier.ts
```diff
@@ -71,6 +71,7 @@ export class Spiderfier {
 
     // only one cluster expanded at a time
     if (this.cluster === cluster || cluster == null) {
+      this.cluster = null;
```
devinleighsmith, line 74: also when the spiderfied results are closed currently it can't be re-opened, this fixes that.

Fosol (approved):

agalbrai (approved):
hurray! no more weird popup

jdtoombs (approved):

PR #759, opened 2021-01-07
PA-1932 Add Status and Risk Value to SPL Report

agalbrai:
image


frontend/src/features/splReports/containers/SplReportContainer.test.tsx
```diff
@@ -133,17 +138,6 @@ describe('Spl Report Container', () => {
       });
     });
 
-    it('Displays correct message when there are no snapshots', async () => {
```
agalbrai, line 136: This was a duplicate test

frontend/src/utils/utils.ts
```diff
@@ -301,3 +301,18 @@ export const isMouseEventRecent = (e?: MouseEvent | null) =>
  * @param squareMeters
  */
 export const squareMetersToHectares = (squareMeters: number) => (squareMeters / 10000).toFixed(2);
+
+/**
+ * Turn riskId from number into associated colour
+ * @param id riskId number
+ */
+export const formatRisk = (id: number) => {
+  switch (id) {
```
goemen, line 310: How do you handle the default case?

agalbrai: It appears I do not haha, I'll fix that

frontend/src/features/projects/common/interfaces.ts
```diff
@@ -194,6 +194,7 @@ export interface IProject {
   removalFromSplRequestOn?: Date | string;
   removalFromSplApprovedOn?: Date | string;
   removalFromSplRationale?: string;
+  riskId?: number;
```
Fosol, line 197: I think if you include `risk` you'll get a string name for each risk.  The `riskId` isn't really useful to the display.

frontend/src/features/splReports/columns.tsx
```diff
@@ -53,6 +53,25 @@ export const columns: any[] = [
     width: spacing.medium,
     minWidth: 80,
   },
+  {
+    Header: 'Status',
```
Fosol, line 57: This is the `Risk` column, not the `Status`.

frontend/src/utils/utils.ts
```diff
@@ -301,3 +301,20 @@ export const isMouseEventRecent = (e?: MouseEvent | null) =>
  * @param squareMeters
  */
 export const squareMetersToHectares = (squareMeters: number) => (squareMeters / 10000).toFixed(2);
+
+/**
+ * Turn riskId from number into associated colour
+ * @param id riskId number
+ */
+export const formatRisk = (id: number) => {
```
Fosol, line 309: Function not required, the name should come from the API `risk` column.

Fosol (changes requested):

frontend/src/features/splReports/columns.tsx
agalbrai: I know, but in the excel export it's labelled as "status"

Fosol: The Excel report regrettably must match an existing format so that the clients other reports will continue to work.  Within PIMS we can label them correctly.

frontend/src/features/splReports/columns.tsx
```diff
@@ -53,6 +53,22 @@ export const columns: any[] = [
     width: spacing.medium,
     minWidth: 80,
   },
+  {
+    Header: 'Risk',
+    accessor: 'project.risk',
```
goemen, line 58: Can we change this so that instead of displaying the color name, we display a badge with that color: https://react-bootstrap.github.io/components/badge/

goemen (changes requested):
I dont think displaying the risk color name conveys the right message, I recommend that we display a https://react-bootstrap.github.io/components/badge/ matching the color

frontend/src/features/splReports/columns.tsx
Fosol: This isn't required

Fosol (approved):

frontend/src/features/splReports/columns.tsx
agalbrai: I think if I add it, that one column would stand out a lot, perhaps too much?

goemen (approved):

PR #761, opened 2021-01-07
PA-2380 OCP4.0 Setup DEV Database

Fosol:
Setting up and configuring network security policies and standing up the DEV DB.


Fosol (commented):
Moved all old openshift files to new folder.

PR #763, opened 2021-01-07
Location filter bug fix

jdtoombs:
Note - I am going to make another ticket as I noticed some of the other fields were not clearing properly (eg. Agency). They clear when you hit search and then reset but if you fill in values without submitting and then hit reset they will not. Will assign ticket to myself for this sprint. 

Includes:
- bug fix for the clearing of location
- adding the `Typeahead` component to the `Surplus Properties` location search to allow for autocomplete

### DIFF
--- a/frontend/src/features/properties/filter/PropertyFilter.tsx
+++ b/frontend/src/features/properties/filter/PropertyFilter.tsx
@@ -1,6 +1,6 @@
 import './PropertyFilter.scss';
 
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import { Col } from 'react-bootstrap';
 import { Formik, getIn } from 'formik';
 import { ILookupCode } from 'actions/lookupActions';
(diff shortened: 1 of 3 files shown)


frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -91,11 +92,11 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
 
   const changeFilter = (values: IPropertyFilter) => {
     const agencyIds = (values.agencies as any)?.value ?? '';
-    setPropertyFilter({ ...values, agencies: agencyIds });
     onChange?.({ ...values, agencies: agencyIds });
   };
 
   const resetFilter = () => {
+    setClear(true);
```
Fosol, line 99: Won't it be stuck as `true` after you click the **reset** button?

jdtoombs: It is setback to false in the Typeahead component!

PR #767, opened 2021-01-09
PA-2244 Add Marina to Building Predominate use

agalbrai:


backend/dal/Migrations/Initial/Up/PostUp/01-BuildingPredominantUses.sql
```diff
@@ -74,6 +74,6 @@ INSERT INTO dbo.[BuildingPredominateUses] (
     , 0
 ), (
     17
-    , 'ZEBRA FEEDING STATION'
+    , 'Marina'
```
agalbrai, line 77: I'm assuming this will work if my other change goes through first?

backend/tests/core/Entities/BuildingPredominateUseHelper.cs
```diff
@@ -41,18 +41,20 @@ public static List<Entity.BuildingPredominateUse> CreateDefaultBuildingPredomina
...
-                new Entity.BuildingPredominateUse(7, "Storage") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(8, "Storage Vehicle") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(9, "Trailer Office") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(10, "Trailer Other") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(11, "Training Center") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(12, "Transportation (Airport / Rail / Bus station)") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(13, "University / Collect") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(14, "Warehouse") { RowVersion = new byte[] { 12, 13, 14 } },
-                new Entity.BuildingPredominateUse(15, "Weigh Station") { RowVersion = new byte[] { 12, 13, 14 } }
+                new Entity.BuildingPredominateUse(4, "Retail") { RowVersion = new byte[] { 12, 13, 14 } },
```
agalbrai, line 44: "Retail" was missing so all the numbers were off, not sure if this affects anything too much

backend/dal/Migrations/Initial/Up/PostUp/01-BuildingPredominantUses.sql
Fosol: This doesn't do anything for production, or any DB that is already running.  But it does add it if a database is refreshed or setup for the first time.

backend/dal/Migrations/v01.05.06/Up/PostUp/01-BuildingPredominateUses.sql
```diff
@@ -0,0 +1,13 @@
+PRINT 'Add Marina to Building Predominate Uses'
+
+--Add Marina
+
+INSERT INTO dbo.[BuildingPredominateUses] (
```
Fosol, line 5: This will update production and any database that is currently setup.  However it will break for any database refresh or the initial setup.  This is because you update the `init` migration to already include this.  When the migration runs this v01.05.06 it will throw an error as the database already has the value.  You need to first make sure there is no Marina value and then insert.

Fosol (changes requested):

devinleighsmith (approved):

devinleighsmith:
For some reason coverlet is failing to install which is why the api github action is failing. I tried re-running that check but it continues to fail

goemen (approved):
Looks good to me

backend/dal/Migrations/v01.05.06/Up/PostUp/01-BuildingPredominateUses.sql
Fosol: You will need to handle rollback and roll forward.  For example, if I apply this migration it will add the new value `Marina`.  Then when I rollback it will remain.  That means if I reapply (roll forward again), it will attempt to re-add the `Marina` and result in a failure.

Fosol (changes requested):
Need to handle rollback scenario

backend/dal/Migrations/v01.05.06/Up/PostUp/01-BuildingPredominateUses.sql
agalbrai: will this change work now?

backend/dal/Migrations/20210114232914_v01.06.01.Designer.cs
```diff
@@ -0,0 +1,3412 @@
...
+using NetTopologySuite.Geometries;
+using Pims.Dal;
+
+namespace Pims.Dal.Migrations
+{
+    [DbContext(typeof(PimsContext))]
+    [Migration("20210114232914_v01.06.01")]
+    partial class v010601
+    {
+        protected override void BuildTargetModel(ModelBuilder modelBuilder)
```
Fosol, line 17: This doesn't look right, looks like you're migration is for some reason trying to do something with Access Requests.

backend/dal/Migrations/v01.06.01/Up/PostUp/01-BuildingPredominateUses.sql
```diff
@@ -0,0 +1,18 @@
+PRINT 'Add Marina to Building Predominate Uses'
+
+--Check in case of rollback/rollforward
+BEGIN
```
Fosol, line 4: Don't think you need to have the extra `BEGIN` and `END` wrapping the whole thing

backend/dal/Migrations/20210114232914_v01.06.01.Designer.cs
agalbrai: I just checked this against other .Designer files from other database changes and they're all the same, isn't this just listing everything in alphabetical order?

Fosol: You're right they all have this stuff.  hmmm, never noticed this before.

Fosol (approved):

PR #768, opened 2021-01-11
PA-775: Be able to type option not just click

jdtoombs:
Allows for the ability to type out the option in our Typeahead component and not just click it from the dropdown menu that appears. Not enabled by default - new flag called `selectClosest` must be set in the `TypeaheadField` component as it may not always be desired.

pa775

### DIFF
--- a/frontend/src/components/common/form/ParentSelect.tsx
+++ b/frontend/src/components/common/form/ParentSelect.tsx
@@ -31,6 +31,8 @@ interface IParentSelect {
   setClearSelected?: Function;
   /** Event occurs when the selection changes. */
   onChange?: (vals: any) => void;
+  /** get the component to select the item with closest label match to the input provided */
+  selectClosest?: boolean;
 }
 
 /** Component used to group children items with their parent.
@@ -48,6 +50,7 @@ export const ParentSelect: React.FC<IParentSelect> = ({
   setClearSelected,
   label,
   onChange,
+  selectClosest,
 }) => {
   const { setFieldValue, values } = useFormikContext();
   const value = getIn(values, field);
@@ -106,6 +109,7 @@ export const ParentSelect: React.FC<IParentSelect> = ({
       <TypeaheadField
         disabled={disabled}
         clearMenu={clear}
+        selectClosest={selectClosest}
         setClear={setClear}
         name={field}
         labelKey={option => `${option.label}`}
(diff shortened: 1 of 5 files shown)


frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -92,7 +105,9 @@ export function TypeaheadField<T extends TypeaheadModel>({
                 setFieldValue(name, getIn(newValues[0], 'value') ?? newValues[0]);
               }
         }
-        onBlur={() => setFieldTouched(name, true)}
+        onBlur={
```
agalbrai, line 108: will clicking Enter trigger the onBlur?

agalbrai (approved):
looks great!

goemen (approved):
Looks good to me

goemen (commented):
I would also look into this property of the Typeahead field `highlightOnlyResult`, https://github.com/ericgio/react-bootstrap-typeahead/blob/HEAD/docs/API.md

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -31,6 +31,8 @@ interface IParentSelect {
   setClearSelected?: Function;
   /** Event occurs when the selection changes. */
   onChange?: (vals: any) => void;
+  /** get the component to select the item with closest label match to the input provided */
+  selectClosest?: boolean;
```
Fosol, line 35: Does this default to `true`?  I'd think this would almost always be the desired behaviour.

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -92,7 +106,9 @@ export function TypeaheadField<T extends TypeaheadModel>({
...
+          selectClosest ? (e: any) => handleBlur(e.target.value) : () => setFieldTouched(name, true)
```
devinleighsmith, line 110: FYI formik has a handleBlur function that is a bit shorter then using setFieldTouched

devinleighsmith: May also be a reason not to use handleBlur as the name of your custom function to avoid confusion.

devinleighsmith (approved):

frontend/src/components/common/form/Typeahead.tsx
jdtoombs: Ooh good point, will adjust

PR #770, opened 2021-01-11
Coverlet console fix

jdtoombs:
Fix for the failing GitHub action currently block PR's. Coverlet console had an update 2 days ago and it looks like it does not support .Net Core 3.1. Reverted back to the previously used version of coverlet to address this problem


Fosol (approved):

goemen (approved):
looks good to me

PR #772, opened 2021-01-12
Using tiles to fetch PIMS Inventory

goemen:
I used a library called `tiles-in-bbox` to generate tiles (size 256), then use the generated tiles to calculate a bbox for each tile and use it to fetch inventory markers features for the tile. The results of each tiles are merged together and passed as props to the PointCluster component for rendering (same behaviour as before). After all the files have loaded/processed, the map zooms to the results.

### DIFF
--- a/frontend/src/components/maps/leaflet/InventoryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/InventoryLayer.tsx
@@ -1,7 +1,7 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import { IPropertyDetail } from 'actions/parcelsActions';
 import { IGeoSearchParams } from 'constants/API';
-import { BBox, Feature } from 'geojson';
+import { BBox } from 'geojson';
 import useDeepCompareEffect from 'hooks/useDeepCompareEffect';
 import { LatLngBounds } from 'leaflet';
 import { useLeaflet } from 'react-leaflet';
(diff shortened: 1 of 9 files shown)


frontend/src/features/projects/spl/forms/SurplusPropertyListForm.tsx
```diff
@@ -65,7 +65,6 @@ const SurplusPropertyListForm = ({
   onClickContractInPlaceUnconditional,
   onClickDisposedExternally,
 }: ISurplusPropertyListFormProps) => {
-  debugger;
```
Fosol, line 68: thanks

frontend/src/constants/API.ts
```diff
@@ -70,6 +70,7 @@ export interface IGeoSearchParams {
   floorCount?: number;
   rentableArea?: number;
   propertyType?: string;
+  tileKey?: string;
```
Fosol, line 73: What's this for?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -67,65 +137,118 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
   onMarkerClick,
   selected,
 }) => {
-  const [features, setFeatures] = React.useState<Array<PointFeature>>([]);
+  // const [features, setFeatures] = React.useState<Array<PointFeature>>([]);
```
Fosol, line 140: Commented out line

Fosol (approved):
Looking forward to seeing this in action

agalbrai (commented):
looking forward to it!

frontend/src/constants/API.ts
goemen: This does not get sent to the api, its a local field used to track the tile used to generate the bbox. This does not affect the outcome of the api request

jdtoombs (approved):

PR #773, opened 2021-01-12
Fixing DevOps Jenkins configuration

Fosol:
Moving jenkins files around to prepare for OCP4.0 broke the 3.11 pipelines.  These updates should resolve them.

### DIFF
--- a/openshift/3.11/templates/jenkins/generic-pipeline.yaml
+++ b/openshift/3.11/templates/jenkins/generic-pipeline.yaml
@@ -49,48 +49,48 @@ objects:
       lastVersion: 0
 parameters:
   - name: NAME
-    displayName: Pipeline Name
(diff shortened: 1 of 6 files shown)


openshift/3.11/templates/jenkins/generic-pipeline.yaml
```diff
@@ -49,48 +49,48 @@ objects:
       lastVersion: 0
 parameters:
   - name: NAME
-    displayName: Pipeline Name
-    description: Name of the pipeline, e.g cicd, promote-to-test, sonarqube. A -pipeline suffix will be appended to the name provided.
+    displayName: "Pipeline Name"
```
Fosol, line 52: Double quotes appear to be required for OCP4.0.  Although it's not entirely consistent behaviour based on my initial experience.

PR #776, opened 2021-01-12
PA-2420 - Default to land on inventory list

jdtoombs:
Inventory list view now functions the same as before (defaulting to land), and `PropertyType` is only set to undefined within the query to download the excel/csv report

### DIFF
--- a/frontend/src/features/properties/list/PropertyListView.tsx
+++ b/frontend/src/features/properties/list/PropertyListView.tsx
@@ -45,7 +45,7 @@ const initialQuery: IPropertyQueryParams = {
 };
 
 const defaultFilterValues: IPropertyFilter = {
-  searchBy: 'address',
+  searchBy: 'name',
   pid: '',
   address: '',
   administrativeArea: '',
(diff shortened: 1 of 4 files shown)


frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -45,7 +45,7 @@ const initialQuery: IPropertyQueryParams = {
 };
 
 const defaultFilterValues: IPropertyFilter = {
-  searchBy: 'address',
+  searchBy: 'name',
```
jdtoombs, line 48: just to keep consistent with filter elsewhere

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -252,10 +248,10 @@ const PropertyListView: React.FC = () => {
   const dispatch = useDispatch();
 
   const fetch = (accept: 'csv' | 'excel') => {
-    const query = getServerQuery(true, { pageIndex, pageSize, filter, agencyIds });
+    const query = getServerQuery({ pageIndex, pageSize, filter, agencyIds });
     return dispatch(
       download({
-        url: getPropertyReportUrl({ ...query, all: true }),
+        url: getPropertyReportUrl({ ...query, all: true, propertyType: undefined }),
```
agalbrai, line 254: So this puts both property types into the excel export?

jdtoombs: Yup, same idea as the previous `ignorePropertyType` but this one only affect the excel/csv reports allowing the normal list view to default to land again

agalbrai: perfect 👍

agalbrai (approved):

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -66,15 +66,12 @@ const defaultFilterValues: IPropertyFilter = {
  * @param ignorePropType - Whether to ignore the PropertyTypes (for doing excel export with all values)
```
devinleighsmith, line 66: Probably can remove this comment then.

devinleighsmith: we don't need the ignorePropertyType anymore?

devinleighsmith (approved):

PR #779, opened 2021-01-13
PA-2460 Adding config for delete properties

Fosol:
Update API to have an endpoint that will delete properties from inventory.

Fixed property popup information.  Now when viewing a property on the map it will initially display the data it has and make a separate AJAX request to fetch the rest of the data.

image

### DIFF
--- a/backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
+++ b/backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
@@ -47,6 +47,82 @@ namespace Pims.Api.Areas.Tools.Helpers
 
         #region Methods
         /// <summary>
+        /// Parse the value to extract the PID number.
+        /// If it's invalid it will return -1.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        private int ParsePid(string value)
+        {
+            if (int.TryParse(value?.Replace("-", ""), out int pid))
+                return pid;
+
+            return -1;
+        }
+
+        /// <summary>
+        /// Delete properties from inventory that match the parcel PID or building name.
+        /// </summary>
+        /// <param name="properties"></param>
+        /// <param name="updatedBefore">Only allow deletes to properties updated before this date.</param>
+        /// <returns></returns>
+        public IEnumerable<Entity.Property> DeleteProperties(IEnumerable<Model.ImportPropertyModel> properties, DateTime? updatedBefore = null)
+        {
+            if (properties == null) throw new ArgumentNullException(nameof(properties));
+
+            var results = new List<Entity.Property>();
+            foreach (var property in properties)
+            {
+                var pid = ParsePid(property.ParcelId ?? property.PID);
+                _logger.LogDebug($"Requesting to remove property pid:{pid}, type:{property.PropertyType}, local:{property.LocalId}");
+
+                // Ignore invalid PID values or '000-000-000'.
+                if (pid < 1) continue;
+
+                switch (property.PropertyType?.ToLower())
+                {
+                    case "parcel":
+                    case "land":
+                        var parcel = ExceptionHelper.HandleKeyNotFound(() => _pimsAdminService.Parcel.GetByPid(pid));
+                        // Only delete the parcel if it exists in inventory.
+                        if (parcel != null && (updatedBefore == null || parcel.UpdatedOn == null || parcel.UpdatedOn < updatedBefore))
+                        {
+                            _pimsAdminService.Parcel.Remove(parcel);
+                            results.Add(parcel);
+                            _logger.LogInformation($"Deleting parcel pid:{pid}");
+                        }
+                        break;
+                    case "building":
+                        var name = $"{(String.IsNullOrWhiteSpace(property.LocalId) ? "" : $"{property.LocalId} ")}{property.Description}";
+                        // Searching for a name may result in multiple matches.
+                        //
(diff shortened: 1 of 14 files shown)


backend/api/appsettings.Local.json
```diff
@@ -22,5 +22,11 @@
     "Notifications": {
       "ThrowExceptions": true
     }
+  },
+  "Keycloak": {
+    "Authority": "http://localhost:8080/auth/realms/pims",
```
Fosol, line 27: This makes it simpler to run the API locally without having to manually change this in your .env file.

backend/dal/Services/Admin/Concrete/BuildingService.cs
```diff
@@ -99,7 +98,7 @@ public IEnumerable<Building> GetByPid(int pid, string name = null)
         /// </summary>
         /// <param name="name"></param>
         /// <returns></returns>
-        public Building GetByName(string name)
+        public IEnumerable<Building> GetByName(string name)
```
Fosol, line 101: As names are not unique it can return multiple properties and be valid.

backend/dal/Services/Admin/Concrete/BuildingService.cs
```diff
@@ -278,8 +296,18 @@ public override void Remove(Building building)
             building.ThrowIfNotAllowedToEdit(nameof(building), this.User, new[] { Permissions.SystemAdmin, Permissions.AgencyAdmin });
 
             var originalBuilding = this.Context.Buildings.Find(building.Id) ?? throw new KeyNotFoundException();
+            this.Context.Entry(originalBuilding).Collection(p => p.Parcels).Load();
```
Fosol, line 299: Don't love the separate requests to load this data, but I don't have time to look into this library https://entityframework.net/ef-plus

frontend/src/components/maps/PopupView.tsx
```diff
@@ -1,38 +1,66 @@
...
   propertyDetail,
   disabled,
   zoomTo,
   onLinkClick,
 }) => {
-  if (propertyTypeId === PropertyTypes.PARCEL) {
+  const api = useApi();
+  const [property, setProperty] = React.useState<IProperty | null>(propertyDetail);
+
+  React.useEffect(() => {
```
Fosol, line 35: Make a request for the property information

frontend/src/components/maps/PopupView.tsx
```diff
@@ -43,5 +71,5 @@ export const PopupView: React.FC<IPopupViewProps> = ({
   if (propertyTypeId === PropertyTypes.DRAFT_BUILDING) {
     return <p>This is a draft marker for building: {propertyDetail?.name ?? 'New Building'}</p>;
   }
-  return null;
+  return <></>;
```
jdtoombs, line 74: Ah, this makes more sense

jdtoombs (approved):

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -46,6 +46,82 @@ public ImportPropertiesHelper(IPimsAdminService pimsAdminService, ILogger logger
...
+                return pid;
+
+            return -1;
+        }
+
+        /// <summary>
+        /// Delete properties from inventory that match the parcel PID or building name.
+        /// </summary>
+        /// <param name="properties"></param>
+        /// <param name="updatedBefore">Only allow deletes to properties updated before this date.</param>
```
devinleighsmith, line 67: should the updatedBefore also apply to the createdOn date in the DB? we have both fields, but only createdOn is populated when a row is initially created.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -46,6 +46,82 @@ public ImportPropertiesHelper(IPimsAdminService pimsAdminService, ILogger logger
...
+        /// <returns></returns>
+        public IEnumerable<Entity.Property> DeleteProperties(IEnumerable<Model.ImportPropertyModel> properties, DateTime? updatedBefore = null)
+        {
+            if (properties == null) throw new ArgumentNullException(nameof(properties));
+
+            var results = new List<Entity.Property>();
+            foreach (var property in properties)
+            {
+                var pid = ParsePid(property.ParcelId ?? property.PID);
```
devinleighsmith, line 76: so the PID can be in either of these fields? Actually I guess based on your ordering most of the time the ParcelId is equal to the PID?

frontend/src/components/maps/PopupView.tsx
```diff
@@ -1,38 +1,67 @@
...
 }) => {
-  if (propertyTypeId === PropertyTypes.PARCEL) {
+  const { getParcel, getBuilding } = useApi();
+  const [property, setProperty] = React.useState<IProperty | null>(propertyDetail);
+  const id = propertyDetail?.id;
+
+  React.useEffect(() => {
+    if (propertyTypeId === PropertyTypes.PARCEL) {
+      getParcel(id as number).then(parcel => {
+        console.log(parcel.name);
```
devinleighsmith, line 39: I'm assuming we don't need these

frontend/src/components/maps/PopupView.tsx
devinleighsmith: as far as I'm aware returning null in a render method is valid jsx, but I don't mind this syntax either.

devinleighsmith (approved):

PR #780, opened 2021-01-13
PA-1742 - Create a Financial Reporter role

LawrenceLau2020:
includes a database migration script


backend/dal/Migrations/v01.05.06/Up/PostUp/01-RoleClaims.sql
```diff
@@ -0,0 +1,54 @@
...
+    );
+
+PRINT 'Adding Claim'
+--Adding new claim for the SRES Financial Manager
+INSERT INTO
+    dbo.[Claims] ([Id], [Name], [Description], [IsDisabled])
+VALUES
+(
+    '81ded21c-ed32-4694-8f33-79ef17833f2b',
+    'reports-spl-admin',
```
Fosol, line 30: Claims have a `KeycloakRoleId` value that maps to keycloak roles.  You should probably set this value too.

backend/dal/Migrations/v01.05.06/Up/PostUp/01-RoleClaims.sql
```diff
@@ -0,0 +1,54 @@
+PRINT 'Adding Role'
```
Fosol, line 1: You will need to do one of the following in the case of a rollback.

- Undo any of the changes you made here
- or, update this script to only perform these updates if they haven't already been done

backend/dal/Security/Permissions.cs
```diff
@@ -70,6 +70,9 @@ public enum Permissions
         ReportsView = 34,
 
         [Display(GroupName = "project", Name = "reports-spl", Description = "Can view/modify/generated project spl reports.")]
```
Fosol, line 72: Might want to update the description to match what you have in the above description

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -179,10 +179,13 @@ public ProjectReport Add(ProjectReport report)
         public ProjectReport Update(ProjectReport report)
         {
             this.User.ThrowIfNotAuthorized(Permissions.ReportsSpl);
+            var isSPLAdmin = this.User.HasPermission(Permissions.ReportsSplAdmin);
 
             var originalReport = this.Context.ProjectReports
                 .SingleOrDefault(p => p.Id == report.Id) ?? throw new KeyNotFoundException();
-
+            
+            if (!originalReport.IsFinal && report.IsFinal && !isSPLAdmin) throw new InvalidOperationException($"You do not have permissions to make a report final.");
```
Fosol, line 187: This exceptions should actually be `NotAuthorizedException` so that they get a 403 instead of a 400.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -179,10 +179,13 @@ public ProjectReport Add(ProjectReport report)
...
+            if (originalReport.IsFinal && !report.IsFinal && !isSPLAdmin) throw new InvalidOperationException($"You do not have permissions to update a final report.");
```
Fosol, line 188: This exceptions should actually be `NotAuthorizedException` so that they get a 403 instead of a 400.

agalbrai (commented):
looks like it was complicated, nice work!

backend/dal/Migrations/v01.05.06/Up/PostUp/01-RoleClaims.sql
LawrenceLau2020: Ok, I think I'm confusing 81ded21c-ed32-4694-8f33-79ef17833f2b with the claim id, when it seems like it's is actually the KeycloakRoleId. How do I find the id for this particular claim, as all I can see is the id field in the url which looks like it's the KeycloakRoleId http://localhost:8080/auth/admin/master/console/#/realms/pims/roles/81ded21c-ed32-4694-8f33-79ef17833f2b?

Fosol: You can make up your own, or just keep the on you have.

LawrenceLau2020:
Please ignore this pull request as this was just to give Devin access to the code. Actually the migration scripts are not completed yet, and one of the tests are failing and will probably need to create another test case for the Financial Reporter.

backend/dal/Migrations/v01.05.06/Up/PostUp/01-RoleClaims.sql
LawrenceLau2020: Is the Claims.KeycloakRoleId field supposed to map to the Roles.Id field? I looked at these values in my local database and they don't seem to map to each other, so I'm wondering what this field actually maps to?

LawrenceLau2020:
Please review this PR, and the migration script changes, thanks!

backend/dal/Migrations/20210122232754_v01.06.01.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010601 : Migration
```
Fosol, line 5: Doesn't inherit from `SeedMigration` so no scripts will be run, as the `PreUp(...)` and `PostUp(...)` are not available unless you use the correct inheritance.  Use the `make db-add n=v01.06.01` to automate the process for you.

backend/dal/Services/Concrete/ProjectReportService.cs
```diff
@@ -181,10 +181,13 @@ public ProjectReport Add(ProjectReport report)
         public ProjectReport Update(ProjectReport report)
         {
             this.User.ThrowIfNotAuthorized(Permissions.ReportsSpl);
+            var isSPLAdmin = this.User.HasPermission(Permissions.ReportsSplAdmin);
 
             var originalReport = this.Context.ProjectReports
                 .SingleOrDefault(p => p.Id == report.Id) ?? throw new KeyNotFoundException();
-
+            
+            if (!originalReport.IsFinal && report.IsFinal && !isSPLAdmin) throw new NotAuthorizedException ($"You do not have permissions to make a report final.");
```
Fosol, line 189: Probably could shrink this down to a single `if` statement.

Fosol (changes requested):

backend/dal/Services/Concrete/ProjectReportService.cs
LawrenceLau2020: I removed the "extra" if statement that wasn't needed.

backend/dal/Migrations/20210122232754_v01.06.01.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+using Pims.Dal.Helpers.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010601 : SeedMigration
```
Fosol, line 6: Just be aware that @devinleighsmith has a migration with the same number in a recent PR

backend/dal/Services/Concrete/ProjectReportService.cs
Fosol: This will allow a non admin to turn `IsFinal` off.  That was captured by the second `if` statement previously.  You should handle both scenarios in a single `if` statement.

Fosol (changes requested):

Fosol (approved):

goemen (approved):
Looks good to me

PR #781, opened 2021-01-13
merge dev-alpha into dev.

devinleighsmith:


frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -7,6 +7,7 @@ import { Highlighter, Menu, MenuItem } from 'react-bootstrap-typeahead';
 import { Label } from '../Label';
```
devinleighsmith, line 7: @jdtoombs please review these changes, I've updated the components that use ParentSelect to handle the change from returning an object to returning an id by default.

backend/api/Areas/Property/Models/Building/LeasedLandMetadataModel.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Pims.Api.Areas.Property.Models.Building
+{
+    public class LeasedLandMetadataModel
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - Whether an exemption was requested for the ERP.
```
Fosol, line 12: Doc

backend/api/Models/Parcel/ParcelModel.cs
```diff
@@ -21,7 +21,7 @@ public class ParcelModel : PropertyModel
 
         public IEnumerable<ParcelFiscalModel> Fiscals { get; set; } = new List<ParcelFiscalModel>();
 
-        public IEnumerable<ParcelBuildingModel> Buildings { get; set; } = new List<ParcelBuildingModel>();
+        public virtual IEnumerable<ParcelBuildingModel> Buildings { get; set; } = new List<ParcelBuildingModel>();
```
Fosol, line 24: Is there a reason it's `virtual`?

backend/dal/Helpers/Extensions/BuildingExtensions.cs
```diff
@@ -24,7 +24,11 @@ public static class BuildingExtensions
         /// <exception type="DbUpdateException">The name within a parcel should be unique.</exception>
         public static void ThrowIfNotUnique(this PimsContext context, Entity.Parcel parcel, Entity.Building building)
         {
-            var parcelBuildings = context.Parcels.Where(p => p.Id == parcel.Id).SelectMany(p => p.Buildings.Select(b => b.Building.Name)).Distinct().ToArray();
+            if(parcel == null || String.IsNullOrEmpty(building.Name))
```
Fosol, line 27: If it's not named or belongs to a parcel don't worry about being unique?

backend/dal/Migrations/20201217172215_v01.05.03.Designer.cs
```diff
@@ -0,0 +1,3415 @@
+﻿// <auto-generated />
```
Fosol, line 1: I think this file should be deleted.

backend/dal/Migrations/20201217172215_v01.05.03.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
Fosol, line 1: This migration needs to be merged into `v01.06.00` and then deleted.

backend/dal/Migrations/20210106182851_v01.06.01.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
```
Fosol, line 1: As you'll need to update `v01.06.00` from merging `v01.05.03` you may as well merge this back into `v01.06.00` too

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -193,14 +227,132 @@ public Building Update(Building building)
...
+            }
+
+            if (allowEdit)
+            {
+                this.Context.Entry(existingBuilding.Address).CurrentValues.SetValues(building.Address);
+                this.Context.Entry(existingBuilding).CurrentValues.SetValues(building);
+                this.Context.SetOriginalRowVersion(existingBuilding);
+            }
+
+            if (allowEdit)
```
Fosol, line 270: Any reason you can't include this if statement with the one above?

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -193,14 +227,132 @@ public Building Update(Building building)
...
+                    // Delete the fiscals from the parcel that have been removed.
+                    if (!building.Fiscals.Any(e => (e.BuildingId == buildingFiscals.BuildingId && e.FiscalYear == buildingFiscals.FiscalYear && e.Key == buildingFiscals.Key)))
+                    {
+                        this.Context.BuildingFiscals.Remove(buildingFiscals);
+                    }
+                }
+            }
+            existingBuilding.LeasedLandMetadata = building.LeasedLandMetadata;
             this.Context.Buildings.Update(existingBuilding); // TODO: Must detach entity before returning it.
             this.Context.CommitTransaction();
```
Fosol, line 354: Have you tested for concurrency with `RowVersion`?

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -212,6 +213,20 @@ public Parcel Add(Parcel parcel)
...
+            return parcel;
+        }
+
+        /// <summary>
+        /// Update the specified parcel in the datasource, but do not commit the transaction.
+        /// </summary>
+        /// <param name="parcel"></param>
+        /// <exception type="KeyNotFoundException">Entity does not exist in the datasource.</exception>
+        /// <returns></returns>
+        public Parcel PendingUpdate(Parcel parcel)
```
Fosol, line 229: I like this naming convention `Pending*`.  We should use it more.

backend/entities/BuildingFiscal.cs
```diff
@@ -26,6 +26,12 @@ public class BuildingFiscal : BaseEntity
         /// <value></value>
         public int FiscalYear { get; set; }
 
+        /// <summary>
+        /// get/set - The effective date of this fiscal value
+        /// </summary>
+        /// <value></value>
+        public DateTime EffectiveDate { get; set; }
```
Fosol, line 33: As this isn't nullable, how will existing data be handled in the migration?

backend/entities/Models/LeasedLandMetadata.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace Pims.Dal.Entities.Models
+{
+    public class LeasedLandMetadata
+    {
+        #region Properties
+        /// <summary>
+        /// get/set - Whether an exemption was requested for the ERP.
```
Fosol, line 9: doc

backend/entities/Models/LeasedLandMetadata.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        public string OwnershipNote { get; set; }
+        /// <summary>
+        /// get/set - An id used to associate this metadata with a parcel.
+        /// </summary>
+        public string ParcelId { get; set; }
+        /// <summary>
+        /// get/set - the lease type
+        /// </summary>
+        public int Type { get; set; }
```
Fosol, line 19: I suspect we should change the `int` to an `enum`

backend/entities/ParcelFiscal.cs
```diff
@@ -26,6 +26,12 @@ public class ParcelFiscal : BaseEntity
         /// <value></value>
         public int FiscalYear { get; set; }
 
+        /// <summary>
+        /// get/set - The effective date of this fiscal value
+        /// </summary>
+        /// <value></value>
+        public DateTime? EffectiveDate { get; set; }
```
Fosol, line 33: `BuildingFiscal` is not nullable, but here a `ParselFiscal` is?  Which is correct?

frontend/package.json
```diff
@@ -124,7 +126,7 @@
     "test": "react-scripts test --env=jsdom-fourteen --ci --watchAll=false --runInBand",
     "test:watch": "react-scripts test --env=jsdom-fourteen",
     "coverage": "react-scripts test --ci --coverage --testResultsProcessor jest-sonar-reporter --env=jsdom-fourteen --watchAll=false --runInBand",
-    "update-snapshots": "react-scripts test -- --updateSnapshot",
+    "update-snapshots": "react-scripts test --env=jsdom-fourteen --updateSnapshot",
```
Fosol, line 129: What's the `env` for?

frontend/src/actions/parcelsActions.ts
```diff
@@ -73,16 +80,52 @@ export interface IBuilding extends IProperty {
...
   buildingTenancy: string;
+  buildingTenancyUpdatedOn?: string;
   rentableArea: number | '';
+  totalArea: number | '';
   agencyId: number | '';
   agency: string;
   agencyCode: string;
   evaluations: IEvaluation[];
   fiscals: IFiscal[];
+  leasedLand: {
```
Fosol, line 96: Should there be an interface for `ILeasedLand`?

jdtoombs (approved):
Looks good re the parent select and typeahead!

backend/api/Models/Parcel/ParcelModel.cs
devinleighsmith: I'll check, I thought I was overriding this somewhere.

backend/dal/Helpers/Extensions/BuildingExtensions.cs
devinleighsmith: Yes, unfortunately we need this as there are parcels with buildings that already violate this rule within the DB. For example
Parcel 1
  Building 1 Name: ''
  Building 2 Name: ''
  Building 3 Name: 'Test'
If I modify Building 3, this check was previously throwing an exception.

backend/dal/Services/Concrete/BuildingService.cs
devinleighsmith: Not that I can see.

backend/dal/Services/Concrete/BuildingService.cs
devinleighsmith: Yes

backend/entities/BuildingFiscal.cs
devinleighsmith: Existing rows would be assigned 01/01/0001, so will probably need to make nullable

frontend/package.json
devinleighsmith: try running the old version, it won't work. the jsdom-fourteen has some required features we need to make the snapshots work.

backend/entities/ParcelFiscal.cs
devinleighsmith: they should both be nullable.

backend/api/Models/Parcel/ParcelModel.cs
devinleighsmith: Yes, overriden in the buildingParcelModel

agalbrai (commented):
A round of applause for you sir, this must have been a huge amount of work. Fingers crossed it goes over well

Fosol (approved):

PR #786, opened 2021-01-15
Autocomplete property name

jdtoombs:
Updated with `AsyncTypeaheadComponent`. Modified so it allows users to select option or send free form text. Only options belonging to the user's agency are rendered. See GIF below.

asynctype


backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -83,6 +83,38 @@ where projectNumbers.Contains(p.Project.ProjectNumber)
             return properties;
         }
 
+
+        /// <summary>
+        /// Get an array of properties within the specified filters.
```
jdtoombs, line 88: will adjust

frontend/src/features/properties/filter/PropertyFilterOptions.tsx
```diff
@@ -9,12 +9,12 @@ import { IPropertyFilter } from './IPropertyFilter';
 export const PropertyFilterOptions: React.FC = () => {
   const state: { options: any[]; placeholders: Record<string, string> } = {
     options: [
-      { label: 'Name', value: 'name' },
+      // { label: 'Name', value: 'name' },
```
jdtoombs, line 12: will cleanup

frontend/src/features/properties/filter/PropertyFilterOptions.tsx
```diff
@@ -28,7 +28,7 @@ export const PropertyFilterOptions: React.FC = () => {
   const desc = state.placeholders[searchBy] || '';
 
   const reset = () => {
-    setFieldValue('name', '');
+    // setFieldValue('name', '');
```
jdtoombs, line 31: will cleanup

frontend/src/features/properties/filter/PropertyFilterOptions.tsx
```diff
@@ -9,12 +9,12 @@ import { IPropertyFilter } from './IPropertyFilter';
 export const PropertyFilterOptions: React.FC = () => {
```
jdtoombs, line 9: This is what I mentioned in scrum this morning, options is a required field in typeahead so I cannot make these all autocomplete components, I can spend some more time making a component that changes its input type based on the selected option but that will be some more time. For now I have pulled out the property name into a separate input, if that is not ok I will adjust and make new PropertyFilterOptions component to conditionally render the input based on what is selected.

backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -83,6 +83,37 @@ where projectNumbers.Contains(p.Project.ProjectNumber)
...
+            {
+                filter.PropertyType = Entities.PropertyTypes.Land;
+            }
+            else if (!parcelFilter.IsValid())
+            {
+                filter.PropertyType = Entities.PropertyTypes.Building;
+            }
+
+            var query = this.Context.GenerateQuery(this.User, filter);
+            var properties = query.Where(x=> !string.IsNullOrWhiteSpace(x.Name)).Select(x => x.Name).ToArray();
```
jdtoombs, line 112: also filter out blank property names

Fosol, line 112: This won't work correctly, as `IsNullOrWhiteSpace(...)` will not get translated to SQL.  If you want to generate SQL  you would need `x.Name != null && x.Name != ""`.  However I don't know if this would actually perform better.

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -51,6 +55,17 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
...
+  const keycloak = useKeycloakWrapper();
+  const [initialLoad, setInitialLoad] = useState(false);
+  useEffect(() => {
+    if (!initialLoad) {
+      dispatch(fetchPropertyNames(keycloak.agencyId!));
+      setInitialLoad(true);
+    }
+  }, [initialLoad, dispatch, keycloak.agencyId]);
+  const propertyNames = useSelector<RootState, String[]>(state => state.propertyNames);
+  console.log(propertyNames);
```
Fosol, line 68: Remove `console.log`

Fosol (approved):

backend/dal/Services/Concrete/PropertyService.cs
jdtoombs: Ah I think I have made this same mistake for, I am getting deja vu. Thanks!

frontend/src/components/common/form/InputGroup.tsx
```diff
@@ -36,6 +37,12 @@ type OptionalAttributes = {
   fast?: boolean;
   outerClassName?: string;
   displayErrorTooltips?: boolean;
+  /** style to pass down to the FastInput or Input */
+  style?: CSSProperties;
+  /** pass options for typeahead component */
+  options?: string[];
+  /** autocomplete flag to determine whether to use typeahed or not */
```
agalbrai, line 44: really small thing but you did spell "typeahead" wrong in the comment

frontend/src/features/properties/common/slices/propertyNameSlice.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Slice to handle storage of property names
+ */
+const propertyNameSlice = createSlice({
+  name: 'propertyNames',
+  initialState: [] as String[],
+  reducers: {},
+  extraReducers: (builder: any) => {
+    builder.addCase(savePropertyNames, (state: any, action: any) => {
+      console.log(action.payload);
```
agalbrai, line 14: What's a slice/what is this doing? Also there's still a console.log here

agalbrai (commented):
Looked really good in sprint review this morning!

frontend/src/components/common/form/InputGroup.tsx
jdtoombs: lol darn it, thank you!

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -126,6 +136,28 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
...
+                    setInitialLoad(false);
+                  });
+                }}
+                options={options}
+                onChange={(newValues: string[]) => {
+                  setFieldValue('name', getIn(newValues[0], 'value') ?? newValues[0]);
+                }}
+                ref={ref}
+                onBlur={(e: any) =>
+                  getIn(values, 'name') === '' && setFieldValue('name', e.target.value)
```
jdtoombs, line 157: let users enter free form text

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -80,7 +80,7 @@ export type LayerPopupInformation = PopupContentConfig & {
 };
 
 const defaultFilterValues: IPropertyFilter = {
-  searchBy: 'name',
+  searchBy: 'address',
```
Fosol, line 83: Is it supposed to be default to `address`?

Fosol (approved):

frontend/src/features/properties/common/slices/propertyNameSlice.ts
jdtoombs: It creates the action creators and action types based off the reducers and state (bunch of redux stuff). It is part of the Redux Toolkit that cleans up a lot of the normal work you would have to do!

frontend/src/components/maps/leaflet/Map.tsx
jdtoombs: Only because I had to move name off that `PropertyOptions `and into its own input

PR #796, opened 2021-01-16
Fixed RowVersion on Second Submit

Fosol:
Presently it is impossible to click the Submit button a second time when saving properties because the `rowVersion` isn't updated correctly.

- Note: this does not fix a more complex problem when attempting to save financial values in the second submit.

### DIFF
--- a/backend/api/Areas/Property/Controllers/ParcelController.cs
+++ b/backend/api/Areas/Property/Controllers/ParcelController.cs
@@ -130,10 +130,8 @@ namespace Pims.Api.Areas.Property.Controllers
         {
             var entity = _mapper.Map<Entity.Parcel>(model);
 
-            _pimsService.Parcel.Update(entity); // TODO: Update related properties (i.e. Address).
-            var parcel = _mapper.Map<Model.ParcelModel>(entity);
-
-            return new JsonResult(parcel);
+            var parcel = _pimsService.Parcel.Update(entity);
+            return new JsonResult(_mapper.Map<Model.ParcelModel>(parcel));
         }
 
         /// <summary>
--- a/backend/dal/Services/Concrete/ParcelService.cs
+++ b/backend/dal/Services/Concrete/ParcelService.cs
@@ -409,7 +409,7 @@ namespace Pims.Dal.Services
 
             this.Context.SaveChanges();
             this.Context.CommitTransaction();
-            return parcel;
+            return Get(parcel.Id);
         }
 
         /// <summary>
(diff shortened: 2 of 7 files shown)


backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -409,7 +409,7 @@ public Parcel Update(Parcel parcel)
 
             this.Context.SaveChanges();
             this.Context.CommitTransaction();
-            return parcel;
+            return Get(parcel.Id);
```
Fosol, line 412: Fetch the latest information for the parcel and return it.

backend/api/Areas/Property/Controllers/ParcelController.cs
```diff
@@ -130,10 +130,8 @@ public IActionResult UpdateParcel(int id, [FromBody] Model.ParcelModel model)
         {
             var entity = _mapper.Map<Entity.Parcel>(model);
 
-            _pimsService.Parcel.Update(entity); // TODO: Update related properties (i.e. Address).
-            var parcel = _mapper.Map<Model.ParcelModel>(entity);
-
-            return new JsonResult(parcel);
+            var parcel = _pimsService.Parcel.Update(entity);
```
Fosol, line 133: Use the latest parcel information to be certain the `rowVersion` is correct.

frontend/src/actionCreators/parcelsActionCreator.ts
```diff
@@ -123,7 +123,6 @@ export const updateParcel = (parcel: IParcel) => async (dispatch: Function) => {
       parcel,
     );
     dispatch(success(actionTypes.UPDATE_PARCEL, status));
-    dispatch(fetchParcelDetail(data));
```
Fosol, line 126: There is no need to make another AJAX request as the returned result will have this.

frontend/src/features/properties/containers/ParcelDetailFormContainer.tsx
```diff
@@ -194,8 +194,10 @@ const ParcelDetailForm: React.FunctionComponent<IParcelPropertyProps> = ({
               if (!values.id) {
                 const data = await createParcel(apiValues)(dispatch);
                 persistCallback(data);
+                values.rowVersion = data.rowVersion;
               } else {
-                await updateParcel(apiValues)(dispatch);
+                const data = await updateParcel(apiValues)(dispatch);
+                values.rowVersion = data.rowVersion;
```
Fosol, line 200: Manually set the `rowVersion` so that the next submit will use it.

Fosol (commented):
This does not fix the issue when financial values are edited and submitted the second time.

PR #805, opened 2021-01-18
Quick fix for Building Assessed value table

agalbrai:
image
image


Fosol (approved):
Not a bad solution.  You probably could have extracted it from the object that is being viewed `PropertyType`, but this works nice.

PR #806, opened 2021-01-19
Location hotfix

jdtoombs:
- look for exact match with users input before select closest
- do not overwrite selected value if chosen from the list
- added display additional back to the location filter when results cap out

hotfix


frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -163,7 +163,6 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
                 name="administrativeArea"
                 placeholder="Enter a location"
                 selectClosest
-                paginate={false}
```
jdtoombs, line 166: This removed the `Display additional results...` option

agalbrai (approved):
Awesome!

devinleighsmith (approved):

PR #808, opened 2021-01-19
PIMS inventory features loading performance optimization

goemen:
- Only make api request when filter is changed and use full bc map bounds to calculate tiles for the filter
- Using map onmoveend to track bbox used to calculate clusters in supercluster (fix bottom 10-15% data not showing and Victoria are data)
- Hide parcel pins that have buildings

### DIFF
--- a/frontend/src/components/maps/leaflet/InventoryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/InventoryLayer.tsx
@@ -1,7 +1,7 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { IPropertyDetail } from 'actions/parcelsActions';
 import { IGeoSearchParams } from 'constants/API';
-import { BBox, Feature } from 'geojson';
+import { BBox } from 'geojson';
 import useDeepCompareEffect from 'hooks/useDeepCompareEffect';
 import { LatLngBounds } from 'leaflet';
 import { useLeaflet } from 'react-leaflet';
@@ -9,9 +9,11 @@ import { toast } from 'react-toastify';
 import { PointFeature } from '../types';
 import PointClusterer from './PointClusterer';
 import { useApi } from 'hooks/useApi';
-import _ from 'lodash';
 import { useSelector } from 'react-redux';
 import { RootState } from 'reducers/rootReducer';
+import { flatten, uniqBy } from 'lodash';
+import { tilesInBbox } from 'tiles-in-bbox';
+import { useFilterContext } from '../providers/FIlterProvider';
 
 export type InventoryLayerProps = {
   /** Latitude and Longitude boundary of the layer. */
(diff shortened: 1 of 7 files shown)


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -80,14 +147,21 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
     throw new Error('<InventoryLayer /> must be used under a <Map> leaflet component');
   }
 
-  const bbox = getBbox(bounds);
+  const bbox = useMemo(() => getBbox(bounds), [bounds]);
+  useEffect(() => {
+    if (filterChanged) {
+      map.fitBounds(defaultBounds, { maxZoom: 5 });
```
Fosol, line 153: Is `defaultBounds` a dependency?

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -176,11 +179,14 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
         groupBounds.isValid() &&
         group.getBounds().isValid() &&
         filterState.changed &&
-        !selected
+        !selected &&
+        tilesLoaded
```
Fosol, line 183: Is `tilesLoaded` a dependency?

Fosol (approved):

frontend/src/components/maps/leaflet/InventoryLayer.tsx
goemen: No, its a constant that never changes

frontend/src/components/maps/leaflet/PointClusterer.tsx
goemen: Should be, nice catch

jdtoombs (approved):

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -80,14 +147,21 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
+  }, [map, filterChanged]);
 
   minZoom = minZoom ?? 0;
   maxZoom = maxZoom ?? 18;
 
-  const params = useMemo<IGeoSearchParams>(
-    () => ({
-      bbox: getApiBbox(bounds ?? map.getBounds()).toString(),
+  const params = useMemo((): any => {
+    const tiles = getTiles(defaultBounds, 5);
```
devinleighsmith, line 161: So whenever the filter changes we get all of tiles based on the default bounds again? wouldn't the result of that function call be the same every time and therefore be memoizable?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -104,45 +178,57 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
-    ),
-    [],
-  );
+  const loadTile = async (filter: IGeoSearchParams) => {
+    return loadProperties(filter);
+  };
+
+  const search = async (filters: IGeoSearchParams[]) => {
+    try {
+      const data = flatten(await Promise.all(filters.map(x => loadTile(x))))
```
devinleighsmith, line 190: So the map won't display updated data unless all of the tiles have loaded?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -104,45 +178,57 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
+        .filter(feature => {
```
devinleighsmith, line 191: Do you need this based on your testing?

PR #809, opened 2021-01-19
Agency filter hotfix

jdtoombs:
Fix that ensures the `agency.value` is being sent on subsequent requests with the filter, ensure that name is changed on subsequent requests when the value is not equal to the existing formik value, account for various other scenarios where subsequent requests were previously causing bugs

### DIFF
--- a/frontend/src/components/common/form/Typeahead.tsx
+++ b/frontend/src/components/common/form/Typeahead.tsx
@@ -70,7 +70,7 @@ export function TypeaheadField<T extends TypeaheadModel>({
     setFieldTouched(name, true);
     if (
       (!getIn(name, 'value') && val !== '' && getIn(values, name) === '') ||
-      getIn(values, name) === null
+      (!getIn(values, name) && getIn(values, name) !== '')
     ) {
       const regex = new RegExp(val, 'i');
       const exact = new RegExp(`^${val}$`, 'i');
--- a/frontend/src/features/properties/filter/PropertyFilter.tsx
+++ b/frontend/src/features/properties/filter/PropertyFilter.tsx
@@ -99,7 +99,9 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
   }, [agencies, propertyFilter, defaultFilter]);
 
   const changeFilter = (values: IPropertyFilter) => {
-    const agencyIds = values.agencies;
+    const agencyIds = (values.agencies as any).value
+      ? (values.agencies as any).value
+      : values.agencies;
     setPropertyFilter({ ...values, agencies: agencyIds });
     onChange?.({ ...values, agencies: agencyIds });
   };
@@ -155,7 +157,7 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
                 }}
                 ref={ref}
                 onBlur={(e: any) =>
-                  getIn(values, 'name') === '' && setFieldValue('name', e.target.value)
+                  getIn(values, 'name') !== e.target.value && setFieldValue('name', e.target.value)
                 }
               />
             </Col>


frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -153,9 +155,10 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
                   setFieldValue('name', getIn(newValues[0], 'value') ?? newValues[0]);
                 }}
                 ref={ref}
-                onBlur={(e: any) =>
-                  getIn(values, 'name') === '' && setFieldValue('name', e.target.value)
-                }
+                onBlur={(e: any) => {
```
jdtoombs, line 158: Found another bug where name was not updating with free form text on subsequent requests, now solved with the following changes

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -70,7 +70,8 @@ export function TypeaheadField<T extends TypeaheadModel>({
     setFieldTouched(name, true);
     if (
       (!getIn(name, 'value') && val !== '' && getIn(values, name) === '') ||
-      getIn(values, name) === null
```
jdtoombs, line 73: There was another bug with selecting closest match on subsequent match where the value was undefined rather than `''` or `null` causing the closest match not to fire. This solves that

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -70,7 +70,8 @@ export function TypeaheadField<T extends TypeaheadModel>({
...
+      getIn(values, name) === null ||
```
devinleighsmith, line 73: I think javascript falsy checks includes null, undefined and '', so you can probably remove this line

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -153,9 +155,10 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
...
+                  (getIn(values, 'name') === '' || getIn(values, 'name') !== e.target.value) &&
```
devinleighsmith, line 159: do you need the first part of this conditional? the second check seems to encompass it.

frontend/src/components/common/form/Typeahead.tsx
jdtoombs: I was wondering about that, I'll double check to confirm. Thanks!

frontend/src/features/properties/filter/PropertyFilter.tsx
jdtoombs: Good point, was getting lost in my logic lol, gonna test it out.

frontend/src/components/common/form/Typeahead.tsx
```diff
@@ -70,7 +70,7 @@ export function TypeaheadField<T extends TypeaheadModel>({
     setFieldTouched(name, true);
     if (
       (!getIn(name, 'value') && val !== '' && getIn(values, name) === '') ||
-      getIn(values, name) === null
+      (!getIn(values, name) && getIn(values, name) !== '')
```
jdtoombs, line 73: Need to make sure name is blank, other wise it will auto select 100 Mile House on blur

agalbrai (commented):
I hope this fixes things!

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -99,7 +99,9 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
   }, [agencies, propertyFilter, defaultFilter]);
 
   const changeFilter = (values: IPropertyFilter) => {
-    const agencyIds = values.agencies;
+    const agencyIds = (values.agencies as any).value
```
Fosol, line 102: Seems odd we have to cast to `any`

Fosol (approved):

agalbrai (approved):

PR #810, opened 2021-01-19
PA-2383 fix maintenance and update caddyfile.

devinleighsmith:


jdtoombs (approved):

maintenance/Caddyfile
```diff
@@ -1,16 +1,19 @@
...
-  regexp .*
-  to /maintenance.html
+:2015
+root * /var/www/html
+file_server {
+  index maintenance.html
 }
 
+log
+
```
agalbrai, line 16: what is a Caddy/Caddyfile?

devinleighsmith: Caddy is another web server stack. You've heard of Apache probably? Caddy is like that. Caddy is nice because it does https automatically for you. Not sure if you've ever had to set up https manually, but its a pain.

https://caddyserver.com/docs/

agalbrai: Ah okay thank you!

maintenance/README.md
```diff
@@ -33,24 +33,52 @@ For the sake of simplicity all examples will use jcxjin-test.
...
        '{ "spec": { "to": { "name": "proxy-caddy" }, "port": { "targetPort": "2015-tcp" }}}'
    ```
 
 ### Build and Deployment
 
-This application's template has been broken down into build and deploy components.
+This application's template has been broken down into build and deploy components. The following commands will create all required bc/dc/service/route objects, assuming that the defaults used in the maintenance script are still correct(see below to confirm:
+
+```
+./maintenance.sh tools s2i
```
Fosol, line 56: I looked at the `maintenance.sh` script, there doesn't appear to be an `s2i` command.

maintenance/README.md
```diff
@@ -33,24 +33,52 @@ For the sake of simplicity all examples will use jcxjin-test.
...
+
+Contains:
+
+- ImageStream
+- BuildConfig
+
+Default vars:
+
+- NAME: s2i-caddy
+- GIT_REPO: https://github.com/BCDevOps/s2i-caddy.git
```
Fosol, line 75: This parameter does not appear to be correctly named.  Additionally the description states that this should be our URL, not the default value (unless experimenting).

maintenance/maintenance.sh
```diff
@@ -88,10 +88,15 @@ then
   oc patch route ${STATIC_PAGE_NAME} -n ${PROJECT} -p \
     '{ "spec": { "to": { "name": "'$( echo ${STATIC_PAGE_NAME} )'" },
     "port": { "targetPort": "'$( echo ${STATIC_PAGE_PORT} )'" }}}'
+elif [ "${COMMAND}" == "s2i" ]
```
devinleighsmith, line 91: s2i command added here.

maintenance/README.md
devinleighsmith: the s2i command is new, and is included in the below changes.

maintenance/README.md
devinleighsmith: That was true when they were actually hosting this image on openshift. Now since they aren't hosting this image we have to build it ourselves.

Fosol: Shouldn't we update the documentation so we don't even point to the BCGov repo then?

Fosol (approved):

maintenance/README.md
devinleighsmith: We would then need to copy all of the build artifacts to our repo as well, so the s2i/bin folder, dockerfile, default caddyfile, etc

Fosol: Hmmm, what's the chance that the existing repo will be maintained?

devinleighsmith: I think the main risk would be if the repo was deleted entirely, as we can always lock our git ref to a specific revision to prevent accidentally taking breaking changes. Seems unlikely but who knows with the OCP 4 migration.

PR #812, opened 2021-01-20
Pa-2239

devinleighsmith:
by default, two state radio check component will not have either radio selected.

### DIFF
--- a/frontend/src/components/common/form/Typeahead.tsx
+++ b/frontend/src/components/common/form/Typeahead.tsx
@@ -68,14 +68,31 @@ export function TypeaheadField<T extends TypeaheadModel>({
   const errorTooltip = error && touch && displayErrorTooltips ? error : undefined;
   const customBlur = (val: string) => {
     setFieldTouched(name, true);
-    if (!getIn(name, 'value') && val !== '') {
+    if (
+      (!getIn(name, 'value') && val !== '' && getIn(values, name) === '') ||
+      getIn(values, name) === null
+    ) {
       const regex = new RegExp(val, 'i');
+      const exact = new RegExp(`^${val}$`, 'i');
+
+      /** check for exact match before returning closest fit, will return undefined if failure */
+      const checkExactMatch = rest.options.find((x: any) =>
+        x.label ? x.label.match(exact) : x.match(exact),
+      );
+
+      /** returns item with the closest matching text, used if exact match is not found */
       const matchedItem = rest.options.find((x: any) =>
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/form/Check.tsx
```diff
@@ -95,12 +95,12 @@ export const Check: React.FC<CheckProps> = ({
             isInvalid={!!touch && !!error}
             type={type}
             {...rest}
-            value={!!checked}
+            value={checked === true}
```
agalbrai, line 98: is there much of a difference between these two?

agalbrai (approved):

frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
```diff
@@ -19,7 +19,7 @@ export const defaultInformationFormValues = {
   landLegalDescription: '',
   classificationId: '',
   agencyId: '',
-  isSensitive: false,
+  isSensitive: '',
```
Fosol, line 22: Why is we always use empty string instead of `undefined` or `null`?

maintenance/Caddyfile
```diff
@@ -1,16 +1,19 @@
-0.0.0.0:2015
-
-root /var/www/html
-
-log stdout
-
-errors stdout
+{
+    storage file_system {
```
Fosol, line 2: Looks like you mixed PRs here.

frontend/src/components/common/form/Check.tsx
devinleighsmith: Yep, javascript truthy/falsey checks encompass a wide range of values https://developer.mozilla.org/en-US/docs/Glossary/Falsy

explicitly checking for true/false narrows the number of accepted values considerably

agalbrai: thanks!

frontend/src/features/properties/components/forms/subforms/InformationForm.tsx
devinleighsmith: Formik requires that all fields are initialized with a non-null, non-undefined value or you will get a warning about changing an uncontrolled input to a controlled one when you enter data into that field at runtime.

so generally, all formik forms should be provided with an initialValues object where all fields of the form are explicitly given a value. in most cases the typings for those values should be:
someField: theActualType | '';

maintenance/Caddyfile
devinleighsmith: oops, will fix.

PR #814, opened 2021-01-20
PA-2481 Surplus Properties search - UX Updates

agalbrai:
image


frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -138,6 +138,7 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
             </Col>
             <Col className="map-filter-typeahead">
               <AsyncTypeahead
+                disabled={findMoreOpen}
```
agalbrai, line 141: I just noticed that this new field wasn't disabling when the search window was open so I fixed it real quick

devinleighsmith (approved):

Fosol (approved):

PR #823, opened 2021-01-21
Reapply filter in list view from map view, query municipality layer and zoom to result

goemen:
- Zoom to municipality using data from municipality layer
- Reapply filter in list view from map view, query municipality layer and zoom to result


Fosol (approved):

agalbrai (commented):
Thanks for doing the property list view fix

jdtoombs (approved):

PR #826, opened 2021-01-21
Do not remove buildings associated to parcels in the building->parcel…

devinleighsmith:
…s->buildings relationship.

### DIFF
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -208,6 +208,7 @@ namespace Pims.Dal.Services
                 .Include(b => b.Address).ThenInclude(a => a.Province)
                 .Include(b => b.Evaluations)
                 .Include(b => b.Fiscals)
+                .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Buildings)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Evaluations)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Classification)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Fiscals)
@@ -233,7 +234,7 @@ namespace Pims.Dal.Services
 
             foreach (var parcel in building.Parcels.Select(pb => pb.Parcel))
             {
-                // Check if the building already exists.
+                // Check if the parcel already exists.
                 var existingAssociatedParcel = existingBuilding.Parcels
                     .FirstOrDefault(pb => pb.ParcelId == parcel.Id)?.Parcel;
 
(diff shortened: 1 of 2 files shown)


backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -377,12 +377,10 @@ public Parcel PendingUpdate(Parcel parcel)
                     var updateBuilding = parcel.Buildings.FirstOrDefault(pb => pb.BuildingId == parcelBuilding.BuildingId);
                     if (updateBuilding == null)
                     {
-                        // TODO: This will delete the building from the datasource.  This isn't entirely ideal as it may be related to multiple parcels.
                         this.ThrowIfNotAllowedToUpdate(parcelBuilding.Building, _options.Project);
 
                         var parcelBuildings = this.Context.ParcelBuildings.Where(pb => pb.BuildingId == parcelBuilding.BuildingId).ToArray();
                         parcelBuildings.ForEach(pb => this.Context.ParcelBuildings.Remove(pb)); // Remove all relationships from other parcels to this building.
-                        this.Context.Buildings.Remove(parcelBuilding.Building);
```
devinleighsmith, line 385: In my opinion this no longer is required, as previously we needed to deleted buildings that were removed from parcels to prevent "orphaned" buildings.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -464,7 +462,6 @@ public void Remove(Parcel parcel)
                 this.Context.ParcelBuildings.Remove(b);
                 this.Context.BuildingEvaluations.RemoveRange(b.Building.Evaluations);
                 this.Context.BuildingFiscals.RemoveRange(b.Building.Fiscals);
-                this.Context.Buildings.Remove(b.Building);
```
devinleighsmith, line 467: Same here, if a parcel is removed, it shouldn't just delete all of the associated buildings (it should just delete the parcel->building relationship.

devinleighsmith:
@Fosol This is a good one for you to look at.

Fosol (approved):

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -246,16 +247,21 @@ public Building Update(Building building)
                 {
                     if (!allowEdit) throw new NotAuthorizedException("User may not add parcels to a property they don't own.");
                     var existingParcel = this.Context.Parcels
+                    .Include(p => p.Buildings)
                     .FirstOrDefault(pb => pb.Id == parcel.Id);
                     if(existingParcel != null)
                     {
+                        parcel.Buildings.Clear(); // Do not modify the list of buildings associated to parcels when performing building updates.
```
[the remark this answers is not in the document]
devinleighsmith: yeah, open to suggestions here. Here is what is going on in abstract:

We don't want to trust this part (highlighted in bold) of the incoming buildings relationships building -> *parcels -> buildings*

1) Clear out all of the parcels associated buildings (since there is no setter on this list). 
2) Add associated buildings such that our incoming parcel's buildings is identical to the existing parcel's buildings in the db.

Since the incoming parcel's buildings are identical to the existing parcel's buildings no add/updates/removals should take place.

jdtoombs (approved):

PR #830, opened 2021-01-21
PA-2451 Clear the parcel layer data when it is consumed, so that futu…

devinleighsmith:
…re "populate" actions


frontend/src/components/maps/hooks/useMarkerZoom.tsx
```diff
@@ -1,61 +0,0 @@
-import { useState, useEffect } from 'react';
```
devinleighsmith, line 1: This is dead code, we solved the marker zoom a different way.

Fosol (approved):

jdtoombs (approved):

agalbrai (approved):

PR #833, opened 2021-01-22
Adjust height to fit more roles in user profile

jdtoombs:
Slight adjustment (not sure how often if ever a user would have this many roles) but increasing the height to accommodate more user roles.

image

### DIFF
--- a/frontend/src/components/layout/Header/UserProfile.tsx
+++ b/frontend/src/components/layout/Header/UserProfile.tsx
@@ -14,7 +14,7 @@ const StyleDropDown = styled(NavDropdown)`
   font-size: 14px;
   .dropdown-menu {
     width: 300px;
-    height: 120px;
+    height: 150px;
   }
   .nav-link {
     color: #fff;


frontend/src/components/layout/Header/UserProfile.tsx
```diff
@@ -14,7 +14,7 @@ const StyleDropDown = styled(NavDropdown)`
   font-size: 14px;
   .dropdown-menu {
     width: 300px;
-    height: 120px;
+    height: 150px;
```
devinleighsmith, line 17: Do we need a static height here? could just set some padding here and just let it grow to accommodate.

jdtoombs: I had that originally, but it kept leaving a little bit of white trim underneath the blue footer. Will try playing around with it again though!

goemen:
I have a feeling that when a user has more roles, this design will not work. Maybe a user account page will be better. Just a thought

Fosol (approved):

agalbrai (approved):

PR #835, opened 2021-01-22
PA-2654 building not populating location when dropping pin.

devinleighsmith:


agalbrai (approved):

frontend/src/features/properties/hooks/useParcelLayerData.tsx
```diff
@@ -134,7 +134,11 @@ const getAdminAreaFromLayerData = (
     const splitLayerMunicipality = layerMunicipality.split(',');
     if (splitLayerMunicipality.length === 2) {
       const formattedLayerMunicipality = `${splitLayerMunicipality[1].trim()} ${splitLayerMunicipality[0].trim()}`;
-      return _.find(administrativeAreas, { name: formattedLayerMunicipality });
+      let match = _.find(administrativeAreas, { name: formattedLayerMunicipality });
+      if (!match) {
+        match = _.find(administrativeAreas, { name: splitLayerMunicipality[0].trim() });
```
Fosol, line 139: What is the `splitLayerMunicipality`?

Fosol (approved):

frontend/src/features/properties/hooks/useParcelLayerData.tsx
devinleighsmith: I'm attempting to match the location from parcel map with our locations as best I can, so I'm splitting the municipality from the layer and using the split string to try and match one of our locations.

PR #841, opened 2021-01-27
PA-2479 Slide out for owning agency, view land

agalbrai:
https://user-images.githubusercontent.com/33818575/105923191-d8dbb800-5ff0-11eb-9900-eededfdf1571.mp4

https://user-images.githubusercontent.com/33818575/105924540-4f79b500-5ff3-11eb-9dda-e4b23ab2ec2c.mp4

I'm sure there's a few things here that could be improved, please let me know!

https://user-images.githubusercontent.com/33818575/106067729-22410b80-60b4-11eb-9730-d3b85a8fc1ae.mp4

Now can view slide out from property list view

### DIFF
--- a/frontend/src/actions/parcelsActions.ts
+++ b/frontend/src/actions/parcelsActions.ts
@@ -71,6 +71,7 @@ export enum LeasedLand {
 
 export interface IBuilding extends IProperty {
   parcelId: number;
+  pid: number;
   address: IAddress;
   buildingFloorCount?: number | '';
   buildingConstructionType?: string;
@@ -92,6 +93,8 @@ export interface IBuilding extends IProperty {
   agencyId: number | '';
   agency: string;
   agencyCode: string;
+  subAgency?: string;
+  assessed: number | '';
   evaluations: IEvaluation[];
   fiscals: IFiscal[];
   leasedLand: {
@@ -160,9 +163,11 @@ export interface IParcel extends IProperty {
   zoning: string;
   zoningPotential: string;
   agency?: string;
+  subAgency?: string;
   agencyId: number | '';
   isSensitive: boolean;
   buildings: IBuilding[];
+  assessed: number | '';
   evaluations: IEvaluation[];
   fiscals: IFiscal[];
   rowVersion?: string;
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
@@ -18,7 +18,7 @@ export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo
   let formatAssessed;
   if (buildingInfo?.assessed) {
     formatAssessed = formatMoney(buildingInfo?.assessed);
-  } else if (buildingInfo?.evaluations[0].value) {
+  } else if (buildingInfo?.evaluations?.length >= 1) {
     formatAssessed = formatMoney(buildingInfo?.evaluations[0].value);
   } else {
     formatAssessed = '$0';
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
@@ -1,0 +1,48 @@
+import * as React from 'react';
+import { ListGroup } from 'react-bootstrap';
+import { IBuilding } from 'actions/parcelsActions';
+import { Label } from 'components/common/Label';
+import './InfoSlideOut.scss';
+import { formatMoney } from 'utils/numberFormatUtils';
+
+interface IBuildingAttributes {
+  /** the selected building information */
+  buildingInfo: IBuilding;
+}
+
+/**
+ * Displays Building specific information needed on the information slide out
+ * @param buildingInfo the selected parcel data
+ */
+export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo }) => {
+  let formatAssessed;
+  if (buildingInfo?.assessed) {
+    formatAssessed = formatMoney(buildingInfo?.assessed);
+  } else if (buildingInfo?.evaluations[0].value) {
+    formatAssessed = formatMoney(buildingInfo?.evaluations[0].value);
+  } else {
+    formatAssessed = '$0';
+  }
+
+  return (
+    <>
+      <ListGroup>
+        <Label className="header">Valuation</Label>
+        <ListGroup.Item>
+          <Label>Assessed value:</Label>
+          {formatAssessed}
+        </ListGroup.Item>
+      </ListGroup>
+      {buildingInfo?.pid && (
+        <ListGroup>
+          <Label className="header">Associated Land</Label>
+          <ListGroup.Item>
+            <Label>{buildingInfo.pid}</Label>
+          </ListGroup.Item>
+        </ListGroup>
+      )}
+    </>
+  );
+};
+
+export default BuildingAttributes;
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
@@ -1,0 +1,117 @@
+import styled from 'styled-components';
+import * as React from 'react';
+import { Row } from 'react-bootstrap';
+import { IBuilding, IParcel, PropertyTypes } from 'actions/parcelsActions';
+import { useLeaflet } from 'react-leaflet';
+import { MAX_ZOOM } from 'constants/strings';
+import { Link, useLocation } from 'react-router-dom';
+import queryString from 'query-string';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import Claims from 'constants/claims';
+
+const LinkMenu = styled(Row)`
+  background-color: #f2f2f2;
+  height: 35px;
+  width: 322px;
+  margin-left: -10px;
+  font-size: 14px;
+  padding: 10px;
+  a {
+    padding: 0px 10px;
+    color: #1a5a96;
+  }
+`;
+
+const VerticalBar = styled.div`
+  border-left: 2px solid rgba(96, 96, 96, 0.2);
+  height: 18px;
+`;
+
+interface IHeaderActions {
+  /** The selected property */
+  propertyInfo: IParcel | IBuilding | null;
+  /** the selected property type */
+  propertyTypeId: PropertyTypes | null;
+  /** additional action to be taken when a link in the menu is clicked */
+  onLinkClick?: () => void;
+}
+
+/**
+ * Actions that can be done on the property info slide out
+ * @param propertyInfo the selected property information
+ * @param propertyTypeId the selected property type
+ * @param onLinkClick additional action on menu item click
+ */
+const HeaderActions: React.FC<IHeaderActions> = ({ propertyInfo, propertyTypeId, onLinkClick }) => {
+  const leaflet = useLeaflet();
+  const location = useLocation();
+  const keycloak = useKeycloakWrapper();
+  const jumpToView = () =>
+    leaflet.map?.setView(
+      [propertyInfo?.latitude as number, propertyInfo?.longitude as number],
+      MAX_ZOOM,
+    );
+  const zoomToView = () =>
+    leaflet.map?.flyTo(
+      [propertyInfo?.latitude as number, propertyInfo?.longitude as number],
+      MAX_ZOOM,
+    );
+
+  const buildingId = propertyTypeId === PropertyTypes.BUILDING ? propertyInfo?.id : undefined;
+  const parcelId = propertyTypeId === PropertyTypes.PARCEL ? propertyInfo?.id : undefined;
+  return (
+    <LinkMenu>
+      Actions:
+      {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
+        keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) && (
+        <>
+          <Link
+            onClick={() => {
+              jumpToView();
+              if (onLinkClick) onLinkClick();
+            }}
+            to={{
+              pathname: `/mapview`,
+              search: queryString.stringify({
+                ...queryString.parse(location.search),
+                sidebar: true,
+                disabled: true,
+                loadDraft: false,
+                buildingId: buildingId,
+                parcelId: parcelId,
+              }),
+            }}
+          >
+            View details
+          </Link>
+          <VerticalBar />
+          <Link
+            onClick={() => {
+              jumpToView();
+              if (onLinkClick) onLinkClick();
+            }}
+            to={{
+              pathname: `/mapview`,
+              search: queryString.stringify({
+                ...queryString.parse(location.search),
+                disabled: false,
+                sidebar: true,
+                loadDraft: false,
+                buildingId: buildingId,
+                parcelId: parcelId,
+              }),
+            }}
+          >
+            Update
+          </Link>
+          <VerticalBar />
+        </>
+      )}
+      <Link to={{ ...location }} onClick={zoomToView}>
+        Zoom map
+      </Link>
+    </LinkMenu>
+  );
+};
+
+export default HeaderActions;
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
@@ -1,0 +1,94 @@
+import * as React from 'react';
+import { ListGroup } from 'react-bootstrap';
+import { IBuilding, IParcel, PropertyTypes } from 'actions/parcelsActions';
+import { Label } from 'components/common/Label';
+import { ParcelPIDPIN } from './ParcelPIDPIN';
+import ParcelAttributes from './ParcelAttributes';
+import './InfoSlideOut.scss';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import Claims from 'constants/claims';
+import BuildingAttributes from './BuildingAttributes';
+
+interface IInfoContent {
+  /** the selected property information */
+  propertyInfo: IParcel | IBuilding | null;
+  /** The property type [Parcel, Building] */
+  propertyTypeId: PropertyTypes | null;
+}
+
+/**
+ * Component that displays the appropriate information about the selected property
+ * in the property info slideout
+ * @param {IInfoContent} propertyInfo the selected property
+ * @param {IInfoContent} propertyTypeId the property type [Parcel, Building]
+ */
+export const InfoContent: React.FC<IInfoContent> = ({ propertyInfo, propertyTypeId }) => {
+  const keycloak = useKeycloakWrapper();
+  return (
+    <>
+      <ListGroup>
+        <Label className="header">Parcel Identification</Label>
+        {propertyTypeId === PropertyTypes.PARCEL && (
+          <ParcelPIDPIN parcelInfo={propertyInfo as IParcel} />
+        )}
+        {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
+          keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) && (
+          <>
+            {propertyInfo?.name && (
+              <ListGroup.Item>
+                <Label>Name:</Label>
+                {propertyInfo?.name}
+              </ListGroup.Item>
+            )}
+            <ListGroup.Item>
+              <Label>Owning Agency:</Label>
+              {propertyInfo?.agency}
+            </ListGroup.Item>
+            {propertyInfo?.subAgency && (
+              <ListGroup.Item>
+                <Label>Sub-agency:</Label>
+                {propertyInfo.subAgency}
+              </ListGroup.Item>
+            )}
+          </>
+        )}
+        <ListGroup.Item>
+          <Label>Classification:</Label>
+          {propertyInfo?.classification}
+        </ListGroup.Item>
+      </ListGroup>
+      <ListGroup>
+        <Label className="header">Location data</Label>
+        <ListGroup.Item>
+          <Label>Civic Address:</Label>
+          {propertyInfo?.address?.line1}
+        </ListGroup.Item>
+        <ListGroup.Item>
+          <Label>Location:</Label>
+          {propertyInfo?.address?.administrativeArea}
+        </ListGroup.Item>
+        <ListGroup.Item>
+          <Label>Latitude:</Label>
+          {propertyInfo?.latitude}
+        </ListGroup.Item>
+        <ListGroup.Item>
+          <Label>Longitude:</Label>
+          {propertyInfo?.longitude}
+        </ListGroup.Item>
+      </ListGroup>
+      {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
+        keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) && (
+        <>
+          {propertyTypeId === PropertyTypes.PARCEL && (
+            <ParcelAttributes parcelInfo={propertyInfo as IParcel} />
+          )}
+          {propertyTypeId === PropertyTypes.BUILDING && (
+            <BuildingAttributes buildingInfo={propertyInfo as IBuilding} />
+          )}
+        </>
+      )}
+    </>
+  );
+};
+
+export default InfoContent;
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.scss
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.scss
@@ -1,0 +1,31 @@
+@import '../../../../colors.scss';
+
+.list-group-item {
+  padding: 0px 0px 5px 0px;
+  border: none;
+  font-family: 'BCSans', Fallback, sans-serif;
+  font-size: 12px;
+  text-align: left;
+  font-weight: bold;
+  color: $text-color;
+
+  .label {
+    color: $light-variant-color;
+    font-weight: lighter;
+    padding-right: 5px;
+  }
+}
+
+.header {
+  border-bottom: 2px solid rgba(96, 96, 96, 0.2);
+  font-weight: 700;
+  font-size: 14px;
+  color: $text-color;
+  padding: 2px 0px;
+  margin: 5px 0px;
+  text-align: left;
+}
+
+.leaflet-control {
+  margin-right: 0px;
+}
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.test.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.test.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react';
+import 'leaflet';
+import { Map as LeafletMap } from 'leaflet';
+import { Map as ReactLeafletMap, MapProps } from 'react-leaflet';
+import Adapter from 'enzyme-adapter-react-16';
+import Enzyme, { mount } from 'enzyme';
+import InfoSlideOut from './InfoSlideOut';
+import 'leaflet/dist/leaflet.css';
+import { waitFor } from '@testing-library/dom';
+import { Button } from 'react-bootstrap';
+
+jest.mock('axios');
+Enzyme.configure({ adapter: new Adapter() });
+
+let mapRef: React.RefObject<ReactLeafletMap<MapProps, LeafletMap>> | undefined;
+
+const MapComponent = () => {
+  const [open, setOpen] = React.useState(false);
+  mapRef = React.useRef<any>();
+  return (
+    <div id="mapid" style={{ width: 500, height: 500 }}>
+      <ReactLeafletMap ref={mapRef} center={[48.423078, -123.360956]} zoom={18}>
+        <InfoSlideOut open={open} setOpen={() => setOpen(!open)} />
+      </ReactLeafletMap>
+    </div>
+  );
+};
+
+describe('InfoSlideOut View', () => {
+  beforeEach(() => {
+    mapRef = undefined;
+  });
+
+  it('Should render the slide out button', () => {
+    const component = mount(<MapComponent />);
+    waitFor(() => expect(mapRef?.current).toBeTruthy(), { timeout: 500 });
+    const infoButton = component.find(Button).first();
+    expect(infoButton).toBeDefined();
+    expect(infoButton.props().id).toBe('slideOutInfoButton');
+  });
+
+  it('Component should be closed by default', () => {
+    const component = mount(<MapComponent />);
+    waitFor(() => expect(mapRef?.current).toBeTruthy(), { timeout: 500 });
+    const infoContainer = component.find('#infoContainer').first();
+    expect(infoContainer).toBeDefined();
+    expect(infoContainer.props().className?.includes('closed')).toBeTruthy();
+  });
+
+  it('Clicking the button should open the info component', () => {
+    const component = mount(<MapComponent />);
+    waitFor(() => expect(mapRef?.current).toBeTruthy(), { timeout: 500 });
+    const infoContainer = component.find('#infoContainer').first();
+    const infoButton = component.find('#slideOutInfoButton').first();
+    expect(infoContainer).toBeDefined();
+    expect(infoButton).toBeDefined();
+    expect(infoContainer.props().className?.includes('closed')).toBeTruthy();
+    infoButton.invoke('onClick');
+    waitFor(() => expect(infoContainer.props().className?.includes('closed')).toBeFalsy(), {
+      timeout: 500,
+    });
+  });
+
+  it('Clicking the button should close the info component', () => {
+    const component = mount(<MapComponent />);
+    waitFor(() => expect(mapRef?.current).toBeTruthy(), { timeout: 500 });
+    const infoContainer = component.find('#infoContainer').first();
+    const infoButton = component.find('#slideOutInfoButton').first();
+    expect(infoContainer).toBeDefined();
+    expect(infoButton).toBeDefined();
+    expect(infoContainer.props().className?.includes('closed')).toBeTruthy();
+    infoButton.invoke('onClick');
+    waitFor(() => expect(infoContainer.props().className?.includes('closed')).toBeFalsy(), {
+      timeout: 500,
+    });
+    infoButton.invoke('onClick');
+    waitFor(() => expect(infoContainer.props().className?.includes('closed')).toBeTruthy(), {
+      timeout: 500,
+    });
+  });
+});
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
@@ -1,0 +1,148 @@
+import * as React from 'react';
+import { Button } from 'react-bootstrap';
+import { FaInfo } from 'react-icons/fa';
+import Control from 'react-leaflet-control';
+import styled from 'styled-components';
+import clsx from 'classnames';
+import * as L from 'leaflet';
+import { useEffect } from 'react';
+import HeaderActions from './HeaderActions';
+import { InfoContent } from './InfoContent';
+import TooltipWrapper from 'components/common/TooltipWrapper';
+import { PropertyPopUpContext } from 'components/maps/providers/PropertyPopUpProvider';
+
+const InfoContainer = styled.div`
+  margin-right: -10px;
+  width: 341px;
+  min-height: 52px;
+  height: auto;
+  max-height: 500px;
+  background-color: #fff;
+  position: relative;
+  border-radius: 4px;
+  box-shadow: -2px 1px 4px rgba(0, 0, 0, 0.2);
+  z-index: 1000;
+  &.closed {
+    width: 0px;
+    height: 0px;
+  }
+`;
+
+const InfoHeader = styled.div`
+  width: 100%;
+  height: 52px;
+  background-color: #1a5a96;
+  color: #fff;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: space-evenly;
+  padding-top: 16px;
+  border-top-right-radius: 4px;
+  border-top-left-radius: 4px;
+`;
+
+const InfoMain = styled.div`
+  width: 100%;
+  max-height: 420px;
+  padding-left: 10px;
+  padding: 0px 10px 5px 10px;
+
+  &.open {
+    overflow-y: scroll;
+  }
+`;
+
+const InfoIcon = styled(FaInfo)`
+  font-size: 30px;
+`;
+
+const InfoButton = styled(Button)`
+  width: 52px;
+  height: 52px;
+  position: absolute;
+  left: -51px;
+  background-color: #fff;
+  color: #1a5a96;
+  border-color: #1a5a96;
+  box-shadow: -2px 1px 4px rgba(0, 0, 0, 0.2);
+  &.open {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    top: 0px;
+  }
+`;
+
+const Title = styled.p`
+  font-size: 18px;
+  color: #ffffff;
+  text-decoration: none solid rgb(255, 255, 255);
+  line-height: 18px;
+  font-weight: bold;
+`;
+
+export type InfoControlProps = {
+  /** whether the slide out is open or closed */
+  open: boolean;
+  /** set the slide out as open or closed */
+  setOpen: () => void;
+  /** additional action for when a link is clicked */
+  onHeaderActionClick?: () => void;
+};
+
+/**
+ * Component to display the popup information of a parcel/building
+ * @param open open/closed state of the slideout
+ * @param setOpen function to set the slideout as open or closed
+ * @param onHeaderActionClick action to be taken when a menu item is clicked
+ */
+const InfoControl: React.FC<InfoControlProps> = ({ open, setOpen, onHeaderActionClick }) => {
+  const popUpContext = React.useContext(PropertyPopUpContext);
+
+  useEffect(() => {
+    const elem = L.DomUtil.get('infoContainer');
+    if (elem) {
+      L.DomEvent.on(elem!, 'mousewheel', L.DomEvent.stopPropagation);
+    }
+  });
+
+  return (
+    <Control position="topright">
+      <InfoContainer id="infoContainer" className={clsx({ closed: !open })}>
+        {open && (
+          <InfoHeader>
+            <Title>Property Info</Title>
+          </InfoHeader>
+        )}
+        <TooltipWrapper toolTipId="info-slideout-id" toolTip="Property Information">
+          <InfoButton
+            id="slideOutInfoButton"
+            variant="outline-secondary"
+            onClick={setOpen}
+            className={clsx({ open })}
+          >
+            <InfoIcon />
+          </InfoButton>
+        </TooltipWrapper>
+        <InfoMain className={clsx({ open })}>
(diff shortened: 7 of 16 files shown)


frontend/src/actions/parcelsActions.ts
```diff
@@ -71,6 +71,7 @@ export enum LeasedLand {
 export interface IBuilding extends IProperty {
   parcelId: number;
   address: IAddress;
+  administrativeArea: string;
```
agalbrai, line 74: I'm not sure if this was the best way to do to access the fields, this was just the most obvious to me

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+  border-width: 2px;
+`;
+
+export type IHeaderActions = {
+  /** The selected property */
+  propertyInfo: IParcel | IBuilding | null;
+  /** the selected property type */
+  propertyTypeId: PropertyTypes | null;
+  /** additional action to be taken when a link in the menu is clicked */
+  onLinkClick?: () => void;
```
agalbrai, line 40: I put this here since I wasn't sure if we wanted the pop out to close when the property detail view is open

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <ListGroup.Item>
+          <Label>Latitude:</Label>
+          {propertyInfo?.latitude}
+        </ListGroup.Item>
+        <ListGroup.Item>
+          <Label>Longitude:</Label>
+          {propertyInfo?.longitude}
+        </ListGroup.Item>
+      </ListGroup>
+      {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
```
agalbrai, line 71: I wasn't sure what would be viewable to owning vs non-owning agencies so I put the check here

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Displays parcel specific information needed on the information slide out
+ * @param parcelInfo the selected parcel data
+ */
+export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo }) => {
+  let formatAssessed;
+  if (parcelInfo?.assessed !== '') {
+    formatAssessed = '$' + new Intl.NumberFormat('en-US').format(parcelInfo?.assessed as number);
```
agalbrai, line 19: format the assessed value into $___,___ format (no cents)

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -27,8 +27,8 @@ export type InventoryLayerProps = {
   maxZoom?: number;
   /** Search filter to apply to properties. */
   filter?: IGeoSearchParams;
-  /** What to do when the point feature is clicked. */
-  onMarkerClick?: (point: PointFeature, position?: [number, number]) => void;
+  /** What to do when the marker is clicked. */
+  onMarkerClick: () => void;
```
agalbrai, line 31: This wasn't being used at all (as far as I could tell) so I changed it

frontend/src/components/maps/leaflet/LayersControl/LayersControl.tsx
```diff
@@ -79,13 +80,18 @@ const ControlButton = styled(Button)`
...
+  open: boolean;
+  /** set the slide out as open or closed */
+  setOpen: () => void;
+};
+
 /**
  * Component to display the layers control on the map
  * @example ./LayersControl.md
  */
-const LayersControl: React.FC = () => {
```
agalbrai, line 86: Moved to Map for flipping between info, Layers slide outs

frontend/src/components/maps/leaflet/LayersControl/LayersControl.tsx
```diff
@@ -102,14 +108,16 @@ const LayersControl: React.FC = () => {
...
         )}
-        <ControlButton
-          id="layersControlButton"
-          variant="outline-secondary"
-          onClick={() => setOpen(!open)}
-          className={clsx({ open })}
-        >
-          <LayersIcon />
-        </ControlButton>
+        <TooltipWrapper toolTipId="layer-control-id" toolTip="Layer Controls">
```
agalbrai, line 111: just added a tooltip

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -100,6 +99,24 @@ const defaultFilterValues: IPropertyFilter = {
...
+  const [propertyTypeID, setPropertyTypeID] = React.useState(null);
+  return (
+    <PropertyPopUpContext.Provider
+      value={{ propertyInfo, setPropertyInfo, propertyTypeID, setPropertyTypeID }}
+    >
+      {children}
+    </PropertyPopUpContext.Provider>
+  );
+};
+
```
agalbrai, line 119: I think I was told to put this definition somewhere else but I've forgotten where

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -289,86 +301,112 @@ const Map: React.FC<MapProps> = ({
     }
   };
 
+  const [infoOpen, setInfoOpen] = React.useState(false);
+  const [layersOpen, setLayersOpen] = React.useState(false);
+
   return (
-    <ReactResizeDetector handleWidth>
```
agalbrai, line 293: I added the context provider to the map, nothing else on the map was changed other than the two slideouts

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -289,86 +301,112 @@ const Map: React.FC<MapProps> = ({
...
+                        />
+                      </Popup>
+                    )}
+                    <LegendControl />
+                    <ZoomOutButton map={mapRef} bounds={defaultBounds} />
+                    <LayersControl
+                      open={layersOpen}
+                      setOpen={() => {
+                        setLayersOpen(!layersOpen);
+                        setInfoOpen(false);
```
agalbrai, line 380: Only one of the slide outs can be open at a time

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -289,86 +301,112 @@ const Map: React.FC<MapProps> = ({
...
+                        setLayersOpen(false);
+                      }}
+                    />
+                    <InventoryLayer
+                      zoom={zoom}
+                      bounds={bounds}
+                      onMarkerClick={() => {
+                        if (!infoOpen) {
+                          setLayersOpen(false);
+                          setInfoOpen(true);
```
agalbrai, line 396: don't close the slide out when a marker is clicked again: if one marker is clicked then a different one is clicked right after, the slide out stays open

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -258,18 +261,12 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
               key={index}
               position={[latitude, longitude]}
               icon={getMarkerIcon(cluster)}
-            >
```
agalbrai, line 261: don't show the popups anymore

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -321,23 +312,7 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
-                        selected.parcelDetail!.latitude as number,
-                        selected.parcelDetail!.longitude as number,
-                      ],
-                      MAX_ZOOM,
-                    )
-                  }
-                />
-              </Popup>
-            </SelectedPropertyMarker>
+            />
```
agalbrai, line 315: From what I can tell, a SelectedPropertyMarker appears after someone clicks "View" from the property list view, in which case the property detail view is open anyway

frontend/src/mocks/filterDataMock.ts
```diff
@@ -144,6 +144,8 @@ export const mockDetails = [
     classification: 'Core Operational',
     name: 'test name',
     description: 'test',
+    administrativeArea: 'Victoria',
+    assessed: 10000,
```
agalbrai, line 148: again, not sure if this was the best way to do this. Also I don't actually know what the filterDataMock is for

frontend/src/actions/parcelsActions.ts
devinleighsmith: I believe that these are all valid fields returned by the API, so seems fine to me.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import Claims from 'constants/claims';
+
+const LinkMenu = styled(Row)`
+  background-color: #f2f2f2;
+  height: 35px;
+  width: 341px;
+  margin-left: -10px;
+  font-size: 14px;
+  font-family: ProximaNova, sans-serif;
```
devinleighsmith, line 18: We use BC-Sans everywhere else, so this concerns me. Was another font requested in the design?

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+  padding: 10px;
+  a {
+    padding: 0 10px;
+    color: #1a5a96;
```
devinleighsmith, line 22: This reminds me that I still haven't figured out if there is a way to access css variables in styled components, will try to remember to research this.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }
+`;
+
+const VerticalBar = styled.div`
+  border-left: 6px solid rgba(96, 96, 96, 0.2);
+  height: 18px;
+  margin-left: 1%;
```
devinleighsmith, line 29: a little surprised to see % margins, any particular reason for this? generally percentages are used when the size relative to the parent is important.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
devinleighsmith: Yeah, not sure on that one either. @bbjornholt ?

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+    );
+
+  const buildingId = propertyTypeId === PropertyTypes.BUILDING ? propertyInfo?.id : undefined;
+  const parcelId = propertyTypeId === PropertyTypes.PARCEL ? propertyInfo?.id : undefined;
+  return (
+    <LinkMenu>
+      Actions:
+      <Link
+        onClick={() => {
+          if (curZoom! < MAX_ZOOM) jumpToView();
```
devinleighsmith, line 72: This is probably fine, but I generally think about using ! in the sense of what does this get me?
Well, generally curZoom should exist as leaflet.map?.getZoom() should generally be available. However in the off chance that is is null or undefined curZoom! will blow up and leave a nasty error.

using some sort of sensible default would prevent the nasty error at least. Alternatively you could just not cull jumpToView if curZoom is not known

devinleighsmith: Also, does this if make sense? It seems like if we are zoomed in more then max zoom we would still want to jump to the property right? we may not want to change the zoom level in that case, but moving the center of the map seems to make sense.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,122 @@
...
+              }),
+            }}
+          >
+            Update
+          </Link>
+          <VerticalBar />
+        </>
+      )}
+      {curZoom! < MAX_ZOOM && (
+        <Link to={{ ...location }} onClick={zoomToView}>
```
devinleighsmith, line 114: Same thought, what if I open your detail view, zoom in all the way, and then navigate the map away from my property, being able to jump back might be nice.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { ListGroup } from 'react-bootstrap';
+import { IBuilding, IParcel, PropertyTypes } from 'actions/parcelsActions';
+import { Label } from 'components/common/Label';
+import { ParcelPIDPIN } from './ParcelPIDPIN';
+import ParcelAttributes from './ParcelAttributes';
+import './InfoSlideOut.scss';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import Claims from 'constants/claims';
+
+export type IInfoContent = {
```
devinleighsmith, line 11: The I makes a bit less sense if you aren't using an interface, but I don't think it really matters.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** the selected property information */
+  propertyInfo: IParcel | IBuilding | null;
+  /** The property type [Parcel, Building] */
+  propertyTypeId: PropertyTypes | null;
+};
+
+/**
+ * Component that displays the appropriate information about the selected property
+ * in the property info slideout
+ * @param propertyInfo the selected property
```
devinleighsmith, line 21: slightly misleading doc since this component takes in a single param, props, with a number of fields. One option is to do this:

```
interface ITenancyProps {
  title: string;
  disabled?: boolean;
  nameSpace?: string;
  index?: any;
  showImprovements?: boolean;
}

/**
 * Display valuation tables for assessed and netbook values.
 * @param {ITenancyProps} param0
 */
```

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
devinleighsmith: We'll want an answer before merging this. Any input @bbjornholt ?

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * @param propertyTypeId the property type [Parcel, Building]
+ */
+export const InfoContent: React.FC<IInfoContent> = ({ propertyInfo, propertyTypeId }) => {
```
devinleighsmith, line 24: I like how you've split out this display component that doesn't have any state.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
bbjornholt: @agalbrai & @devinleighsmith - It's a good question. I don't think there's a case where we'd need to have both open at the same time. That would probably cramp the map size. My thought is that the pop-out panel should close it when the detail view is opened, but it could be opened manually even if the detail view is open. So it would close, but then the user could pop it open if they need to.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  it('Should render the slide out button', () => {
+    const component = mount(<MapComponent />);
+    waitFor(() => expect(mapRef?.current).toBeTruthy(), { timeout: 500 });
+    const infoButton = component.find(Button).first();
+    expect(infoButton).toBeDefined();
+    expect(infoButton.props().id).toBe('slideOutInfoButton');
+  });
+
+  it('Should component to be closed by default', () => {
```
devinleighsmith, line 42: nit: Component should be closed by default

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -0,0 +1,145 @@
...
+import clsx from 'classnames';
+import * as L from 'leaflet';
+import { useEffect } from 'react';
+import { PropertyPopUpContext } from '../Map';
+import HeaderActions from './HeaderActions';
+import { InfoContent } from './InfoContent';
+import TooltipWrapper from 'components/common/TooltipWrapper';
+
+const InfoContainer = styled.div`
+  margin-right: -10px;
```
devinleighsmith, line 15: What is this negative margin being used for?

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -0,0 +1,145 @@
...
+ * @param setOpen function to set the slideout as open or closed
+ * @param onHeaderActionClick action to be taken when a menu item is clicked
+ */
+const InfoControl: React.FC<IInfoControl> = ({ open, setOpen, onHeaderActionClick }) => {
+  const popUpContext = React.useContext(PropertyPopUpContext);
+
+  useEffect(() => {
+    const elem = L.DomUtil.get('infoContainer');
+    if (elem) {
+      L.DomEvent.on(elem!, 'mousewheel', L.DomEvent.stopPropagation);
```
devinleighsmith, line 102: you probably want to turn this event listener off when this component is cleaned up. useeffect returns a function that is invoked when the component is destroyed, so you can put the L.DomEvent.off(...) there.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -0,0 +1,145 @@
...
+  top: 0;
+  left: -51px;
+  background-color: #fff;
+  color: #1a5a96;
+  border-color: #1a5a96;
+  box-shadow: -2px 1px 4px rgba(0, 0, 0, 0.2);
+  &.open {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+    top: 0px;
```
devinleighsmith, line 69: do you need this top 0 and the top 0 in the parent?

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
devinleighsmith: Id like to see you either use one of the util functions for currency formatting or add one if you have a new use case.

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,47 @@
...
+  } else {
+    formatAssessed = '';
+  }
+  return (
+    <>
+      <ListGroup>
+        <Label className="header">Parcel attributes</Label>
+        <ListGroup.Item>
+          <Label>Lot size:</Label>
+          {parcelInfo?.landArea + ' hectares'}
```
devinleighsmith, line 29: Didn't this get updated to sq m?

frontend/src/components/maps/leaflet/InventoryLayer.tsx
devinleighsmith: Yes, I think it was used back when we had a completely different page when submitting properties, so this is fine.

frontend/src/components/maps/leaflet/Map.tsx
devinleighsmith: Ideally in it's own file, trying to keep the map code from getting more bloated.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -289,86 +301,112 @@ const Map: React.FC<MapProps> = ({
     }
   };
 
+  const [infoOpen, setInfoOpen] = React.useState(false);
```
devinleighsmith, line 304: could put these in the context as well to keep the impact on the map code small.

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: SelectedPropertyMarkers show up whenever the sidebar is open in view or update mode.

frontend/src/mocks/filterDataMock.ts
devinleighsmith: its just a test utility that contains mocked objects, prevents us from having to recreate these every time we create a new test that needs parcels/buildings etx.

devinleighsmith (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
goemen: Styled components has away the wrap the app with the theme provider that allows us to access the theme colors

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.scss
```diff
@@ -0,0 +1,32 @@
...
+  .label {
+    color: $light-variant-color;
+    font-weight: lighter;
+    padding-right: 5px;
+  }
+}
+
+.header {
+  border-bottom: 2px solid rgba(96, 96, 96, 0.2);
+  font-family: 'BCSANS-Bold', 'BC Sans';
```
goemen, line 21: No need to do this here

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo }) => {
+  let formatAssessed;
+  if (parcelInfo?.assessed !== '') {
+    formatAssessed = '$' + new Intl.NumberFormat('en-US').format(parcelInfo?.assessed as number);
+  } else {
+    formatAssessed = '';
+  }
+  return (
+    <>
```
goemen, line 24: You can remove the fragment since ListGroup is the only node here

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelPIDPIN.tsx
```diff
@@ -0,0 +1,34 @@
...
+  parcelInfo: IParcel | null;
+};
+
+/**
+ * Displays PID/PIN information in property popout for selected parcel
+ * @param parcelInfo parcel data
+ */
+export const ParcelPIDPIN: React.FC<IParcelPIDPIN> = ({ parcelInfo }) => {
+  return (
+    <>
```
goemen, line 17: Same here, remove fragment tag

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -289,86 +301,112 @@ const Map: React.FC<MapProps> = ({
...
-                  closePopupOnClick={interactive}
-                  onzoomend={e => setZoom(e.sourceTarget.getZoom())}
-                  onmoveend={handleBounds}
-                >
-                  {activeBasemap && (
-                    <TileLayer
-                      attribution={activeBasemap.attribution}
-                      url={activeBasemap.url}
-                      zIndex={0}
+    <PropertyPopUpContextProvider>
```
goemen, line 308: You should only be wrapping the map with the provider, the filter does not need access to the popup context

goemen (approved):
Few nitpicking comments, but looks good

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: I didn't realize we had some, I'll do that!

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
agalbrai: This one was used in the design, not sure if it matters? @bbjornholt ?

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
agalbrai: I just copied this code from elsewhere, it doesn't really affect it much so I'll pull it out

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
agalbrai: I hear ya, that makes sense 👍

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
agalbrai: ooooh I getcha

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.scss
agalbrai: right you are!

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.test.tsx
agalbrai: woops

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
agalbrai: it keeps the container against the edge of the screen, it floats on the map otherwise

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
agalbrai: fixed it, I only need the top 0 when it's open, it keeps the button at the top edge of the container

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
agalbrai: I just copied this over from the layers component, I'm not actually sure what this is doing, what do you mean by "cleaned up"?

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
bbjornholt: @agalbrai There is a separate mockup for the non-owning agencies indicating the fields they should be able to see. I'll send you a link.

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
bbjornholt: I think that's my mistake. This can/should be BCSans. thanks!

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: The parcel details page says hectares, I thought it was building area that was changed to square meters?

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: there's multiple ListGroups though

frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
```
agalbrai, line 1: My ticket just says for parcels/land, but I added this as a sort of placeholder so at least there's something

frontend/src/components/maps/leaflet/InfoSlideOut/HeaderActions.tsx
```diff
@@ -0,0 +1,120 @@
...
+      MAX_ZOOM,
+    );
+
+  const buildingId = propertyTypeId === PropertyTypes.BUILDING ? propertyInfo?.id : undefined;
+  const parcelId = propertyTypeId === PropertyTypes.PARCEL ? propertyInfo?.id : undefined;
+  return (
+    <LinkMenu>
+      Actions:
+      {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
+        keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) && (
```
agalbrai, line 69: Non-owning agencies can only see "Zoom map" link

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const keycloak = useKeycloakWrapper();
+  return (
+    <>
+      <ListGroup>
+        <Label className="header">Parcel Identification</Label>
+        {propertyTypeId === PropertyTypes.PARCEL && (
+          <ParcelPIDPIN parcelInfo={propertyInfo as IParcel} />
+        )}
+        {(keycloak.hasAgency(propertyInfo?.agencyId as number) ||
+          keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) && (
```
agalbrai, line 35: Non-owning agencies can only see PID/PIN and Classification in this section

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <ListGroup>
+        <Label className="header">Valuation</Label>
+        <ListGroup.Item>
+          <Label>Assessed value:</Label>
+          {formatAssessed}
+        </ListGroup.Item>
+      </ListGroup>
+      {parcelInfo?.buildings && (
+        <ListGroup>
+          <Label className="header">Associated Buildings</Label>
```
agalbrai, line 43: I'm displaying only three buildings and then "+ __ more" because some parcels (like Uvic) have upwards of 60 buildings

frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Displays Building specific information needed on the information slide out
+ * @param buildingInfo the selected parcel data
+ */
+export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo }) => {
+  const formatAssessed = buildingInfo?.assessed
+    ? formatMoney(buildingInfo?.assessed)
+    : formatMoney(buildingInfo?.evaluations[0].value);
```
agalbrai, line 20: for whatever reason the property detail sent from the SelectedPropertyMarker is in a different format so sometimes it gets the single assessed value and sometimes it doesn't

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+ * Component that displays the appropriate information about the selected property
+ * in the property info slideout
+ * @param {IInfoContent} propertyInfo the selected property
+ * @param {IInfoContent} propertyTypeId the property type [Parcel, Building]
+ */
+export const InfoContent: React.FC<IInfoContent> = ({ propertyInfo, propertyTypeId }) => {
+  const keycloak = useKeycloakWrapper();
+  return (
+    <>
+      <ListGroup>
```
devinleighsmith, line 29: This is pretty long for a screen on a small device. You may need to provide a max-length and scroll: auto in the case that not all of the details are visible on the screen.

jdtoombs (approved):
Looks good Alex!!

frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
devinleighsmith: I think if buildinginfo.assessed and buildingInfo.evaluations are both [] this will throw an arrayindexoutofbounds

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <ListGroup.Item>
+            <Label>{parcelInfo.buildings[0].name}</Label>
+          </ListGroup.Item>
+          {parcelInfo.buildings[1] && (
```
devinleighsmith, line 47: You don't want to manually iterating through a list like this, this is unsafe.
if you want to limit the maximum, create a new array of just the top four elements and then do a .map of that array.

devinleighsmith: checking if parcelInfo.buildings[1] exists doesn't work, because if it doesn't exist javascript will throw an error to check if an array element exists use the .length property or use getIn(someArray, someIndex)

frontend/src/components/maps/providers/PropertyPopUpProvider.tsx
```diff
@@ -0,0 +1,25 @@
+import { noop } from 'lodash';
```
devinleighsmith, line 1: This one is hard to explain, so let me just provide some code:

```
import { noop } from 'lodash';
import React from 'react';
import { IParcel, IBuilding } from 'actions/parcelsActions';

interface IPopupContext {
  propertyInfo: IParcel | IBuilding | null;
  setPropertyInfo: (propertyInfo: IParcel | IBuilding | null) => void;
  propertyTypeID: number | null;
  setPropertyTypeID: (propertyTypeId: number) => void;
}

export const PropertyPopUpContext = React.createContext<IPopupContext>({
  propertyInfo: null,
  setPropertyInfo: noop,
  propertyTypeID: null,
  setPropertyTypeID: noop,
});

/**
 * Allows for the property information to be sent to the map
 * when the user clicks on a marker
 */
export const PropertyPopUpContextProvider: React.FC = ({ children }) => {
  const [propertyInfo, setPropertyInfo] = React.useState<IParcel | IBuilding | null>(null);
  const [propertyTypeID, setPropertyTypeID] = React.useState<number | null>(null);
  return (
    <PropertyPopUpContext.Provider
      value={{ propertyInfo, setPropertyInfo, propertyTypeID, setPropertyTypeID }}
    >
      {children}
    </PropertyPopUpContext.Provider>
  );
};

```

devinleighsmith: by using generics, you can provide the expected typings for context and state hooks, which make things much cleaner.

devinleighsmith: If you don't do this, typescript/react infers the type based on the initialized value, which is null in this case. We don't want our context fields typings to be null.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
Fosol: What feature is this for?

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
Fosol: Need to ask Bobbi

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelPIDPIN.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import { ListGroup } from 'react-bootstrap';
+import { IParcel } from 'actions/parcelsActions';
+import { Label } from 'components/common/Label';
+
+interface IParcelPIDPIN {
+  /** the selected parcel information */
+  parcelInfo: IParcel | null;
```
Fosol, line 8: I always wonder if we can pass `null` do we really want use this component?

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
agalbrai: The slideout itself. Can I just remove this useEffect, @devinleighsmith ? It doesn't seem to effect functionality if I do

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: @bbjornholt should Lot Size be in hectares or square meters?

frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
agalbrai: Okay, that makes a bit more sense thank you.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
agalbrai: on it 👍

frontend/src/components/maps/providers/PropertyPopUpProvider.tsx
agalbrai: Thank you so much, that's really helpful

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
devinleighsmith: If this is reused code I'll just log some tech debt for this.

frontend/src/actions/parcelsActions.ts
```diff
@@ -70,7 +70,9 @@ export enum LeasedLand {
 
 export interface IBuilding extends IProperty {
   parcelId: number;
+  pid: number;
   address: IAddress;
+  administrativeArea: string;
```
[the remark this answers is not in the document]
agalbrai: Currently this is what's being returned by the map (address is just a string rather than an IAddress), is this what your PR is changing? 
image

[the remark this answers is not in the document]
Fosol: The map uses GeoJSON which is a flat model of values for the property.  However when displaying the information in the sidebar or on Inventory List View it should be in a more standard format `IParcel` or `IBuilding`.  These have an `IAddress` property `property.address.administrativeArea`.  My PR has a converter that ensures the GeoJSON is changed to match the correct `IParcel` or `IBuilding` interface.  It calls this converter before sending the data to the Popup.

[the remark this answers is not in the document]
agalbrai: I've copied over your converter, thank you!

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: it's hectares 👍

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <ListGroup.Item>
+          <Label>Classification:</Label>
+          {propertyInfo?.classification}
+        </ListGroup.Item>
+      </ListGroup>
+      <ListGroup>
+        <Label className="header">Location data</Label>
+        <ListGroup.Item>
+          <Label>Civic Address:</Label>
+          {propertyInfo?.address?.line1}
```
agalbrai, line 64: fixed the address section

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -0,0 +1,148 @@
...
+`;
+
+const InfoMain = styled.div`
+  width: 100%;
+  max-height: 420px;
+  padding-left: 10px;
+  padding: 0px 10px 5px 10px;
+
+  &.open {
+    overflow-y: scroll;
```
agalbrai, line 52: added a max height and scroll

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Label>Assessed value:</Label>
+          {formatAssessed}
+        </ListGroup.Item>
+      </ListGroup>
+      {parcelInfo?.buildings && (
+        <ListGroup>
+          <Label className="header">Associated Buildings</Label>
+          {buildingsCopy.map((building, buildingId) => (
+            <ListGroup.Item key={buildingId}>
+              <Label>{building.name}</Label>
```
agalbrai, line 55: how's this @devinleighsmith ?

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -30,10 +37,40 @@ export type PointClustererProps = {
...
-  onMarkerClick?: (point: PointFeature, position?: [number, number]) => void;
+  /** Action when a marker is clicked */
+  onMarkerClick: () => void;
   tilesLoaded: boolean;
 };
 
+/**
+ * Converts the flat list of properties into the correct type of inventory property
+ * @param property A flat list of property values (from a Feature).
+ */
```
agalbrai, line 48: I borrowed @Fosol 's converter from his PR

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Displays parcel specific information needed on the information slide out
+ * @param parcelInfo the selected parcel data
+ */
+export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo }) => {
+  let formatAssessed;
+  if (parcelInfo?.assessed) {
+    formatAssessed = formatMoney(parcelInfo?.assessed);
+  } else if (parcelInfo?.evaluations[0].value) {
```
devinleighsmith, line 21: I think this will throw an exception if parcelInfo.assessed is undefined and parcelInfo.evaluations is defined but is equal to [].

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
devinleighsmith: Yes, this is much better, thanks.

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
agalbrai: is this where I would use getIn()?

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -0,0 +1,69 @@
...
+ */
+export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo }) => {
+  let formatAssessed;
+  if (parcelInfo?.assessed) {
+    formatAssessed = formatMoney(parcelInfo?.assessed);
+  } else if (parcelInfo?.evaluations?.length >= 1) {
+    formatAssessed = formatMoney(parcelInfo?.evaluations[0].value);
+  } else {
+    formatAssessed = '$0';
+  }
```
agalbrai, line 25: hopefully last one @devinleighsmith

devinleighsmith (approved):

PR #846, opened 2021-01-28
PA-283 View all properties

Fosol:
This enhancement provides a way for users to search all properties in PIMS on the map.  Please note that this feature is not meant to be fully implemented this sprint.   Properties not owned by the user's agency or sub-agency will only provide minimal information.

Fixed admin agency bugs that would cause user access issues if and when agencies were updated.

The following work remains to be completed to fully enable viewing all properties.

- Right toolbar to display property information.
- Don't allow searching other agencies for their properties.
- Removing fiscal and evaluation information from entities and models
- Correctly implementing assessed information in entities and models
- Providing a way to search for all properties in the Inventory List View

### Implementation Details

There is now a URL query parameter `includeAllProperties=true` that controls whether all properties are returned by the GeoJSON endpoint.

image


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "dotnet-test-explorer.testProjectPath": "./test/BackendApi.Test",
+  "dotnet-test-explorer.testProjectPath": "./backend",
```
Fosol, line 2: VSCode unit-test plugin works now (kinda).

backend/api/Areas/Admin/Controllers/AgencyController.cs
```diff
@@ -98,11 +104,18 @@ public IActionResult GetAgencies(EModel.AgencyFilter filter)
...
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "admin-agency" })]
-        public IActionResult AddAgency([FromBody] Model.AgencyModel model)
+        public async Task<IActionResult> AddAgencyAsync([FromBody] Model.AgencyModel model)
         {
             var entity = _mapper.Map<Entity.Agency>(model);
             _pimsAdminService.Agency.Add(entity);
 
+            // TODO: This isn't ideal as the db update may be successful but this request may not.
+            await entity.Users.ForEachAsync(async u =>
```
Fosol, line 113: Need to tell Keycloak to update the users that were affected by the agency change.

backend/api/Areas/Admin/Controllers/AgencyController.cs
```diff
@@ -118,11 +131,18 @@ public IActionResult AddAgency([FromBody] Model.AgencyModel model)
...
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "admin-agency" })]
-        public IActionResult UpdateAgency([FromBody] Model.AgencyModel model)
+        public async Task<IActionResult> UpdateAgencyAsync([FromBody] Model.AgencyModel model)
         {
             var entity = _mapper.Map<Entity.Agency>(model);
             _pimsAdminService.Agency.Update(entity);
 
+            // TODO: This isn't ideal as the db update may be successful but this request may not.
+            await entity.Users.ForEachAsync(async u =>
```
Fosol, line 140: Need to tell Keycloak to update the users that were affected by the agency change.

backend/api/Areas/Admin/Controllers/AgencyController.cs
```diff
@@ -137,11 +157,18 @@ public IActionResult UpdateAgency([FromBody] Model.AgencyModel model)
         [ProducesResponseType(typeof(Model.AgencyModel), 200)]
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "admin-agency" })]
-        public IActionResult DeleteAgency([FromBody] Model.AgencyModel model)
+        public async Task<IActionResult> DeleteAgencyAsync([FromBody] Model.AgencyModel model)
         {
             var entity = _mapper.Map<Entity.Agency>(model);
             _pimsAdminService.Agency.Remove(entity);
 
+            // TODO: This isn't ideal as the db update may be successful but this request may not.
```
Fosol, line 165: Need to tell Keycloak to update the users that were affected by the agency change.

backend/api/Areas/Property/Mapping/Search/GeoJsonMap.cs
```diff
@@ -9,6 +9,67 @@ public class GeoJsonMap : IRegister
...
+
+            config.NewConfig<Entity.Models.ParcelModel, Model.PropertyModel>()
+                .Inherits<Entity.Models.PropertyModel, Model.PropertyModel>()
+                .Map(dest => dest.PID, src => src.PID)
+                .Map(dest => dest.PIN, src => src.PIN)
+                .Map(dest => dest.LandLegalDescription, src => src.LandLegalDescription)
+                .Map(dest => dest.LandArea, src => src.LandArea)
+                .Map(dest => dest.Zoning, src => src.Zoning)
+                .Map(dest => dest.ZoningPotential, src => src.ZoningPotential);
+            // TODO: Fix assessment values.
```
Fosol, line 52: Intentional, as I still need to figure this stuff out at some point.

backend/api/Areas/Property/Models/Search/PropertyModel.cs
```diff
@@ -52,12 +43,12 @@ public class PropertyModel
         /// <summary>
         /// get/set - The GIS latitude location of the property.
         /// </summary>
-        public double Latitude { get; set; }
+        public double? Latitude { get; set; }
```
Fosol, line 46: Nulled values will not get returned in the JSON from the API by default.  Which will reduce the clutter of the flattening process of returning parcel and building properties in a single model.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -146,6 +147,146 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
...
+        /// <summary>
+        /// Generate a query for the specified 'filter'.
+        /// Returns all properties, even properties that don't belong to the user's agency or sub-agencies.
+        /// The results of this query must be 'cleaned' so that only appropriate data is returned to API.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="user"></param>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public static IQueryable<Entity.Views.Property> GenerateAllPropertyQuery(this PimsContext context, ClaimsPrincipal user, Entity.Models.AllPropertyFilter filter)
```
Fosol, line 163: A lot of this is duplicated from the other queries.  Tech-debt to extract duplication in an extension method at some point.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -146,6 +147,146 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
...
+            var isAdmin = user.HasPermission(Permissions.AdminProperties);
+            var viewSensitive = user.HasPermission(Permissions.SensitiveView);
+
+            // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
+            var query = context.Properties
+                .AsNoTracking()
+                .Where(p => p.ClassificationId != 4); // Disposed properties are not visible.
+
+            // Users are not allowed to view sensitive properties outside of their agency or sub-agencies.
+            if (!isAdmin)
```
Fosol, line 179: This is important to get right.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -146,6 +147,146 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
...
+                query = query.Where(p => !p.IsSensitive || (p.IsSensitive && userAgencies.Contains(p.AgencyId)));
+
+            // Only return properties owned by user's agency or sub-agencies.
+            if (!filter.IncludeAllProperties)
```
Fosol, line 183: Here's the magic new part that allows finding properties owned by other agencies.

Fosol: It however has an unintended consequence of essentially allowing other agencies to know who owns what.  Which is what SRES wants to avoid, as they want to be called instead of the owning agency...

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -31,7 +31,7 @@ public class AgencyService : BaseService<Agency>, IAgencyService
         #endregion
 
         #region Methods
-        
+
```
Fosol, line 34: A lot of ugly updates to this service...  Essentially we need to either update the users affected by agency updates, or we need to return a list of users who will be affected by the agency update so that keycloak can be updated.

backend/dal/Services/Concrete/UserService.cs
```diff
@@ -245,9 +245,13 @@ public AccessRequest UpdateAccessRequest(AccessRequest request)
...
         {
-            var user = this.Context.Users.Include(u => u.Agencies).ThenInclude(a => a.Agency).ThenInclude(a => a.Children).Single(u => u.Id == userId) ?? throw new KeyNotFoundException();
+            var user = this.Context.Users
+                .Include(u => u.Agencies)
+                .ThenInclude(a => a.Agency)
+                .ThenInclude(a => a.Children)
+                .Single(u => u.Id == userId) ?? throw new KeyNotFoundException();
             var agencies = user.Agencies.Select(a => a.AgencyId).ToList();
-            agencies.AddRange(user.Agencies.SelectMany(a => a.Agency?.Children).Select(a => a.Id));
+            agencies.AddRange(user.Agencies.SelectMany(a => a.Agency?.Children.Where(ac => !ac.IsDisabled)).Select(a => a.Id));
```
Fosol, line 254: Don't return agencies that have been disabled.

backend/entities/Models/PropertyModel.cs
```diff
@@ -0,0 +1,206 @@
+using System;
+using System.Linq;
+using System.Security.Claims;
+using NetTopologySuite.Geometries;
+using Pims.Core.Extensions;
+
+namespace Pims.Dal.Entities.Models
```
Fosol, line 7: A handful of new models that the DAL can return.  These specific ones are used to remove information from properties that are not owned by the user's agency or sub-agencies.

backend/entities/Views/Property.cs
```diff
@@ -121,7 +121,7 @@ public class Property
         /// get/set - The most recent market value.
         /// </summary>
         [Column(TypeName = "MONEY")]
-        public decimal Market { get; set; }
+        public decimal? Market { get; set; }
```
Fosol, line 124: Nullable so that we don't get default values.

frontend/src/actions/parcelsActions.ts
```diff
@@ -13,6 +13,7 @@ export enum PropertyTypes {
 export interface IProperty {
   id: number | '';
   propertyTypeId?: PropertyTypes;
+  agencyId?: number | '';
```
Fosol, line 16: I don't like this implementation, but based current requirements properties not owned by the user's agency or sub-agencies are only allowed to share certain information.  The owning agency of the property is not currently supposed to be shared.

frontend/src/components/maps/BuildingPopupView.tsx
```diff
@@ -57,21 +57,27 @@ export const BuildingPopupView: React.FC<IBuildingPopupViewProps> = (
...
-              <ListGroup>
-                <ListGroup.Item>
-                  <Label>Assessed Value:</Label>
-                  {buildingDetail?.evaluations?.length &&
-                    '$' +
-                      buildingDetail?.evaluations
-                        ?.find(e => e.key === EvaluationKeys.Assessed)
-                        ?.value?.toLocaleString('en-US', { minimumFractionDigits: 2 })}
-                </ListGroup.Item>
+                {buildingDetail?.evaluations?.length && (
```
Fosol, line 71: Stop trying to display values not included in the JSON

frontend/src/components/maps/PopupView.tsx
```diff
@@ -32,18 +29,22 @@ export const PopupView: React.FC<IPopupViewProps> = ({
...
   React.useEffect(() => {
-    if (propertyTypeId === PropertyTypes.PARCEL) {
-      getParcel(id as number).then(parcel => {
-        setProperty(parcel);
-      });
-    } else if (propertyTypeId === PropertyTypes.BUILDING) {
-      getBuilding(id as number).then(building => {
-        setProperty(building);
-      });
+    if ((agencyId as number) > 0) {
```
Fosol, line 36: Only make a request to fetch more information if the property is owned by the user's agency or sub-agencies.

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -299,7 +304,10 @@ const Map: React.FC<MapProps> = ({
               <Container fluid className="px-0 map-filter-container">
                 <Container className="px-0">
                   <PropertyFilter
-                    defaultFilter={defaultFilterValues}
+                    defaultFilter={{
+                      ...defaultFilterValues,
+                      includeAllProperties: keycloak.hasClaim(Claims.ADMIN_PROPERTIES),
```
Fosol, line 309: Default to displaying all properties for SRES.

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -34,6 +41,40 @@ export type PointClustererProps = {
   tilesLoaded: boolean;
 };
 
+/**
+ * Converts the flat list of properties into the correct type of inventory property.
+ * @param property A flat list of property values (from a Feature).
+ */
+export const convertToProperty = (property: any): IParcel | IBuilding | null => {
```
Fosol, line 48: This will convert the GeoJSON to a normalized shape for the use of the Popup.

backend/api/Areas/Property/Mapping/Search/GeoJsonMap.cs
```diff
@@ -9,6 +9,67 @@ public class GeoJsonMap : IRegister
     {
         public void Register(TypeAdapterConfig config)
         {
+            config.NewConfig<Entity.Models.PropertyModel, Model.GeoJson<Model.PropertyModel>>()
+                .Map(dest => dest.Type, src => "Feature")
+                .Map(dest => dest.Geometry.Type, src => src.Location.GeometryType)
```
devinleighsmith, line 14: No possibility of nullreference exceptions here?

Fosol: Technically I suppose yes as objects are nullable `Location`.  But this data is from EF and the `Location` will always have a value.

backend/core/Extensions/EnumerableExtensions.cs
```diff
@@ -39,6 +40,36 @@ public static void ForEach<T>(this IEnumerable<T> items, Action<T, int> action)
             }
         }
 
+        /// <summary>
+        /// Iterates through the enumerable collection and performs the specified action.
+        /// </summary>
+        /// <param name="items"></param>
+        /// <param name="action"></param>
+        /// <typeparam name="T"></typeparam>
+        public static async Task ForEachAsync<T>(this IEnumerable<T> items, Func<T, Task> action)
```
devinleighsmith, line 49: This will invoke all of these actions serially correct? to we want a Parallel version?

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -146,6 +147,146 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
...
+        /// <param name="user"></param>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public static IQueryable<Entity.Views.Property> GenerateAllPropertyQuery(this PimsContext context, ClaimsPrincipal user, Entity.Models.AllPropertyFilter filter)
+        {
+            filter.ThrowIfNull(nameof(filter));
+            filter.ThrowIfNull(nameof(user));
+
+            // Check if user has the ability to view sensitive properties.
+            var userAgencies = user.GetAgenciesAsNullable();
```
devinleighsmith, line 169: I don't think the above comment is relevant for userAgencies, up to you if you want to move it.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
devinleighsmith: Could do a weird join with an inner table where you replace the actual agency with SRES maybe?

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                this.Context.UpdateRange(users);
+            }
+            else if (!original.ParentId.HasValue && agency.ParentId.HasValue)
+            {
+                // This agency has become a sub-agency, all original users need their list of agencies reduced to only this agency.
+                updatedUsers.AddRange(this.Context.Users.Include(u => u.Agencies).Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
+            }
+            else if (original.ParentId != agency.ParentId)
+            {
+                // Remove the sub-agency from curretly linked users and add it to the users who belong to the new parent agency.
```
devinleighsmith, line 154: curretly

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+
+            var original = this.Context.Agencies.Find(agency.Id) ?? throw new KeyNotFoundException();
+            var updatedUsers = new List<User>();
+
+            // If the agency has become a sub-agency, or a parent-agency then users will need to be updated.
+            if (original.ParentId.HasValue && !agency.ParentId.HasValue)
+            {
+                // This agency has become a parent agency, all users associated with it will be removed
+                // because users should only be associated with a single parent.
+                updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == original.ParentId)));
```
devinleighsmith, line 140: I found this comment misleading because updatedUsers are the users that the agency will be added to, but the comment specifically refers to removing associations. I'd prefer it if the comment for a line specifies the functionality of that line. If you put a single line of whitespace between the comment and the new line I'd be fine with this. Or even better, add a comment specific to line 140 and keep the 138-139 comment above that.

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+
+                // Remove any user who are linked to the agency.
+                var users = this.Context.Users.Include(u => u.Agencies).Where(u => u.Agencies.Any(ua => ua.AgencyId == agency.Id));
+                users.ForEach(u => u.Agencies.Clear());
+                this.Context.UpdateRange(users);
+            }
+            else if (!original.ParentId.HasValue && agency.ParentId.HasValue)
+            {
+                // This agency has become a sub-agency, all original users need their list of agencies reduced to only this agency.
```
devinleighsmith, line 149: only to this agency -> So should this logic be clearing the unrelated agencies then?

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                updatedUsers.AddRange(this.Context.Users.Include(u => u.Agencies).Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
+            }
+            else if (original.ParentId != agency.ParentId)
+            {
+                // Remove the sub-agency from curretly linked users and add it to the users who belong to the new parent agency.
+                updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == original.ParentId)));
```
devinleighsmith, line 155: I must be missing something, doesn't updatedUsers just add the agency being updated to all the users in the list? if so, how is this removing the sub-agency from users who are currently linked?

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                }
+            }
 
-            this.Context.Entry(agency).CurrentValues.SetValues(entity);
-            base.Update(agency);
+            this.Context.Entry(original).CurrentValues.SetValues(agency);
+            base.Update(original);
+
+            // Mutate original entity.
+            this.Context.Entry(agency).CurrentValues.SetValues(original);
```
devinleighsmith, line 186: is this just setting the rowversion/metadata?

devinleighsmith: because you do the context.entry twice, which is a pattern I haven't seen before.

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                    // This will result in the user not belonging to an agency.
+                    var users = this.Context.Users.Include(u => u.Agencies).Where(u => u.Agencies.Any(a => a.AgencyId == agency.Id));
+                    users.ForEach(u => u.Agencies.Clear());
+                    updatedUsers.AddRange(users);
+                    this.Context.UpdateRange(users);
+                }
+                else if (agency.ParentId.HasValue)
+                {
+                    // Add the sub-agency to users who are associated with the parent.
+                    updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
```
devinleighsmith, line 178: yeah, so this code that is supposed to be doing the adding of the agency is identical to the code that does the removal in my above comment?

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: It technically doesn't.  The DB DAL doesn't have access to the Keycloak DAL.
Users are only ever currently assigned to a single agency.  So the **User -> Agencies** relationship will currently only have the single agency they're assigned to.  However if they are assigned to a parent agency, they will **inherit** through logic all sub-agencies.  But the DB will only have the parent agency listed.
Most of the code here is simply returning all the users affected by the agency update.  In other words the DB doesn't need to be updated but keycloak does.  This is because in keycloak we keep a list of the parent agency and all sub-agencies the user has access to.
Now we return the list of users affected so that a separate routine can update keycloak.

I've noted in the comments that this isn't ideal, but we are dealing with two separate systems that need to be updated.

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
+            }
 
-            var agency = this.Context.Agencies.Find(entity.Id) ?? throw new KeyNotFoundException();
+            if (original.IsDisabled != agency.IsDisabled)
+            {
+                if (agency.ParentId.HasValue && agency.IsDisabled)
+                {
+                    // Remove the sub-agency from users.
+                    updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
```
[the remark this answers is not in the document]
Fosol: Same answer as above.

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: Save answer as above.

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+
+            this.Context.Entry(original).CurrentValues.SetValues(agency);
+            base.Remove(original);
+
+            // Mutate original entity.
+            this.Context.Entry(agency).CurrentValues.SetValues(original);
+            updateUsers.ForEach(u =>
+            {
+                this.Context.Entry(u).State = EntityState.Detached;
+                agency.Users.Add(new UserAgency(u, agency));
```
devinleighsmith, line 228: I don't think I understand what this is doing. If you detach the user and then add a UserAgency does that even modify the DB?

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: I just need to provide additional comments for clarification.  The current comment is correct in what needs to be done.  Line 140 is the logic to ensure that the comment is fulfilled.  However the full answer is below.

backend/dal/Services/Admin/Concrete/AgencyService.cs
devinleighsmith: Also, wouldn't this be adding a UserAgency for the agency that just got deleted?

backend/entities/Models/ParcelModel.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            var userAgencies = user.GetAgenciesAsNullable();
+
+            // The property belongs to the user's agency or sub-agency, so include these properties.
+            if (userAgencies.Contains(property.AgencyId))
+            {
+                this.Zoning = property.Zoning;
+                this.ZoningPotential = property.ZoningPotential;
+
+                // this.AssessedLand = property.AssessedLand;
```
devinleighsmith, line 103: on purpose?

Fosol: yes, there is still work to be done on linking assessed values.

devinleighsmith (approved):

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: This code is ugly, but required either here or somewhere else.  If I leave the user attached to context then further operations on the user with existing code has conflicts regrettably.  In an ideal world any entity we returned from the DAL would be detached to ensure no unintentional tampering.
Adding the users to the agency is a simple way to return the users that were affected by the update.  

The best solution would probably be to give the DAL access to Keycloak...  But this also has complications and will take too much time.

backend/dal/Services/Admin/Concrete/AgencyService.cs
devinleighsmith: maybe just put them both in the same if statement then?

backend/api/Areas/Admin/Controllers/AgencyController.cs
```diff
@@ -118,11 +131,18 @@ public IActionResult AddAgency([FromBody] Model.AgencyModel model)
         [ProducesResponseType(typeof(Model.AgencyModel), 200)]
         [ProducesResponseType(typeof(Api.Models.ErrorResponseModel), 400)]
         [SwaggerOperation(Tags = new[] { "admin-agency" })]
-        public IActionResult UpdateAgency([FromBody] Model.AgencyModel model)
+        public async Task<IActionResult> UpdateAgencyAsync([FromBody] Model.AgencyModel model)
         {
             var entity = _mapper.Map<Entity.Agency>(model);
             _pimsAdminService.Agency.Update(entity);
```
devinleighsmith, line 137: I'm pretty against using side effects to modify values like this (in this case the list of users) can we return something instead?

backend/core/Extensions/EnumerableExtensions.cs
Fosol: Good suggestion, I'll update.

backend/dal/Services/Admin/Concrete/AgencyService.cs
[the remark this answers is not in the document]
devinleighsmith: So if I disable Fraser Health, then this will add all of the users who belong to the MOH (because MOH is my parent agency) to this updatedUsers list. All of the users in that list then get the disabled Fraser Health agency added to them (but not in the DB) with this code:

```
updatedUsers.Distinct().ForEach(u =>
            {
                this.Context.Entry(u).State = EntityState.Detached;
                agency.Users.Add(new UserAgency(u, agency));
            });
``` 
and then that list of users gets sent to keycloak right? 

Presumably for all of the users in my agency that just removes the agency from their keycloak user.

How does this show up in keycloak for a user in another MOH subagency? I guess for them this is just a no-op?

For users that belong to the parent agency, I guess we aren't worried because our Logic based inheritence should prevent them from using the disabled agency?

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
+                {
+                    // Remove the sub-agency from users.
+                    updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
+                }
+                else if (!agency.ParentId.HasValue && agency.IsDisabled)
+                {
+                    // Remove the agency from the users.
+                    // This will result in the user not belonging to an agency.
+                    var users = this.Context.Users.Include(u => u.Agencies).Where(u => u.Agencies.Any(a => a.AgencyId == agency.Id));
+                    users.ForEach(u => u.Agencies.Clear());
```
devinleighsmith, line 171: So if we disable an agency with no parent we actually remove that agency from the users in the DB. Why can't we do this if the agency does have a parent?

backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -100,35 +100,133 @@ public Agency Get(int id)
...
 
-            var agency = this.Context.Agencies.Find(entity.Id) ?? throw new KeyNotFoundException();
+            if (original.IsDisabled != agency.IsDisabled)
+            {
+                if (agency.ParentId.HasValue && agency.IsDisabled)
+                {
+                    // Remove the sub-agency from users.
+                    updatedUsers.AddRange(this.Context.Users.Where(u => u.Agencies.Any(a => a.AgencyId == agency.ParentId)));
+                }
+                else if (!agency.ParentId.HasValue && agency.IsDisabled)
```
devinleighsmith, line 166: Do we need an if statement for the opposite of this case? That way we can re-add the agency to the users when an agency is re-enabled.

backend/api/Areas/Admin/Controllers/AgencyController.cs
Fosol: Technically the DAL is doing more than it should.  All it cares about is the PIMS DB.  I've added returning the affected users so that we can keep Keycloak in sync.

In an ideal solution I would through dependency injection (DI) add an interface in the DAL that subsequently informs DI interface that an update has been made to this agency and allow DI to figure out how to talk to other services like Keycloak.  But the time to do this is extreme.  

So I've gone the functional route.  I've let the Controller handle the updating of Keycloak after it receives a list of affected users.

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: An agency with no parent is a parent-agency (or a ministry).  If it becomes disabled all users linked to that agency will need to lose their access to that agency.

An agency with a parent is a sub-agency.  If it becomes disabled, all we need to do is remove that single sub-agency from the users Keycloak credentials.

Fosol:
I have updated the comments in `AgencyService` to explain what's going on better.

backend/api/Areas/Admin/Controllers/AgencyController.cs
devinleighsmith: All I'm asking is that you return an object and operate on that rather then depending on your function to modify values in one of the function parameters.

If for some reason you need to do this, then add an out variable so that at least it is clear which values the function modifies.

backend/dal/Services/Admin/Concrete/AgencyService.cs
devinleighsmith: although now the if statement on 169-170 can be cleaned up by removing the && agency.isDisabled and the && !agency.isDisabled

devinleighsmith: so just agency.parent.hasvalue

devinleighsmith (approved):

backend/dal/Services/Admin/Concrete/AgencyService.cs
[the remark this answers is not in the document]
Fosol: If you disable a sub-agency, all users who belong to it or the parent-agency will be returned by this function.  The Controller will then update Keycloak for all users in the list.

Users who belong to the parent agency will be updated to no longer have access to the disabled agency.
Users who belong only to the sub-agency will no longer have access to data.

Keycloak keeps an attribute array of all agencies the user has access to (i.e. if the user is linked to a parent agency `[1, 5, 20, 30]`, or if the user is linked only to a sub-agency `[5]`).

There are still issues when an agency is disabled for a user currently logged in.  This is because their lookup list of agencies still contains the disabled agency.  If they refresh it will be gone.  Additionally their token will contain the agency that was disabled until it is refreshed.

PR #854, opened 2021-01-29
Invalidate cache after changing financial values

goemen:


frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -501,11 +501,20 @@ const PropertyListView: React.FC = () => {
...
+                      netBook: item.netBook,
+                      market: item.market,
+                    } as any;
                   }
                   return item;
                 });
               }
+
+              // Set key in localstorage to force the properties search to invalidate cached cache
+              localStorage.setItem(propertiesListViewChangedKey, 'true');
```
devinleighsmith, line 517: is there no way to invalidate the cache directly? this seems a bit hacky.

goemen: I could not find a better way, but I welcome better ideas

Fosol: Possibly you could include an extra query string parameter in the request. The cache i believe is based on that.  Then just make a request with a new value.  But you would need to test if the cache header will use a random query string

Fosol (approved):

frontend/src/features/properties/list/PropertyListView.tsx
devinleighsmith: Couple of ideas here: 
https://stackoverflow.com/questions/49263559/using-javascript-axios-fetch-can-you-disable-browser-cache

agalbrai:
why do we want to do this?

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -14,7 +14,7 @@ import { ILookupCodeState } from 'reducers/lookupCodeReducer';
 import { IPropertyQueryParams, IProperty } from '.';
```
agalbrai, line 14: why do we want to invalidate the cache?

Fosol: We cache to improve performance.  Essentially we don't want to keep asking the API to give us the same thing when the last copy is still good enough.  However when that data has been updated, we need a way to say 'no we really need a fresh copy'.  Otherwise what the client sees is old data and often incorrect.

jdtoombs (approved):

PR #855, opened 2021-01-29
Pa 2584 insert "add buildings" button on new detail slideout.

devinleighsmith:
Includes a fix for the properties related api failure in dev.
Also fixes the slideout's view/update/zoom buttons.
Also fixes the map zooming out when the sidebar is closed (filter changes re-introduced this bug).


backend/api/Areas/Property/Mapping/Search/GeoJsonMap.cs
```diff
@@ -22,7 +39,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.PropertyTypeId, src => src.PropertyTypeId)
                 .Map(dest => dest.Name, src => src.Name)
                 .Map(dest => dest.Description, src => src.Description)
-                .Map(dest => dest.ProjectNumbers, src => src.ProjectNumbers)
+                .Map(dest => dest.ProjectNumbers, src => JsonSerializer.Deserialize<IEnumerable<string>>(src.ProjectNumbers ?? "[]", _serializerOptions))
```
devinleighsmith, line 42: this was the issue affecting dev.

frontend/src/actions/parcelsActions.ts
```diff
@@ -242,3 +243,20 @@ export const storeBuildingDetail = (
     position,
   },
 });
+
+export interface IStoreAssociatedBuildingDetail {
```
devinleighsmith, line 247: decided to go with a new field in redux for this as reusing the parcelDetail field causes undesirable side effects.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -98,13 +107,72 @@ export type InfoControlProps = {
  */
 const InfoControl: React.FC<InfoControlProps> = ({ open, setOpen, onHeaderActionClick }) => {
   const popUpContext = React.useContext(PropertyPopUpContext);
+  const leaflet = useLeaflet();
+  const propertyInfo = popUpContext.propertyInfo;
+  const location = useLocation();
+  const history = useHistory();
+  const dispatch = useDispatch();
+  const { getParcel } = useApi();
+  const jumpToView = () =>
```
devinleighsmith, line 116: moved some of this action logic up a level since the associatedBuilding link needs it as well.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -98,13 +107,72 @@ export type InfoControlProps = {
...
+        to={{
+          pathname: `/mapview`,
+          search: queryString.stringify({
+            ...queryString.parse(location.search),
+          }),
+        }}
+        onClick={() => {
+          toast.dark('Creating new associated building...');
+          getParcel(propertyInfo?.id as number)
+            .then((parcel: IParcel) => {
```
devinleighsmith, line 146: we need the parcel in the inventory format with all metadata so that the update doesn't trigger unintended changes.

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -62,6 +67,9 @@ export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo }) =>
           )}
         </ListGroup>
       )}
+      <ListGroup>
+        <ListGroup.Item>{addAssociatedBuildingLink}</ListGroup.Item>
```
devinleighsmith, line 71: just pass in the jsx so this display component doesn't need any action logic.

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -279,7 +279,14 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
               icon={getMarkerIcon(cluster)}
               onclick={() => {
                 onMarkerClick(); //open information slideout
-                popUpContext.setPropertyInfo(convertToProperty(cluster.properties));
+                const [longitude, latitude] = cluster.geometry.coordinates;
+                popUpContext.setPropertyInfo(
+                  convertToProperty({
+                    ...cluster.properties,
+                    latitude: latitude,
```
devinleighsmith, line 286: the cluster doesn't contain the lat/lng which is required for the view/update/zoom actions in the slide out.

frontend/src/features/mapSideBar/hooks/useQueryParamSideBar.tsx
```diff
@@ -62,7 +62,11 @@ const useQueryParamSideBar = (formikRef?: any): IMapSideBar => {
   useDeepCompareEffect(() => {
     setShowSideBar(searchParams.sidebar === 'true');
     setParcelId(searchParams.parcelId ? +searchParams.parcelId || undefined : undefined);
-    setBuildingId(searchParams.buildingId ? +searchParams.buildingId || undefined : undefined);
+    setBuildingId(
```
devinleighsmith, line 65: differentiate 0 from null/undefined.

frontend/src/features/mapSideBar/hooks/useSideBarBuildingLoader.tsx
```diff
@@ -44,30 +51,36 @@ const useSideBarBuildingLoader = ({
...
+          sideBarContext,
+        )
+      ) {
+        const newContext = disabled
+          ? SidebarContextType.VIEW_BUILDING
+          : SidebarContextType.UPDATE_BUILDING;
+        newContext !== sideBarContext && setSideBarContext(newContext);
+      }
+    } else if (!!cachedBuildingDetail && !cachedBuildingDetail?.id) {
+      setSideBarContext(SidebarContextType.ADD_BUILDING);
```
devinleighsmith, line 83: technically we are adding a building even though we are passing initial building data.

frontend/src/hooks/useRouterFilter.ts
```diff
@@ -21,7 +21,15 @@ import { PropertyTypes } from 'constants/propertyTypes';
 const extractProps = (props: string[], source: any): any => {
   var dest = {} as any;
   props.forEach(p => {
-    if (source[p] !== undefined) dest[p] = source[p];
+    if (source[p] !== undefined) {
```
devinleighsmith, line 24: 'true' vs true was causing the filter .equals function to fail.

If we keep having problems with this will need to implement whitelist style comparison on the filter to avoid new additions breaking the comparison.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -98,13 +107,72 @@ export type InfoControlProps = {
...
+      MAX_ZOOM,
+    );
+  const zoomToView = () =>
+    leaflet.map?.flyTo(
+      [propertyInfo?.latitude as number, propertyInfo?.longitude as number],
+      MAX_ZOOM,
+    );
 
   useEffect(() => {
     const elem = L.DomUtil.get('infoContainer');
```
Fosol, line 103: What's this stuff for?

Fosol (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
devinleighsmith: unrelated to my changes, but this prevents scrolling on the slideout from zooming the map.

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -8,12 +8,15 @@ import './InfoSlideOut.scss';
...
 import BuildingAttributes from './BuildingAttributes';
+import { ReactElement } from 'react';
 
 interface IInfoContent {
   /** the selected property information */
   propertyInfo: IParcel | IBuilding | null;
   /** The property type [Parcel, Building] */
   propertyTypeId: PropertyTypes | null;
+  /** ReactElement used to display a link for adding an associated building */
+  addAssociatedBuildingLink: ReactElement;
```
agalbrai, line 19: I didn't realize you could pass a whole element, that's pretty neat

frontend/src/components/maps/leaflet/PointClusterer.tsx
agalbrai: huh, weird that it wasn't giving me any issues before

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -98,13 +107,72 @@ export type InfoControlProps = {
...
+          getParcel(propertyInfo?.id as number)
+            .then((parcel: IParcel) => {
+              jumpToView();
+              dispatch(
+                storeAssociatedBuilding({
+                  ...defaultBuildingValues,
+                  parcelId: parcel?.id,
+                  parcels: [{ ...parcel }],
+                }),
+              );
```
agalbrai, line 154: is there anyway this could lead to duplicate buildings being created? (like if someone wants to attach an existing building but clicks the link on the slide out and creates a new one instead) or do we even care/check for that in general?

agalbrai (commented):
thanks for fixing the links!

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
devinleighsmith: It would create a duplicate if they went all the way through the inventory form and saved it. However, theoretically when the looked at the map they would see the pin on the building already and wouldn't add it again. Also, eventually we'll be showing all of the existing associated buildings on the slideout, so they'll see their building already exists and hopefully won't feel the need to add it again.

goemen (approved):

PR #859, opened 2021-01-29
Show all land pins

Fosol:
Bare land wasn't showing up because buildings no longer have a single link to a parcel `building.parcelId`.  We're going back to showing all land parcel pins.  We can reevaluate when we have more time.


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -213,12 +213,11 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
-        return items.find(
-          ({ properties }: any) =>
-            properties.propertyTypeId === 1 && properties.parcelId === property.parcelId,
-        );
-      };
+      /**
+       * Whether the land has buildings on it.
+       * @param property PIMS property
+       */
+      const hasBuildings = (property: IParcel | IBuilding) => false;
```
Fosol, line 220: Leaving function in as we'll probably need to go back to it.  For now it force all land to show up as pins.

jdtoombs (approved):

PR #861, opened 2021-01-29
Ensure that the admin properties flag is included in the filter durin…

devinleighsmith:
…g initial load.

### DIFF
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -153,7 +153,10 @@ const Map: React.FC<MapProps> = ({
 }) => {
   const keycloak = useKeycloakWrapper();
   const dispatch = useDispatch();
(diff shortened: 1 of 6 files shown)


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -153,7 +153,10 @@ const Map: React.FC<MapProps> = ({
 }) => {
   const keycloak = useKeycloakWrapper();
   const dispatch = useDispatch();
-  const [geoFilter, setGeoFilter] = useState<IGeoSearchParams>({});
+  const [geoFilter, setGeoFilter] = useState<IGeoSearchParams>({
```
devinleighsmith, line 156: This ensures that the include all flag is set on initial map load.

Fosol (approved):

agalbrai (approved):

PR #862, opened 2021-01-29
Small slide out fixes

agalbrai:
- Legal land description is occasionally empty so conditionally render that section
- Classification value disappeared in conversion so fixed that
- Change title for "Building" vs "Land" Identification


backend/api/Areas/Property/Mapping/Search/GeoJsonMap.cs
```diff
@@ -43,6 +43,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.AddressId, src => src.AddressId)
                 .Map(dest => dest.Address, src => src.Address)
                 .Map(dest => dest.AdministrativeArea, src => src.AdministrativeArea)
+                .Map(dest => dest.Classification, src => src.Classification)
```
agalbrai, line 46: this is fixing every situation EXCEPT for viewing a building from the property list view, and I can't figure out why

Fosol (approved):

devinleighsmith (approved):

PR #865, opened 2021-01-30
PA-2612 correct error display in SPL forms.

devinleighsmith:


frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -159,8 +161,8 @@ const SplStep = ({ formikRef }: IStepProps) => {
             formikRef,
             submitStatusCode,
             getStatusTransitionWorkflow(submitStatusCode),
-          ).then((project: IProject) => {
-            actions.setValues(project);
+          ).then((project: IApiProject) => {
+            actions.resetForm({ values: toFlatProject(project) });
```
devinleighsmith, line 165: two problems fixed here.
1) we want to set the initial state to whatever we get from the api, so reset form does that for us.
2) the format of the data returned by the api needs to be transformed before we inject it into formik.

frontend/src/components/common/form/FastDatePicker.tsx
```diff
@@ -74,7 +75,7 @@ const FormikDatePicker: FunctionComponent<FastDatePickerProps> = ({
   if (typeof value === 'string') {
     value = moment(value, 'YYYY-MM-DD').toDate();
   }
-  if (value && value !== initialValue) {
+  if (value && dequal(moment(initialValue, 'YYYY-MM-DD'), moment(value, 'YYYY-MM-DD'))) {
```
devinleighsmith, line 78: this is what was triggering all of the extra validations.

jdtoombs: Good catch

jdtoombs (approved):

PR #867, opened 2021-01-30
PA-2640 Added Delete Property

Fosol:
Users with `admin-properties` claim, or users with `property-delete` and the property is owned by their agency can now delete properties.

image


frontend/src/actionCreators/parcelsActionCreator.ts
```diff
@@ -167,17 +167,23 @@ const parcelDeletingToasts: LifecycleToasts = {
   errorToast: pimsToasts.parcel.PARCEL_DELETING_ERROR,
 };
 
-export const deleteParcel = (parcel: IParcel) => (dispatch: Function) => {
+/**
+ * Make an AJAX request to delete the specified 'parcel' from inventory.
+ * @param parcel IParcel object to delete from inventory.
+ */
+export const deleteParcel = (parcel: IParcel) => async (dispatch: Function) => {
```
Fosol, line 174: Converted to async/await.  For some reason our `customAxios` doesn't return an error object.  It's always `undefined` in this `try/catch`...

frontend/src/features/mapSideBar/containers/MapSideBarContainer.tsx
```diff
@@ -244,15 +255,33 @@ const MapSideBarContainer: React.FunctionComponent<IMapSideBarContainerProps> =
...
   const ConditionalEditButton = () => (
     <>
       {disabled &&
         (keycloak.hasClaim(Claims.ADMIN_PROPERTIES) ||
-          keycloak.hasClaim(Claims.PROPERTY_EDIT) ||
-          keycloak.agencyId === parcelDetail?.agencyId ||
-          keycloak.agencyId === buildingDetail?.agencyId) && (
-          <EditButton onClick={() => setDisabled(false)} />
-        )}
+          (keycloak.hasClaim(Claims.PROPERTY_EDIT) &&
```
Fosol, line 265: Corrected logic here.  `admin-properties` or a user with `property-edit` and the property belongs to their agency.  This however is still wrong, as a user should be able to view and edit sub-agency properties.  I'll look into this later.

frontend/src/features/mapSideBar/containers/MapSideBarContainer.tsx
```diff
@@ -244,15 +255,33 @@ const MapSideBarContainer: React.FunctionComponent<IMapSideBarContainerProps> =
...
+    </>
+  );
+
+  /**
+   * Only display the delete button if the user has the correct claim and the property is owned by their agency.
+   */
+  const ConditionalDeleteButton = () => (
+    <>
+      {disabled &&
+        (keycloak.hasClaim(Claims.ADMIN_PROPERTIES) ||
```
Fosol, line 279: `admin-properties` or a user with `property-edit` and the property belongs to their agency.  This however is still wrong, as a user should be able to view and edit sub-agency properties.  I'll look into this later.

frontend/src/features/mapSideBar/containers/MapSideBarContainer.tsx
```diff
@@ -410,6 +442,33 @@ const MapSideBarContainer: React.FunctionComponent<IMapSideBarContainerProps> =
...
+              case SidebarContextType.VIEW_BUILDING:
+                await deleteBuilding(buildingDetail as IBuilding)(dispatch);
+                break;
+              default:
+                await deleteParcel(parcelDetail as IParcel)(dispatch);
+                break;
+            }
+            setShowSideBar(false, undefined, undefined, true);
+          } catch (error) {
+            debugger;
```
Fosol, line 465: We need to handle errors like concurrency, but for some reason the error is always `undefined`.

jdtoombs (approved):

frontend/src/actionCreators/parcelsActionCreator.ts
devinleighsmith: Custom axios may be doing some error handling itself? or you need to do something different when doing an await ie pass a parameter to axios or something.

devinleighsmith (approved):

PR #875, opened 2021-02-02
PA-2662 Fix up slide out styling

agalbrai:
- Still needs a fix, associated land/buildings are not being pulled properly and not populating
image

### DIFF
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -131,6 +131,7 @@ namespace Pims.Dal.Services
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Fiscals)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Address).ThenInclude(a => a.Province)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Classification)
+                .Include(b => b.Classification)
                 .Include(p => p.BuildingPredominateUse)
                 .Include(p => p.BuildingConstructionType)
                 .Include(p => p.BuildingOccupantType)
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
@@ -4,7 +4,8 @@ import { IBuilding } from 'actions/parcelsActions';
 import { Label } from 'components/common/Label';
 import './InfoSlideOut.scss';
 import { formatMoney } from 'utils/numberFormatUtils';
-import { compareDate } from './InfoContent';
+import { compareDate, OuterRow } from './InfoContent';
+import { ThreeColumnItem } from './ThreeColumnItem';
 
 interface IBuildingAttributes {
   /** the selected building information */
@@ -31,17 +32,18 @@ export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo
     <>
       <ListGroup>
         <Label className="header">Valuation</Label>
-        <ListGroup.Item>
-          <Label>Assessed value:</Label>
-          {formatAssessed}
-        </ListGroup.Item>
+        <OuterRow>
+          <ThreeColumnItem leftSideLabel={'Assessed value:'} rightSideItem={formatAssessed} />
+        </OuterRow>
       </ListGroup>
       {buildingInfo?.pid && (
         <ListGroup>
           <Label className="header">Associated Land</Label>
-          <ListGroup.Item>
-            <Label>{buildingInfo.pid}</Label>
-          </ListGroup.Item>
+          <OuterRow>
+            <ListGroup.Item>
+              <Label>{buildingInfo.pid}</Label>
+            </ListGroup.Item>
+          </OuterRow>
         </ListGroup>
       )}
     </>
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
@@ -28,6 +28,9 @@ export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo
     formatAssessed = '';
   }
 
+  const newLength = buildingInfo.parcels?.length > 3 ? 3 : buildingInfo.parcels?.length;
+  const parcelsCopy = buildingInfo.parcels?.slice(0, newLength);
+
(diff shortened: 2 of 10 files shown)


frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -48,12 +47,12 @@ const InfoHeader = styled.div`
 
 const InfoMain = styled.div`
   width: 100%;
-  max-height: 420px;
   padding-left: 10px;
   padding: 0px 10px 5px 10px;
 
   &.open {
     overflow-y: scroll;
+    max-height: calc(100vh - 380px);
```
agalbrai, line 55: This allows the slide out to stretch to the edge of the screen if need be, but shrink for smaller screens

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+`;
+
+const RightCol = styled(Col)`
+  padding-left: 10px;
+  padding-right: 0px;
+`;
+
+interface IThreeColItem {
+  leftSideLabel: string;
+  rightSideItem: any;
```
agalbrai, line 31: not sure if this "any" is an issue, I think I can change this to "string | undefined" if need be

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+  width: 110px;
+  max-width: 110px;
+  padding-right: 10px;
+  padding-left: 0px;
+`;
+
+const CenterCol = styled(Col)`
+  max-width: 1px;
+  padding: 0px;
+  background-color: rgba(96, 96, 96, 0.2);
```
agalbrai, line 21: this creates the vertical bar that will stretch the height of the column

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
devinleighsmith: does this take into account the height of the footer?

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
devinleighsmith: I'd always prefer the attempt to be as close to the expected value as possible.

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import { Col, ListGroup, Row } from 'react-bootstrap';
+import { Label } from 'components/common/Label';
+
+const InnerRow = styled(Row)`
+  margin: 0px;
+  width: 300px;
```
devinleighsmith, line 8: assuming the outer container has a fixed width, you should be able to use col-md-x to size columns based on a percentage of the overall row. This is a bit less brittle then specifying multiple fixed widths, but may not always be an option depending on the situation.

What I'm trying to say is, please use col-md-x where possible. if not, feel free to use fixed widths.

devinleighsmith (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
agalbrai: yes, I added up (or tried to get close to) the header, filter bar and footer

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
agalbrai: can you do that with styled components?

agalbrai: I played around with it, I couldn't seem to get it to work

jdtoombs (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -238,6 +239,23 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+          popUpContext.setPropertyInfo(parcel);
+        });
+      } else if (propertyTypeId === PropertyTypes.BUILDING) {
+        getBuilding(id as number).then(building => {
+          popUpContext.setPropertyInfo(building);
+        });
+      }
+    },
+    [getParcel, getBuilding, popUpContext],
+  );
```
agalbrai, line 257: now can get all appropriate fields

frontend/src/components/maps/leaflet/InfoSlideOut/BuildingAttributes.tsx
```diff
@@ -27,21 +28,25 @@ export const BuildingAttributes: React.FC<IBuildingAttributes> = ({ buildingInfo
     formatAssessed = '';
   }
 
+  const newLength = buildingInfo.parcels?.length > 3 ? 3 : buildingInfo.parcels?.length;
+  const parcelsCopy = buildingInfo.parcels?.slice(0, newLength);
+
```
agalbrai, line 33: I know most buildings won't have more than 1 parcel, but i thought I'd do the same as associated buildings just in case

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -131,6 +131,7 @@ public Building Get(int id)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Fiscals)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Address).ThenInclude(a => a.Province)
                 .Include(b => b.Parcels).ThenInclude(pb => pb.Parcel).ThenInclude(b => b.Classification)
+                .Include(b => b.Classification)
```
agalbrai, line 134: buildings were missing classification

Fosol (approved):
Not to make your life difficult, but the design provided around agencies is incorrect.  The owning agency is the agency that owns the property.  But currently you have that listed often as the parent ministry.  In reality the parent ministry doesn't own the property, the sub-agency owns it.  The implementation and design should be as follows;
- If the property is owned by a sub-agency, then display `ministry: name` `owning agency: name`
- If the property is owned by a ministry, then display `owning ministry: name`

You don't have to fix it in this PR.  Probably should create a new story for this, but if you want feel free to update in this PR.

PR #882, opened 2021-02-04
Pa 2659

devinleighsmith:
ensure that parcels/buildings using current year not most recent year.

### DIFF
--- a/frontend/src/features/mapSideBar/SidebarContents/subforms/BuildingReviewPage.tsx
+++ b/frontend/src/features/mapSideBar/SidebarContents/subforms/BuildingReviewPage.tsx
@@ -47,7 +47,6 @@ export const BuildingReviewPage: React.FC<any> = (props: IReviewProps) => {
     },
     [props.nameSpace],
   );
-  const netBookYear = getIn(formikProps.values, withNameSpace('financials.0.netbook.fiscalYear'));
   const defaultEditValues = useMemo(
     () => ({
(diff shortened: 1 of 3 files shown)


frontend/src/features/mapSideBar/SidebarContents/subforms/BuildingReviewPage.tsx
```diff
@@ -45,7 +45,6 @@ export const BuildingReviewPage: React.FC<any> = (props: IReviewProps) => {
     },
     [props.nameSpace],
   );
-  const netBookYear = getIn(formikProps.values, withNameSpace('financials.0.netbook.fiscalYear'));
```
devinleighsmith, line 48: Whilst this technically works since all 10 years of history from the current year are created, and in sorted order, and currently only contain netbook values, this doesn't work if another type of fiscal is added. Switching to searching based on the programmatically determined fiscal index is safer.

jdtoombs (approved):

Fosol (approved):

frontend/src/features/mapSideBar/SidebarContents/subforms/BuildingReviewPage.tsx
agalbrai: it took my brain a minute to read "programmatically" lol

agalbrai (approved):

PR #884, opened 2021-02-04
Add SPP Number to inventory list view

agalbrai:
image


frontend/src/features/properties/list/PropertyListView.test.tsx
```diff
@@ -197,7 +197,6 @@ describe('Property list view', () => {
           netBook: 0.0,
           assessedLand: 958000.0,
           assessedLandDate: '2018-01-01T00:00:00',
-          appraised: 0.0,
```
agalbrai, line 200: this field must have been removed from the type

frontend/src/features/properties/list/columns.tsx
```diff
@@ -33,6 +41,19 @@ const spacing = {
   xxlarge: unit * 8,
 };
 
+const getProjectLinkNoDrafts = (namespace: string = 'properties') => (cellInfo: any) => {
+  const projectNumbers = _.filter(cellInfo.value, (p: string) => !p.includes('DRAFT'));
```
agalbrai, line 45: don't include draft projects

Fosol: Is there a reason not to include the draft projects?

agalbrai: The story says "only show for approved surplus projects" which I assumed meant no drafts?

Fosol: hmmmm, I doubt that is correct, but Mikaela did put that in as a requirement.  We can get a new story if she wants it different.

Fosol (approved):

jdtoombs (approved):

devinleighsmith:
The screenshot just shows the one project number, does this look ok when there is multiple?

devinleighsmith (approved):

agalbrai:
> The screenshot just shows the one project number, does this look ok when there is multiple?

image
seems to work just fine :)

PR #892, opened 2021-02-06
PA-2674 and PA-2676 Add Many to Many for subdivisions. Add new subdiv…

devinleighsmith:
…ision property type. Update parcel service to handle subdivisions.


backend/api/Areas/Property/Mapping/Parcel/ParcelParcelMap.cs
```diff
@@ -0,0 +1,86 @@
...
+        public ParcelParcelMap(IOptions<JsonSerializerOptions> serializerOptions)
+        {
+            _serializerOptions = serializerOptions.Value;
+        }
+        #endregion
+
+        public void Register(TypeAdapterConfig config)
+        {
+            //all mappings for mapping between a subdivision parcel that has divided parcels. Users cannot edit parcels/subdivisions through entity relationships so just pass ids.
+            config.NewConfig<Entity.Parcel, Model.SubdivisionParcelModel>()
```
devinleighsmith, line 34: Since we have no intention of allowing users to modify parcels using the Parcels or Subdivisions association, there is no point in passing/mapping all of those fields since they won't be used.

devinleighsmith: (So just use the ones that are going to be used, so pid/pin and id).

backend/api/Areas/Property/Models/Parcel/ParcelModel.cs
```diff
@@ -22,6 +22,16 @@ public class ParcelModel : PropertyModel
         public IEnumerable<ParcelFiscalModel> Fiscals { get; set; } = new List<ParcelFiscalModel>();
 
         public IEnumerable<ParcelBuildingModel> Buildings { get; set; } = new List<ParcelBuildingModel>();
+
+        /**
+         * All of the parcels related to this subdivision
+         */
+        public IEnumerable<SubdivisionParcelModel> Parcels { get; set; } = new List<SubdivisionParcelModel>();
```
devinleighsmith, line 29: in other words, a subdivision parcel will only populate the Parcels array, and a Parcel with subdivisions will only populate the Subdivisions array.

backend/dal/Configuration/ParcelParcelsConfiguration.cs
```diff
@@ -0,0 +1,35 @@
...
+            builder.Property(m => m.ParcelId).IsRequired();
+            builder.Property(m => m.ParcelId).ValueGeneratedNever();
+
+            builder.Property(m => m.SubdivisionId).IsRequired();
+            builder.Property(m => m.SubdivisionId).ValueGeneratedNever();
+
+
+            //I had some trouble getting this to work as A->A Many to Many relationships aren't documented well on the fluent efcore site. Based on:
+            // https://stackoverflow.com/questions/49214748/many-to-many-self-referencing-relationship/49219124#49219124
+            builder.HasOne(m => m.Parcel).WithMany(m => m.Subdivisions).HasForeignKey(m => m.ParcelId).OnDelete(DeleteBehavior.Restrict);
```
devinleighsmith, line 28: the combination of parcel->subdivisions->parcelid is critical so that ONLY the parcels or subdivisions array is populated at the expected time. Changing this leads to a swap where the subdivisions are populated instead of the parcels, or sometimes both subdivisions and parcels were populated when just inserting elements into one of the arrays.

backend/dal/Migrations/v01.07.01/Up/PreUp/01-ViewProperties.sql
```diff
@@ -0,0 +1,213 @@
+PRINT 'Updating View_Properties - Updating Assessment'
+GO
+-- This script provides a way to union both parcels and buildings into a single result of properties.
+-- Updating view to support many-to-many parcels and buildings.TEMPORARY
+CREATE OR ALTER VIEW dbo.[View_Properties] AS
+SELECT
+    p.[Id]
+    , p.[RowVersion]
+    , [PropertyTypeId] =  (SELECT CASE WHEN sp.[ParcelId] IS NOT NULL THEN 2 ELSE 0 END)
```
devinleighsmith, line 9: this injects the new property type of 2

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -285,7 +285,6 @@ public Building Update(Building building)
                     {
                         this.ThrowIfNotAllowedToUpdate(parcelBuilding.Building, _options.Project);
 
-                        var parcelBuildings = this.Context.ParcelBuildings.Where(pb => pb.ParcelId == parcelBuilding.ParcelId).ToArray();
```
devinleighsmith, line 288: looks like dead code.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -176,11 +178,13 @@ public Parcel Get(int id)
         public Parcel Add(Parcel parcel)
         {
             parcel.ThrowIfNull(nameof(parcel));
-            this.User.ThrowIfNotAuthorized(Permissions.PropertyAdd);
```
devinleighsmith, line 179: not allowing a user with admin but no add?

backend/tests/unit/dal/Services/ParcelServiceTest.cs
```diff
@@ -649,6 +649,209 @@ public void Get_Sensitive_AsAdmin()
         }
         #endregion
 
+        #region Add Parcel
```
devinleighsmith, line 652: Missing Add* tests, so added them.

backend/api/Areas/Property/Models/Parcel/ParcelModel.cs
Fosol: Is this more intuitive than having the property type provide the type?

backend/dal/Migrations/v01.07.02/Up/PostUp/01-PropertyTypes.sql
```diff
@@ -0,0 +1,14 @@
+PRINT 'Updating PropertyTypes'
+
+-- Ensure the insert succeeds.
+DELETE FROM dbo.[PropertyTypes] where Id = 2;
```
Fosol, line 4: In our case this works fine, but not ideal. It theoretically could remove an existing property type in a database.  Generally it's best practice to only do the insert here and on down do the delete.  I wouldn't change it though, as currently we're 100% in control of the production data.

backend/dal/Migrations/v01.07.02/Up/PostUp/01-ViewProperties.sql
```diff
@@ -0,0 +1,213 @@
+PRINT 'Updating View_Properties - Updating Assessment'
+GO
+-- This script provides a way to union both parcels and buildings into a single result of properties.
+-- Updating view to support many-to-many parcels and buildings.TEMPORARY
+CREATE OR ALTER VIEW dbo.[View_Properties] AS
+SELECT
+    p.[Id]
+    , p.[RowVersion]
+    , [PropertyTypeId] =  (SELECT CASE WHEN sp.[ParcelId] IS NOT NULL THEN 2 ELSE 0 END)
```
Fosol, line 9: I thought we were going to add `Property.PropertyType`.  Which would allow for this value to be extracted from that property.  I don't think I clearly articulated this in any story however :(.  Might need to be thought through and considered though.

backend/dal/Migrations/20210206053021_v01.07.02.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using Pims.Dal.Helpers.Migrations;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010702 : SeedMigration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
```
Fosol, line 9: I thought we were going to add `Property.PropertyType`, which would result in both the `Parcel.PropertyType` and `Building.PropertyType`.  I suppose in some ways it's an imperfect relationship, as some property types don't apply to either building or parcel.

backend/dal/Migrations/v01.07.02/Up/PostUp/01-ViewProperties.sql
```diff
@@ -0,0 +1,213 @@
...
+    , [AssessedLand] = eas.[Value]
+    , [AssessedLandDate] = eas.[Date]
+    , [AssessedBuilding] = eim.[Value]
+    , [AssessedBuildingDate] = eim.[Date]
+    , [Market] = ISNULL(fe.[Value], 0)
+    , [MarketFiscalYear] = fe.[FiscalYear]
+    , [NetBook] = ISNULL(fn.[Value], 0)
+    , [NetBookFiscalYear] = fn.[FiscalYear]
+FROM dbo.[Parcels] p
+LEFT JOIN dbo.[ParcelParcels] sp ON p.[Id] = sp.[ParcelId]
```
Fosol, line 70: I think this will actually have unintentional effect of creating multiple copies of the same parcel and will result in not correctly apply the correct `PropertyType` to the parent parcel.

backend/dal/Services/Concrete/BuildingService.cs
Fosol: Just make sure it wasn't being used to lazy-load data for some model

Fosol: Bad way to do it if it was though.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -176,11 +178,13 @@ public Parcel Get(int id)
...
+            parcel.ThrowIfNotAllowedToEdit(nameof(parcel), this.User, new[] { Permissions.PropertyAdd, Permissions.AdminProperties });
 
             var agency = this.User.GetAgency(this.Context) ??
                 throw new NotAuthorizedException("User must belong to an agency before adding parcels.");
 
+            if (parcel.Parcels.Count() > 0 && parcel.Subdivisions.Count() > 0) throw new InvalidOperationException("Parcel may only have assocatiated parcels or subdivisions, not both.");
```
Fosol, line 186: spelling `assocatiated`

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -192,6 +196,17 @@ public Parcel Add(Parcel parcel)
             parcel.Classification = this.Context.PropertyClassifications.Find(parcel.ClassificationId);
             parcel.IsVisibleToOtherAgencies = false;
 
+            parcel.Parcels.ForEach(pp =>
```
Fosol, line 199: My brain finds the two `ForEach(...)` statements hard to read.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -287,6 +308,8 @@ public Parcel PendingUpdate(Parcel parcel)
             // Only administrators can dispose a property.
             if (!isAdmin && parcel.ClassificationId == 4) throw new NotAuthorizedException("Parcel classification cannot be changed to disposed."); // TODO: Classification '4' should be a config settings.
 
+            if (parcel.Parcels.Count() > 0 && parcel.Subdivisions.Count() > 0) throw new InvalidOperationException("Parcel may only have assocatiated parcels or subdivisions, not both.");
```
Fosol, line 311: spelling `assocatiated`

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -278,6 +297,8 @@ public Parcel PendingUpdate(Parcel parcel)
             var ownsABuilding = originalParcel.Buildings.Any(pb => userAgencies.Contains(pb.Building.AgencyId.Value));
             if (!allowEdit && !ownsABuilding) throw new NotAuthorizedException("User may not edit parcels outside of their agency.");
 
+            if (!isAdmin && originalParcel?.ProjectNumbers?.Contains("SPP") == true) throw new NotAuthorizedException("User may not edit parcels that are in a SPP Project.");
```
Fosol, line 300: Don't change, but I suspect this will cause issues down the road.  The `SPP` value is a configurable setting for projects.  In the future there could be all different types of project numbers.

backend/api/Areas/Property/Models/Parcel/ParcelModel.cs
devinleighsmith: I hadn't really considered adding the property type here since that would allow for parcels to be given building designations and buildings to be given parcel designations. It would work, although it would continue our not great trend of enforcing our data model in logic and not in the database. If someone forgets to add/update the property type on these objects it would lead to invalid data, same with forgetting to update the project number array.

backend/dal/Migrations/v01.07.02/Up/PostUp/01-ViewProperties.sql
devinleighsmith: This should be LEFT JOIN dbo.[ParcelParcels] sp ON p.[Id] = sp.[SubdivisionId].

Parcels that have their id being the same as a subdivisionId are subdivisions.

I suppose the duplication would happen if there were multiple parcelParcel rows with the same subdivisionId. Could get around this by joining to a select distinct subdivisionId from parcelParcel instead of just joining to the parcelParcel table.

backend/api/Areas/Property/Models/Parcel/SubdivisionParcelModel.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+using Model = Pims.Api.Models;
+
+namespace Pims.Api.Areas.Property.Models.Parcel
+{
+    public class SubdivisionParcelModel : Model.BaseModel
```
goemen, line 7: How is this different from `ParcelSubdivisionModel `?

backend/api/Areas/Property/Models/Parcel/ParcelModel.cs
Fosol: You theoretically can still enforce this behaviour in the DB, but I won't recommend triggers.  It's easier to manage in the DAL.  The more explicit implementation is different types `ParcelType` and `BuildingType`

backend/api/Areas/Property/Mapping/Parcel/ParcelParcelMap.cs
```diff
@@ -0,0 +1,86 @@
...
+                .EnableNonPublicMembers(true)
+                .Map(dest => dest.Id, src => src.Id)
+                .Map(dest => dest.PID, src => src.PID)
+                .Map(dest => dest.PIN, src => src.PIN)
+                .Inherits<Entity.BaseEntity, BModel.BaseModel>();
+
+            config.NewConfig<Model.SubdivisionParcelModel, Entity.Parcel>()
+                .EnableNonPublicMembers(true)
```
agalbrai, line 42: what is this for?

backend/dal/Migrations/v01.07.02/Up/PostUp/01-ViewProperties.sql
Fosol: I think you can avoid this by not using this relationship to decide on the `ProperytType`.  Just include the `PropertyType` to the table, or add a more complex relationship `ParcelType` and `BuildingType`

backend/dal/Services/Concrete/ParcelService.cs
devinleighsmith: I'll add some comments, and split this logic up with an if statement since there should never be a time that both for loops run for the same parcel.

backend/api/Areas/Property/Models/Parcel/SubdivisionParcelModel.cs
devinleighsmith: I could reuse the same model, but I'm trying to convey the fact that the service returns a divided parcel with a list of subdivisions or a subdivision parcel with a list of divided parent parcels.

For now there is no difference between a subdivision parcel model and a parent divided parcel, but that may not always be the case.

backend/api/Areas/Property/Mapping/Parcel/ParcelParcelMap.cs
devinleighsmith: This allows mapping of non public fields

backend/entities/ParcelParcel.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+
+namespace Pims.Dal.Entities
+{
+    /// <summary>
+    /// ParcelParcels class, provides the many-to-many relationship between parcels and parcels (generally for subdivisions).
+    /// </summary>
+    public class ParcelParcel : BaseEntity
```
agalbrai, line 8: Every time I read "parcelparcel" my brain imagines some strange pokemon lol

backend/api/Areas/Property/Models/Parcel/ParcelModel.cs
devinleighsmith: Although this method requires that you correctly manage the type by inserting your data into the correct array, so it isn't perfect either.

agalbrai (commented):
my SQL is not too strong so I don't have much of value to add, but nice work!

goemen (approved):
Just one minor comment

backend/dal/Migrations/v01.07.02/Up/PostUp/01-ViewProperties.sql
devinleighsmith: I'll add in the property type, but I think I'll keep the two arrays. Only using one array works, but ends up being similarly confusing and isn't worth the time investment to rewrite everything in my opinion.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -385,7 +467,7 @@ public Parcel PendingUpdate(Parcel parcel)
                 }
             }
 
-            return Get(parcel.Id);
+            return originalParcel;
```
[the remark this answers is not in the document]
Fosol: I prefer what you're doing here, less round trips to the DB if they are not required.

backend/dal/Migrations/20210209065005_v01.07.03.cs
```diff
@@ -0,0 +1,98 @@
...
+                name: "propertyTypeId",
+                table: "Parcels",
+                nullable: false,
+                defaultValue: 0);
+
+            migrationBuilder.AddColumn<int>(
+                name: "propertyTypeId",
+                table: "Buildings",
+                nullable: false,
+                defaultValue: 0);
```
Fosol, line 22: Wonder if the default value should be 1 here.  Would that reduce the need for your additional script to set it to 1?

backend/dal/Migrations/20210209065005_v01.07.03.cs
```diff
@@ -0,0 +1,98 @@
...
+    public partial class v010703 : SeedMigration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            PreUp(migrationBuilder);
+            migrationBuilder.AddColumn<int>(
+                name: "propertyTypeId",
+                table: "Parcels",
+                nullable: false,
+                defaultValue: 0);
```
Fosol, line 16: Since it defaults to 0, do you need the additional script to set it to 0?

backend/entities/Property.cs
```diff
@@ -15,6 +15,11 @@ public abstract class Property : BaseEntity
         /// </summary>
         public int Id { get; set; }
 
+        /// <summary>
+        /// get/set - The type of the property. Type is managed by business logic.
+        /// </summary>
+        public PropertyTypes propertyTypeId { get; set; }
```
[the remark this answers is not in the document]
Fosol: Shouldn't this be a foreign key to `PropertyTypes` now?

Fosol (changes requested):

backend/dal/Services/Concrete/ParcelService.cs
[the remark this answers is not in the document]
devinleighsmith: This actually was technically incorrect before as this PendingUpdate method doesn't update the DB, so calling Get() would give unintended results. The Get method should only be called after the pending updates are committed.

backend/entities/Property.cs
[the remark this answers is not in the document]
devinleighsmith: Oh, I actually forgot we had a table for that, will fix.

backend/dal/Migrations/20210206053021_v01.07.02.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using Pims.Dal.Helpers.Migrations;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Pims.Dal.Migrations
+{
+    public partial class v010702 : SeedMigration
```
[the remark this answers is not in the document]
devinleighsmith: I think you made this comment on an outdated file, I upgraded to 010703 last week.

Fosol (approved):
Awesome

PR #893, opened 2021-02-06
PA-2742 Transferring properties

Fosol:
Currently most school district properties are assigned to the Ministry of Education.  They however need to be transferred to their owning sub-agency (School District ##).

Added DB migration `v01.07.02`.

> I think there may be a few PRs currently with this version number, so whomever gets to merge first will get to keep the number.


openshift/4.0/templates/api/build.yaml
```diff
@@ -20,6 +20,16 @@ parameters:
     description: The name of the application component role (e.g ui, web, api, database).
     required: true
     value: api
+  - name: PROJECT_NAMESPACE
```
Fosol, line 23: Some cleanup and applying some consistent standards across templates.  Not related to this PR, but doesn't particular hurt to include it here either.

backend/dal/Migrations/v01.07.02/Up/PreUp/01-Buildings.sql
```diff
@@ -0,0 +1,423 @@
+PRINT 'Transfer Capital Planning Branch Buildings to School Districts'
```
Fosol, line 1: Identified all `AdministrativeAreas` within our DB and migrated them to an appropriate school district.  This took forever...

devinleighsmith (approved):

goemen (approved):
lgtm

PR #896, opened 2021-02-08
PA-1910 Allow SPL to Transfer within GRE

Fosol:
Allow SRES to **Transfer within GRE** when a project is in the **SPL** workflow.

The SPL tab will now have a way to transfer within GRE.

image

Then the user can select which agency that the properties will be transferred to.

image


frontend/src/features/projects/spl/tabs/SplTabs.tsx
```diff
@@ -76,7 +79,11 @@ export const SplTabs: React.FunctionComponent<ISplTabsProps> = ({
           title="Surplus Properties List"
           tabClassName={isTabInError(errors, SPPApprovalTabs.spl)}
         >
-          <SplTab isReadOnly={isReadOnly} setSubmitStatusCode={setSubmitStatusCode} />
+          <SplTab
+            isReadOnly={isReadOnly}
+            setSubmitStatusCode={setSubmitStatusCode}
+            goToGreTransferred={() => submitForm().then(() => goToGreTransferred())}
```
jdtoombs, line 85: Oh duh! The other day I was trying to use `handleSubmit` as a promise and forgot about `submitForm` 🤦

jdtoombs (approved):

frontend/src/features/projects/spl/steps/GreTransferStep.tsx
```diff
@@ -0,0 +1,151 @@
...
+const CenterBoldText = styled.div`
+  text-align: center;
+  font-family: 'BCSans-Bold';
+  margin-bottom: 1.5rem;
+`;
+
+const FlexRight = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: row-reverse;
```
agalbrai, line 41: why row-reverse?

frontend/src/features/projects/spl/tabs/SplTab.tsx
```diff
@@ -5,6 +5,7 @@ import { ReviewWorkflowStatus } from 'features/projects/common';
 interface ISplTabProps {
   isReadOnly?: boolean;
   setSubmitStatusCode: Function;
+  goToGreTransferred: Function;
```
agalbrai, line 8: is this different than declaring a prop as () => void ?

jdtoombs: Same thing - just a different way I believe

frontend/src/features/projects/spl/forms/GreTransferForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+        if (agency !== undefined) {
+          setFieldValue(`properties.${index}.agencyCode`, agency.code);
+          setFieldValue(`properties.${index}.subAgency`, agency.name);
+        }
+      }
+    });
+  }, [setFieldValue, values.properties, values.agencyId, agencies, touched, initialAgencyId]);
+  return (
+    <Fragment>
+      <ProjectDraftForm isReadOnly={false} />
```
devinleighsmith, line 48: I did a diff on this form and the original GreTransferForm and the only difference is that this line is false instead of true. Intentional?

I'd rather you pass that in as a prop, and then move the GreTransferForm to /projects/common/forms rather then 1-1 duplicating this form.

I get the desire to avoid overly complicate components with lots of conditionals to handle different uses, but if you only need to differ by a single boolean value its not really reasonable to duplicate this entire file.

frontend/src/features/projects/spl/steps/GreTransferStep.tsx
```diff
@@ -0,0 +1,151 @@
+import React, { useState } from 'react';
```
devinleighsmith, line 1: Same comment here.

frontend/src/features/projects/spl/tabs/SplTabs.tsx
devinleighsmith: we may need to enhance the error handling for this, but probably fine for now.

frontend/src/features/projects/spl/steps/GreTransferStep.tsx
Fosol: I've copied this from the nearly identical implementation for **ERP** to **Transfer within GRE**.  @devinleighsmith may be able to answer this question.

devinleighsmith: basically this provides a flex that grows right -> left instead of left -> right.

frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
```diff
@@ -124,14 +123,15 @@ describe('GRE Transfer Step', () => {
...
     });
-    it('form fields are disabled', () => {
-      const component = render(getGreTransferStep());
-      const textboxes = component.queryAllByRole('textbox');
-      textboxes.forEach(textbox => {
-        expect(textbox).toBeVisible();
-        expect(textbox).toBeDisabled();
-      });
-    });
+    // TODO: Unable to get this to work.
```
Fosol, line 126: After an hour of attempts to get a simple test to work...  I won't waste any more time on this.  If you know the answer, feel free to let me know.  Otherwise the value of this test is so low that it isn't a priority.

frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
```diff
@@ -146,14 +146,16 @@ describe('GRE Transfer Step', () => {
...
     });
-    it('form fields are disabled', () => {
-      const component = render(getGreTransferStep(getStore(project)));
-      const textboxes = component.queryAllByRole('textbox');
-      textboxes.forEach(textbox => {
-        expect(textbox).toBeVisible();
-        expect(textbox).toBeDisabled();
-      });
-    });
+    // TODO: Unable to get this to work.
```
Fosol, line 149: After an hour of attempts to get a simple test to work...  I won't waste any more time on this.  If you know the answer, feel free to let me know.  Otherwise the value of this test is so low that it isn't a priority.

frontend/src/features/projects/spl/steps/GreTransferStep.tsx
Fosol: Component has been placed in common shared components

frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
devinleighsmith: If this test isn't working, can you validate that the logic that is being tested here does work if tested manually? so the project name field is enabled, the description is disabled, and the agency is enabled?

devinleighsmith (approved):

frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
Fosol: I've tested manually for both ERP and SPL.  It works as expected.  But the unit-test library we're using doesn't appear to be intuitive enough for my limited time with it to be able to both find the element and validate it.

PR #897, opened 2021-02-08
PA-2700, 2341: Demolished/ Subdivided

jdtoombs:
Add demolished and subdivided classifications.

- only buildings can select classification of demolished 
- only SRES can select demolished and subdivided 
- demolished and subdivided not displayed on map or list view
- only SRES can filter Demolished/Subdivided

### DIFF
--- a/backend/dal/Helpers/Constants/ClassificationTypes.cs
+++ b/backend/dal/Helpers/Constants/ClassificationTypes.cs
@@ -1,0 +1,27 @@
+namespace Pims.Dal.Helpers.Constants
+{
+    /// <summary>
+    /// ClassificationTypes static class, provides constant values for classifications.
+    /// </summary>
+    public static class ClassificationTypes
+    {
+        #region Classifications
+        public enum Classifications {
+            /** The property is currently being used. */
+            CoreOperational = 0,
+            /** The property is not currently being used but has a strategic purpose. */
+            CoreStrategic = 1,
+            /** The property is surplus. */
+            SurplusActive = 2,
+            /** The property is surplus with an encumberance. */
+            SurplusEncumbered = 3,
+            /** The property has been disposed. */
+            Disposed = 4,
+            /** The property has been demolished */
+            Demolished = 5,
+            /** The property has been subdivided */
+            Subdivided = 6,
+        }
+        #endregion
+    }
+}
--- a/backend/dal/Helpers/Extensions/PropertyExtensions.cs
+++ b/backend/dal/Helpers/Extensions/PropertyExtensions.cs
@@ -36,7 +36,7 @@ namespace Pims.Dal.Helpers.Extensions
             // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
             var query = context.Properties
                 .AsNoTracking()
-                .Where(p => p.ClassificationId != 4); // Disposed properties are not visible.
+                .Where(p => p.ClassificationId != 4 && p.ClassificationId != 5 && p.ClassificationId != 6); // Disposed and subdivided/demolished properties are not visible.
 
             // Only allowed to see user's own agency properties.
             if (!isAdmin)
@@ -173,7 +173,7 @@ namespace Pims.Dal.Helpers.Extensions
             // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
             var query = context.Properties
                 .AsNoTracking()
-                .Where(p => p.ClassificationId != 4); // Disposed properties are not visible.
+                .Where(p => p.ClassificationId != 4 && p.ClassificationId != 5 && p.ClassificationId != 6); // Disposed/subdivided/demolished properties are not visible.
 
             // Users are not allowed to view sensitive properties outside of their agency or sub-agencies.
             if (!isAdmin)
--- a/backend/dal/Helpers/Extensions/PropertyExtensions.cs
+++ b/backend/dal/Helpers/Extensions/PropertyExtensions.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
 using Entity = Pims.Dal.Entities;
+using Pims.Dal.Helpers.Constants;
 
 namespace Pims.Dal.Helpers.Extensions
 {
@@ -36,7 +37,7 @@ namespace Pims.Dal.Helpers.Extensions
             // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
             var query = context.Properties
                 .AsNoTracking()
-                .Where(p => p.ClassificationId != 4 && p.ClassificationId != 5 && p.ClassificationId != 6); // Disposed and subdivided/demolished properties are not visible.
+                .Where(p => p.ClassificationId != (int)ClassificationTypes.Classifications.Disposed && p.ClassificationId != (int)ClassificationTypes.Classifications.Demolished && p.ClassificationId != (int)ClassificationTypes.Classifications.Subdivided); // Disposed and subdivided/demolished properties are not visible.
 
             // Only allowed to see user's own agency properties.
             if (!isAdmin)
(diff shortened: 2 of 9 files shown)


backend/dal/Helpers/Constants/ClassificationTypes.cs
```diff
@@ -0,0 +1,21 @@
+namespace Pims.Dal.Helpers.Constants
```
jdtoombs, line 1: Not sure if this was necessary or not...

backend/dal/Helpers/Constants/ClassificationTypes.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Classification Types static class, provides constant values for Classifications.
+    /// </summary>
+    public static class ClassificationTypes
```
Fosol, line 6: While having the file isn't bad, we didn't have it because classifications technically come from the database and their `Id` could be anything.  Regrettably due to time we haven't always applied business rules with this in-mind.  Additionally you shouldn't use capitalization in C# in an `enum`.  The normal coding standard for c# is Pascal Case (refer to other enum examples).

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -36,7 +37,7 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
             // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
             var query = context.Properties
                 .AsNoTracking()
-                .Where(p => p.ClassificationId != 4); // Disposed properties are not visible.
+                .Where(p => p.ClassificationId != (int)ClassificationTypes.Classifications.DISPOSED && p.ClassificationId != (int)ClassificationTypes.Classifications.SUBDIVIDED && p.ClassificationId != (int)ClassificationTypes.Classifications.DEMOLISHED); // Disposed and subdivided/demolished properties are not visible.
```
Fosol, line 40: You don't have to change, but `!=` are actually poor performing queries in SQL.  It is always better to be explicit `=` or `in`.  In code however it is easier to read and write, especially if the condition is for a single value.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
```diff
@@ -86,17 +90,37 @@ export const ClassificationForm: React.FC<IClassificationFormProps> = ({
...
+    /** non SRES users cannot see subdivided */
+    if (!keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) {
+      return (
+        val !== Classifications.Subdivided &&
+        val !== Classifications.Disposed &&
+        val !== Classifications.Demolished
+      );
+    } else {
+      /** only buildings can be demolished */
+      if (!fieldLabel?.includes('Building')) {
```
Fosol, line 112: This logic should also exist in the DB to ensure the user doesn't submit invalid JSON.  Additionally you should ensure only a SRES user `admin-properties` can selected these classifications in the backend.  Remember to never trust the frontend.

backend/dal/Helpers/Constants/ClassificationTypes.cs
jdtoombs: Ooops totally knew that too, just in TS mode lol. Thanks!

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
jdtoombs: Gotchya, makes sense

backend/dal/Helpers/Constants/ClassificationTypes.cs
jdtoombs: Going to remove based off of Jeremy's comments

backend/dal/Helpers/Extensions/PropertyExtensions.cs
devinleighsmith: Did we have requirements around admin users being able to view subdivided parcels? actually, I think that was just in the generated reports.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
```diff
@@ -86,17 +90,37 @@ export const ClassificationForm: React.FC<IClassificationFormProps> = ({
...
+        return val !== Classifications.Disposed;
+      }
+    }
+  };
+
   let filteredClassifications = classifications;
   /** users not allowed to select disposed, but display these values if one of these classifications has already been selected. */
   filteredClassifications = classifications.filter(
-    c => Number(c.value) !== Classifications.Disposed || +c.value === +classId,
+    c => determineFilter(+c.value) || +c.value === +classId,
```
devinleighsmith, line 123: Not sure if this is an issue or not, but be careful with unary + as it will convert some values to numbers unexpectedly.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

For example if classId is '' (formik default) +classId will be converted to 0, which is equivalent to Core Operational.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
```diff
@@ -86,17 +90,37 @@ export const ClassificationForm: React.FC<IClassificationFormProps> = ({
...
     getIn(formikProps.values, field) === Classifications.SurplusEncumbered ||
     getIn(formikProps.values, field) === Classifications.SurplusActive;
 
   /** classId based on current formik values to determine which classsification information box to display */
   let classId = getIn(formikProps.values, field);
 
+  /** determine how to filter the classifications */
+  const determineFilter = (val: Number) => {
+    /** non SRES users cannot see subdivided */
+    if (!keycloak.hasClaim(Claims.ADMIN_PROPERTIES)) {
```
devinleighsmith, line 104: just a small style comment, if possible if true else false is generally easier for people to read. Not a hard and fast rule of course.

backend/dal/Helpers/Constants/ClassificationTypes.cs
devinleighsmith: Ideally we would use code values instead of ids for everything, but yeah, constants are quick and dirty.

devinleighsmith (approved):

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
jdtoombs: Ah makes sense, will adjust to keep with best practices!

agalbrai (approved):
looks good!

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static IQueryable<Entity.Views.Property> GenerateQuery(this PimsContext c
             // Users may only view sensitive properties if they have the `sensitive-view` claim and belong to the owning agency.
             var query = context.Properties
                 .AsNoTracking()
-                .Where(p => p.ClassificationId != 4); // Disposed properties are not visible.
+                .Where(p => p.ClassificationId != 4 && p.ClassificationId != 5 && p.ClassificationId != 6); // Disposed and subdivided/demolished properties are not visible.
```
devinleighsmith, line 39: Sorry, I think something got lost in translation here. If we are going to be using ids in a conditional like this, I'd rather we use an enum like what you were doing earlier to prevent magic numbers and make the code more readable. It isn't the best way to handle codified values (thus my quick and dirty comment from before), but it is at least readable.

The other option is to use an enum corresponding to the classification code values. For that you'd need to modify the view to return the code, currently it returns the classificationId and the classification name.

jdtoombs: Ahhh oops haha I did misread the comments, adding back now.

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -234,6 +234,12 @@ public Building Update(Building building)
             // Only administrators can dispose a property.
             if (building.ClassificationId == 4 && !isAdmin) throw new NotAuthorizedException("Building classification cannot be changed to disposed.");
 
+            // Only administrators can set a building to demolished
+            if(building.ClassificationId == 5 && isAdmin) throw new NotAuthorizedException("Building classification cannot be changed to demolished.");
```
devinleighsmith, line 238: Thanks for adding back the enum, but could you apply it here as well? sorry to be a pain.

jdtoombs: No problem at all, forgot about this area. Thanks for catching it!

PR #901, opened 2021-02-09
PA-1898 Updated Project Import

Fosol:
Updating Converter tool to include `ExemptionRequested` and **Exemption from ERP** status.

### DIFF
--- a/tools/converters/excel/Converters/StringConverters.cs
+++ b/tools/converters/excel/Converters/StringConverters.cs
@@ -31,7 +31,7 @@ namespace Pims.Tools.Converters.ExcelConverter.Converters
         }
 
         /// <summary>
-        ///
+        /// Convert the 'value' to a date value, or null if allowed.
         /// </summary>
         /// <param name="value"></param>
         /// <param name="nullable"></param>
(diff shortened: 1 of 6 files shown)


tools/converters/excel/Converters/StringConverters.cs
```diff
@@ -68,6 +68,25 @@ public static class StringConverters
             throw new InvalidOperationException($"Unable to convert value '{value}' to date.");
         }
 
+        /// <summary>
+        /// Convert 'value' to boolean, or null if allowed.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <param name="nullable"></param>
+        /// <returns></returns>
+        public static Boolean? ConvertToBoolean(this string value, bool nullable = false)
```
devinleighsmith, line 77: it might be nice to have support for 'true' and 'false' as well, but not required ofc.

devinleighsmith: oh wait, this is just for the tools, so nvm.

devinleighsmith (approved):

tools/converters/excel/Converters/StringConverters.cs
Fosol: It does handle `true` and `false` too

jdtoombs (approved):

devinleighsmith (approved):

PR #903, opened 2021-02-09
PA-2544 "Last Updated By" on detail view

agalbrai:
image


jdtoombs (approved):

frontend/src/features/properties/components/LastUpdatedBy.tsx
```diff
@@ -14,35 +14,43 @@ const UpdateText = styled.p`
...
-  color: #069;
+  color: #18548c;
   text-decoration: underline;
 `;
 
 /**
  * Display formatted last update user name, email, and date/time.
  */
-const LastUpdatedBy: React.FunctionComponent<ILastUpdatedByProps> = ({
+const LastUpdatedBy: React.FC<ILastUpdatedByProps> = ({
```
jdtoombs, line 28: 😎

agalbrai: I think that was an accident lol, I couldn't figure out what an error meant so I was just trying stuff

Fosol (approved):

frontend/src/features/properties/components/LastUpdatedBy.tsx
```diff
@@ -14,35 +14,43 @@ const UpdateText = styled.p`
...
-      </TooltipWrapper>
-      &nbsp;
-      {formatApiDateTime(updatedOn ?? createdOn)}
-    </UpdateText>
-  ) : null;
+  if (updatedOn) {
+    return (
+      <UpdateText>
+        Last Updated By:&nbsp;
+        <TooltipWrapper toolTipId="lastUpdatedEmail" toolTip={updatedByEmail}>
```
devinleighsmith, line 38: If I remember correctly, part of the reason for the 'unknown' is that we don't have a firm requirement on users in the system having an email associated, or even a name. Those are both optional fields.

agalbrai: Ah, I gotcha. I'll put that back

PR #904, opened 2021-02-09
PA-2679: Fix backend erroneously rounding values

jdtoombs:
`float` has a precision of 6-9 digits while `decimal` has 28-29 which will handle the large financial values correctly. Could always go with a `double` as well if that is preferred (15-17)


agalbrai (approved):
I'm guessing storage space isn't a huge deal here?

devinleighsmith (approved):

Fosol (approved):
Always nice when the fix is fairly straight forward.

PR #906, opened 2021-02-10
Subdivision Frontend updates

devinleighsmith:
PA-2690 display of subdivisions in slide out.
PA-2699 display of subdivisions in inventory list view.
PA-2696 restrict display of subdivisions
PA-2689 display of subdivisions on the map.


backend/api/Areas/Property/Mapping/Search/GeoJsonMap.cs
```diff
@@ -42,6 +42,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.ProjectNumbers, src => JsonSerializer.Deserialize<IEnumerable<string>>(src.ProjectNumbers ?? "[]", _serializerOptions))
                 .Map(dest => dest.AddressId, src => src.AddressId)
                 .Map(dest => dest.Address, src => src.Address)
+                .Map(dest => dest.ProjectWorkflow, src => src.ProjectWorkflow)
```
devinleighsmith, line 45: This is only tangentially related, but I noticed that the erp/spl search results weren't triggering the correct PIN colour. Returning the workflow code seems like the easiest way to resolve this.

backend/dal/Services/Concrete/PropertyService.cs
```diff
@@ -141,7 +142,20 @@ public IEnumerable<PropertyModel> Search(AllPropertyFilter filter)
             }
 
             var query = this.Context.GenerateAllPropertyQuery(this.User, filter);
-            var properties = query.Select(p => p.PropertyTypeId == Entities.PropertyTypes.Land ? new ParcelModel(p, this.User) as PropertyModel : new BuildingModel(p, this.User)).ToArray();
+            var properties = query.Select(p => new[] { Entities.PropertyTypes.Land, Entities.PropertyTypes.Subdivision }.Contains(p.PropertyTypeId) ? new ParcelModel(p, this.User) as PropertyModel : new BuildingModel(p, this.User)).ToArray();
```
devinleighsmith, line 145: This code is reused from another area of the application to get the workflow code. Note that this code doesn't address the case where there are multiple projects. We need some logic that can determine the most advanced project workflow and return that to the frontend.

backend/entities/Models/ParcelModel.cs
```diff
@@ -85,7 +85,7 @@ public class ParcelModel : PropertyModel
         /// <param name="user"></param>
         public ParcelModel(Views.Property property, ClaimsPrincipal user) : base(property, user)
         {
-            this.PropertyTypeId = PropertyTypes.Land;
+            this.PropertyTypeId = property.PropertyTypeId;
```
devinleighsmith, line 88: these are sprinkled throughout the application which has been fun.

frontend/src/components/Table/PropertyTypeCell.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ReactComponent as SubdivisionSvg } from 'assets/images/project-diagram-solid.svg';
+
+import { IProperty } from 'actions/parcelsActions';
+import { PropertyTypes } from 'actions/parcelsActions';
+
+/**
+ * Display an icon based on the property type.
+ * @param {CellProps<IProperty, number>} param0
+ */
+export const PropertyTypeCell = ({ cell: { value } }: CellProps<IProperty, number>) => {
```
devinleighsmith, line 14: made this into a reusable component since it is used in tables all over the place.

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -105,22 +155,32 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
  * Get an icon type for the specified cluster property details (type, draft, erp, spp etc)
  */
 export const getMarkerIcon = (feature: ICluster) => {
-  const { propertyTypeId, projectStatus, classificationId } = feature?.properties;
-  if (
-    [
```
devinleighsmith, line 110: this list of statuses is incomplete and is at constant risk of being out of sync with the db. We really want to know what workflow the property is in anyway not what status.

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -140,31 +140,34 @@ export const toFlatProject = (project?: IApiProject) => {
...
   evaluations.push({
-    parcelId: property.propertyTypeId === 0 ? property.id : undefined,
-    buildingId: property.propertyTypeId === 1 ? property.id : undefined,
-    value:
-      property.propertyTypeId === 0 ? property.assessedLand || 0 : property.assessedBuilding || 0,
-    date:
-      property.propertyTypeId === 0
-        ? property.assessedLandDate ?? formatDate(new Date())
-        : property.assessedBuildingDate ?? formatDate(new Date()),
+    parcelId: isParcelOrSubdivision(property) ? property.id : undefined,
```
devinleighsmith, line 150: this file made a lot of assumptions about only two types of properties, so updating all of that here.

frontend/src/components/Table/PropertyTypeCell.tsx
Fosol: So the component requires a `cell` parameter with a `value` property that is an `IProperty`?

frontend/src/constants/workflows.ts
```diff
@@ -0,0 +1,7 @@
+/** Property type enum */
+export enum Workflows {
```
Fosol, line 2: Might need to include the `ASSESS-EX-DISPOSAL` workflow too

Fosol (approved):

backend/dal/Services/Concrete/PropertyService.cs
devinleighsmith: @Fosol maybe not out of scope. with this logic, if the property is in an SPP project and a draft project the project workflow will be set to the workflowCode of DRAFT and then won't display properly on the map (depending on the order of the projects in the DB).

frontend/src/components/Table/PropertyTypeCell.tsx
devinleighsmith: Yeah, so essentially it can only be used in tables that are listing rows of IProperty

devinleighsmith: I think that is adequately described in the file name and the above comment, but I could add more doc.

frontend/src/constants/workflows.ts
devinleighsmith: ok, presumably that should trigger the ERP red map markers?

jdtoombs (approved):

frontend/src/constants/workflows.ts
devinleighsmith: Ill just include this in my next PR.

PR #909, opened 2021-02-10
PA-2482 Allow row increase in select properties

jdtoombs:
Originally had just implemented a **Show All** option like the story suggested; however, there was no way to get back from Show All state to reduce the list size (based off UI) so I thought it may be better to just allow them to choose from multiple options with the `PageSelector's` existing functionality in case they want to jump between sizes. 

Confirmed this was ok with Bobbi!

PageSizeSelection

### DIFF
--- a/frontend/src/components/Table/Table.tsx
+++ b/frontend/src/components/Table/Table.tsx
@@ -119,7 +119,7 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   filterable?: boolean;
   filter?: { [key in keyof T]?: any };
   onFilterChange?: (values: any) => void;
-  /** have menu drop-up to avoid contianer growing in some scenraios */
+  /** have page selection menu drop-up to avoid contianer growing in some scenraios */
   dropUp?: boolean;
 }
 
--- a/frontend/src/components/Table/Table.tsx
+++ b/frontend/src/components/Table/Table.tsx
@@ -120,7 +120,7 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   filter?: { [key in keyof T]?: any };
   onFilterChange?: (values: any) => void;
   /** have page selection menu drop-up to avoid container growing in some scenarios */
-  dropUp?: boolean;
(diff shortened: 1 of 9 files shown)


frontend/src/components/Table/Table.tsx
```diff
@@ -119,6 +119,8 @@ export interface TableProps<T extends object = {}> extends TableOptions<T> {
   filterable?: boolean;
   filter?: { [key in keyof T]?: any };
   onFilterChange?: (values: any) => void;
+  /** have page selection menu drop-up to avoid contianer growing in some scenraios */
+  dropUp?: boolean;
```
agalbrai, line 123: nit picky but "contianer", "scenraios"

agalbrai (approved):
👍

frontend/src/components/Table/Table.tsx
jdtoombs: Lol nit picky is good, rather not have a million typos from myself :)

frontend/src/features/projects/common/components/PropertyListViewSelect.tsx
```diff
@@ -155,14 +159,15 @@ export const PropertyListViewSelect: React.FC<InputProps> = ({
             name="SelectPropertiesTable"
             columns={columns}
             data={data}
-            lockPageSize
+            dropUp
```
Fosol, line 162: What's a `dropUp`?

Fosol (approved):

frontend/src/features/projects/common/components/PropertyListViewSelect.tsx
jdtoombs: Just has the menu options grow upwards rather than a 'drop-down menu'. Playing off of Goeme's existing functionality. Helps prevent the menu from being cut off / and or the container unnecessarily growing

jdtoombs: Going to rename this prop so it fits the description better

PR #910, opened 2021-02-10
PA-2388 Updated Caddy Config

Fosol:
Caddy configuration doesn't handle if the URL has a path.  We'll need to update.


openshift/4.0/templates/maintenance/README.md
```diff
@@ -68,21 +67,49 @@ You may need to manually run the build config.
...
 ```conf
-PROJECT_NAMESPACE=354028
-ROLE_NAME=proxy
 ENV_NAME=dev
+IMAGE_TAG=dev
 APP_DOMAIN=proxy-caddy-pims-dev.apps.silver.devops.gov.bc.ca
 APP_PORT=2015
+CPU_REQUEST=100m
+CPU_LIMIT=1
+MEMORY_REQUEST=500Mi
```
jdtoombs, line 84: Today I learned what a mebibyte is... lol

jdtoombs (approved):

openshift/4.0/templates/maintenance/README.md
agalbrai: me too haha

devinleighsmith (approved):

devinleighsmith:
Wait, so this doesn't fix the caddyfile issue?

Fosol:
No it doesn't. I spent an hour trying a few different configurations but couldn't find one that would return the maintenance page no matter what the original path was.

PR #911, opened 2021-02-10
PA 2459 Highlight active pin

agalbrai:
image

- This currently only works for single pins


frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -154,39 +244,70 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
...
-  } else if (propertyTypeId === PropertyTypes.PARCEL) {
-    return parcelIcon;
-  } else if (propertyTypeId === PropertyTypes.SUBDIVISION) {
-    return subdivisionIcon;
-  } else if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
+  if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
     return draftParcelIcon;
   } else if (propertyTypeId === PropertyTypes.DRAFT_BUILDING) {
     return draftBuildingIcon;
+  } else if (selected) {
```
agalbrai, line 253: I wasn't sure if it would be better to do one global check here, or check in the return statement for each type

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -154,39 +244,70 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
...
-        return subdivisionSppIcon;
-      default:
-        return landSppIcon;
-    }
-  } else if (propertyTypeId === PropertyTypes.PARCEL) {
-    return parcelIcon;
-  } else if (propertyTypeId === PropertyTypes.SUBDIVISION) {
-    return subdivisionIcon;
-  } else if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
+  if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
```
agalbrai, line 249: do we want a draft pin to be highlighted? it's already blinking

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -361,7 +361,7 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
           !currentClusterIds.includes(+selected?.parcelDetail?.id) && (
             <SelectedPropertyMarker
               {...selected.parcelDetail}
-              icon={getMarkerIcon({ properties: selected } as any)}
+              icon={getMarkerIcon({ properties: selected } as any, true)}
```
agalbrai, line 364: I'm stumped for what to do for the spiderfied clusters

Fosol: Do you mean highlighting the selected cluster?  Is that a requirement?  I'd lean towards not doing that, as the value doesn't seem to justify it.

Fosol (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
agalbrai: No, I mean highlighting a pin within a cluster, as this only will highlight single pins

Fosol: Ah, that would be more complicated yes.

frontend/src/components/maps/leaflet/mapUtils.tsx
devinleighsmith: I don't think so, but I have yet to see any UI for draft marker functionality so who knows?

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: I think just highlighting the **selected** pin is fine, I've made some updates that should make it appear more consistently.

devinleighsmith (approved):

PR #914, opened 2021-02-11
PA-2635: ERP or SPL Checkbox now required

jdtoombs:
Validation for ERP or SPL Checkboxes now make them required **ONLY** when searching with the Surplus Properties Filter

### DIFF
--- a/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
+++ b/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
@@ -7,7 +7,7 @@ import { mapLookupCode } from 'utils';
 import * as API from 'constants/API';
 import styled from 'styled-components';
 import { Container, Form, Row } from 'react-bootstrap';
-import { Button, Check, Input, Select } from 'components/common/form';
+import { Button, Check, DisplayError, Input, Select } from 'components/common/form';
 import _ from 'lodash';
 import { getIn, useFormikContext } from 'formik';
 import { TypeaheadField } from 'components/common/form/Typeahead';
@@ -61,13 +61,21 @@ const SearchButton = styled(props => <Button {...props} />)`
   margin-left: 665px;
 `;
 
-const ErrorMessage = styled.p`
-  color: ${variables.dangerColor};
-`;
-
 const StyledLocation = styled(props => <TypeaheadField {...props} />)`
   width: 250px;
   margin-left: 60px;
+`;
+
+/** the area where invalid feedback or ERP/SPL selection will be displayed */
+const InvalidFeedback = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: row-reverse;
+  margin: 0.5rem 0;
+  text-align: center;
+  .invalid-feedback {
+    display: block;
+  }
 `;
 /** This form is triggered by the FindMorePropertiesButton and contains additional filter fields for the PropertiesFilter */
 const FindMorePropertiesForm = <T extends any>(props: any) => {
@@ -86,11 +94,12 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
     return lookupCode.type === API.AMINISTRATIVE_AREA_CODE_SET_NAME;
   });
   const adminAreas = (administrativeAreas ?? []).map(c => mapLookupCode(c, null));
+
+  /** attempt submission of search, display errors if present */
   const handleSearch = () => {
-    if (errors.inEnhancedReferralProcess || errors.inSurplusPropertyProgram) {
+    handleSubmit();
+    if (errors) {
       setDisplayError(true);
-    } else {
-      handleSubmit();
     }
   };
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/common/GenericModal.tsx
```diff
@@ -103,7 +103,12 @@ const GenericModal = (props: ModalProps) => {
 
   return (
     <Container>
-      <Modal show={show} onHide={close} dialogClassName={classNames(props.size)}>
+      <Modal
+        style={{ zIndex: 1080 }}
```
jdtoombs, line 107: had to make the z index > than the popover

frontend/src/features/properties/filter/IPropertyFilter.ts
```diff
@@ -50,4 +50,6 @@ export interface IPropertyFilter {
   maxMarketValue?: '';
   /** Whether to return properties owned by other agencies. */
   includeAllProperties?: boolean;
+  /** when the surplus properties filter is open */
```
jdtoombs, line 53: Needed to add another field value to the filter so that I could check within the `YupSchema` whether it was being called from the Surplus Filter because otherwise we don't care if they are null

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -154,13 +154,19 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
         setSubmitting(false);
       }}
     >
-      {({ isSubmitting, handleReset, handleSubmit, setFieldValue, values }) => (
```
jdtoombs, line 157: unused

frontend/src/utils/YupSchema.ts
```diff
@@ -239,18 +239,49 @@ export const ParcelSchema = Yup.object()
...
+        }
+      },
+      then: Yup.boolean().nullable(),
+      otherwise: Yup.boolean().required(
+        'ERP or SPL Properties required when using the Surplus Properties filter.',
+      ),
+    }),
+    inSurplusPropertyProgram: Yup.boolean().when(['inEnhancedReferralProcess', 'surplusFilter'], {
+      is: (inEnhancedReferralProcess, surplusFilter) => {
+        if (!surplusFilter) {
```
jdtoombs, line 270: Always valid when not using the surplus filter

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
```diff
@@ -102,6 +112,13 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
         </a>
       </p>
       <FormSection>
+        {displayError && (
```
jdtoombs, line 115: I was having some difficulty with out existing components so just used a <p> instead. If preferred I can try to work the other one out

frontend/src/components/common/GenericModal.tsx
jdtoombs: Outdated removing

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
devinleighsmith: the other one requires that an error be in the touched array and in the errors array so that may be all that you are missing.

devinleighsmith: also check that .invalid-feedback is not set do display: none

devinleighsmith: hmm, can't you just trigger handlesubmit and then if there is a validation error, display it?

Fosol (approved):

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
jdtoombs: the display being none was the culprit!

devinleighsmith: Do you need the displayError && part? or is it enough that the DisplayError component won't display unless the touched and error values within formik are present?

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
```diff
@@ -102,7 +121,13 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
         </a>
       </p>
       <FormSection>
-        <StyledRow style={{ marginLeft: 140, paddingTop: 10 }}>
+        {displayError && (
+          <InvalidFeedback>
```
devinleighsmith, line 125: Do you need the displayError && part? or is it enough that the DisplayError component won't display unless the touched and error values within formik are present?

jdtoombs: It seems so because the user may not be interacting with those checkboxes so they will never be touched. And with the errorPrompt prop enabled it will display even without submit so I need a little bit of extra logic

jdtoombs: (if they try to submit without choosing an option)

PR #916, opened 2021-02-12
PA-1906: Add SRES to BCC for all notifications

jdtoombs:
Add SRES to BCC for all notification templates, already included in the first one (id=1)


devinleighsmith (approved):

Fosol (commented):
Make sure that when we have emails configured to go to the user who triggered them for DEV and TEST that the bcc doesn't get triggered.  Otherwise they will start getting spammed when we're testing

agalbrai (approved):

Fosol (changes requested):
Need to make sure we don't spam SRES when testing

jdtoombs:
Looks like @Fosol already [did the work for me](https://github.com/bcgov/PIMS/pull/427) for ensuring Bcc is not sent in testing environments. Tested with personal email address as well

Fosol (approved):

PR #919, opened 2021-02-12
PA-2115: Project status filter Fix

jdtoombs:
So there was a line tucked away that was checking the value of the status against the `parentId` instead of the value of the item's parent. It just happened to work still because those values shared the same number id's as the parent. But caused that weird bug we were seeing before. The headers will now correctly stay in place.

Highlights the readability issues of this component, not a huge fan of my implementation. Adjusted some variable names to try and be more clear.

statuses

### DIFF
--- a/frontend/src/components/common/form/ParentSelect.tsx
+++ b/frontend/src/components/common/form/ParentSelect.tsx
@@ -146,7 +146,7 @@ export const ParentSelect: React.FC<IParentSelect> = ({
                     onClick={() =>
                       enableMultiple
                         ? handleMultiSelectHeaderClick(
-                            results.filter(x => x.parentId?.toString() === parentId),
+                            results.filter(x => x.parentId === +parentId),
                           )
                         : handleMenuHeaderClick(results.find(x => x.value === parentId)!)
                     }
@@ -154,7 +154,7 @@ export const ParentSelect: React.FC<IParentSelect> = ({
                     <b style={{ cursor: 'pointer' }}>
                       {/* project statuses has a different way of selecting parent value*/}
                       {field === 'statusId'
-                        ? results.find(x => x.parentId?.toString() === parentId)?.parent
+                        ? results.find(x => x.parentId === +parentId)?.parent
                         : results.find(x => x.value === parentId)?.label}
                     </b>
                   </Menu.Header>
--- a/frontend/src/components/common/form/ParentSelect.tsx
+++ b/frontend/src/components/common/form/ParentSelect.tsx
@@ -127,7 +127,8 @@ export const ParentSelect: React.FC<IParentSelect> = ({
         hideValidation
         required={required}
         renderMenu={(results, menuProps) => {
-          const parents = groupBy(
+          /** group the results by their desired parents */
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -136,28 +137,30 @@ export const ParentSelect: React.FC<IParentSelect> = ({
             }),
             x => x.parentId,
           );
-          const items = Object.keys(parents)
+          const items = Object.keys(resultGroup)
             .sort()
-            .map(parent => (
-              <Fragment key={parent}>
-                {!!results.find((x: SelectOption) => x.value === parent) && (
```
jdtoombs, line 143: The culprit 😢 🤦

agalbrai (approved):

devinleighsmith:
we may be able to partially address some of the readability issues with more robust tests.

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -136,28 +137,30 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
+            .map(parentId => (
+              <Fragment key={parentId}>
+                {!!results.find((x: SelectOption) => x.parentId === +parentId) && (
                   <Menu.Header
                     onClick={() =>
                       enableMultiple
                         ? handleMultiSelectHeaderClick(
-                            results.filter(x => x.parentId?.toString() === parent),
+                            results.filter(x => x.parentId?.toString() === parentId),
```
devinleighsmith, line 149: its a bit confusing how here we are doing string comparison of parentIds but above we are doing number comparisons.

Fosol (approved):

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -136,28 +137,30 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
+                        : handleMenuHeaderClick(results.find(x => x.value === parentId)!)
                     }
                   >
                     <b style={{ cursor: 'pointer' }}>
+                      {/* project statuses has a different way of selecting parent value*/}
                       {field === 'statusId'
-                        ? results.find(x => x.parentId?.toString() === parent)?.parent
-                        : results.find(x => x.value === parent)?.label}
+                        ? results.find(x => x.parentId?.toString() === parentId)?.parent
+                        : results.find(x => x.value === parentId)?.label}
```
devinleighsmith, line 158: so x.value is a string but x.parentId isn't and so we need the toString()?

jdtoombs: There's something strange with out typing somewhere, going to look into it more after I finish off my last story

jdtoombs: But yes correct

frontend/src/components/common/form/ParentSelect.tsx
jdtoombs: Oh woops changed one and forgot to do the other

PR #921, opened 2021-02-13
PA-2528: Export additional fields for SRES

jdtoombs:
New export button and report for SRES/SRES Financial Manager. This report contains some additional fields that the basic report does not.

Few things to note: 

- The story asked for all information we had, I did not include the numeric values for classifications etc as I assumed this was not desired. Please correct me otherwise!
- I believe I have pulled in all relevant fields outside of what I mentioned above - hard to confirm as I did not have a list to reference other than the property entity itself (did my best to choose what is relevant). So if one looks like it's missing let me know

exportall

### DIFF
--- a/backend/api/Areas/Reports/Controllers/PropertyController.cs
+++ b/backend/api/Areas/Reports/Controllers/PropertyController.cs
@@ -2,7 +2,6 @@ using MapsterMapper;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
-using Pims.Api.Areas.Reports.Models.Property;
 using Pims.Api.Helpers.Constants;
 using Pims.Api.Helpers.Exceptions;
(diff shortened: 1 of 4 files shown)


backend/api/Areas/Reports/Controllers/PropertyController.cs
```diff
@@ -66,9 +66,27 @@ public IActionResult ExportProperties(bool all = false)
             return ExportProperties(new Property.Models.Search.PropertyFilterModel(query), all);
         }
 
+        /// <summary>
+        /// Exports properties with all fields as an Excel file, only available for SRES
+        /// Include 'Accept' header to request the appropriate export
+        /// </summary>
+        /// <param name="all"></param>
+        /// <returns></returns>
+        [HttpGet("allfields")]
```
jdtoombs, line 75: Not sure about the naming convention for this

Fosol (approved):
I have updated this PR endpoint route and fixed the endpoint permission bug.

PR #923, opened 2021-02-14
PA-2742 Disabled some schools

Fosol:
Disabled school districts that are no longer in use.


devinleighsmith:
I don't know if I can provide much value here, presumably you based this off a list so it should be correct.

devinleighsmith (approved):

PR #927, opened 2021-02-16
Fix disabled agency

Fosol:
Disabled agencies and other lookup codes shouldn't be selectable when they have been disabled.  There is a lot of tech-debt related to this issue, as we haven't correctly applied the rules.

- Disabled lookup codes shouldn't be displayed as a selection option
- If an existing entity uses the disabled lookup code, then it should still be able to show that lookup code.

### DIFF
--- a/frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
+++ b/frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
@@ -3,7 +3,7 @@ import renderer from 'react-test-renderer';
 import { createMemoryHistory } from 'history';
 import { Provider } from 'react-redux';
 import { Router } from 'react-router-dom';
-import { render, act, screen, wait } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { useKeycloak } from '@react-keycloak/web';
 import { Claims } from 'constants/claims';
 import MockAdapter from 'axios-mock-adapter';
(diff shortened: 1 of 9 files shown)


devinleighsmith:
I think this PR is missing the useLookupCodes hook which is used many places in the App.

devinleighsmith:
Might be better to switch all of these to use the useLookupCodes hook, which applies the disabled filter.

frontend/src/features/projects/erp/steps/GreTransferStep.test.tsx
```diff
@@ -161,39 +161,42 @@ describe('GRE Transfer Step', () => {
     beforeAll(() => {
       mockKeycloak([Claims.ADMIN_PROJECTS]);
     });
-    it('enables update button when new agency is selected', async () => {
-      const project = _.cloneDeep(mockProject);
-      const { container } = render(getGreTransferStep(getStore(project)));
-      const updateButton = screen.getByText(/Update Property Information Management System/);
-      await fillInput(container, 'agencyId', 'BC Transit', 'typeahead');
+    // TODO: Stopped working after apply `useLookupCodes` hook.
```
Fosol, line 164: I couldn't figure out why these unit-test stopped working after introducing the `useLookupCodes` hook

agalbrai (approved):

jdtoombs (approved):

PR #934, opened 2021-02-16
Tech debt and small parent select bug

jdtoombs:
- removed `AutoCompleteText.tsx` 
- removed outdated `ParentGroupedFilter.tsx` as it was replaced by `ParentSelect.tsx`
- fixed bug of agency parent disappearing after filtering the menu


frontend/src/features/properties/list/columns.tsx
```diff
@@ -69,7 +68,7 @@ export const columns = (
     sortable: true,
     filterable: true,
     filter: {
-      component: ParentGroupedFilter,
+      component: TypeaheadField,
```
jdtoombs, line 71: Outdated parent grouped filter, as well as we do not need it as we are splitting up the agency and sub agencies here

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -148,14 +148,14 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
-                        : handleMenuHeaderClick(results.find(x => x.value === parentId)!)
+                        : handleMenuHeaderClick(options.find(x => x.value === parentId)!)
                     }
                   >
                     <b style={{ cursor: 'pointer' }}>
                       {/* project statuses has a different way of selecting parent value*/}
                       {field === 'statusId'
                         ? results.find(x => x.parentId === +parentId)?.parent
-                        : results.find(x => x.value === parentId)?.label}
+                        : results.find(x => x.parentId === +parentId)?.parent}
```
jdtoombs, line 158: Was previously looking through the results for the parent; however, the result could have been rendered out. Now correctly looks at parentId and puts the string value as the header. Then when clicked it will browse the options for the value as it will contain it even if it has been rendered out of the results

jdtoombs: Woops just noticed I can get rid of ternary

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -148,14 +148,11 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
-                        : handleMenuHeaderClick(results.find(x => x.value === parentId)!)
+                        : handleMenuHeaderClick(options.find(x => x.value === parentId)!)
                     }
                   >
                     <b style={{ cursor: 'pointer' }}>
-                      {/* project statuses has a different way of selecting parent value*/}
-                      {field === 'statusId'
-                        ? results.find(x => x.parentId === +parentId)?.parent
-                        : results.find(x => x.value === parentId)?.label}
+                      {results.find(x => x.parentId === +parentId)?.parent}
```
jdtoombs, line 155: Now operates the same for both

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -148,14 +148,11 @@ export const ParentSelect: React.FC<IParentSelect> = ({
                         ? handleMultiSelectHeaderClick(
                             results.filter(x => x.parentId === +parentId),
                           )
-                        : handleMenuHeaderClick(results.find(x => x.value === parentId)!)
+                        : handleMenuHeaderClick(options.find(x => x.value === parentId)!)
```
jdtoombs, line 151: Have to get the value from `options` as the results may no longer have it as it could possibly be filtered out at this point

Fosol (approved):
cheer

agalbrai (approved):

PR #936, opened 2021-02-16
Prevent validation form removing the display of errors.

devinleighsmith:


frontend/src/features/projects/spl/steps/SplStep.tsx
```diff
@@ -157,7 +157,6 @@ const SplStep = ({ formikRef }: IStepProps) => {
     <Container fluid className="splStep">
       <Formik
         initialValues={initialValues}
-        validateOnMount={true}
```
devinleighsmith, line 160: This one gave me a heck of a time. Essentially what was happening when there is a submit is this
submitStatusCode gets set to a value
this triggers a submit
the submit fails do to a validation error
submitStatusCode gets reset after submission

The reset was triggering the mount validation because the state submitStatusCode exists above this Formik component, which causes the validation to run again after the submit logic, overwriting any errors thrown by the submit validation.

devinleighsmith: turning off the validateOnMount fixes this for now.

Eventually I'd like to rewrite this logic to stop using the submitStatusCode state var and instead just pass down a submit function that has the validation baked in (bypassing formik's submit logic). Until then this should work.

Fosol (approved):

agalbrai (approved):
Nice catch! I'm sure that was frustrating

PR #939, opened 2021-02-17
PA-2394 Updating Keycloak Sync Config

Fosol:
Updating configuration for the Keycloak Sync Tool so that it will work with the new environment.


tools/keycloak/sync/realm.json
```diff
@@ -485,9 +497,15 @@
...
         "AuthorizationServicesEnabled": true,
         "RootUrl": "https://pims.gov.bc.ca/api",
-        "RedirectUris": ["https://pims.gov.bc.ca/api/*"],
+        "RedirectUris": [
+          "https://pims.gov.bc.ca/api/*",
+          "https://pims.apps.silver.devops.gov.bc.ca/api/*"
+        ],
         "BaseUrl": "/",
-        "WebOrigins": ["*"],
+        "WebOrigins": [
```
devinleighsmith, line 505: Why is this changing from * to two specific urls? have you tested this?

devinleighsmith (approved):

jdtoombs (approved):

PR #940, opened 2021-02-17
PA-2827 ensure pid/pin search returns building information. remove logic that is removing associated buildings from land.

devinleighsmith:


devinleighsmith:
The bug is:
If a user is submitting parcels using the update parcel inventory form, and they drop the parcel marker on a parcel that is already in PIMS, the parcel's data would be loaded into the form but not the associated buildings. If the user saved the form it would then remove all of the building associations. See comments on code for more details.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -84,13 +84,39 @@ public static IQueryable<Entity.Parcel> GenerateQuery(this PimsContext context,
             {
                 var pidValue = filter.PID.Replace("-", "").Trim();
                 if (Int32.TryParse(pidValue, out int pid))
-                    query = query.Where(p => p.PID == pid || p.PIN == pid);
+                    query = query.Include(p => p.Buildings)
```
devinleighsmith, line 87: This includes the building data if we are doing a pid/pin search, since a pid/pin search is what the FE uses to find a parcel match for map clicks and pid/pin searches.

Alternative would be to make one request to get the id of the pid/pin match, and another followup to get the details.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -431,15 +431,6 @@ public Parcel PendingUpdate(Parcel parcel)
                 foreach (var parcelBuilding in originalParcel.Buildings)
                 {
                     var updateBuilding = parcel.Buildings.FirstOrDefault(pb => pb.BuildingId == parcelBuilding.BuildingId);
-                    if (updateBuilding == null)
```
devinleighsmith, line 434: as far as I'm aware, there is no FE workflow that allows a user to **disassociate** one or more buildings from a parcel. So even though this code is correct, it serves no purpose other then to potentially harm the land->building relationships if the frontend makes a bad request.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
Fosol: We'll need to do a hot-fix to prod for this.

backend/dal/Services/Concrete/ParcelService.cs
devinleighsmith: Also worth noting that I've left in the inverse of this logic (if a user submits a building with fewer associated land than what is in the DB, all **leftover** associations will be removed. I've done that because there is a FE workflow to remove associated land from a building.

Fosol (approved):

jdtoombs (approved):

agalbrai (approved):

PR #942, opened 2021-02-17
PA-2589: Purple pin appearing when it shouldn't

jdtoombs:
Purple pin now checks for the existence of project number(s) as well as the Surplus Active/Encumbered classification

### DIFF
--- a/frontend/src/components/maps/leaflet/mapUtils.tsx
+++ b/frontend/src/components/maps/leaflet/mapUtils.tsx
@@ -244,7 +244,13 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
  * Get an icon type for the specified cluster property details (type, draft, erp, spp etc)
  */
 export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
-  const { propertyTypeId, projectWorkflow, classificationId } = feature?.properties;
+  const {
+    propertyTypeId,
+    projectWorkflow,
+    classificationId,
+    parcelDetail,
+    projectNumbers,
+  } = feature?.properties;
   if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
     return draftParcelIcon;
   } else if (propertyTypeId === PropertyTypes.DRAFT_BUILDING) {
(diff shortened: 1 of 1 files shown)


agalbrai (approved):

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -264,8 +270,9 @@ export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
           return landErpIconSelect;
       }
     } else if (
-      selectedClassificationId === Classifications.SurplusActive ||
-      selectedClassificationId === Classifications.SurplusEncumbered
+      (selectedClassificationId === Classifications.SurplusActive ||
+        selectedClassificationId === Classifications.SurplusEncumbered) &&
+      projectNumbers.length
```
devinleighsmith, line 275: Unfortunately I still think this is not quite right because projectNumbers may have one or more draft projects which shouldn't change the display of the parcel.

My suggestion would be to see if we can use the workflowcode similar to how erp is working, or if not just check for the presence of SPP in the projectNumbers array.

jdtoombs: Ahh ok, makes sense. Will adjust - thanks!

Fosol (commented):
Ideally a property will not be part of multiple projects.  We probably shouldn't even allow it, but oh well.   So you'll just need the workflow of one of those projects that isn't in draft.  This isn't great either, but we can't do everything.

devinleighsmith (approved):

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -264,8 +264,10 @@ export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
           return landErpIconSelect;
       }
     } else if (
-      selectedClassificationId === Classifications.SurplusActive ||
-      selectedClassificationId === Classifications.SurplusEncumbered
+      (selectedClassificationId === Classifications.SurplusActive ||
+        selectedClassificationId === Classifications.SurplusEncumbered) &&
+      projectWorkflow &&
+      [Workflows.SPL].includes(projectWorkflow)
```
jdtoombs, line 270: After talking with Devin and Bobbi, we determined that we only want the pin to show as purple it has been approved and is in the SPL workflow

Fosol (approved):

PR #944, opened 2021-02-17
PA 2459 highlight active pin

agalbrai:
- Now highlights selected pin in a cluster, and any pin on click rather than just when the detail view is open

https://user-images.githubusercontent.com/33818575/108274048-3d3df300-7129-11eb-973e-e3bfb7d48810.mp4

### DIFF
--- a/frontend/src/components/maps/leaflet/PointClusterer.tsx
+++ b/frontend/src/components/maps/leaflet/PointClusterer.tsx
@@ -18,6 +18,8 @@ import { useApi } from 'hooks/useApi';
 import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
 import { PropertyTypes } from 'constants/propertyTypes';
 import SelectedPropertyMarker from './SelectedPropertyMarker/SelectedPropertyMarker';
+import * as parcelsActions from 'actions/parcelsActions';
+import { useDispatch } from 'react-redux';
 
 export type PointClustererProps = {
   points: Array<PointFeature>;
@@ -75,7 +77,6 @@ export const convertToProperty = (
   ) {
     return property;
   }
-
   return null;
 };
 
@@ -204,18 +205,6 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
   );
 
   /**
-   * Update the map view to center on the selected parcel when it changes.
-   */
-  useDeepCompareEffect(() => {
-    if (!!selected?.parcelDetail?.latitude && !!selected?.parcelDetail?.longitude) {
-      map.setView(
-        { lat: selected.parcelDetail.latitude, lng: selected.parcelDetail.longitude },
-        Math.max(maxZoom as number, map.getZoom()),
-      );
-    }
-  }, [selected]);
-
-  /**
    * Update the map bounds and zoom to make all draft properties visible.
    */
   useDeepCompareEffect(() => {
@@ -272,7 +261,10 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
     },
     [getParcel, getBuilding, popUpContext],
   );
+
   const keycloak = useKeycloakWrapper();
+  const dispatch = useDispatch();
+
   return (
     <>
       <FeatureGroup ref={featureGroupRef}>
@@ -305,7 +297,7 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
                     iconSize: [40, 40],
                   })
                 }
-              ></Marker>
+              />
             );
           }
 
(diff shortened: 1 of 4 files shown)


frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -203,18 +204,6 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
-   * Update the map view to center on the selected parcel when it changes.
-   */
-  useDeepCompareEffect(() => {
-    if (!!selected?.parcelDetail?.latitude && !!selected?.parcelDetail?.longitude) {
-      map.setView(
-        { lat: selected.parcelDetail.latitude, lng: selected.parcelDetail.longitude },
-        Math.max(maxZoom as number, map.getZoom()),
-      );
-    }
-  }, [selected]);
```
agalbrai, line 216: The view is set whenever a pin is clicked, don't need it repeated here. Also the zoom is done when "view details" or "Update" is clicked, this would trigger the map to zoom any time a pin is clicked now

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -317,13 +309,25 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+                  cluster.properties,
+                  latitude,
+                  longitude,
+                );
+                //sets this pin as currently selected
+                if (
+                  cluster.properties.propertyTypeId === PropertyTypes.PARCEL ||
+                  cluster.properties.propertyTypeId === PropertyTypes.SUBDIVISION
+                ) {
+                  dispatch(parcelsActions.storeParcelDetail(convertedProperty as IParcel));
```
agalbrai, line 322: Correct me if I'm wrong that a Subdivision is still a Parcel

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -338,8 +342,21 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
             {...(m.properties as any)}
             key={index}
             position={m.position}
-            icon={getMarkerIcon(m)}
+            //highlight pin if currently selected
+            icon={getMarkerIcon(
+              m,
+              (m.properties.id as number) === (selected?.parcelDetail?.id as number),
```
agalbrai, line 348: Check if the current pin in the cluster is the selected pin

frontend/src/components/maps/leaflet/SelectedPropertyMarker/SelectedPropertyMarker.tsx
```diff
@@ -12,7 +12,6 @@ const SelectedPropertyMarker: React.FC<MarkerProps & { map: Map }> = props => {
   useDeepCompareEffect(() => {
     if (ref.current.leafletElement) {
       props.map.setView(props.position, props.map.getZoom());
-      ref.current.leafletElement.openPopup();
```
agalbrai, line 15: we no longer use the leaflet popup on properties so this could go

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -244,17 +244,13 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
  * Get an icon type for the specified cluster property details (type, draft, erp, spp etc)
  */
 export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
-  const { propertyTypeId, projectWorkflow, classificationId, parcelDetail } = feature?.properties;
+  const { propertyTypeId, projectWorkflow, classificationId } = feature?.properties;
```
agalbrai, line 247: parcelDetail was always usually undefined for some reason, already had the info needed anyway

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: this is correct.

frontend/src/components/maps/leaflet/mapUtils.tsx
devinleighsmith: can you verify this works properly when the inventory view is open? that is normally when parcelDetail is valued.

agalbrai: Looks pretty good to me I think 

https://user-images.githubusercontent.com/33818575/108283775-c4469780-7138-11eb-9bd9-871e1e6cebeb.mp4

Fosol (approved):
Cheer

frontend/src/components/maps/leaflet/PointClusterer.tsx
devinleighsmith: I think this will take away the logic that jumps back to a map marker after it is saved. I'm not sure that there is actually a requirement for that however so maybe that is fine.

devinleighsmith (approved):

frontend/src/components/maps/leaflet/PointClusterer.tsx
agalbrai: I just tried, it doesn't seem to zoom back out after the sidebar closes if that's what you mean

PR #946, opened 2021-02-17
PA 2767 Add text to building valuation

agalbrai:
image
- Also swapped the tables as per the ticket

### DIFF
--- a/frontend/src/features/mapSideBar/SidebarContents/BuildingForm.test.tsx
+++ b/frontend/src/features/mapSideBar/SidebarContents/BuildingForm.test.tsx
@@ -89,7 +89,7 @@ describe('Building Form', () => {
     await wait(() => {
       fireEvent.click(getByText(/Continue/i));
     });
-    expect(getByText(/Building Valuation/i)).toBeInTheDocument();
+    expect(getByText('Building Valuation', { exact: true })).toBeInTheDocument();
     await wait(() => {
       fireEvent.click(getByText(/Continue/i));
     });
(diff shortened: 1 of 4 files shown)


frontend/src/features/mapSideBar/SidebarContents/BuildingForm.test.tsx
```diff
@@ -89,7 +89,7 @@ describe('Building Form', () => {
     await wait(() => {
       fireEvent.click(getByText(/Continue/i));
     });
-    expect(getByText(/Building Valuation/i)).toBeInTheDocument();
+    expect(getByText('Building Valuation', { exact: true })).toBeInTheDocument();
```
agalbrai, line 92: changed this because there's text that says "building valuation" as well as the title, so this only gets the title

frontend/src/features/mapSideBar/SidebarContents/subforms/ValuationForm.scss
```diff
@@ -14,6 +14,9 @@
         margin-right: -20px;
       }
     }
+    &.assessed {
+      width: 600px;
```
agalbrai, line 18: to match the netbook table

Fosol (approved):

frontend/src/features/mapSideBar/SidebarContents/BuildingForm.test.tsx
jdtoombs: I believe just taking out the `i` from Regex will also get you the desired behaviour. It sets the Regex to case insensitive so `/Building Valuation/` would be case-sensitive. But this way/ regex way both accomplish the same thing, just wanted to pass it on!

jdtoombs (approved):

PR #947, opened 2021-02-18
Delete Project Adjustments

jdtoombs:
Only projects that are still in the `SUBMIT-DISPOSAL` workflow can be deleted. The delete icon will appear next to a project if that condition is satisfied AND the user 

- has a `project-edit` claim
- or an `admin-projects` claim
- or they created that project

### DIFF
--- a/backend/dal/Helpers/Extensions/ProjectExtensions.cs
+++ b/backend/dal/Helpers/Extensions/ProjectExtensions.cs
@@ -43,6 +43,7 @@ namespace Pims.Dal.Helpers.Extensions
                 .Include(p => p.TierLevel)
                 .Include(p => p.Risk)
                 .Include(p => p.Agency)
+                .Include(p => p.Workflow)
                 .Include(p => p.Agency).ThenInclude(a => a.Parent)
                 .Include(p => p.Notes)
                 .AsNoTracking();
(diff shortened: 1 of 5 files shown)


frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -222,6 +222,7 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
       project.status = projectStatuses.find((x: any) => x.name === project.status)!;
       await service.deleteProject(project);
       setData(data?.filter(p => p.projectNumber !== project.projectNumber));
+      setDeleteId(undefined);
```
jdtoombs, line 225: Allow for multiple deletions if necessary

backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -43,6 +43,7 @@ public static IQueryable<Entity.Project> GenerateQuery(this PimsContext context,
                 .Include(p => p.TierLevel)
                 .Include(p => p.Risk)
                 .Include(p => p.Agency)
+                .Include(p => p.Workflow)
```
jdtoombs, line 46: So the list view is able to access this information to determine if delete icon will be displayed based off of `workflowCode`

frontend/src/features/projects/list/ProjectListView.tsx
agalbrai: why?

jdtoombs: As in they delete one project and goto delete another one after, previously it would have kept the old id in the state variable and not reset

Fosol (approved):

agalbrai (approved):

PR #948, opened 2021-02-18
Logic fix for filter changed

jdtoombs:
`isEmpty(true)` returns `true` which caused an issue for when we set our filter changed state. The issue would occur when either the ERP or SPL checkbox was checked as even if the first half of code resulted in them not being equal, the second half would evaluate to true even though they are not "empty". i.e) `isEmpty(true) && isEmpty(undefined)` evaluates to true

### DIFF
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -237,11 +237,7 @@ const Map: React.FC<MapProps> = ({
   const handleMapFilterChange = async (filter: IPropertyFilter) => {
     const compareValues = (objValue: any, othValue: any) => {
       return whitelistedFilterKeys.reduce((acc, key) => {
-        return (
-          (isEqual(objValue[key], othValue[key]) ||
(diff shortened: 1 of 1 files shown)


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -237,11 +237,7 @@ const Map: React.FC<MapProps> = ({
   const handleMapFilterChange = async (filter: IPropertyFilter) => {
     const compareValues = (objValue: any, othValue: any) => {
       return whitelistedFilterKeys.reduce((acc, key) => {
-        return (
-          (isEqual(objValue[key], othValue[key]) ||
-            (isEmpty(objValue[key]) && isEmpty(othValue[key]))) &&
-          acc
-        );
+        return (isEqual(objValue[key], othValue[key]) || (!objValue[key] && !othValue[key])) && acc;
```
jdtoombs, line 240: Just replaced `isEmpty(obj[key])` with `!obj[key]` which handles the case of the value being true correctly

Fosol (approved):

agalbrai (approved):
nice catch!

PR #952, opened 2021-02-18
Remove properties validation in close out form

goemen:


frontend/src/features/projects/spl/forms/splYupSchema.ts
```diff
@@ -100,19 +100,4 @@ export const CloseOutFormValidationSchema = Yup.object().shape({
   offerAcceptedOn: Yup.date(),
   preliminaryFormSignedOn: Yup.date(),
   finalFormSignedOn: Yup.date(),
-  properties: Yup.array().of(
```
Fosol, line 103: The appraised value validation would be a required value and should be moved over to the appropriate tab.

Fosol (changes requested):

frontend/src/features/projects/spl/forms/splYupSchema.ts
goemen: This is removing the validation for the project properties, and we already the `appraised` field on the project information tab and is not a required field.

devinleighsmith: yep, I don't think appraised on the property array is required. Also dubious if it is required on the project. Maybe if it is in a certain tier?

Fosol (approved):

jdtoombs (approved):

PR #955, opened 2021-02-19
PA-2707 Fixed Property Classifications

Fosol:
Property Classifications will now use the commonly shared hook `useLookupCodes()`.  Additionally based on permissions the correct classifications will be displayed.

### DIFF
--- a/frontend/src/components/maps/leaflet/Map.tsx
+++ b/frontend/src/components/maps/leaflet/Map.tsx
@@ -61,7 +61,6 @@ export type MapProps = {
   zoom?: number;
   properties: IProperty[];
   agencies: ILookupCode[];
-  propertyClassifications: ILookupCode[];
   administrativeAreas: ILookupCode[];
   lotSizes: number[];
   mapRef: React.RefObject<ReactLeafletMap<LeafletMapProps, LeafletMap>>;
@@ -161,7 +160,6 @@ const Map: React.FC<MapProps> = ({
   lng,
   zoom: zoomProp,
   agencies,
-  propertyClassifications,
   administrativeAreas,
   lotSizes,
   selectedProperty,
@@ -339,7 +337,6 @@ const Map: React.FC<MapProps> = ({
                     }}
                     agencyLookupCodes={agencies}
                     adminAreaLookupCodes={administrativeAreas}
-                    propertyClassifications={propertyClassifications}
                     onChange={handleMapFilterChange}
                     showAllAgencySelect={true}
                   />
--- a/frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.test.tsx
+++ b/frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.test.tsx
@@ -1,16 +1,102 @@
 import React from 'react';
-import { ClassificationForm } from './ClassificationForm';
 import renderer from 'react-test-renderer';
-import { Formik } from 'formik';
-import { Classifications } from 'constants/classifications';
 import { cleanup, fireEvent, render, wait } from '@testing-library/react';
+import { Formik } from 'formik';
 import { noop } from 'lodash';
+import axios from 'axios';
+import thunk from 'redux-thunk';
+import configureMockStore from 'redux-mock-store';
+import { Provider } from 'react-redux';
+import { Router } from 'react-router-dom';
+import { createMemoryHistory } from 'history';
+import { useKeycloak } from '@react-keycloak/web';
+import { ClassificationForm } from './ClassificationForm';
+import { Classifications } from 'constants/classifications';
+import { SelectOption, SelectOptions } from 'components/common/form';
+import * as API from 'constants/API';
+import * as reducerTypes from 'constants/reducerTypes';
+import { ILookupCode } from 'actions/lookupActions';
 
+jest.mock('axios');
+jest.mock('@react-keycloak/web');
+
+const mockedAxios = axios as jest.Mocked<typeof axios>;
 const mockClassifications = [
-  { value: Classifications.CoreOperational, label: 'Core Operational' },
-  { value: Classifications.CoreStrategic, label: 'Core Strategic' },
-  { value: Classifications.SurplusEncumbered, label: 'Surplus Encumbered' },
-];
+  { value: Classifications.CoreOperational, label: 'Core Operational' } as SelectOption,
+  { value: Classifications.CoreStrategic, label: 'Core Strategic' } as SelectOption,
+  { value: Classifications.SurplusEncumbered, label: 'Surplus Encumbered' } as SelectOption,
+] as SelectOptions;
+
+const mockKeycloak = (claims: string[]) => {
+  (useKeycloak as jest.Mock).mockReturnValue({
+    keycloak: {
+      subject: 'test',
+      userInfo: {
+        roles: claims,
+        agencies: ['1'],
+      },
+    },
+  });
+};
+const mockStore = configureMockStore([thunk]);
+let history = createMemoryHistory();
+
+const lCodes = {
+  lookupCodes: [
+    { name: 'agencyVal', id: '1', isDisabled: false, type: API.AGENCY_CODE_SET_NAME },
+    { name: 'disabledAgency', id: '2', isDisabled: true, type: API.AGENCY_CODE_SET_NAME },
+    { name: 'roleVal', id: '1', isDisabled: false, type: API.ROLE_CODE_SET_NAME },
+    { name: 'disabledRole', id: '2', isDisabled: true, type: API.ROLE_CODE_SET_NAME },
+    {
+      name: 'Core Operational',
+      id: '0',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Core Strategic',
+      id: '1',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Surplus Active',
+      id: '2',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Surplus Encumbered',
+      id: '3',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Disposed',
+      id: '4',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Demolished',
+      id: '5',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+    {
+      name: 'Subdivided',
+      id: '6',
+      isDisabled: false,
+      type: API.PROPERTY_CLASSIFICATION_CODE_SET_NAME,
+    },
+  ] as ILookupCode[],
+};
+
+const getStore = () =>
+  mockStore({
+    [reducerTypes.PROPERTY_NAMES]: ['test'],
+    [reducerTypes.LOOKUP_CODE]: lCodes,
+  });
 
 it('renders correctly', () => {
   const tree = renderer.create(form).toJSON();
@@ -18,12 +104,25 @@ it('renders correctly', () => {
 });
 
 const form = (
-  <Formik onSubmit={noop} initialValues={{ classificationId: '' }}>
-    <ClassificationForm classifications={mockClassifications} field="classificationId" />
-  </Formik>
+  <Provider store={getStore()}>
+    <Router history={history}>
+      <Formik onSubmit={noop} initialValues={{ classificationId: '' }}>
+        <ClassificationForm
+          classifications={mockClassifications}
+          field="classificationId"
+          encumbranceField="test"
+        />
+      </Formik>
+    </Router>
+  </Provider>
 );
 
 describe('renders definitions correctly', () => {
+  mockedAxios.get.mockImplementationOnce(() => Promise.resolve({}));
+
+  beforeEach(() => {
+    mockKeycloak([]);
+  });
   afterEach(() => {
     cleanup();
     jest.clearAllMocks();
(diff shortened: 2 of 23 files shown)


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -61,7 +61,6 @@ export type MapProps = {
   zoom?: number;
   properties: IProperty[];
   agencies: ILookupCode[];
-  propertyClassifications: ILookupCode[];
```
Fosol, line 64: No longer required because they are retrieved via the `useLookupCodes()` hook.  Ideally I'd replace the others too, but keeping this PR as small as possible.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
```diff
@@ -101,20 +103,28 @@ export const ClassificationForm: React.FC<IClassificationFormProps> = ({
...
-      } else {
-        return val !== Classifications.Disposed;
-      }
-    } else {
-      return (
-        val !== Classifications.Subdivided &&
-        val !== Classifications.Disposed &&
-        val !== Classifications.Demolished
-      );
+    switch (context) {
```
Fosol, line 106: Based on the context display the correct classification options.

frontend/src/hooks/useKeycloakWrapper.tsx
```diff
@@ -49,7 +49,7 @@ export interface IKeycloak {
 /**
  * Provides extension methods to interact with the `keycloak` object.
  */
-function useKeycloakWrapper(): IKeycloak {
+export function useKeycloakWrapper(): IKeycloak {
```
Fosol, line 52: The general recommendation these days is to not use default exports, but I'm not in the know on whether it truly matters.  Either way, we can do both.

frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -46,8 +55,34 @@ function useCodeLookups() {
...
 
   const getOptionsByType = (type: string) => getByType(type).map(mapLookupCode);
+
+  /**
+   * Return an array of SelectOptions containing property classifications.
+   * @function
+   * @param filter - A filter to determine which classifications will be returned.
+   * @returns
+   */
+  const getPropertyClassificationOptions = (
```
Fosol, line 65: Here's the real change.  We now enforce user based claims and can override.

frontend/src/utils/utils.ts
```diff
@@ -78,17 +78,6 @@ export const isPositiveNumberOrZero = (input: string | number | undefined | null
   return !isNaN(Number(input)) && Number(input) > -1;
 };
 
-export const mapLookupCode = (
```
Fosol, line 81: Moved to separate file.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.test.tsx
```diff
@@ -1,29 +1,128 @@
...
 import renderer from 'react-test-renderer';
-import { Formik } from 'formik';
-import { Classifications } from 'constants/classifications';
 import { cleanup, fireEvent, render, wait } from '@testing-library/react';
+import { Formik } from 'formik';
 import { noop } from 'lodash';
+import axios from 'axios';
+import thunk from 'redux-thunk';
+import configureMockStore from 'redux-mock-store';
+import { Provider } from 'react-redux';
```
devinleighsmith, line 9: If I get some time I'd like to write a test component that provides a wrapper for the router/memory router and the provider/store, since some tests don't really care about setting that stuff up but they have to because of a useRouter or redux call somewhere.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.test.tsx
```diff
@@ -1,29 +1,128 @@
...
 
 describe('renders definitions correctly', () => {
+  mockedAxios.get.mockImplementationOnce(() => Promise.resolve({}));
+
+  beforeEach(() => {
+    mockKeycloak([]);
+  });
   afterEach(() => {
     cleanup();
     jest.clearAllMocks();
```
devinleighsmith, line 128: this might clear your axios mock above, just be careful.

devinleighsmith: remember that the order that tests are run in isn't guaranteed, so if more then one test needs the axios mock your tests may pass right now but break later when the execution order changes.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
devinleighsmith: this does functionally tie this classification form to the inventory sidebar. I'm ok with this given where we are in the project however.

devinleighsmith: ugh, the use of val as a variable name hurts me, not your code though.

frontend/src/features/mapSideBar/hooks/useQueryParamSideBar.tsx
```diff
@@ -52,7 +52,7 @@ interface IMapSideBar {
 }
 
 /** control the state of the side bar via query params. */
-const useQueryParamSideBar = (formikRef?: any): IMapSideBar => {
+export const useQueryParamSideBar = (formikRef?: any): IMapSideBar => {
```
devinleighsmith, line 55: this is already being exported as a default export.

frontend/src/features/projects/common/components/FilterBar.tsx
```diff
@@ -44,26 +46,22 @@ export interface IFilterBarState {
...
-    label: code.name,
-    value: code.id.toString(),
-  });
+const FilterBar: React.FC<FilterBarProps> = ({ onChange }) => {
+  const lookupCode = useCodeLookups();
   //restrict available agencies to user agencies.
-  const agencies = (agencyLookupCodes ?? []).map(c => mapLookupCode(c));
-  const classifications = (propertyClassifications ?? []).map(c => mapLookupCode(c));
+  const agencies = lookupCode.getOptionsByType(API.AGENCY_CODE_SET_NAME);
+  const classifications = lookupCode.getPropertyClassificationOptions(
```
devinleighsmith, line 59: this does make the assumption that this filter bar should never use these values regardless of where it is used, I'm fine with this because of where the project is at.

frontend/src/features/properties/filter/PropertyFilter.test.tsx
```diff
@@ -24,13 +26,76 @@ jest.mock('actionCreators/propertyActionCreator');
 (fetchPropertyNames as any).mockImplementation(jest.fn(() => () => ['test']));
 
 const mockedAxios = axios as jest.Mocked<typeof axios>;
+const mockKeycloak = (claims: string[]) => {
```
devinleighsmith, line 29: We want this in a shared function, but not strictly required for this PR.

frontend/src/hooks/useLookupCodes.tsx
devinleighsmith: Does the @function provide value?

devinleighsmith: Same with @returns if it is left blank.

frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -46,8 +55,34 @@ function useCodeLookups() {
...
+    filter?: (value: SelectOption, index: number, array: SelectOption[]) => unknown,
+  ) => {
+    const classifications = getByType(API.PROPERTY_CLASSIFICATION_CODE_SET_NAME);
+    return filter
+      ? (classifications ?? []).map(c => mapLookupCode(c)).filter(filter)
+      : !keycloak.hasClaim(Claims.ADMIN_PROPERTIES)
+      ? (classifications ?? [])
```
devinleighsmith, line 72: Not a fan of nested ternarys unless they are extremely simple. Suggesting this get simplified.

agalbrai (commented):
hurray for simplifying things!

frontend/src/hooks/useLookupCodes.tsx
devinleighsmith: possibly could just remove the filter entirely and allow users to use es6 builtin .map and .filter functions

frontend/src/features/projects/common/components/FilterBar.tsx
devinleighsmith: Also doesn't getPropertyClassificationOptions already do this?

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.test.tsx
jdtoombs: Oooh that would be handy

frontend/src/utils/mapLookupCode.ts
```diff
@@ -0,0 +1,20 @@
+import { ILookupCode } from 'actions/lookupActions';
+import { SelectOption } from 'components/common/form';
+
+/**
+ * Convert an ILookupCode into a SelectOption.
+ *
+ * @param {ILookupCode} code - The code value to identify this item.
+ * @param {(number | string | null)} [defaultId] - The default ID.
```
devinleighsmith, line 8: what does the defaultId do? would prefer comments that indicate funtionality or intended use, or skip if self explanatory and.

frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -46,8 +55,34 @@ function useCodeLookups() {
...
+   * @param filter - A filter to determine which classifications will be returned.
+   * @returns
+   */
+  const getPropertyClassificationOptions = (
+    filter?: (value: SelectOption, index: number, array: SelectOption[]) => unknown,
+  ) => {
+    const classifications = getByType(API.PROPERTY_CLASSIFICATION_CODE_SET_NAME);
+    return filter
+      ? (classifications ?? []).map(c => mapLookupCode(c)).filter(filter)
+      : !keycloak.hasClaim(Claims.ADMIN_PROPERTIES)
```
devinleighsmith, line 71: slight preference to structuring these as if true then false. optional.

frontend/src/hooks/useKeycloakWrapper.tsx
devinleighsmith: I'll look into this, I don't really like any recommendation if I don't understand the rationale behind it.

devinleighsmith:
Just to provide some explanation for some of my comments, While moving the logic to get the lookup codes into the functions that use them does allow us to remove some props, it also reduces the portability of those components. Not necessarily a problem, but something to keep in mind.

jdtoombs (approved):

devinleighsmith:
Thanks for doing this.

frontend/src/features/mapSideBar/hooks/useQueryParamSideBar.tsx
devinleighsmith: I see your other comment now, this is fine.

frontend/src/features/mapSideBar/SidebarContents/subforms/ClassificationForm.tsx
Fosol: I'll update to `id` and include a better description for the function.  I was going to add a prop to the component to control, but I couldn't figure out how it is called differently for subdivisions and gave up.  This works for now.

frontend/src/hooks/useKeycloakWrapper.tsx
Fosol: Adding it this way doesn't change the rest of our code, it just lets us use the `import { name } from name` and `import name from name`

frontend/src/hooks/useLookupCodes.tsx
Fosol: I've updated.

frontend/src/features/projects/common/components/FilterBar.tsx
Fosol: `getPropertyClassificationOptions(...)` will by default only return options that your claims allow.  However I've extended it to allow for a custom filter when required.  There are a few places where it is easier to implement if we have full control of the results.

frontend/src/features/properties/filter/PropertyFilter.test.tsx
Fosol: I think I extracted it from some other test and modified for my use-case.  I'd prefer a shared function too, but at this point I'm leaning toward adding technical debt, as I don't have enough hours to complete critical work.

frontend/src/hooks/useLookupCodes.tsx
Fosol: I also agree with nested ternarys being difficult to read. I gave into this one as it was the quickest way to add the overload with the existing code.

frontend/src/utils/mapLookupCode.ts
Fosol: This was inherited functionality.  I simply wanted to keep consistent with intellisense.  Self explanatory doesn't limit itself to no comment.  Something obvious to one person isn't obvious to another.

frontend/src/hooks/useLookupCodes.tsx
Fosol: I think it does follow that.  `filter == true` use the filter, otherwise do the claim.

Fosol:
> Just to provide some explanation for some of my comments, While moving the logic to get the lookup codes into the functions that use them does allow us to remove some props, it also reduces the portability of those components. Not necessarily a problem, but something to keep in mind.

I agree, the component is more tightly coupled.  That being said, within our solution the property classifications are tightly coupled with the component you created, so the benefits of centralizing the code enables us to make future changes more efficiently and with less chance of forgetting an instance.  Of course it also increases the number of touch points for any given change, which increases risk in a counter way.  Pros and cons to everything...

frontend/src/hooks/useLookupCodes.tsx
Fosol: Removing the filter entirely would remove the ability to override, as you would only get results based on your claims.  The purpose of the override is to allow for this hook to be used in other scenarios.

PR #956, opened 2021-02-19
PA-2809, 2808, 2813, 2811, 2812, 2810: Project number link related stories 

jdtoombs:
Display appropriate project number(s) in property info panel

- if an admin clicks the link it will go to the project status route
- link goes to summary if project has been submitted and user is in the same agency
- link goes to project status route if it is still in draft and user is a part of the agency
- if user is outside of the agency and not an admin the numbers will not be clickable 

Updated:

- land, associated land, and building review pages now have project number(s) with appropriate links
- also included when viewing the info screen
- fixed bug where lat and long was not being stored when a marker was clicked
- fixed another bug where the project workflow information was not being read correctly because the data shape was changing 

image

### DIFF
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
@@ -11,6 +11,8 @@ import { ThreeColumnItem } from './ThreeColumnItem';
 import variables from '_variables.module.scss';
 import { PropertyTypes } from 'constants/propertyTypes';
 import useCodeLookups from 'hooks/useLookupCodes';
+import { useState } from 'react';
+import { ProjectNumberLink } from './ProjectNumberLink';
 
 /**
  * Compare two dates to evaluation which is earlier.
@@ -83,6 +85,7 @@ export const InfoContent: React.FC<IInfoContent> = ({
     [PropertyTypes.PARCEL, PropertyTypes.SUBDIVISION].includes(propertyTypeId);
 
   const lookupCodes = useCodeLookups();
+  const [privateProject, setPrivateProject] = useState<boolean>(false);
 
   return (
     <>
@@ -118,8 +121,23 @@ export const InfoContent: React.FC<IInfoContent> = ({
             leftSideLabel={'Classification'}
             rightSideItem={propertyInfo?.classification}
           />
+          {!!propertyInfo?.projectNumbers?.length && (
+            <ThreeColumnItem
+              leftSideLabel={
+                propertyInfo.projectNumbers.length > 1 ? 'Project Numbers' : 'Project Number'
+              }
+              rightSideItem={propertyInfo.projectNumbers.map((projectNum: string) => (
+                <ProjectNumberLink
+                  setPrivateProject={setPrivateProject}
+                  privateProject={privateProject}
+                  agencyId={propertyInfo.agencyId}
+                  projectNumber={projectNum}
+                />
+              ))}
+            />
+          )}
         </OuterRow>
-        {!canViewDetails && (
+        {(!canViewDetails || privateProject) && (
           <ContactSres>
             <em>
               For more information on this property, contact{' '}
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
@@ -126,10 +126,12 @@ export const InfoContent: React.FC<IInfoContent> = ({
               leftSideLabel={`Project Number${propertyInfo.projectNumbers.length > 1 ? 's' : ''}`}
               rightSideItem={propertyInfo.projectNumbers.map((projectNum: string) => (
                 <ProjectNumberLink
+                  key={projectNum}
                   setPrivateProject={setPrivateProject}
                   privateProject={privateProject}
                   agencyId={propertyInfo.agencyId}
                   projectNumber={projectNum}
+                  breakLine
                 />
               ))}
             />
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
@@ -123,9 +123,7 @@ export const InfoContent: React.FC<IInfoContent> = ({
           />
           {!!propertyInfo?.projectNumbers?.length && (
             <ThreeColumnItem
-              leftSideLabel={
-                propertyInfo.projectNumbers.length > 1 ? 'Project Numbers' : 'Project Number'
-              }
+              leftSideLabel={`Project Number${propertyInfo.projectNumbers.length > 1 ? 's' : ''}`}
               rightSideItem={propertyInfo.projectNumbers.map((projectNum: string) => (
                 <ProjectNumberLink
                   setPrivateProject={setPrivateProject}
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
@@ -1,0 +1,88 @@
+import Claims from 'constants/claims';
+import { Workflows } from 'constants/workflows';
+import { fetchProject, IProject } from 'features/projects/common';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+import React, { useEffect, useState } from 'react';
+import { useDispatch } from 'react-redux';
+import { Link } from 'react-router-dom';
+
+interface IProjectNumberLinkProps {
+  /** The project number to analyze */
+  projectNumber: string;
+  /** The agency ID of the property being viewed */
+  agencyId: number | '';
+  /** To determine whether the project is private to the user or not */
+  setPrivateProject: Function;
+  /** The result of setPrivateProject used in both the parent and child component */
+  privateProject: boolean;
+}
+
+/**
+ * This component is used to display project numbers and appropriate links if the user has permission.
+ * A user who belongs to the same agency and a project that has not been submitted will grant user access to a link that goes directly to the status.
+ * An admin will have access to a link that goes directly to the status.
+ * If a user belongs to the agency but the project has been submitted it will redirect you to the summary page.
+ * If a user is outside the agency they will not contain a link.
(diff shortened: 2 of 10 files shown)


frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+  agencyId,
+  setPrivateProject,
+  privateProject,
+}) => {
+  const dispatch = useDispatch();
+  const keycloak = useKeycloakWrapper();
+  /** Stores the various states of the project route (summary page, project status route) */
+  const [projectRoute, setProjectRoute] = useState<string>('');
+  /** State used to determine whether further project info has been fetcehd already */
+  const [loaded, setLoaded] = useState<boolean>(false);
```
jdtoombs, line 42: Don't really like doing this, but it was making me add keycloak to the dependency array and that would cause an infinite loop so created a state variable that will execute and determine whether the component still needs to get the project status route

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
```diff
@@ -28,7 +28,7 @@ const RightCol = styled(Col)`
 
 interface IThreeColItem {
   leftSideLabel: string;
-  rightSideItem: string | number | undefined;
+  rightSideItem: string | number | React.ReactNode | undefined;
```
jdtoombs, line 31: So the new `ProjectNumberLink` component can be used

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -113,8 +117,23 @@ export const InfoContent: React.FC<IInfoContent> = ({
             leftSideLabel={'Classification'}
             rightSideItem={propertyInfo?.classification}
           />
+          {!!propertyInfo?.projectNumbers?.length && (
+            <ThreeColumnItem
+              leftSideLabel={
+                propertyInfo.projectNumbers.length > 1 ? 'Project Numbers' : 'Project Number'
```
Fosol, line 123: fancy

Fosol: `'Project Number${propertyInfo.projectNumbers.length && 's'}'`  might be fancier

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -113,8 +117,23 @@ export const InfoContent: React.FC<IInfoContent> = ({
...
+              }
+              rightSideItem={propertyInfo.projectNumbers.map((projectNum: string) => (
+                <ProjectNumberLink
+                  setPrivateProject={setPrivateProject}
```
Fosol, line 127: What is the private project part about?

frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
Fosol: keycloak "shouldn't" cause infinite loop.  But once again the dep array is killing us.  I suspect the deepcompare would solve it, but Devin will know more.

Fosol (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
jdtoombs: Oooh yes I like that lol

frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
jdtoombs: I tried `useDeepCompareEffect` as well.. I must be missing something somewhere. Just not sure of what..

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
jdtoombs: Was just to see if the project was outside of the agency so it can display the contract SRES block in the parent

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
jdtoombs: Gonna steal that and add it in, much cleaner :D

agalbrai: "how unreadable can we make this to normal people for literally no reason" lol

agalbrai (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/InfoContent.tsx
```diff
@@ -113,8 +117,21 @@ export const InfoContent: React.FC<IInfoContent> = ({
             leftSideLabel={'Classification'}
             rightSideItem={propertyInfo?.classification}
           />
+          {!!propertyInfo?.projectNumbers?.length && (
+            <ThreeColumnItem
+              leftSideLabel={`Project Number${propertyInfo.projectNumbers.length > 1 ? 's' : ''}`}
```
devinleighsmith, line 122: I think I'd be fine with just Project Number(s) in the future. I don't think the logic here is worth the overhead.

devinleighsmith: not wrong ofc.

frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
```diff
@@ -0,0 +1,88 @@
...
+      return (
+        <>
+          <Link to={`${projectRoute}?projectNumber=${projectNumber}`}>{projectNumber}</Link>
+          <br />
+        </>
+      );
+    }
+  };
+
+  useEffect(() => {
```
devinleighsmith, line 63: Hey Jonny, I put a comment in your story for this, but I probably should have talked to you about this.
There is an endpoint in the ProjectRouter /projects that already does this logic to figure out the routing for you. ALL external routing to the project logic should use that endpoint to ensure consistency. This includes new functionality, that logic should be updated to support anything rather then adding new paths.

ProjectRouter line 62 has that code. please double check it to make sure that it works for your use case (it should). Should allow you to drastically simplify this component.

Sorry about the communication misfire.

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
devinleighsmith: Good article about passing components around that helped me:
https://medium.com/better-programming/typescript-reactjs-the-element-vs-reactelement-vs-htmlelement-vs-node-confusion-6cda21315ddd

frontend/src/components/maps/leaflet/InfoSlideOut/ProjectNumberLink.tsx
jdtoombs: No problem at all! Thanks for the heads up, will take a look and adjust now

frontend/src/components/maps/leaflet/InfoSlideOut/ThreeColumnItem.tsx
jdtoombs: Thanks! Will give it a read

frontend/src/components/maps/leaflet/MapProperties.test.tsx
```diff
@@ -6,8 +6,6 @@ import Map from './Map';
 import { Map as LeafletMap } from 'leaflet';
```
jdtoombs, line 6: Came across a few unused things in my debugging

frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -349,13 +348,18 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
...
+                m.properties,
+                m.position.lat,
+                m.position.lng,
+              );
               if (
                 m.properties.propertyTypeId === PropertyTypes.PARCEL ||
                 m.properties.propertyTypeId === PropertyTypes.SUBDIVISION
               ) {
-                dispatch(parcelsActions.storeParcelDetail(m.properties as IParcel));
+                dispatch(parcelsActions.storeParcelDetail(convertedProperty as IParcel));
```
jdtoombs, line 360: This was causing the error as when it tried to reload the selected property marker there was no lat/long present

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -244,13 +244,16 @@ export const pointToLayer = (feature: ICluster, latlng: LatLngExpression): Layer
...
 export const getMarkerIcon = (feature: ICluster, selected?: boolean) => {
-  const { propertyTypeId, projectWorkflow, classificationId } = feature?.properties;
+  const { propertyTypeId, projectWorkflow, classificationId, parcelDetail } = feature?.properties;
   if (propertyTypeId === PropertyTypes.DRAFT_PARCEL) {
     return draftParcelIcon;
   } else if (propertyTypeId === PropertyTypes.DRAFT_BUILDING) {
     return draftBuildingIcon;
   } else if (selected) {
-    if ([Workflows.ERP, Workflows.ASSESS_EX_DISPOSAL].includes(projectWorkflow)) {
+    if (
```
jdtoombs, line 253: There's something funky going on with selecting markers, the shape of the data is inconsistent, spent some time debugging but need to work on my other tasks. This does "solve" the issue though. `parcelDetail` is normally stored on marker click, but for some reason selecting certain markers they instantly deselect (even though select is set to true) so it would then have to look at the `projectWorkflow` instead of the `parcelDetail`. Creating a separate bug as these were found during my story to add project number to the panel..

agalbrai: mikaela noticed that when clicking on a pin it will change to blue/green no matter what colour it used to be. Would this fix that?

jdtoombs: Yes this will! @agalbrai

PR #958, opened 2021-02-19
PA 2562 Fix inventory list header alignment

agalbrai:
Pictures of a couple different screen sizes for reference:
image
image


frontend/src/features/properties/list/columns.tsx
```diff
@@ -239,7 +239,7 @@ export const columns = (
     },
   },
   {
-    Header: 'Lot Size (in ha)',
+    Header: 'Lot Size (in\u00A0ha)',
```
agalbrai, line 242: unicode for non-breaking space

devinleighsmith:
In the image reference the top image appears to be missing sort arrows between classification and type

agalbrai:
> In the image reference the top image appears to be missing sort arrows between classification and type

You're right, I'll fix that

goemen (approved):

Fosol (approved):

PR #959, opened 2021-02-19
PA 2607 Align SPL report headers and fields

agalbrai:
Rather than try to build some workaround to get only the headers to align right, I just aligned the columns left
image

### DIFF
--- a/frontend/src/components/Table/Table.tsx
+++ b/frontend/src/components/Table/Table.tsx
@@ -499,8 +499,8 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
               >
                 {actions => (
                   <>
-                    <div className={'th reset-filter svg-btn'}>
-                      {filterable && (
+                    {filterable && (
+                      <div className={'th reset-filter svg-btn'}>
                         <TooltipWrapper
                           toolTipId="properties-list-filter-reset-tooltip"
                           toolTip="Reset Filter"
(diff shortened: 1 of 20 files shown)


devinleighsmith:
The approach makes sense to me, just have 2 questions:
1) It looks like the th have some padding or something that is messing with the left alignment, can we fix that?
2) do @bbjornholt or @Fosol have any concerns with this? I believe they have driven conversations about column alignment for financials in the past.

Fosol (approved):

devinleighsmith:
actually, it seems like the columns get pushed left slowly after each column? thats odd.

agalbrai:
> The approach makes sense to me, just have 2 questions:
> 
> 1. It looks like the th have some padding or something that is messing with the left alignment, can we fix that?
> 2. do @bbjornholt or @Fosol have any concerns with this? I believe they have driven conversations about column alignment for financials in the past.
From what I can see there's an extra column on the far left in the header, and I have tried everything I can think of but I can't find where it's coming from or how to remove it

devinleighsmith:
Simple solution may be to just hide the reset-filter column in the table:
image

Also, while you are in there something seems to have broken the resize logic on the top bar, on small screens that bar is being cut off.

agalbrai:
> Simple solution may be to just hide the reset-filter column in the table:
> image
> 
> Also, while you are in there something seems to have broken the resize logic on the top bar, on small screens that bar is being cut off.

I'm confused as to which bar you mean, could you highlight it for me?

jdtoombs (approved):

PR #964, opened 2021-02-20
PA-2816, PA-2815

devinleighsmith:
Ensure that all incoming subdivisions are:
1) given the proper property type
2) use the pid of the first passed in parcel
3) generate a unique pin when a new subdivision is being added.
4) don't display pid/pin on frontend


backend/api/Areas/Property/Mapping/Parcel/ParcelMap.cs
```diff
@@ -67,6 +67,7 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Id, src => src.Id)
                 .Map(dest => dest.PID, src => ParcelConverter.ConvertPID(src.PID))
                 .Map(dest => dest.PIN, src => src.PIN)
+                .Map(dest => dest.PropertyTypeId, src => src.PropertyTypeId)
```
devinleighsmith, line 70: This is for an edge case - it is currently possible to create a subdivision with no parent by creating a subdivision and then deleting the parent. In that case, this logic will ensure that a subdivision is not converted to a parcel during an update just because it doesn't have a parent parcel.

backend/api/Areas/Property/Mapping/Parcel/ParcelParcelMap.cs
```diff
@@ -53,6 +48,7 @@ public void Register(TypeAdapterConfig config)
             config.NewConfig<Model.SubdivisionParcelModel, Entity.ParcelParcel>()
                 .EnableNonPublicMembers(true)
                 .Map(dest => dest.ParcelId, src => src.Id)
+                .Map(dest => dest.Parcel.PID, src => ParcelConverter.ConvertPID(src.PID))
```
devinleighsmith, line 51: We need to pass through the parent parcel pid so that it can be used in place of whatever pid is currently associated to the subdivision.

backend/dal/Helpers/Extensions/ParcelExtensions.cs
```diff
@@ -190,6 +190,19 @@ public static IQueryable<Entity.Parcel> GenerateQuery(this PimsContext context,
             return parcel?.Id;
         }
 
+        /// <summary>
+        /// Get unique a pin that will uniquely identify a parcel using a pre-existing pid.
+        /// </summary>
+        /// <param name="existingPid"></param>
+        /// <returns></returns>
+        public static int GetUniquePidPin(this PimsContext context, int existingPid)
```
devinleighsmith, line 198: This should ensure that a new unique PIN is generated when a subdivision is added with a pre-existing PID. Will leave unused pins if a subdivision is deleted, and start counting from a high number if a parcel already has a pin but I think that both of these things should be fine.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -283,6 +290,14 @@ public Parcel PendingUpdate(Parcel parcel)
             parcel.ThrowIfNotAllowedToEdit(nameof(parcel), this.User, new[] { Permissions.PropertyEdit, Permissions.AdminProperties });
             var isAdmin = this.User.HasPermission(Permissions.AdminProperties);
 
+            parcel.PropertyTypeId = (int)(parcel.Parcels.Count > 0 ? PropertyTypes.Subdivision : PropertyTypes.Land);
```
devinleighsmith, line 293: Need the same logic for updates. the pid on the subdivision parcel should always be replaced by whatever is on the parent parcel.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -181,6 +181,13 @@ public Parcel Add(Parcel parcel)
         {
             parcel.ThrowIfNull(nameof(parcel));
             parcel.PropertyTypeId = (int)(parcel.Parcels.Count > 0 ? PropertyTypes.Subdivision : PropertyTypes.Land);
+            if(parcel.PropertyTypeId == (int)PropertyTypes.Subdivision)
+            {
+                var parentPid = parcel.Parcels.FirstOrDefault()?.Parcel?.PID;
+                if (parentPid == null) throw new InvalidOperationException("Invalid parent parcel associated to subdivision, parent parcels must contain a valid PID");
+                parcel.PID = parentPid.Value;
```
devinleighsmith, line 188: use the pid of the first parent parcel.

frontend/src/features/mapSideBar/SidebarContents/subforms/ParcelIdentificationForm.tsx
```diff
@@ -113,12 +113,14 @@ export const ParcelIdentificationForm: React.FC<IIdentificationProps> = ({
           <h5>Parcel Details</h5>
         </Col>
         <Col md={6}>
-          <PidPinForm
-            nameSpace={nameSpace}
-            handlePidChange={noop}
-            handlePinChange={noop}
-            disabled={disabled}
-          />
+          {propertyTypeId !== PropertyTypes.SUBDIVISION && (
```
devinleighsmith, line 116: Since the api handles supplying the subdivision with the correct PID/PIN, simply hide those form controls.

Fosol (approved):

goemen (approved):

agalbrai (commented):
looks good!

PR #973, opened 2021-02-23
Separate filter form and body form in table

goemen:
- Fix financial value filters and adding tooltips


frontend/src/components/Table/Table.tsx
```diff
@@ -498,7 +512,7 @@ const Table = <T extends object>(props: PropsWithChildren<TableProps<T>>): React
                 innerRef={filterFormRef as any}
               >
                 {actions => (
-                  <>
+                  <Form>
```
agalbrai, line 515: How does this change the component? Does it add behaviour of some kind?

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -490,113 +501,135 @@ const PropertyListView: React.FC = () => {
...
+                const changedRows = dirtyRows.map(change => {
+                  const data = { ...values.properties![change.rowId] };
+
+                  return { rowId: change.rowId, data } as any;
+                });
+                if (changedRows.length > 0) {
+                  for (const change of changedRows) {
+                    const apiProperty = toApiProperty(change.data as any, true);
+                    const callApi = apiProperty.parcelId ? updateParcel : updateBuilding;
+                    const response: any = await callApi(apiProperty.id, apiProperty);
```
devinleighsmith, line 608: does this perform ok if the user changes a relatively large number of rows at the same time? What happens if one of the requests fails?

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -490,113 +501,135 @@ const PropertyListView: React.FC = () => {
...
+          }}
+          renderBodyComponent={({ body }) => (
+            <Formik
+              innerRef={tableFormRef as any}
+              initialValues={{ properties: data || [] }}
+              enableReinitialize
+              onSubmit={async values => {
+                let nextProperties = [...values.properties];
+                const changedRows = dirtyRows.map(change => {
+                  const data = { ...values.properties![change.rowId] };
```
devinleighsmith, line 600: this works properly with paging, right?

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -490,113 +501,135 @@ const PropertyListView: React.FC = () => {
...
+            onExpand: loadBuildings,
+            getRowId: row => row.id,
+          }}
+          filter={appliedFilter}
+          onFilterChange={values => {
+            setFilter({ ...filter, ...values });
+          }}
+          renderBodyComponent={({ body }) => (
+            <Formik
+              innerRef={tableFormRef as any}
```
devinleighsmith, line 594: to we not have any validation on this form? note that while chrome doesn't allow you to type non-numeric characters into numeric fields other browsers will.

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -490,113 +501,135 @@ const PropertyListView: React.FC = () => {
...
-                } else {
-                  return false;
-                }
-              }}
-              detailsPanel={{
-                render: val => {
-                  if (expandData[val.id]) {
-                    return <Buildings hideHeaders={true} data={expandData[val.id]} />;
+          canRowExpand={(val: any) => {
+            if (val.values.propertyTypeId === 0) {
```
devinleighsmith, line 568: This probably needs to include subdivisions now.

frontend/src/features/properties/list/PropertyListView.tsx
```diff
@@ -490,113 +501,135 @@ const PropertyListView: React.FC = () => {
...
+          }}
+          filter={appliedFilter}
+          onFilterChange={values => {
+            setFilter({ ...filter, ...values });
+          }}
+          renderBodyComponent={({ body }) => (
+            <Formik
+              innerRef={tableFormRef as any}
+              initialValues={{ properties: data || [] }}
+              enableReinitialize
```
devinleighsmith, line 596: probably will want to make sure that leaving the enableReinitialize isn't refreshing the form if a user is in edit mode for  a while ( a few minutes) also want to make sure that changes in the filter bar affect edit mode in a reasonable way.

devinleighsmith (approved):
Approving, but I left some comments on areas where I'm concerned about possible issues.

frontend/src/features/properties/list/PropertyListView.tsx
Fosol: Will subdivisions have buildings on them?

Fosol: Don't we have an `enum` for `PropertyTypes`?

frontend/src/features/properties/list/PropertyListView.tsx
Fosol: You need to handle if one or more of the requests fail

Fosol (approved):

frontend/src/components/Table/Table.tsx
goemen: This allows to submit the form by clicking ENTER

frontend/src/features/properties/list/PropertyListView.tsx
goemen: Only the current page is part of the form, yes is works with paging

frontend/src/features/properties/list/PropertyListView.tsx
goemen: Added a toast display and added row fields validation error display

frontend/src/features/properties/list/PropertyListView.tsx
goemen: Added validation

frontend/src/features/properties/list/PropertyListView.tsx
goemen: @devinleighsmith can you verify if this change need to be made, not too familiar with subdivisions in the table?

frontend/src/features/properties/list/PropertyListView.tsx
goemen: This seems to be not a problem

PR #974, opened 2021-02-23
SPL Filter now working correctly 

jdtoombs:
Mikaela made a comment that the SPL filter was returning SPP results. Renamed the SPP filter accordingly and it now only return SPL properties. Backend now reflects frontend!

Also my [old PR that has been approved](https://github.com/bcgov/PIMS/pull/956) has been updated and I have re-requested some reviews as there has been some changes.

splfix


backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -95,8 +95,6 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
                 query = query.Where(p => p.ProjectNumbers.Contains(filter.ProjectNumber));
             if (filter.IgnorePropertiesInProjects == true)
                 query = query.Where(p => p.ProjectNumbers == null || p.ProjectNumbers == "[]");
-            if (filter.InSurplusPropertyProgram == true)
```
jdtoombs, line 98: I think at one point this was the desired filter but it evolved over time and only got changed on the FE

goemen (approved):
Looks good

frontend/src/components/common/form/SppButton.tsx
```diff
@@ -29,7 +29,7 @@ const SppButton: React.FC<ISppButtonProps> = ({
 
   const iconColor = inEnhancedReferralProcess
```
agalbrai, line 30: I think this carried over from the old button... The colour doesn't actually change anymore

agalbrai: I just checked, also the tooltip doesn't work anymore

agalbrai (approved):
Thank goodness for "Find and Replace" haha

frontend/src/components/common/form/SppButton.tsx
jdtoombs: Oh you're right, I think I vaguely remember the old button lol, looks like it is not used anywhere but the test now! Will remove

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -159,6 +157,18 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
                             && property.ProjectNumbers.Contains(project.ProjectNumber)));
             }
 
+            if (filter.InSurplusPropertyList.HasValue && filter.InSurplusPropertyList.Value)
```
Fosol, line 160: I think you can just go 
```
filter.InSuplusPropertyList == true
```

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -159,6 +157,18 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
...
+            {
+                var statuses = context.Workflows.Where(w => w.Code == "SPL")
+                    .SelectMany(w => w.Status).Where(x => !x.Status.IsTerminal)
+                    .Select(x => x.StatusId).Distinct().ToArray();
+
+                query = query.Where(property =>
```
Fosol, line 166: This is a very expensive query.   If when we search for ERP or SPL I think we should invert the original query.  Start with projects and return properties form that list.  Then we can apply the other filter values to the properties.

jdtoombs: Ahh ok, I'll switch the ERP one as well

PR #977, opened 2021-02-24
PA-2691 PA-2692

devinleighsmith:
Add Delete/Deny logic for subdivisions, with warning messages on the FE.


backend/api/Areas/Project/Controllers/DisposeController.cs
```diff
@@ -143,8 +143,8 @@ public async Task<IActionResult> UpdateProjectAsync(string projectNumber, Projec
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0060:Remove unused parameter", Justification = "To support standardized routes (/{projectNumber})")]
         public async Task<IActionResult> DeleteProjectAsync(string projectNumber, ProjectModel model)
         {
-            await _pimsService.Project.RemoveAsync(_mapper.Map<Entity.Project>(model));
-            return new JsonResult(model);
+            var project = await _pimsService.Project.RemoveAsync(_mapper.Map<Entity.Project>(model));
+            return new JsonResult(_mapper.Map<ProjectModel>(project));
```
devinleighsmith, line 147: The delete project was an interesting case. Didn't make sense to include all properties on a project within the project list view, which would provide the required metadata to know a subdivision exists on a project.

So instead what I went with was updating the delete api to return the properties. This allows the FE to display a warning after the delete occurs, which I think is fine in this case.

frontend/src/features/projects/assess/components/ReviewApproveActions.test.tsx
```diff
@@ -36,6 +38,7 @@ describe('Review Approve actions', () => {
   describe('approval button', () => {
     (useFormikContext as jest.Mock).mockReturnValue({
```
devinleighsmith, line 39: I want to comment on this, because I don't think this works the way the author thought it does.

Anything not included in a before() after() it or similar block runs simultaneously (not only during the test) In addition, mocking a value is a global action. 

This test mocks useFormikContext multiple times. All of those mocks affect all of the tests if you want to add something to the mocked formik object, you need to add id to all instances of this mock.

I have plans to circle back and fix this test, but just sharing what I learned here.

frontend/src/features/projects/assess/components/ReviewApproveActions.tsx
```diff
@@ -108,9 +114,21 @@ export const ReviewApproveActions = ({
...
-                : 'Are you sure you want to deny this project with the request for exemption? Please ensure to provide reasoning in the shared notes prior to clicking deny.'
+              <>
+                {!values.exemptionRequested
+                  ? `Are you sure you want to deny the project for Enhanced
+                Referral Process? Please ensure to provide reasoning in the shared notes prior to
+                clicking deny.`
+                  : `Are you sure you want to deny this project with the request for
+                exemption? Please ensure to provide reasoning in the shared notes prior to clicking
+                deny.`}
+                {hasSubdivisions && (
```
devinleighsmith, line 125: when denying, display the subdivision warning in the original warning message. We could decide to display this after the deny if we wanted, I just wanted to keep the # of clicks down.

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -97,14 +106,18 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
   const agencies = useMemo(() => lookupCodes.getByType(API.AGENCY_CODE_SET_NAME), [lookupCodes]);
   const projectStatuses = useSelector<RootState, IStatus[]>(state => state.statuses as any);
   const keycloak = useKeycloakWrapper();
-  const [deleteId, setDeleteId] = React.useState<string | undefined>();
+  const [deleteProjectNumber, setDeleteProjectNumber] = React.useState<string | undefined>();
```
devinleighsmith, line 109: not actually changing how this works, just didn't like the variable name.

frontend/src/features/projects/common/strings.ts
```diff
@@ -49,3 +49,6 @@ export const appraisalDateWarning =
   'You have entered an appraisal date that is older then the current appraisal date stored in PIMS. Are you sure that you would like to update PIMS to use this older date?';
 export const tabErrorWarning =
   'Please correct all of the highlighted errors within the red coloured tab(s) before continuing.';
+export const deleteProjectWarning = 'Are you sure that you want to delete this project?';
+export const deletePotentialSubdivisionParcels =
+  'There are Potential Subdivision Parcels associated with this project in the PIMS Inventory. Please delete or re-use for another Disposal Project.';
```
agalbrai, line 54: This wording may be a bit ambiguous, when I read this I'm not sure if the second sentence is asking me to delete/re-use the project or the subdivision(s)

frontend/src/features/projects/list/ProjectListView.tsx
```diff
@@ -97,14 +106,18 @@ const ProjectListView: React.FC<IProps> = ({ filterable, title, mode }) => {
...
+  const [deletedProject, setDeletedProject] = React.useState<IProjectDetail | undefined>();
   const agencyIds = useMemo(() => agencies.map(x => parseInt(x.id, 10)), [agencies]);
   const agencyOptions = (agencies ?? []).map(c => mapLookupCodeWithParentString(c, agencies));
   const statuses = (projectStatuses ?? []).map(c => mapStatuses(c));
   const columns = useMemo(() => cols, []);
 
   // We'll start our table without any data
   const [data, setData] = useState<IProject[] | undefined>(undefined);
+  const hasSubdivisions = deletedProject?.properties?.some(
```
agalbrai, line 118: Looked into the "some" function, neat!

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -555,19 +555,18 @@ public async System.Threading.Tasks.Task<Project> UpdateAsync(Project project)
...
         {
             project.ThrowIfNotAllowedToEdit(nameof(project), this.User, new[] { Permissions.ProjectDelete, Permissions.AdminProjects });
 
             var userAgencies = this.User.GetAgencies();
             var isAdmin = this.User.HasPermission(Permissions.AdminProjects);
             var originalProject = this.Context.Projects
                 .Include(p => p.Status)
-                .Include(p => p.Properties)
-                .ThenInclude(p => p.Parcel)
+                .Include(p => p.Properties).ThenInclude(p => p.Parcel).ThenInclude(p => p.Parcels)
```
Fosol, line 566: Why do you need to include `Parcels`?  Where you didn't need to include `Buildings` below?

Fosol (approved):

backend/dal/Services/Concrete/ProjectService.cs
devinleighsmith: this is to get the parent parcels of subdivisions, not relevant for buildings.

frontend/src/features/projects/common/strings.ts
devinleighsmith: This is just copy/pasted from the story.

jdtoombs (approved):

PR #979, opened 2021-02-24
PA-2825 Split Assessed into two values

Fosol:
Doesn't fix editing/display issue

image

image


frontend/src/components/Table/MoneyCell.tsx
```diff
@@ -39,4 +50,41 @@ export const EditableMoneyCell = ({
   );
 };
 
-export const MoneyCell = ({ cell: { value } }: CellProps<IProperty, number>) => formatMoney(value);
+/**
+ * Provides a currency editing input if the user is allowed to edit the property and if the property is land/subdivision.
+ * @param param0 Configuration properties for this editable money cell.
+ */
+export const EditableLandMoneyCell = ({
```
Fosol, line 57: This was required to handle when a building row shouldn't allow editing of `Land Assessment`

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -184,15 +184,11 @@ const getApiEvaluations = (property: IProperty): IEvaluation[] => {
...
-    date: isParcelOrSubdivision(property)
-      ? property.assessedLandDate ?? formatDate(new Date())
-      : property.assessedBuildingDate ?? formatDate(new Date()),
-    rowVersion: property.assessedLandRowVersion,
+    value: property.assessedBuilding || 0,
+    date: property.assessedBuildingDate ?? formatDate(new Date()),
+    rowVersion: property.assessedBuildingRowVersion,
     key: EvaluationKeys.Improvements,
-    firm: property.assessedLandFirm ?? '',
+    firm: property.assessedBuildingFirm ?? '',
```
Fosol, line 191: An `Improvement` should always be `assessedBuilding*`.  @devinleighsmith I'm not certain this is implemented correctly through the application.

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -236,7 +232,7 @@ export const toApiProperty = (
       {
         parcelId: isParcelOrSubdivision(property) ? property.id : undefined,
         buildingId: property.propertyTypeId === PropertyTypes.BUILDING ? property.id : undefined,
-        value: property.netBook,
+        value: property.netBook || 0,
```
Fosol, line 235: Without this it will attempt to return `""` back to be deserialized which will result in a failure in the API.

frontend/src/features/projects/common/projectConverter.ts
```diff
@@ -246,7 +242,7 @@ export const toApiProperty = (
       {
         parcelId: isParcelOrSubdivision(property) ? property.id : undefined,
         buildingId: property.propertyTypeId === PropertyTypes.BUILDING ? property.id : undefined,
-        value: property.market,
+        value: property.market || 0,
```
Fosol, line 245: Without this it will attempt to return `""` back to be deserialized which will result in a failure in the API.

Fosol:
@devinleighsmith This PR still requires you to resolve the Assessed values in pa-2667

frontend/src/features/projects/common/projectConverter.ts
devinleighsmith: In my PR I made the assumption that improvements were specific to land, and that a regular assessment evaluation on a building would be the assessedbuilding.

devinleighsmith: sounds like the opposite is true, that an improvement is always an assessedbuilding and that buildings aren't allowed to use EvaluationKeys.assessed

devinleighsmith: My assumption was based around how assessments for buildings previously worked. Did we create a migration to move all of the old building assess values over to improvements?

frontend/src/components/Table/MoneyCell.tsx
```diff
@@ -39,4 +50,41 @@ export const EditableMoneyCell = ({
...
+  return canEdit ? (
+    <FastCurrencyInput
+      formikProps={context}
+      suppressValidation={suppressValidation}
+      field={`${namespace}.${cell.row.id}.${cell.column.id}`}
+    ></FastCurrencyInput>
+  ) : (
+    <TooltipWrapper
+      toolTipId={`${namespace}.${cell.row.id}.${cell.column.id}`}
+      toolTip="You may only edit a property if it belongs to your agency and it is not part of an SPP Project."
```
agalbrai, line 78: should this tooltip also mention that it can't be edited because it's a building on land?

frontend/src/utils/numberFormatUtils.ts
```diff
@@ -1,13 +1,23 @@
...
     minimumFractionDigits: 0,
     maximumFractionDigits: 0,
   });
-  return formatter.format(value || 0);
+  return typeof value === 'undefined' || value === '' ? '' : formatter.format(value || 0);
 };
 
+/**
+ * Formats the specified 'value' as US currency.
+ * @param value The valut to format as money.
```
agalbrai, line 18: nitpicky but "valut"

PR #980, opened 2021-02-24
remove oc check as it breaks the jenkins slaves.

devinleighsmith:


devinleighsmith:
Currently the OCP 4 pipeline fails consistently because the jenkins slaves can't run OC commands. Those slaves don't actually need to run OC commands so this check isn't necessary for them.

The downside of removing this is that if the master jenkins instance loses oc connectivity we will get less informative error messages, but I don't have a solution for that at the moment. Tech Debt.

Fosol (approved):

devinleighsmith:
@goemen not sure how you were able to get the pipeline to pass with this issue, if you have an alternative solution let me know. Be aware that the currently released version of the pipeline is failing due to this issue.

PR #982, opened 2021-02-24
Added separate deploy configs

Fosol:
Added separate route deployment configuration files for the API and APP.  These are not required, as they already exist in the current deployment configuration files.  They make it easier to remove/replace/update the routes independent of all the other objects.

### DIFF
--- a/openshift/4.0/templates/api/deploy-alt-routes.yaml
+++ b/openshift/4.0/templates/api/deploy-alt-routes.yaml
@@ -1,0 +1,188 @@
+kind: Template
+apiVersion: v1
+metadata:
+  name: pims-api-deploy
+  annotations:
+    openshift.io/display-name: .NET Core 3.1 Deployment Template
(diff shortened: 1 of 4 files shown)


openshift/4.0/templates/api/deploy-alt-routes.yaml
```diff
@@ -0,0 +1,188 @@
+kind: Template
```
Fosol, line 1: Alternative routes for the production API.

openshift/4.0/templates/app/deploy-alt-routes.yaml
```diff
@@ -0,0 +1,132 @@
+apiVersion: v1
```
Fosol, line 1: Alternative routes for the application in prod.

jdtoombs (approved):

devinleighsmith (approved):

PR #984, opened 2021-02-24
Added Jenkins RoleBinding Config

Fosol:
The rolebinding configuration was missing.  I have added it but am uncertain if it is correct.


jdtoombs (approved):

openshift/4.0/templates/jenkins/rolebinding.yaml
```diff
@@ -0,0 +1,75 @@
+kind: "Template"
```
devinleighsmith, line 1: looks correct to me.

devinleighsmith (approved):

PR #985, opened 2021-02-24
PA-2830: Change "Submit Bare Land" to "Submit Land"

LawrenceLau2020:
replacing "bare land" with "land"

### DIFF
--- a/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
+++ b/frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
@@ -156,7 +156,7 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
           <span style={{ marginTop: 5, marginLeft: 5, marginRight: 5 }}>-</span>
           <NumberInput field="maxLotSize" placeholder="max" />
           <VerticalLine />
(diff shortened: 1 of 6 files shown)


frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
```diff
@@ -156,7 +156,7 @@ const FindMorePropertiesForm = <T extends any>(props: any) => {
           <span style={{ marginTop: 5, marginLeft: 5, marginRight: 5 }}>-</span>
           <NumberInput field="maxLotSize" placeholder="max" />
           <VerticalLine />
-          <Check label="Bare land only" field="bareLandOnly" />
+          <Check label="Land only" field="bareLandOnly" />
```
agalbrai, line 159: I'm not sure if this is right, does the search only return bare land @jdtoombs ?

agalbrai (approved):

frontend/src/components/SearchBar/FindMorePropertiesForm.tsx
Fosol: I believe that's the purpose of this feature.

Fosol (approved):

PR #987, opened 2021-02-25
Updated SPL/ERP Query

Fosol:
This query will be more efficient and should correctly return properties in ERP and/or SPL.

Jonny, please test once merged to determine if it handles all scenarios.


jdtoombs (approved):
Thank you :) - got a little more complex than I had expected!

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -167,6 +155,56 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
             return query;
         }
 
+        /// <summary>
+        /// Generates a query that returns properties in ERP and/or SPL.
+        /// Note - this does not filter properties that a user shouldn't not be able to view.
```
agalbrai, line 160: nitpicky but "shouldn't not"

PR #989, opened 2021-02-26
PA-2879 Fixed Deny Project Validation

Fosol:
image

### DIFF
--- a/frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
+++ b/frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
@@ -7,6 +7,7 @@ import {
   ProjectDraftStepYupSchema,
   SelectProjectPropertiesStepYupSchema,
   ApproveExemptionRequestSchema,
+  DenyProjectYupSchema,
 } from '../../dispose/forms/disposalYupSchema';
 import { fetchProjectTasks } from '../../common/projectsActionCreator';
 import _ from 'lodash';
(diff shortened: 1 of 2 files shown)


frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
```diff
@@ -46,6 +47,35 @@ export const validateTasks = (project: IProject) => {
...
+export const validateDeny = (project: IProject) => {
+  return new Promise(resolve => {
+    validateYupSchema(project, DenyProjectYupSchema).then(
+      () => resolve({}),
+      (errors: any) => resolve(yupToFormErrors(errors)),
+    );
+  });
+};
+
+export const validateApprove = (project: IProject) => {
```
devinleighsmith, line 63: I'm not sure this will work properly. The validateYupSchema method here will run async, so the overall validateApprove method will return the yupErrors and never wait for the promise to resolve. There is a validatesync variant you could use instead I believe. Or just switch to using an await.

Fosol: It does work because of the `Promise(resolve => ...)`, however I can rewrite with an `await`

jdtoombs (approved):

frontend/src/features/projects/assess/steps/ReviewApproveStep.tsx
```diff
@@ -46,6 +51,36 @@ export const validateTasks = (project: IProject) => {
...
+  } catch (errors) {
+    return Promise.resolve(yupToFormErrors(errors));
+  }
+};
+
+/**
+ * Validate the approve project action.
+ * @param project The project to validate.
+ */
+export const validateApprove = async (project: IProject) => {
```
devinleighsmith, line 71: looks good.

devinleighsmith (approved):

PR #992, opened 2021-02-26
enhance error modal to allow user to go back to home page or reload c…

devinleighsmith:
…urrent page.

May need better repro instructions, as I wasn't seeing what was posted in the jira.

The jira states:

Steps to Reproduce

Login as Real Estate Manager

Enter the following URL with a non-existing project number https://pims.apps.silver.devops.gov.bc.ca/dispose/projects/documentation?projectNumber=DRAFT-00034

this gives me the following:

image

this is an ok (not great message) but clicking ok will just try and reload the page which will never work.

I added a button to the generic error modal that allows the user to navigate back to the home page. That should allow them to recover from routes like this.


devinleighsmith:
image

updated modal

Fosol (approved):
The repo instructions provided a shortcut way to replicate.  The real world example is when a Disposal Project is deleted but the user was last viewing the project.  Which means when the user opens PIMS they are immediately redirected to the delete project.  Which throws the error and has no menu navigation on the page.  Screenshot was included.  The user is essentially stuck

jdtoombs (approved):

devinleighsmith:
> The repo instructions provided a shortcut way to replicate. The real world example is when a Disposal Project is deleted but the user was last viewing the project. Which means when the user opens PIMS they are immediately redirected to the delete project. Which throws the error and has no menu navigation on the page. Screenshot was included. The user is essentially stuck

Can you do me a favour and try and repro this again then @Fosol ? I tried both the original repro instructions as well as what you describe above, don't see the blank screen in either case.

agalbrai (approved):

PR #998, opened 2021-02-26
Make properties not viewable after project termination

devinleighsmith:


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -279,6 +281,17 @@ public static void ReleaseProjectProperties(this PimsContext context, Entity.Pro
             project.Properties.ForEach(p =>
             {
                 context.Update(p.RemoveProjectNumber(project.ProjectNumber));
+                switch (p.PropertyType)
+                {
+                    case (Entity.PropertyTypes.Land):
+                        if (p.Parcel == null) throw new InvalidOperationException("Unable to update parcel status.");
+                        p.Parcel.IsVisibleToOtherAgencies = false;
```
devinleighsmith, line 288: Oh crap, but properties can be in multiple projects now. Probably need to check that before setting this.

Fosol: This property may not even be relevant anymore.  I'd need to review again, but I don't think we even check it anymore when determining whether to return properties.  The main reason is all properties (other than sensitive) are now visible to all agencies.

Fosol (approved):

jdtoombs (approved):

goemen (approved):

PR #1000, opened 2021-02-26
PA-2752 move the exemption request.

devinleighsmith:
I'll admit that I'm not very familiar with the exemption process/business rules.

From a code perspective moving the component and adding the validation seemed to be all that was required.


Fosol (approved):
Yes it was only a move required

goemen (approved):

PR #1002, opened 2021-02-28
Adding direct link to project

jdtoombs:
Just needed to direct the user to the project route with the appropriate project number and Devin's changes to the `ProjectRouter` will handle the rest. Putting up as draft for now so I can test the link (need to get open-id token working.


Fosol (approved):
I've tested locally and it works.

devinleighsmith (approved):

PR #1003, opened 2021-03-01
PA-2823 property list enhancements.

devinleighsmith:
Need to run some of this by Mikaela, as this was based on changes requested verbally, don't want to release this and then have it kicked back for not matching a design that I was told was no longer desirable.


frontend/src/components/Table/MoneyCell.tsx
```diff
@@ -25,11 +25,13 @@ export const EditableMoneyCell = ({
   const canEdit = useKeycloakWrapper().canUserEditProperty(property);
 
   return canEdit ? (
-    <FastCurrencyInput
-      formikProps={context}
-      suppressValidation={suppressValidation}
-      field={`${namespace}.${cell.row.id}.${cell.column.id}`}
-    ></FastCurrencyInput>
+    <>
```
devinleighsmith, line 28: need to remove these.

frontend/src/features/properties/list/PropertyListView.scss
```diff
@@ -82,5 +82,8 @@ $contentheading-height: 60px;
     table {
       background-color: #fff;
     }
+    .table {
```
devinleighsmith, line 85: all that is required for x scroll

frontend/src/features/properties/list/buildings/Buildings.tsx
```diff
@@ -12,21 +11,31 @@ export interface IProps {
...
+  hideHeaders,
+  columns: externalColumns,
+  onRowClick,
+}) => {
   return (
     <Table<IProperty>
       hideHeaders={hideHeaders}
       name="nestedPropertiesTable"
-      columns={columns}
+      columns={externalColumns}
```
devinleighsmith, line 34: Yeah, so having a different set of columns for a table isn't great for % based columns, it makes it very hard to get the columns to align. If this was a html based table it would also be invalid html, but our react tables aren't actually html tables so its fine.

Anyways, updated this to allow the parent to pass in the columns. This allows the columns to be the same for the parent and child in the case of nested building rows.

Fosol (approved):

goemen (approved):
Looks good to me

jdtoombs (approved):
Thank you for doing this Devin, helped me out a ton on Friday.

PR #1004, opened 2021-03-01
PA-2848 Fixed Not in SPL Workflow

Fosol:
Had issues with exempt projects moving through the workflow.

Added DB migration `v01.08.07`


backend/dal/Migrations/v01.07.00/Down/PreDown/01-RemoveNewClassifications.sql
```diff
@@ -1 +1,10 @@
+-- Migration any properties to Disposed Classifications
+UPDATE dbo.[Parcels]
```
Fosol, line 2: Fixed issue with rollback.

jdtoombs (approved):
Just merged my migration so you should be good to go!

goemen (approved):
Looks good to me

PR #1011, opened 2021-03-03
PA-2829 Import Project Location

Fosol:
When importing projects from SIS data it will copy the `Location` information into the `Description`.


jdtoombs (approved):

backend/api/Areas/Tools/Helpers/ImportProjectsHelper.cs
```diff
@@ -173,7 +173,7 @@ private Entity.Project Merge(Entity.Project project, Model.ImportProjectModel mo
 
             project.ProjectNumber = String.IsNullOrWhiteSpace(model.ProjectNumber) ? $"TEMP-{DateTime.UtcNow.Ticks:00000}" : model.ProjectNumber;
             project.Name = model.Description.Truncate(100);
-            project.Description = model.Description;
+            project.Description = model.Description + (String.IsNullOrWhiteSpace(model.Location) ? null : Environment.NewLine + model.Location);
```
devinleighsmith, line 176: forget how c# appends strings. Just wanting to confirm that "test" + null doesn't return something awful like "testnull"

maybe for safety this should be "" instead of null?

devinleighsmith (approved):

PR #1013, opened 2021-03-03
enhance error handling so that async errors trigger the error boundary.

devinleighsmith:
Related to PA-2915, the map side bare isn't displaying errors nicely, this change allows the side bar to use our react error boundary to display errors, which allows the user to reload the page or navigate back to the home page.


frontend/src/hooks/useAsyncError.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * allow an async action to throw a regular error that will be caught be our react error boundary. example:
+ * const throwError useAsyncError();
+ * try{
+ *  await someAsyncFunction();
+ * } catch(error) {
+ *  throwError(error);
+ * }
+ */
+export const useAsyncError = () => {
```
devinleighsmith, line 12: Note, this should only be used on errors that leave the page in a non-usable. Ie. if the property list view page fails to load property data, or similar.

Fosol: Is `Async` relevant?  Seems to me that it's just throwing an `any` error.
I also don't understand the use of the state here?  It looks like you're returning a function that calls the set state, which throws the error.

Fosol (approved):

frontend/src/features/mapSideBar/SidebarContents/AssociatedLandForm.tsx
```diff
@@ -56,7 +56,7 @@ const FormContentWrapper = styled.div`
   flex-direction: column;
```
devinleighsmith, line 56: Note that the side bar was trying to use box display to handle the 100% vh of the form content. This was actually setting the form to 100% of the sidebar, even though other content above the form was present, meaning the form + other content was > 100% of the parent.

Switched the parent to a flex, this ensures that the children will not exceed the height of the parent. No need for really long margins anymore to get around some of the form being outside the parent. Need slightly awkward negative absolute position now to get around the padding of the parent, but still seems much better overall.

The scroll bar now displays correctly as well (was getting cut off the bottom of the screen before).

frontend/src/hooks/useAsyncError.ts
devinleighsmith: I'll link you to the article that I got this from, probably do a better job of explaining.

devinleighsmith: https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971

devinleighsmith: in summary, the only way to rethrow an async action in a manner that an error boundary will catch is by converting an async error to a regular js error. How do you do this? You set a state var in an async function. This causes a re-render in regular, non-async react js. We then use a useEffect such that when the error state var is set we just throw the error

jdtoombs (approved):

PR #1014, opened 2021-03-03
PA-2708: User Management Is Not Updating Keycloak Roles Correctly

LawrenceLau2020:
Based on the error logs from Kibana, there was an unhandled exception "System.Threading.Tasks.TaskCanceledException"  in the PimsKeycloakUserService.cs file.

I'm hoping this will catch the error and output it to the toast UI when it happens again in DEV.

### DIFF
--- a/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
+++ b/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
@@ -142,7 +142,14 @@ namespace Pims.Dal.Keycloak
                 var role = _pimsAdminService.Role.Find(r.RoleId) ?? throw new KeyNotFoundException("Cannot remove a role from a user, when the role does not exist.");
                 if (role.KeycloakGroupId == null) throw new KeyNotFoundException("PIMS has not been synced with Keycloak.");
                 euser.Roles.Remove(r);
-                await _keycloakService.RemoveGroupFromUserAsync(user.Id, role.KeycloakGroupId.Value);
+                try
+                {
+                    await _keycloakService.RemoveGroupFromUserAsync(user.Id, role.KeycloakGroupId.Value);
+                }
+                catch (System.Threading.Tasks.TaskCanceledException ex)
+                {
+                    throw new Exception(ex.Message);
+                }
             });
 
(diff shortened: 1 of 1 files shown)


backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -142,7 +142,14 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
                 var role = _pimsAdminService.Role.Find(r.RoleId) ?? throw new KeyNotFoundException("Cannot remove a role from a user, when the role does not exist.");
                 if (role.KeycloakGroupId == null) throw new KeyNotFoundException("PIMS has not been synced with Keycloak.");
                 euser.Roles.Remove(r);
-                await _keycloakService.RemoveGroupFromUserAsync(user.Id, role.KeycloakGroupId.Value);
+                try
+                {
+                    await _keycloakService.RemoveGroupFromUserAsync(user.Id, role.KeycloakGroupId.Value);
+                }
+                catch (System.Threading.Tasks.TaskCanceledException ex)
```
Fosol, line 149: If this is already throwing a `TaskCanceledException` then I'm not sure there is any value in wrapping it in a generic `Exception`.

You'll get more mileage in our API middleware `ErrorHandlingMiddleware.cs` logic.  I believe it may be somehow ignoring the `TaskCanceledException`.  You'll need to test to confirm however.  When reviewing the code it would appear that at a minimum it should be logging the generic error message "Middleware caught unhandled exception".

Fosol (changes requested):

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
devinleighsmith: @LawrenceLau2020 if you know roughly where the TaskCanceledException occurs in the API, just update the api to always throw that exception from that location. That should let you do all the backend/frontend testing you need. Then just make sure to take out the always thrown exception afterwords ;-)

Fosol:
I've submitted a PR that should resolve some or all of these issues.

PR #1015, opened 2021-03-03
Fix slideout improvements/assessed

jdtoombs:
Making sure the slide out is displaying correct information. Had to verify the keys were the appropriate type before pulling that value.

Additionally, it used to sum up each building and display the cumulative value but now it takes straight from the `Improvements` of the developed land. 

Updated: 

- only take values from current year if present (previously most recent)

### DIFF
--- a/frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
+++ b/frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react';
 import { ListGroup } from 'react-bootstrap';
-import { IBuilding, IEvaluation, IParcel } from 'actions/parcelsActions';
+import { IParcel } from 'actions/parcelsActions';
 import { Label } from 'components/common/Label';
 import './InfoSlideOut.scss';
 import { formatMoney } from 'utils/numberFormatUtils';
-import { compareDate, OuterRow } from './InfoContent';
+import { OuterRow } from './InfoContent';
 import { ThreeColumnItem } from './ThreeColumnItem';
+import { EvaluationKeys } from 'constants/evaluationKeys';
+import { getCurrentYearEvaluation } from 'features/projects/common/projectConverter';
 
 interface IParcelAttributes {
   /** the selected parcel information */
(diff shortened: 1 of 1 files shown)


Fosol (commented):
Your screenshot would imply it sums the most recent values rather than only sums the current year.  
If there are no values in the current year, it shouldn't have a value.

jdtoombs:
> Your screenshot would imply it sums the most recent values rather than only sums the current year.
> If there are no values in the current year, it shouldn't have a value.

Updated so it only displays values for current year if present

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -23,25 +25,20 @@ export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo, canV
...
     formatAssessed = formatMoney(
-      parcelInfo?.evaluations.sort((a, b) => compareDate(a.date, b.date)).reverse()[0].value,
+      getCurrentYearEvaluation(parcelInfo?.evaluations, EvaluationKeys.Assessed)?.value,
     );
   } else {
     formatAssessed = '';
   }
-
-  let improvements = 0;
+  let improvements = '';
```
devinleighsmith, line 35: good call.

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -1,11 +1,13 @@
...
-import { IBuilding, IEvaluation, IParcel } from 'actions/parcelsActions';
+import { IParcel } from 'actions/parcelsActions';
 import { Label } from 'components/common/Label';
 import './InfoSlideOut.scss';
 import { formatMoney } from 'utils/numberFormatUtils';
-import { compareDate, OuterRow } from './InfoContent';
+import { OuterRow } from './InfoContent';
 import { ThreeColumnItem } from './ThreeColumnItem';
+import { EvaluationKeys } from 'constants/evaluationKeys';
+import { getCurrentYearEvaluation } from 'features/projects/common/projectConverter';
```
devinleighsmith, line 10: I need to extract this into a util (as well as some other financial logic). Fine in this review.

frontend/src/components/maps/leaflet/InfoSlideOut/ParcelAttributes.tsx
```diff
@@ -23,25 +25,20 @@ export const ParcelAttributes: React.FC<IParcelAttributes> = ({ parcelInfo, canV
...
-    parcelInfo.buildings.forEach((building: IBuilding) => {
-      if (building.evaluations?.length >= 1) {
-        improvements += +building.evaluations
-          .sort((a: IEvaluation, b: IEvaluation) => compareDate(a.date, b.date))
-          .reverse()[0].value;
-      }
-    });
+    improvements = formatMoney(parcelInfo?.assessedBuilding);
+  } else if (!!getCurrentYearEvaluation(parcelInfo?.evaluations, EvaluationKeys.Improvements)) {
+    improvements = formatMoney(
```
devinleighsmith, line 39: I guess this doesn't really hurt anything, but I'm pretty sure the assessedBuilding IS the improvement from the current year.

devinleighsmith (approved):

goemen (approved):

PR #1016, opened 2021-03-03
PA-2854 Disposal Project Clearance Notification

Fosol:
Disposal Projects will only require the `ClearanceNotificationSentOn` date if they are in ERP.

Disposal Projects that are in other status like **Not in SPL** will not require the value.

This should allow some historical projects to bypass this validation.

### DIFF
--- a/frontend/src/features/projects/erp/forms/EnhancedReferralCompleteForm.tsx
+++ b/frontend/src/features/projects/erp/forms/EnhancedReferralCompleteForm.tsx
@@ -20,6 +20,7 @@ import {
 } from '../../common';
 import GenericModal from 'components/common/GenericModal';
 import { validateFormikWithCallback } from 'utils';
+import { clearanceNotificationSentOnRequired } from './erpYupSchema';
 
 const OrText = styled.div`
   margin: 0.75rem 2rem 0.75rem 2rem;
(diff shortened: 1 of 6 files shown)


frontend/src/features/projects/erp/forms/EnhancedReferralCompleteForm.tsx
```diff
@@ -50,6 +51,9 @@ const EnhancedReferralCompleteForm = ({
   const [proceedToSpl, setProceedToSpl] = useState(false);
   const [notInSpl, setNotInSpl] = useState(false);
   const [disposeExternally, setDisposeExternally] = useState(false);
+  const isClearanceNotificationSentOnRequired = clearanceNotificationSentOnRequired(
```
devinleighsmith, line 54: since this is only ever used in the below checks for disabled, would probably make more sense to include the !formikProps.values.clearanceNotificationSentOn in this boolean, so you can reduce the amount of duplicated code in the disabled checks.

frontend/src/features/projects/erp/forms/erpYupSchema.ts
```diff
@@ -1,6 +1,18 @@
 import * as Yup from 'yup';
+import _ from 'lodash';
+import { IStatus } from 'features/projects/common';
+
+/**
+ * Determine if the specified project status code requires a clearance notification sent on value.
+ * @param statusCode The project status code.
+ */
+export const clearanceNotificationSentOnRequired = (statusCode: string) => {
+  return _.includes(['ERP-ON', 'ERP-OH'], statusCode); // ERP-ON, ERP-OH
```
devinleighsmith, line 10: /nit, try and use the es6 includes when possible. ie ['ERP-ON', 'ERP-OH'].includes(statusCode)

devinleighsmith: also not sure if I agree with putting this check here as I think we normally just use yup schema files for yup schema. Not a big deal I suppose.

jdtoombs (approved):

frontend/src/features/projects/erp/forms/erpYupSchema.ts
Fosol: I've moved to utils

PR #1019, opened 2021-03-03
Default to includeAllProperties in Surplust Filter

jdtoombs:
- Non sres user cannot specify which agency just set it to undefined and `includeAllProperties` when searching surplus filter 
- If sres user they can adjust the agency filter with `includeAllProperties` set to true

Something to confirm, will a user (ex. REM) want to do the SPL/ERP filter with `includeAllProperties = false`? Will have to adjust if this is the case. The Surplus Filter and Property Filter have the same initial values so some more logic will have to be added.

### DIFF
--- a/frontend/src/features/properties/filter/PropertyFilter.tsx
+++ b/frontend/src/features/properties/filter/PropertyFilter.tsx
@@ -114,7 +114,7 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
   }, [defaultFilter, propertyFilter, agencies]);
 
   const changeFilter = (values: IPropertyFilter) => {
-    const agencyIds = (values.agencies as any).value
+    const agencyIds = (values.agencies as any)?.value
       ? (values.agencies as any).value
       : values.agencies;
     setPropertyFilter({ ...values, agencies: agencyIds });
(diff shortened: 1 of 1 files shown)


frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -149,6 +149,8 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
               onEnter={() => {
                 setFindMoreOpen(true);
                 setFieldValue('surplusFilter', true);
+                setFieldValue('includeAllProperties', true);
```
jdtoombs, line 152: Will have to adjust this if we want them to be able to have this as false

Fosol: After closing the SPP form, can they then alter the agency and research?

frontend/src/features/properties/filter/PropertyFilter.tsx
```diff
@@ -149,6 +149,8 @@ export const PropertyFilter: React.FC<IPropertyFilterProps> = ({
...
+                !keycloak.hasClaim(Claims.ADMIN_PROPERTIES) && setFieldValue('agencies', undefined);
```
Fosol, line 153: This removes any current agency filter?

Fosol (approved):

frontend/src/features/properties/filter/PropertyFilter.tsx
jdtoombs: Yup, when they close they can alter as they please, and I had to set it to undefined otherwise when the user opened the Surplus Filter it would correctly switch to includeAllProperties but still have their agency applied against it

jdtoombs: But left it so SRES can run queries against specific agencies

goemen (approved):

PR #1020, opened 2021-03-03
Add filtering to SPL reports

goemen:
Hnet-image (10)

### DIFF
--- a/frontend/src/components/Table/Table.scss
+++ b/frontend/src/components/Table/Table.scss
@@ -59,6 +59,10 @@
     &.reset-filter {
       display: flex;
       align-items: center;
+
+      &.svg-btn {
+        display: block;
+      }
     }
     &.clickable {
       cursor: pointer;
@@ -74,6 +78,26 @@
       justify-content: flex-start;
     }
     .filter-wrapper {
+      div {
+        div {
+          display: flex;
+          .form-group {
+            width: 100%;
+            .rbt {
+              width: 100%;
+              div {
+                input {
+                  width: 100% !important;
+                }
+              }
+              .rbt-menu {
+                width: 370px !important;
+              }
+            }
+          }
+        }
+      }
+
       &.active {
         color: $active-color;
       }
--- a/frontend/src/components/maps/leaflet/LayersControl/data.ts
+++ b/frontend/src/components/maps/leaflet/LayersControl/data.ts
@@ -34,7 +34,7 @@ export const layersTree: ILayerItem[] = [
       {
         key: 'municipalities',
         label: 'Municipalities',
-        on: true,
+        on: false,
         layers: 'pub:WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP',
         url:
           'https://openmaps.gov.bc.ca/geo/pub/WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP/ows?',
--- a/frontend/src/features/splReports/components/ReportForm.tsx
+++ b/frontend/src/features/splReports/components/ReportForm.tsx
@@ -1,7 +1,12 @@
 import * as React from 'react';
 import Table from 'components/Table/Table';
-import { IReport, ISnapshot } from '../interfaces';
+import { IReport, ISnapshot, ISnapshotFilter } from '../interfaces';
 import { columns } from '../columns';
+import useCodeLookups from 'hooks/useLookupCodes';
+import * as API from 'constants/API';
+import { mapSelectOptionWithParent } from 'utils';
+import { useSplReportContext } from '../containers/SplReportContainer';
+import { SortDirection } from 'components/Table/TableSort';
 
 interface IReportFormProps {
   currentReport?: IReport;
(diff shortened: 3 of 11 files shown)


frontend/src/components/Table/Table.scss
```diff
@@ -59,6 +59,10 @@
     &.reset-filter {
       display: flex;
       align-items: center;
+
+      &.svg-btn {
+        display: flex;
```
devinleighsmith, line 64: huh, why do we need flex on a button?

frontend/src/components/Table/Table.scss
```diff
@@ -74,6 +78,26 @@
...
+      div {
+        div {
+          display: flex;
+          .form-group {
+            width: 100%;
+            .rbt {
+              width: 100%;
+              div {
+                input {
+                  width: 100% !important;
```
devinleighsmith, line 90: can we restructure something so that this is not necessary? the width 100 -> width 100 -> width 100 important seems not great.

frontend/src/features/splReports/components/ReportForm.tsx
```diff
@@ -13,11 +17,26 @@ interface IReportFormProps {
...
+
   return (
     <Table
       name="spl report table"
       noRowsMessage="No Reports Available"
-      columns={columns}
+      filterable
+      filter={snapshotFilter as any}
+      onFilterChange={filter => setSnapshotFilter({ ...snapshotFilter, ...filter })}
+      onSortChange={(field: string, direction: any) => {
```
devinleighsmith, line 35: does direction need to be an any? I feel like discrete values make sense here.

frontend/src/features/splReports/containers/SplReportContainer.tsx
```diff
@@ -33,17 +46,52 @@ const SplReportContainer: React.FunctionComponent<ISplReportContainerProps> = ()
...
+    fiscalYear: '',
+    sortBy: { projectId: 'asc', snapshotOn: 'desc' },
+  });
 
   const id = currentReport?.id;
   useEffect(() => {
     const fetch = async () => {
       if (id) {
-        const data = await getProjectReportSnapshotsById(id);
+        let data = await getProjectReportSnapshotsById(id);
```
devinleighsmith, line 60: /nit would prefer to always use descriptive variable names where possible. So in this case snapshots or projectReportSnapshots is preferred over just data (already had issues in the past with naming conflicts between data and the axios data object from axios responses.

frontend/src/features/splReports/containers/SplReportContainer.tsx
```diff
@@ -12,6 +12,19 @@ import { useRouterReport } from '../hooks/useRouterReport';
 
 interface ISplReportContainerProps {}
 
+const SplReportContainerContext = React.createContext<{
```
devinleighsmith, line 15: I think we both would rather this be in its own file, but up to you.

devinleighsmith (approved):

frontend/src/components/Table/Table.scss
goemen: Oooops

frontend/src/components/maps/leaflet/LayersControl/data.ts
```diff
@@ -34,7 +34,7 @@ export const layersTree: ILayerItem[] = [
       {
         key: 'municipalities',
         label: 'Municipalities',
-        on: true,
+        on: false,
```
Fosol, line 37: Ideally keep your PRs specific to a single related story when possible.

frontend/src/features/splReports/components/ReportForm.tsx
```diff
@@ -13,11 +17,26 @@ interface IReportFormProps {
...
+  );
+  const { snapshotFilter, setSnapshotFilter } = useSplReportContext();
+
   return (
     <Table
       name="spl report table"
       noRowsMessage="No Reports Available"
-      columns={columns}
+      filterable
+      filter={snapshotFilter as any}
```
Fosol, line 33: Filter as `any` also feels like a missed opportunity to create an interface.

frontend/src/features/splReports/containers/SplReportContainer.tsx
```diff
@@ -33,17 +46,52 @@ const SplReportContainer: React.FunctionComponent<ISplReportContainerProps> = ()
...
     const fetch = async () => {
       if (id) {
-        const data = await getProjectReportSnapshotsById(id);
+        let data = await getProjectReportSnapshotsById(id);
+        data = data.map(
+          s =>
+            ({
+              ...s,
+              project: { ...s.project, agencyName: s.project?.agency || s.project?.subAgency },
+            } as any),
```
Fosol, line 66: A lot of `any` in this component.  It would be nice to not lose the reason we're using typescript.

frontend/src/features/splReports/containers/SplReportContainer.tsx
```diff
@@ -123,54 +171,60 @@ const SplReportContainer: React.FunctionComponent<ISplReportContainerProps> = ()
...
+        setShowSidebar={setShowSidebar}
+        onOpen={report => setCurrentReport(report)}
+        onDelete={onDelete}
+        onFinal={onFinal}
+        onRefresh={refreshSnapshots}
+        onSave={onSave}
+        onAdd={addReport}
+        onExport={exportReport}
+        onFromChange={(id: number) => {
+          changeFrom({ ...currentReport, from: _.find(reports, { id: id })?.to } as IReport);
```
Fosol, line 188: Do we need lodash anymore?  Seems like array's already have most of this functionality.

Fosol (approved):

frontend/src/features/splReports/components/ReportForm.tsx
goemen: The table by default uses a filter of T = type of data display, so was a way to force it to use a custom filter type

goemen: Addressed

PR #1021, opened 2021-03-03
Blur Page until Map Requests Complete

goemen:
https://user-images.githubusercontent.com/4842061/109880807-3d131c80-7c2c-11eb-952b-2cf0323b4adf.mp4

### DIFF
--- a/frontend/src/components/maps/leaflet/FilterBackdrop/index.tsx
+++ b/frontend/src/components/maps/leaflet/FilterBackdrop/index.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react';
+import { Spinner } from 'react-bootstrap';
+import styled from 'styled-components';
+
+export interface FilterBackdropProps {
+  show?: boolean;
+}
+
+const Backdrop = styled.div`
+  width: 100%;
+  height: 100%;
+  position: fixed;
+  z-index: 999;
+  left: 0;
+  background-color: rgba(0, 0, 0, 0.4);
+  display: flex;
+  align-items: center;
+  align-content: center;
+  justify-items: center;
+  justify-content: center;
+`;
+
+const FilterBackdrop: React.FC<FilterBackdropProps> = ({ show }) => {
+  return show ? (
+    <Backdrop>
+      <Spinner animation="border" variant="warning" />
+    </Backdrop>
+  ) : null;
+};
+
+export default FilterBackdrop;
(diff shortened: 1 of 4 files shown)


jdtoombs (approved):
Nice!

frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -246,16 +251,19 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
...
+        draftPoints={draftProperties}
+        zoom={zoom}
+        bounds={bbox}
+        onMarkerClick={onMarkerClick}
+        zoomToBoundsOnClick={true}
+        spiderfyOnMaxZoom={true}
+        selected={selected}
+        tilesLoaded={!loadingTiles}
+      />
+      <FilterBackdrop show={showFilterBackdrop} />
```
Fosol, line 266: Why not always include the <FilterBackdrop/>, but only display it if set to true?

Fosol (approved):

frontend/src/components/maps/leaflet/FilterBackdrop/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Spinner } from 'react-bootstrap';
+import styled from 'styled-components';
+
+export interface FilterBackdropProps {
+  show?: boolean;
+}
+
+const Backdrop = styled.div`
+  width: 100vw;
+  height: calc(100vh - 117px);
```
devinleighsmith, line 11: could use the css variables here in case the header changes height.

devinleighsmith: Also, does this work with a small screen? in case the filter is pushed to multiple lines and becomes taller then normal.

frontend/src/components/maps/leaflet/InventoryLayer.tsx
goemen: Does not really matter, but this just makes is easy for reusability

frontend/src/components/maps/leaflet/FilterBackdrop/index.tsx
goemen: There is a slight difference in the height of the cover of the backdrop, and we can solve it by making the backdrop cover the whole viewport @devinleighsmith

devinleighsmith: I'm a little nervous about doing that, does that completely cover all of the navigation controls as well? I guess my concern would be that if the map is taking a while to load the user can't navigate to any other screens.

Fosol: I'm not a fan of locking down the whole app when ever we make a request.  It ends up having the opposite effect on users experience.

PR #1023, opened 2021-03-03
Pa 2899

devinleighsmith:
Transition from Aporeto NSPs to Kubernetes NPs.
Remove Aporeto NSPs to avoid confusion.
Remove unused kibana/grafana bcs/dcs


devinleighsmith:
The main takeaway:
before:
pods needed NSPs to be able to allow egress to other pods, the internet, etc. All NSP's written from the perspective of the pod initiating the communication.
after:
pods need NPs to allow ingress via other pods, the internet, etc. All NP's are written from the perspective of the pod being communicated with.
(egress NPs not available at this time).

openshift/4.0/templates/api/deploy.yaml
```diff
@@ -139,37 +139,25 @@ parameters:
     description: "Maximum amount of memory the container can use."
     value: "4Gi"
 objects:
-  # Allow the api to make requests to the internet.
```
devinleighsmith, line 142: nps only act on ingress, and so no np is required to allow pod-to-internet communication.

openshift/4.0/templates/logging/README.md
```diff
@@ -1,49 +0,0 @@
-# Monitor Logs
```
devinleighsmith, line 1: This may be contentious, but if we aren't using kibana/grafana etc I vote we remove the configs.

openshift/4.0/templates/network-security-policy/internal-k8s-api.yaml
```diff
@@ -1,25 +0,0 @@
-kind: Template
```
devinleighsmith, line 1: Similarly, not everyone may agree with this call, but it seemed logical to delete the aporeto nsps if they are being deprecated.

openshift/4.0/templates/network-security-policy/internet-to-routes.yaml
```diff
@@ -0,0 +1,25 @@
+kind: Template
```
devinleighsmith, line 1: This simplifies ingress by allowing internet communication to pods that have published routes. No need to create one NSP per pod with route anymore.

openshift/4.0/templates/network-security-policy/nsp-tools.yaml
```diff
@@ -1,56 +1,51 @@
...
-      description: Allow service accounts to run builds/deployments.
-      source:
-        - - $namespace=${PROJECT_NAMESPACE}-${ENV_NAME}
-          - "@app:k8s:serviceaccountname=builder"
-        - - $namespace=${PROJECT_NAMESPACE}-${ENV_NAME}
-          - "@app:k8s:serviceaccountname=deployer"
-      destination:
-        - - int:network=internal-cluster-api-endpoint
-        - - ext:network=any
+      # Allow all pods within the current namespace to communicate
```
devinleighsmith, line 40: in the case of the tools namespace this seemed acceptable to me, let me know if you don't agree.

goemen (approved):

openshift/4.0/templates/api/deploy.yaml
```diff
@@ -139,37 +139,26 @@ parameters:
...
-        - - ext:network=any
-
   # Allow the api to communicate with the database.
-  - kind: NetworkSecurityPolicy
-    apiVersion: security.devops.gov.bc.ca/v1alpha1
+  - kind: NetworkPolicy
+    apiVersion: networking.k8s.io/v1
     metadata:
-      name: pod-to-pod-permit-${APP_NAME}-${ROLE_NAME}-${DATABASE_ROLE_NAME}
+      name: pod-from-pod-permit-${NAME}-${ROLE_NAME}-${DATABASE_ROLE_NAME}
```
Fosol, line 146: The ${NAME} parameter doesn't appear to be correct as the template has ${APP_NAME}.

openshift/4.0/templates/logging/deploy-elasticsearch.yaml
```diff
@@ -1,227 +0,0 @@
-kind: Template
```
Fosol, line 1: If we're removing eleasticsearch then the API needs to be updated, as it is currently attempting to write to elasticsearch.

openshift/4.0/templates/migrations/migration-job.yaml
```diff
@@ -29,7 +29,41 @@ parameters:
       [Development, Staging, Production].
     required: true
     value: Production
+  - name: DATABASE_ROLE_NAME
```
Fosol, line 32: Is this used by Jenkins to run the database migrations when going to prod?
Also I noticed references to the old project namespace in this file.

openshift/4.0/templates/monitor/deploy-grafana.yaml
```diff
@@ -1,1298 +0,0 @@
-kind: Template
```
Fosol, line 1: If we're removing grafana and prometheus, have you deleted them from the tools namespace too?

openshift/4.0/templates/network-security-policy/deny-all-traffic.yaml
```diff
@@ -0,0 +1,20 @@
+kind: Template
+apiVersion: v1
+name: internet-to-routes
+parameters:
+  - name: NAME
```
Fosol, line 5: Any value in having a parameter?

openshift/4.0/templates/network-security-policy/internet-to-routes.yaml
```diff
@@ -0,0 +1,25 @@
...
+apiVersion: v1
+name: internet-to-routes
+parameters:
+  - name: NAME
```
Fosol, line 5: Any value in the `NAME` parameter?

openshift/4.0/templates/network-security-policy/nsp-tools.yaml
```diff
@@ -1,56 +1,51 @@
...
+      # to one another.
+      podSelector:
+      ingress:
+        - from:
+            - podSelector: {}
+parameters:
+  - name: NAMESPACE
```
Fosol, line 47: Not seeing this parameter used

openshift/4.0/templates/network-security-policy/service-account-to-k8s-api.yaml
```diff
@@ -1,28 +0,0 @@
-kind: Template
```
Fosol, line 1: This is no longer needed?

Fosol (approved):

openshift/4.0/templates/network-security-policy/service-account-to-k8s-api.yaml
devinleighsmith: this is an egress policy which isn't supported in KNPs at the moment.

openshift/4.0/templates/monitor/deploy-grafana.yaml
devinleighsmith: no, I'd do that after we agreed to remove them here.

openshift/4.0/templates/migrations/migration-job.yaml
devinleighsmith: yes, this creates a separate container to run ef just for migrations.

openshift/4.0/templates/logging/README.md
Fosol: I think it might be helpful to add them to a different folder so they are available for future possible use.  I realize they would exist in the git history (unless it was purged at some point), but they're much harder to look for when that occurs.

PR #1029, opened 2021-03-04
Project property Name/PID column and update support email

jdtoombs:
Some of the property names are a bit long which makes the columns a bit obnoxious. Not sure if we want to cut off after a certain length?


Fosol (approved):

devinleighsmith (approved):

goemen (approved):
lgtm

PR #1032, opened 2021-03-05
PA-2875 Updated Not in SPL to Dispose

Fosol:
Removed some of the required attributes on inputs to allow **Not in SPL** to go to **Disposed**.

image


devinleighsmith:
Just to make sure that we are on the same page, that required prop just handles the display of the blue outline on those fields, it shouldn't **allow** or **deny** anything

Fosol:
It just displays blue around `Disposed On`.  The button requires it to be set before it is enabled.

devinleighsmith (approved):

goemen (approved):
Looks good to me

PR #1036, opened 2021-03-06
Land only filter fix

jdtoombs:
This filter was working on its own (ie. without SPL/ERP filter applied); however, started failing when the two were used together.

I spent a couple of hours trying to get the join to work but it seems as if it was failing due to the attempt to join an in-memory collection with EF. 

This was the only working solution that I could find, let me know if it is efficient enough.

### DIFF
--- a/backend/dal/Helpers/Extensions/PropertyExtensions.cs
+++ b/backend/dal/Helpers/Extensions/PropertyExtensions.cs
@@ -56,12 +56,13 @@ namespace Pims.Dal.Helpers.Extensions
                 query = query.Where(p => p.RentableArea <= filter.RentableArea);
 
             if (filter.BareLandOnly == true)
-                query = (from p in query
-                         join pb in context.ParcelBuildings
-                            on p.Id equals pb.ParcelId into ppbGroup
-                         from pb in ppbGroup.DefaultIfEmpty()
-                         where pb == null && p.PropertyTypeId == Entity.PropertyTypes.Land
-                         select p);
+            {
+                query = from p in query
+                        join pa in context.Parcels on new { p.Id, PropertyTypeId = (int)p.PropertyTypeId } equals new { pa.Id, PropertyTypeId = (int)pa.PropertyTypeId }
+                        where p.PropertyTypeId == Entity.PropertyTypes.Land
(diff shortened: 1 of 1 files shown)


backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -56,12 +56,14 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
...
 
             if (filter.BareLandOnly == true)
-                query = (from p in query
-                         join pb in context.ParcelBuildings
-                            on p.Id equals pb.ParcelId into ppbGroup
-                         from pb in ppbGroup.DefaultIfEmpty()
-                         where pb == null && p.PropertyTypeId == Entity.PropertyTypes.Land
-                         select p);
+            {
+                var parcelIds = context.Parcels
```
Fosol, line 60: This will make a separate request for every land in the db were it has no associated building.  Not ideal if we can avoid it.  I'll provide a possible alternate later.

backend/dal/Helpers/Extensions/PropertyExtensions.cs
```diff
@@ -56,12 +56,13 @@ private static IQueryable<Entity.Views.Property> GenerateCommonQuery(this PimsCo
...
 
             if (filter.BareLandOnly == true)
-                query = (from p in query
-                         join pb in context.ParcelBuildings
-                            on p.Id equals pb.ParcelId into ppbGroup
-                         from pb in ppbGroup.DefaultIfEmpty()
-                         where pb == null && p.PropertyTypeId == Entity.PropertyTypes.Land
-                         select p);
+            {
+                query = from p in query
```
Fosol, line 60: I've updated with the following.

Fosol (approved):
Let me know if you have any questions about my change?

backend/dal/Helpers/Extensions/PropertyExtensions.cs
jdtoombs: Thank you for this!

devinleighsmith (approved):

PR #1041, opened 2021-03-08
Pa-2896 Updated Delete Property

Fosol:
Removing the `property-delete` claim from **Real Estate Manager** and **Real Estate Analyst**.  Only **SRES** or administrators will be able to delete properties.

The owning agency will still be able to delete **Subdivisions** if they are not part of a project.

Add DB Migration `v01.09.04` to update role claims.

Fixed **Keycloak Sync Tool** so that it would remove Keycloak roles from groups.


backend/api/Areas/Property/Controllers/ParcelController.cs
```diff
@@ -203,7 +203,7 @@ public IActionResult UpdateParcelFinancials(int id, [FromBody] Model.ParcelModel
         /// <param name="model"></param>
         /// <returns></returns>
         [HttpDelete("{id}")]
-        [HasPermission(Permissions.PropertyDelete)]
+        [HasPermission(Permissions.PropertyDelete, Permissions.PropertyEdit)]
```
Fosol, line 206: Edit claim `property-edit` is required to delete subdivisions.

backend/api/Policies/HasPermissionAttribute.cs
```diff
@@ -29,12 +24,13 @@ public HasPermissionAttribute(Permissions permission) : base(typeof(PermissionFi
...
         /// <summary>
         /// Creates a new instance of a HasPermissionAttribute, initializes it with the specified permissions.
+        /// This will ensure the user has at least one of the specified permissions.
         /// </summary>
         /// <param name="permissions"></param>
         /// <returns></returns>
         public HasPermissionAttribute(params Permissions[] permissions) : base(typeof(PermissionFilter))
         {
-            this.Arguments = permissions.Select(p => p).Cast<object>().ToArray();
+            this.Arguments = new object[] { permissions };
```
Fosol, line 33: Fixed implementation so that DI would work with multiple permissions.

PR #1042, opened 2021-03-08
PA-2396

devinleighsmith:
this change causes the selected marker to pulse when the sidebar is open.


jdtoombs (approved):

Fosol (approved):

devinleighsmith:
Need to talk to Mikaela about this, the active marker may change as the user clicks around the map. Don't want the marker to flash in that case.

PR #1047, opened 2021-03-09
During initial login the state of the selected parameter is undefined…

devinleighsmith:
…. Under certain conditions this will then be updated to null, which is technically a different value which triggers a new search.

In certain cases, this extra search will be triggered right before a normal search causing a race condition.

Rather then trying to enforce the null/undefined every possible place the selector is used, just updating the effect to treat all falsy values the same.


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -247,10 +247,11 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
   };
 
   // Fetch the geoJSON collection of properties.
+  const parcelDetail = !!selected?.parcelDetail ?? undefined;
```
devinleighsmith, line 250: collate all falsey expressions to undefined.

goemen: Should this not in a 'useMemo'

goemen (approved):
Just one nit, but looks good

Fosol (approved):

PR #1048, opened 2021-03-09
Updated API Routes Deployment Config

Fosol:
Removed the routes from the standard API deployment configuration to eliminate issue when deploying to production.  Making it a separate step to ensure the SSL certificate information is correctly added.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -25,6 +25,16 @@
       "stopAtEntry": false
     },
     {
+      "name": "Tool-Import",
+      "type": "coreclr",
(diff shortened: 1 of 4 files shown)


.vscode/launch.json
```diff
@@ -24,6 +24,16 @@
       "cwd": "${workspaceRoot}/tools/converters/excel",
       "stopAtEntry": false
     },
+    {
```
Fosol, line 27: We can debug the importer tool now.

devinleighsmith (approved):

PR #1052, opened 2021-03-09
PA-2948 Debugging Notifications

Fosol:
Adding a configurable delay to all notifications.

Providing a way to test `to`, `cc`, and `bcc` notifications.  This may allow us to debug production email issues as we can send emails and place them in a queue for a length of time where support resources can review the test emails.


goemen (approved):
All good

jdtoombs (approved):

PR #1063, opened 2021-03-11
PA-2708 Fixed Keycloak User Administration

Fosol:
While still not ideal, I am hoping the removal of all groups and then adding back the relevant ones will resolve the issues we're seeing.  Ideally if the requests to keycloak fail PIMS will also not be updated.  Which will allow us to keep them in sync better.

### DIFF
--- a/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
+++ b/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
@@ -9,6 +9,8 @@ using System.Linq;
 using System.Threading.Tasks;
 using Entity = Pims.Dal.Entities;
 using KModel = Pims.Keycloak.Models;
+using Microsoft.Extensions.Logging;
+using Pims.Core.Exceptions;
 
 namespace Pims.Dal.Keycloak
 {
@@ -116,7 +118,7 @@ namespace Pims.Dal.Keycloak
         /// <returns></returns>
         public async Task<Entity.User> UpdateUserAsync(Entity.User user)
         {
-            var kuser = await _keycloakService.GetUserAsync(user.Id) ?? throw new KeyNotFoundException();
+            var kuser = await _keycloakService.GetUserAsync(user.Id) ?? throw new KeyNotFoundException("User does not exist in Keycloak");
             var euser = _pimsAdminService.User.Get(user.Id);
 
             if (user.Username != kuser.Username) throw new InvalidOperationException($"Cannot change the username from '{kuser.Username}' to '{user.Username}'.");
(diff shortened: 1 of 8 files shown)


backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -129,20 +131,32 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
             // Update PIMS
             _mapper.Map(user, euser);
 
-            // Update Roles.
-            addRoles.ForEach(async r =>
+            // Remove all keycloak groups from user.  // TODO: Only add/remove the ones that should be removed.
```
Fosol, line 134: If I had more time I would only remove the ones that need to be removed, but for simplicity I'm removing them all before adding back the relevant ones.

backend/tests/core/TestHelper.cs
```diff
@@ -153,6 +153,16 @@ public T GetService<T>()
             return this.BuildServiceProvider().GetService<T>();
         }
 
+        /// <summary>
+        /// Get the mocked service for the specified 'T' type.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public Mock<T> GetMock<T>() where T : class
```
Fosol, line 161: Just a quicker way to get the mock.

jdtoombs (approved):

goemen (approved):

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -129,20 +131,32 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
...
+            var userGroups = await _keycloakService.GetUserGroupsAsync(euser.Id);
+            foreach (var group in userGroups)
+            {
+                await _keycloakService.RemoveGroupFromUserAsync(user.Id, group.Id);
```
devinleighsmith, line 138: out of curiosity, what happens when one or more of these requests fails? does it break out of the rest of this method and skip the remaining steps, or does it ignore the failure and keep going?

devinleighsmith (approved):

backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
Fosol: It would throw an exception that would stop further requests.  Which would leave PIMS out of sync in the best way possible.  As PIMS will show the original state and failure, so the admin will want to try and save again.  Which would hopefully resolve the issue and sync the user with keycloak fully.

PR #1066, opened 2021-03-11
Add missing childless agencies to agency list

jdtoombs:
Open to feedback here

I assigned a `parentId` of `-1` to parent agencies that have no children. This helps the component group them together and ensure they are in the list. 

Is there a better value than -1 I could use for the id? 

These childless parent agencies show as bold like the other parents in the list.

parentselect

### DIFF
--- a/frontend/src/components/common/form/ParentSelect.tsx
+++ b/frontend/src/components/common/form/ParentSelect.tsx
@@ -59,6 +59,8 @@ export const ParentSelect: React.FC<IParentSelect> = ({
   const [clear, setClear] = useState(false);
   /** controls the multi selections displayed to the user */
   const [multiSelections, setMultiSelections] = React.useState<any>([]);
+  /** parent id used to identify parent agencies with no children */
+  const CHILDLESS_PARENT_ID = -1;
 
   /** wipe the selection from input on reset */
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -131,21 +131,34 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
+                : x.parent,
+            };
+          });
+
+          /** group the results by their desired parents */
+          const resultGroup = groupBy(results, x => x.parentId);
+
           const items = Object.keys(resultGroup)
             .sort()
+            .reverse()
```
jdtoombs, line 158: Just added this since the `parentId` is -1 it would show all the grouped childless parent agencies first which was not visually appealing

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -131,21 +131,34 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
+          // This assigns a specific id to childless parent agencies
+          // Headers will not be displayed for these childless parent agencies but they will be displayed as regular options and grouped
+          results = results.map(x => {
+            return {
+              ...x,
+              parentId: !!childless.find(c => Number(c.value) === Number(x.value))
+                ? -1
+                : x.parentId,
+              parent: !!childless.find(c => Number(c.value) === Number(x.value))
+                ? 'Childless'
```
jdtoombs, line 148: Doesn't actually appear to the user, used by the component logic

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -131,21 +131,34 @@ export const ParentSelect: React.FC<IParentSelect> = ({
...
+          /** group the results by their desired parents */
+          const resultGroup = groupBy(results, x => x.parentId);
+
           const items = Object.keys(resultGroup)
             .sort()
+            .reverse()
             .map(parentId => (
               <Fragment key={parentId}>
-                {!!results.find((x: SelectOption) => x.parentId === +parentId) && (
+                {!!results.find((x: SelectOption) => x.parentId === +parentId) && +parentId !== -1 && (
```
jdtoombs, line 161: Do not show a header for the childless parent agencies

frontend/src/components/common/form/ParentSelect.tsx
```diff
@@ -165,15 +178,16 @@ export const ParentSelect: React.FC<IParentSelect> = ({
                   if (i.parent) {
                     return (
                       <MenuItem key={index + 1} option={i} position={index + 1}>
-                        <Highlighter search="">{i.label}</Highlighter>
+                        <Highlighter search="">
+                          {i.parentId === -1 ? <b>{i.label}</b> : i.label}
```
jdtoombs, line 182: Bold the childless parent agencies so they appear like the other parent agencies

Fosol (approved):
The grouping code looks confusing and probably isn't ideal to be scanning the array over and over for each agency, but I'm okay with it.

frontend/src/components/common/form/ParentSelect.tsx
jdtoombs: Made a variable instead of hard coding -1, so it is tagged as outdated but logic is the same.

frontend/src/components/common/form/ParentSelect.tsx
devinleighsmith: Does this sort everything by revers chronological then?

PR #1068, opened 2021-03-12
update frontend dependencies.

devinleighsmith:
support jest 26 and eslint 7.21


jdtoombs (approved):

devinleighsmith:
well, that was fun.

frontend/src/components/maps/leaflet/mapUtils.tsx
```diff
@@ -1,73 +1,82 @@
...
 import { ICluster, PointFeature } from '../types';
 import { IProperty } from 'actions/parcelsActions';
 import Supercluster from 'supercluster';
 import { Classifications, Workflows, PropertyTypes } from 'constants/index';
 
 // parcel icon (green)
-export const parcelIcon = new Icon({
-  iconUrl: require('assets/images/pins/land-reg.png'),
-  shadowUrl: require('assets/images/pins/marker-shadow.png'),
+export const parcelIcon = L.icon({
```
Fosol, line 8: Curious about this change?

Fosol (approved):

frontend/src/components/maps/leaflet/mapUtils.tsx
devinleighsmith: breaking change in react scripts changes how requires resolve. All requires resolve resources as modules, thus the need for .default. Also breaking change in JEST26 changes the default rootdir of test execution. this causes all of the requires to fail, so just returning the string url in that case.

PR #1072, opened 2021-03-15
disable clustering until after zoom is done

goemen:
Allow zoom to the map results after search and avoid cluster splitting before zoom is done.

### DIFF
--- a/frontend/src/components/maps/hooks/useSupercluster.ts
+++ b/frontend/src/components/maps/hooks/useSupercluster.ts
@@ -5,11 +5,15 @@ import deepEqual from 'dequal';
 import Supercluster from 'supercluster';
 import { ICluster } from '../types';
 
+interface SuperclusterOptions<P, C> extends Supercluster.Options<P, C> {
+  /** Optionally enable clusters recalculation */
+  enableClustering?: boolean;
+}
 export interface UseSuperclusterProps<P, C> {
   points: Array<Supercluster.PointFeature<P>>;
   bounds?: BBox;
   zoom: number;
-  options?: Supercluster.Options<P, C>;
+  options?: SuperclusterOptions<P, C>;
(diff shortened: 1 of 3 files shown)


frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -290,6 +284,7 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
   );
 
   const keycloak = useKeycloakWrapper();
+  console.log(clusters);
```
devinleighsmith, line 287: whoops.

frontend/src/components/maps/hooks/useSupercluster.ts
```diff
@@ -9,7 +9,7 @@ export interface UseSuperclusterProps<P, C> {
   points: Array<Supercluster.PointFeature<P>>;
   bounds?: BBox;
   zoom: number;
-  options?: Supercluster.Options<P, C>;
+  options?: Supercluster.Options<P, C> & { enableClustering?: boolean };
```
Fosol, line 12: I'm not familiar with the syntax.  

My guess below.
- options can be `undefined`
- options is a `Supercluter.Options<P, C>` and must also have a `enableClustering` property that can be `undefined` or `boolean`.

If the above is true, I'm not sure I understand the complexity in the statement?  Wouldn't it be easier to understand by creating an interface with the appropriate inheritance?

Fosol (approved):

frontend/src/components/maps/hooks/useSupercluster.ts
goemen: This is just a shortcut, will switch to using an interface with inheritance.

devinleighsmith (approved):

PR #1074, opened 2021-03-15
Correct some associated land issues.

devinleighsmith:
1) ensure that default agency is populated.
2) if there are multiple new parcels, do not take non-added parcel into account.
3) fix possible issue with bad associated building data.


backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -251,7 +251,7 @@ public Building Update(Building building)
             foreach (var parcel in building.Parcels.Select(pb => pb.Parcel))
             {
                 // Check if the parcel already exists.
-                var existingAssociatedParcel = existingBuilding.Parcels
+                var existingAssociatedParcel = existingBuilding.Parcels.Where(pb => pb.Parcel.Id != 0)
```
devinleighsmith, line 254: in the case of multiple new parcels, this logic was erroneously allowing pending parcels that have yet to be added to be treated as existing.

backend/dal/Services/Concrete/ParcelService.cs
```diff
@@ -218,7 +218,7 @@ public Parcel Add(Parcel parcel)
                 pp.Subdivision = this.Context.Parcels.Find(pp.ParcelId);
             });
 
-            parcel.Buildings.ForEach(pb =>
+            parcel.Buildings.Where(pb => pb.Building != null).ForEach(pb =>
```
devinleighsmith, line 221: I ran into a case where bad data in my local was causing this to occur.

Fosol (approved):

jdtoombs (approved):

PR #1075, opened 2021-03-16
PA-2895 update maintenance page to support SPA style routing.

devinleighsmith:


Fosol (approved):
Finally!  Thanks, I couldn't find the right syntax when I tried to solve this with an hour or two a month ago.

jdtoombs (approved):

PR #1080, opened 2021-03-16
PA-3002

devinleighsmith:
Ensures that requests to refresh the map only occur when absolutely necessary (ie during an update or a delete).
Also always makes a request to get the full details of the property.


frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
```diff
@@ -243,6 +249,31 @@ const InfoControl: React.FC<InfoControlProps> = ({ open, setOpen, onHeaderAction
             id="slideOutInfoButton"
             variant="outline-secondary"
             onClick={() => {
+              const propertyTypeId = popUpContext.propertyTypeID;
```
Fosol, line 252: Not sure if the `popUpContext.propertyTypeID` is correct case on `ID`.  Might be fine.

Fosol (approved):

frontend/src/components/maps/leaflet/InfoSlideOut/InfoSlideOut.tsx
devinleighsmith: it is in this case (had it the other way originally and got a compile warning).

goemen (approved):
Very clean

PR #1081, opened 2021-03-16
Validation bug

jdtoombs:
Now that these values are initialized to false we need to check if one of them is true now instead of the default required check. 
validationbug

### DIFF
--- a/frontend/src/components/layout/Header/Header.tsx
+++ b/frontend/src/components/layout/Header/Header.tsx
@@ -98,7 +98,7 @@ const Header = () => {
             src={BClogoUrl}
             width="156"
             height="43"
-            alt="Go to the Government of British Columbia website"
+            alt="Government of BC logo"
           />
         </a>
         <VerticalBar />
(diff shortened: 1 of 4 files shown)


frontend/src/components/layout/Header/Header.tsx
```diff
@@ -98,7 +98,7 @@ const Header = () => {
             src={BClogoUrl}
             width="156"
             height="43"
-            alt="Go to the Government of British Columbia website"
+            alt="Government of BC logo"
```
jdtoombs, line 101: As per Bobbi's recommendation

frontend/src/components/maps/leaflet/LayersControl/data.ts
```diff
@@ -76,6 +76,7 @@ export const layersTree: ILayerItem[] = [
         format: 'image/png',
```
jdtoombs, line 76: Adding colour for legend when applicable

Fosol (approved):

devinleighsmith (approved):

PR #1086, opened 2021-03-17
Bug fixes for submitting on behalf on child agency

jdtoombs:
Validation now more detailed for type error if it occurs, and allow REM's to submit on behalf of a child agency if they are a part of a parent agency.

### DIFF
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -147,7 +147,7 @@ namespace Pims.Dal.Services
                     && (isAdmin
                         || b.IsVisibleToOtherAgencies
                         || ((!b.IsSensitive || viewSensitive)
-                            && (userAgencies.Contains(b.AgencyId) || userAgencies.Contains(b.Agency.ParentId))))) ?? throw new KeyNotFoundException();
+                            && userAgencies.Contains(b.AgencyId)))) ?? throw new KeyNotFoundException();
 
             return building;
         }
@@ -167,7 +167,7 @@ namespace Pims.Dal.Services
 
             // A building should have a unique name within the parcel it is located on.
             building.Parcels.ForEach(pb => this.Context.ThrowIfNotUnique(pb.Parcel, building));
-            // SRES users allowed to overwrite, or if user belongs to parent ministry
+            // If the user is not an admin, and their agency is not a parent override to their user agency
             if (!this.User.HasPermission(Permissions.AdminProperties) && agency.ParentId != null)
             {
                 building.AgencyId = agency.Id;
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -147,7 +147,7 @@ namespace Pims.Dal.Services
                     && (isAdmin
                         || b.IsVisibleToOtherAgencies
                         || ((!b.IsSensitive || viewSensitive)
-                            && userAgencies.Contains(b.AgencyId)))) ?? throw new KeyNotFoundException();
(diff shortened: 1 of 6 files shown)


backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -167,8 +167,8 @@ public Building Add(Building building)
 
             // A building should have a unique name within the parcel it is located on.
             building.Parcels.ForEach(pb => this.Context.ThrowIfNotUnique(pb.Parcel, building));
-            // SRES users allowed to overwrite
-            if (!this.User.HasPermission(Permissions.AdminProperties))
+            // SRES users allowed to overwrite, or if user belongs to parent ministry
+            if (!this.User.HasPermission(Permissions.AdminProperties) && agency.ParentId != null)
```
jdtoombs, line 171: Want to confirm this isn't a security issue, need to allow parent ministries to override their user agency to submit on behalf of another

frontend/src/components/Table/ColumnFilter.tsx
```diff
@@ -21,7 +21,6 @@ const Wrapper = styled('div')`
   align-items: center;
   justify-items: center;
   position: relative;
-  font-size: 12px;
```
jdtoombs, line 24: This just keeps the text for the filterable columns the same size as the rest, other option is we reduce the size of the other ones to 12.

backend/dal/Services/Concrete/BuildingService.cs
Fosol: If I'm reading this correctly, you allow a **SRES** user or a user who is assigned to a child agency to modify the building's owning agency.  This would be invalid for two reasons.
1) Any user belonging to a parent agency can change the agency of the property **as long as it is an agency they are assigned**
2) A user with `admin-properties` claim can change the agency to anything

backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -147,7 +147,7 @@ public Building Get(int id)
                     && (isAdmin
                         || b.IsVisibleToOtherAgencies
                         || ((!b.IsSensitive || viewSensitive)
-                            && userAgencies.Contains(b.AgencyId)))) ?? throw new KeyNotFoundException();
+                            && (userAgencies.Contains(b.AgencyId) || userAgencies.Contains(b.Agency.ParentId))))) ?? throw new KeyNotFoundException();
```
Fosol, line 150: This doesn't look valid.  The user's claims should include all agencies they have access to.  Ensure your claims are correct for the user you are testing.

Fosol (changes requested):

Fosol:
Wouldn't you also need to update the `ParcelService` if there was an issue adding buildings to other agencies?  Additionally would you also need to update all the `Update` methods?

backend/dal/Services/Concrete/BuildingService.cs
jdtoombs: This is saying it will override it to the user's agency if the user is not an admin and their agency is not a parent agency, originally it always overrided to the user's agency but a while ago I added the check to only do so if they are not an admin based off an older story to allow SRES to update. Will update the documentation so it's less confusing

backend/dal/Services/Concrete/BuildingService.cs
jdtoombs: Ah must of been how I had a user set up locally, thank you

jdtoombs: > Wouldn't you also need to update the `ParcelService` if there was an issue adding buildings to other agencies? Additionally would you also need to update all the `Update` methods?

Update looks like it has the correct implementation, checks if `userAgencies` contains the original agency or they are an admin. Which should be fine if the users claims include all agencies they have access to. But I do need to add it to `ParcelService.cs`

devinleighsmith (approved):

Fosol (approved):

PR #1087, opened 2021-03-18
add tests to increase coverage.

devinleighsmith:
fix legend markers.
fix deleting agencies with users.


backend/dal/Services/Admin/Concrete/AgencyService.cs
```diff
@@ -207,7 +207,7 @@ public override void Remove(Agency agency)
         {
             agency.ThrowIfNull(nameof(agency));
 
-            var original = this.Context.Agencies.Find(agency.Id) ?? throw new KeyNotFoundException();
+            var original = this.Context.Agencies.Include(a => a.Users).FirstOrDefault(a => a.Id == agency.Id) ?? throw new KeyNotFoundException();
```
devinleighsmith, line 210: not including the useragencies causes the cascade to fail.

frontend/src/components/maps/leaflet/Legend/Legend.tsx
```diff
@@ -6,23 +6,23 @@ export const Legend = () => {
   const keys = React.useMemo(() => {
     return [
       {
-        pin: require('assets/images/pins/land-reg.png'),
+        pin: require('assets/images/pins/land-reg.png').default,
```
devinleighsmith, line 9: react scripts 4 loads requires images as modules.

devinleighsmith:
also a lot of the test warnings have been re-introduced, will fix that in a follow-up.

backend/dal/Services/Admin/Concrete/AgencyService.cs
Fosol: The downside to this is `include(...)` statements are expensive operations.

Fosol (approved):

goemen (approved):

backend/dal/Services/Admin/Concrete/AgencyService.cs
devinleighsmith: this should be fine since it is finding by an indexed id.

jdtoombs (approved):

PR #1091, opened 2021-03-18
PA-3009 CHES Domain Change

Fosol:


backend/ches/ChesService.cs
```diff
@@ -224,8 +224,8 @@ public async Task<EmailResponseModel> SendEmailAsync(IEmail email)
 
             // Make sure there are no blank CC or BCC;
             email.To = email.To.NotNullOrWhiteSpace();
-            email.Cc = email.Cc.NotNullOrWhiteSpace();
-            email.Bcc = email.Bcc.NotNullOrWhiteSpace();
+            email.Cc = email.Cc?.NotNullOrWhiteSpace();
```
Fosol, line 227: Fix a bug when passing null values to these properties.

devinleighsmith: bcc and cc get this but not To? is to always provided?

devinleighsmith (approved):

jdtoombs (approved):

backend/ches/ChesService.cs
Fosol: If they don't provide a `to` it would have thrown an argument exception earlier.

PR #1096, opened 2021-03-19
Fix accidental globcal scss. Remove leaflet-draw cdn call.

devinleighsmith:


frontend/public/index.html
```diff
@@ -1,21 +1,19 @@
...
+    <link rel="icon" href="%PUBLIC_URL%/favicon-32x32.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta name="description" content="Property Inventory Management System" />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+    <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
-  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/leaflet.draw/1.0.4/leaflet.draw.css" />
```
devinleighsmith, line 15: we don't use leaflet draw, so we shouldn't be making cdn requests during init.

frontend/src/features/admin/edit-user/EditUserPage.scss
```diff
@@ -1,41 +1,41 @@
 @import '../../../variables';
-.userInfo {
-  width: 40%;
-  vertical-align: middle;
-  text-align: left;
-  display: table;
-}
+.user-edit-form-container {
```
devinleighsmith, line 2: this scss was applying to the app globally since this isn't a css module. Manually only applying these rules to children of a certain class prevents this.

jdtoombs (approved):

Fosol (approved):

PR #1102, opened 2021-04-26
Pa 8 - Database migration to fix 3 notification templates

LawrenceLau2020:
There is a syntax issue with three notification templates in PROD (template IDs 2, 3, & 4), and 10 in the TEST environment.
Just creating a database migration to resolve these templates.


Fosol (approved):
Make sure you locally test the scenario by entering rows with the production values.  Also test the `make db-rollback n=v01.09.0#` then `make db-update` to ensure rollback and roll forward work as expected.

PR #1108, opened 2021-06-02
Fix TRAN agencies

Fosol:
## Summary

I have rolled back DB migration `v01.10.04` in **DEV** and **TEST**.   This PR updates this DB migration so that when it is applied again it will do the following;

1) Change ownership of properties from "TRAN", "PLMB", "TIC" to "BCTFA"
2) Change ownership of projects from "TRAN", "PLMB", "TIC" to "BCTFA"
3) Move all TRAN users to the "TRAN" ministry instead of sub-agency
4) Delete sub-agencies "PLMB" and "TIC"

> Note that users will need to be updated manually after this migration so that they reference the correct agencies (see below).

## Note

The only issue in the prior bug fix was that it was also moving properties from "BC Transit" to "BC Transportation Financing Authority".  I have removed that part of the migration.

### DIFF
--- a/backend/dal/Migrations/v01.10.04/Up/PostUp/01-Agencies.sql
+++ b/backend/dal/Migrations/v01.10.04/Up/PostUp/01-Agencies.sql
@@ -1,5 +1,16 @@
 PRINT 'Updating TRAN Agencies'
 
+----------------------------------------------------------------
+-- Summary
+-- 1) Change ownership of properties from "TRAN", "PLMB", "TIC" to "BCTFA"
+-- 2) Change ownership of projects from "TRAN", "PLMB", "TIC" to "BCTFA"
+-- 3) Move all TRAN users to the "TRAN" ministry instead of sub-agency
+-- 4) Delete sub-agencies "PLMB" and "TIC"
+--
+-- Note that users will need to be updated manually after this
+-- migration so that they reference the correct agencies (see below).
+----------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


backend/dal/Migrations/v01.10.04/Up/PostUp/01-Agencies.sql
```diff
@@ -44,23 +55,23 @@ DROP TABLE #TRANUsers
 -- Move properties to BCTFA
 UPDATE dbo.[Parcels]
 SET [AgencyId] = @BCTFA
-WHERE [AgencyId] IN (@TRAN, @PLMB, @TIC, @BCT)
+WHERE [AgencyId] IN (@TRAN, @PLMB, @TIC)
```
LawrenceLau2020, line 58: I'm not sure if we need to update existing "TRAN" agencies to the subagency "BCTFA", as I thought Maggie mentioned only properties currently reporting as either parent agency "TRAN" and subagency PLMB or parent agency TRAN and subagency is blank or parent agency is TRAN and subagency is TIC should be updated to subagency BCTFA?

LawrenceLau2020 (approved):

backend/dal/Migrations/v01.10.04/Up/PostUp/01-Agencies.sql
LawrenceLau2020: I forgot that a property can belong to a subagency and a parent agency using one agency id.

PR #1111, opened 2021-06-10
Keep Agency on Dispose

Fosol:
Add DB migration v01.05.00

Previously properties agency was set to `null` when they were disposed.  Going forward they will retain their original agency.


LawrenceLau2020 (approved):

Sharala-Perumal (approved):

LawrenceLau2020:
Pushing PR to DEV

PR #1113, opened 2021-06-16
PSP-6 Fix Cancel Project Bug

Fosol:
# Summary

- Fix Cancel Project when in "Not in SPL"
- Add DB migration v01.10.06
- Modify developer environment configuration

For some reason the production database doesn't have all the allowed project status transitions.  This would indicate a migration wasn't run correctly in production at some point in time (I think v01.05.00).  It's unclear how that could have transpired.

To resolve this issue I've added another migration that will only apply the project status transitions if they don't exist.

## Environment Configuration Changes

Additionally to support running multiple instances of PIMS locally I have updated the developer environment configuration.  This will allow developers to customize the ports they are using.

Developers will need to do the following to get their environments up and running again.  Essentially remove their local environment and rebuild it.  By default it will continue to use the same ports as previously, but the docker images will have different names.

Create a `.env` file in the root with the following configuration values.

```conf
KEYCLOAK_PORT=8080
DATABASE_PORT=5433
API_HTTP_PORT=5000
API_HTTPS_PORT=5001
APP_HTTP_PORT=3000
```

```bash
make down
make up
make keycloak-sync
make db-seed
```


Makefile
```diff
@@ -43,7 +43,7 @@ refresh: | down build up ## Recreates local docker environment
 
 up: ## Runs the local containers (n=service name)
 	@echo "$(P) Running client and server..."
-	@docker-compose up -d $(n)
+	@docker-compose --env-file .env up -d $(n)
```
Fosol, line 46: Technically by default `docker-compose` will automatically look for a `.env` file in the same directory, but I've hardcoded it to ensure the user adds the file.

```conf
KEYCLOAK_PORT=8080
DATABASE_PORT=5433
API_HTTP_PORT=5000
API_HTTPS_PORT=5001
APP_HTTP_PORT=3000
```

backend/dal/Migrations/v01.10.06/Up/PostUp/01-ProjectStatusTransitions.sql
```diff
@@ -0,0 +1,71 @@
+PRINT 'Adding ProjectStatusTransitions'
+
+MERGE INTO dbo.[ProjectStatusTransitions] dest
```
Fosol, line 3: Only insert these project transitions if they don't already exist.

docker-compose.yml
```diff
@@ -12,37 +12,37 @@ services:
     volumes:
       - ./auth/keycloak/config/realm-export.json:/tmp/realm-export.json
     ports:
-      - 8080:8080
+      - ${KEYCLOAK_PORT:-8080}:8080
```
Fosol, line 15: Use the `.env` configuration or default values.

frontend/src/index.tsx
```diff
@@ -10,19 +10,43 @@ import React from 'react';
...
+/**
+ * Displays LoginLoading until Keycloak connection is ready.
+ * @returns Index component.
+ */
 const Index = () => {
-  return (
+  const [loading, setLoading] = React.useState(true);
+  const [keycloak, setKeycloak] = React.useState(Keycloak);
+
+  React.useEffect(() => {
```
Fosol, line 31: Provide a way to override the keycloak configuration file locally through environment variables.

frontend/src/utils/createKeycloakInstance.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @returns Promise<KeycloakInstance>
+ */
+export const createKeycloakInstance = async () => {
+  const response = await fetch('/keycloak.json');
+  const data = await response.json();
+  const config = data as IKeycloakConfig;
+
+  //@ts-ignore
+  return new Keycloak({
+    url: process.env.REACT_APP_KEYCLOAK_AUTH_SERVER_URL ?? config.authServerUrl,
```
Fosol, line 16: We can now override Keycloak configuration values locally with environment variables.

scripts/gen-env-files.sh
```diff
@@ -15,6 +15,19 @@ passvar=$(date +%s | sha256sum | base64 | head -c 29)A8!
 echo $passvar
 
 # Set environment variables.
+# Docker Compose
+if test -f "./.env"; then
```
Fosol, line 19: Add the `.env` for docker-compose`

LawrenceLau2020 (approved):
Looks good to me, Jeremy, but you might want to have a second reviewer who is more familiar with the frontend and some of the other files.

Sharala-Perumal:
Hi Jeremy, is it possible to get a developer on your side to get a second approval for this PR? Bruce is away today.

Fosol:
I’ve asked Devin if he can find some time.

Sharala-Perumal:
That would be great. Thanks.

docker-compose.yml
```diff
@@ -12,37 +12,37 @@ services:
...
     networks:
       - pims
 
   ####################### Database #######################
   database:
     restart: on-failure:1
-    container_name: api-db
+    container_name: pims-database
```
devinleighsmith, line 22: I'm always suspicious of doing renames like this, because if an old reference is left somewhere it could break something important.

Have you tried this with a fresh install?

frontend/src/utils/createKeycloakInstance.ts
devinleighsmith: Why is this desirable? this seems unrelated to the bugfix.

devinleighsmith (approved):

frontend/src/utils/createKeycloakInstance.ts
Fosol: It's unrelated to the bug, but required to support different port configurations.  Which is required to run two different versions of PIMS.

docker-compose.yml
Fosol: I have done a fresh install.

PR #1115, opened 2021-06-18
PIMS-6 Fix Cancel Notifications

Fosol:
When notification cancellations fail, so does cancelling a project.  Currently the main issue for this is because there are projects that created notifications in the prior version of CHES, which do not exist in the current version.

This fix will ignore the specific error response from CHES for missing notifications (404).  This isn't an ideal solution, but one that will resolve current issues.

I have notified the Common Services team of their design implementation issue regarding returning a 404 for a valid endpoint and a missing message.  They should be returning a 204, or a 200 with an empty array like they do for other endpoints.

### DIFF
--- a/backend/ches/ChesService.cs
+++ b/backend/ches/ChesService.cs
@@ -103,7 +103,7 @@ namespace Pims.Ches
             }
             catch (HttpClientRequestException ex)
             {
-                _logger.LogError(ex, $"Failed to send/receive request: ${url}");
+                _logger.LogError(ex, $"Failed to send/receive request: {ex.StatusCode} {url}");
                 var response = await this.Client?.DeserializeAsync<Ches.Models.ErrorResponseModel>(ex.Response);
                 throw new ChesException(ex, this.Client, response);
             }
@@ -131,7 +131,7 @@ namespace Pims.Ches
             }
             catch (HttpClientRequestException ex)
             {
-                _logger.LogError(ex, $"Failed to send/receive request: ${url}");
+                _logger.LogError(ex, $"Failed to send/receive request: {ex.StatusCode} {url}");
                 var response = await this.Client?.DeserializeAsync<Ches.Models.ErrorResponseModel>(ex.Response);
                 throw new ChesException(ex, this.Client, response);
             }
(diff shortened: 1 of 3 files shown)


backend/ches/ChesService.cs
```diff
@@ -103,7 +103,7 @@ private async Task<string> SendAsync(string endpoint, HttpMethod method)
             }
             catch (HttpClientRequestException ex)
             {
-                _logger.LogError(ex, $"Failed to send/receive request: ${url}");
+                _logger.LogError(ex, $"Failed to send/receive request: {ex.StatusCode} {url}");
```
Fosol, line 106: Adding status code to response and removing extra dollar sign.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -840,10 +840,36 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
...
-            return page.Items;
+            var notifications = new List<NotificationQueue>();
+            try
+            {
+                var qty = 100;
+                var pos = 1;
+                var total = 101;
+
+                // Page through pending notifications to cancel them.
+                while (pos * qty < total)
```
Fosol, line 851: This should page through as long as the total is greater than the amount returned.

backend/dal/Services/Concrete/ProjectService.cs
```diff
@@ -840,10 +840,36 @@ public async System.Threading.Tasks.Task<Project> SetStatusAsync(Project project
...
+                    total = page.Total;
+
+                    await _notifyService.CancelAsync(page);
+                    notifications.AddRange(page.Items);
+                }
+            }
+            catch (Exception ex)
+            {
+                this.Logger.LogError(ex, $"Failed to cancel all notifications.");
+                if (_options.Notifications.ThrowExceptions) throw;
```
Fosol, line 870: Only throw exceptions that will cause the project to fail to cancel if configured to do so.

backend/notifications/NotificationService.cs
```diff
@@ -171,7 +172,18 @@ public async Task CancelAsync(IEnumerable<Entity.NotificationQueue> notification
...
             {
-                await CancelAsync(notification);
+                try
+                {
+                    await CancelAsync(notification);
+                }
+                catch (HttpClientRequestException ex)
+                {
+                    // CHES returns a 404 when a message doesn't exist.  This isn't a good practice, but now I have to assume 404's are successful and I'm sending invalid message Ids.
+                    if (ex.StatusCode != HttpStatusCode.NotFound) throw;
```
Fosol, line 182: A 404 regrettably is just a missing message (caused by the CHES migration to OCP4).  We'll ignore these for now and continue cancelling all other messages.

LawrenceLau2020 (approved):

PR #1117, opened 2021-06-21
PIMS-18: Allow SPL tab to be accessible by other statuses 

jdtoombs:
Allow the SPL tab to be accessible by other statuses for the case of a project transitioning to another status and needing to view previous data.

- Width was slightly adjusted to accommodate for the extra tab (was previously stacking tabs when closeout form tab was included)

image


Fosol (approved):
Nice and simple.  Show the SPL tab all the time so that the user can see and edit the properties.

LawrenceLau2020 (approved):

PR #1118, opened 2021-06-22
Increase prod pipeline timeout

Fosol:


LawrenceLau2020 (approved):

LawrenceLau2020:
I'm assuming we don't need to modify the other Jenkinsfile.cicd pipeline?

Fosol:
We already did that one yesterday.

PR #1119, opened 2021-06-24
PIMS-21 - Project View for SPL Projects

LawrenceLau2020:


LawrenceLau2020:
I had a very strange issue earlier today, not sure what the cause was, had to do an npm-refresh twice as well as restart VSCode, hopefully it doesn't occur with the build.

bug

Fosol:
I run into issues every once in a while and have to recreate the docker container too.  Not 100% what causes it.

frontend/src/components/layout/AppNavBar/AppNavBar.tsx
```diff
@@ -136,6 +136,11 @@ function DisposeProjectsDropdown() {
           View Projects
         </NavDropdown.Item>
       )}
+      {(keycloak.hasClaim(Claims.PROJECT_VIEW) || keycloak.hasClaim(Claims.ADMIN_PROJECTS)) && (
```
Fosol, line 139: I think  you could have just included it in the above section with a `<><NavDropdown.Item/><NavDropdown.Item/></>`

frontend/src/features/projects/list/SPLProjectListView.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+import thunk from 'redux-thunk';
+import { ILookupCode } from 'actions/lookupActions';
+import * as API from 'constants/API';
+import { Provider } from 'react-redux';
+import * as reducerTypes from 'constants/reducerTypes';
+import service from '../apiService';
+import { noop } from 'lodash';
+import { Formik } from 'formik';
+
+const testData = {
```
Fosol, line 16: There may already be test data kept in a central location.  If there isn't might be a nice chance to make it.  Then other tests could reference this mocked data.

frontend/src/features/projects/list/SPLProjectListView.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+describe('Project Approval Request list view', () => {
+  // clear mocks before each test
+  beforeEach(() => {
+    mockedService.getProjectList.mockClear();
+    mockedService.deleteProject.mockClear();
+  });
+
+  afterEach(() => {
+    cleanup();
```
Fosol, line 129: Does this also clear the other mocks in this file like the `apiService`?

Fosol (approved):

frontend/src/features/projects/list/SPLProjectListView.test.tsx
LawrenceLau2020: I'm not sure, how can I test that it does that?

LawrenceLau2020: a question for Devin, perhaps, as I basically copied the ProjectApprovalRequest.test.tsx file.

LawrenceLau2020: Should we add `mockedService.getProjectList.mockClear();
    mockedService.deleteProject.mockClear();` after the cleanup() in the afterEach() method?

frontend/src/features/projects/list/SPLProjectListView.test.tsx
LawrenceLau2020: Sounds like a good idea, but is there already a central location with test data that I can use, if not do we need to implement this in this ticket?

frontend/src/features/projects/list/SPLProjectListView.test.tsx
Fosol: @devinleighsmith Can you provide insight here?

frontend/src/features/projects/list/SPLProjectListView.test.tsx
Fosol: You don't need to implement in this ticket.  Just feedback on ways we might be able to improve.  If there is a central location for keeping this stuff already then it would make sense to use it though.

jdtoombs (approved):

PR #1120, opened 2021-06-25
SPP Number on Inventory Export

LawrenceLau2020:
There might be a better way to get the project number into the report without the brackets, as it's currently an array.
[pims-inventory (11).xlsx](https://github.com/bcgov/PIMS/files/6718314/pims-inventory.11.xlsx)

### DIFF
--- a/backend/api/Areas/Reports/Mapping/Property/PropertyMap.cs
+++ b/backend/api/Areas/Reports/Mapping/Property/PropertyMap.cs
@@ -1,5 +1,7 @@
 using Mapster;
 using Entity = Pims.Dal.Entities;
+using System.Text.Json;
+using System.Collections.Generic;
 using Model = Pims.Api.Areas.Reports.Models.Property;
 
 namespace Pims.Api.Areas.Reports.Mapping.Property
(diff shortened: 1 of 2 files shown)


backend/api/Areas/Reports/Mapping/Property/PropertyMap.cs
```diff
@@ -39,7 +45,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.RentableArea, src => src.RentableArea)
                 .Map(dest => dest.OccupantName, src => src.OccupantName)
                 .Map(dest => dest.LeaseExpiry, src => src.LeaseExpiry)
-                .Map(dest => dest.TransferLeaseOnSale, src => src.TransferLeaseOnSale);
+                .Map(dest => dest.TransferLeaseOnSale, src => src.TransferLeaseOnSale)
+                .Map(dest => dest.ProjectNumbers, src => src.ProjectNumbers);
```
Fosol, line 49: If you're not deserializing it probably don't need to include the `_serializerOptions`

jdtoombs (approved):

Fosol (approved):

PR #1123, opened 2021-06-29
Fix Remove from SPL fields not appearing

jdtoombs:
An out of place bracket was breaking the logic


LawrenceLau2020 (approved):
Looks good to me!

Fosol (approved):

PR #1125, opened 2021-07-05
PIMS-27 Change Approval Email

Fosol:
The business wants approved projects to send a notification to the user who created the project.  Currently it sends it to the owning agency, which is causing confusion.

Testing this is complicated because we override emails in DEV and TEST to always go to the user who generated them.  Locally I have changed my environment configuration so that it works correctly.  

> DB Migration Version v01.11.00

## Testing

`/backend/api/.env`

```conf
Ches__EmailAuthorized=true
```

Make sure you update all agencies and notification templates in the DB to have an appropriate email address.  I turned off all emails for every agency other than the one I was testing.  Then I changed their email to one I could use.  I also updated notification templates with [To] email address to on I could use.


backend/dal/Migrations/v01.11.00/Down/PostDown/01-NotificationTemplates.sql
```diff
@@ -0,0 +1,5 @@
+PRINT N'Update [NotificationTemplates]'
+
+UPDATE dbo.[NotificationTemplates]
```
devinleighsmith, line 3: This doesn't undo the below notification template changes, does it need to?

devinleighsmith (approved):

backend/dal/Migrations/v01.11.00/Down/PostDown/01-NotificationTemplates.sql
Fosol: Doesn't need to in this case.

jdtoombs (approved):

PR #1128, opened 2021-07-12
PIMS-25: Issue with ERP Notifications

LawrenceLau2020:


backend/dal/Migrations/v01.11.02/Up/PostUp/01-NotificationTemplates.sql
```diff
@@ -0,0 +1,16 @@
+PRINT 'Updating NotificationTemplates'
+
+-- to fix the single case for template 1
+UPDATE dbo.[NotificationTemplates]
+SET [Body] = REPLACE([Body], '@property.Parcel.Address?.ToString()', '@property.Parcel.Address.Address1.ToString()')
```
Fosol, line 5: If you wanted to do this without a DB migration, you could update the `Address` model/entity and override the `.ToString()` method to return an appropriate value.

LawrenceLau2020: That actually sounds like a better way of doing it....just wondering in the original code, why only the parcel made use of the .ToString() method of the Address entity, the building didn't.

Fosol: Not certain, most likely a mistake, or different people working on it.

Fosol (approved):

jdtoombs (approved):

PR #1134, opened 2021-07-21
fixes highlighting polygon matching a spiderfied marker.

devinleighsmith:


frontend/src/components/maps/leaflet/PointClusterer.tsx
```diff
@@ -378,8 +379,8 @@ export const PointClusterer: React.FC<PointClustererProps> = ({
               //sets this pin as currently selected
               const convertedProperty = convertToProperty(
                 m.properties,
-                m.position.lat,
-                m.position.lng,
+                getIn(m.geometry.coordinates, '1') ?? m.position.lat,
```
devinleighsmith, line 382: m.position is the position of the marker, however, during spiderfication the position of the marker on the map is offset from the actual lat/lng of the property. Using the actual coordinates from the geometry correctly highlights the property.

devinleighsmith: using getIn to avoid throwing an exception if geometry is not populated for some reason.

jdtoombs (approved):

Fosol (approved):
Thanks for this

PR #1135, opened 2021-07-22
PIMS-32: Change Description for the Tier's

LawrenceLau2020:
Removing net from tier tool tips


LawrenceLau2020:
I only changed the tooltips strings, but there are also references to the "net" value in the field "Descriptions" in the database table TierLevels.

Fosol (approved):

jdtoombs (approved):

PR #1136, opened 2021-07-22
PIMS-30: Issue with Fiscal Year reporting on excel extract

LawrenceLau2020:


backend/api/Areas/Reports/Models/Project/ProjectModel.cs
```diff
@@ -27,12 +27,12 @@ public class ProjectModel
         /// <summary>
         /// get/set - The reported fiscal year this project.
         /// </summary>
-        public int ReportedFiscalYear { get; set; }
+        public string ReportedFiscalYear { get; set; }
```
Fosol, line 30: Any reason for the string requirement?  Isn't it just an `int`?

Fosol (approved):

backend/api/Areas/Reports/Models/Project/ProjectModel.cs
LawrenceLau2020: The output for the reported fiscal year is actually a string: "21/22"

LawrenceLau2020: Ok, it looks like my change has affected some of the backend tests....

jdtoombs (approved):

LawrenceLau2020:
Please review my changes when you have a chance. Because the project model stores the fiscal year as an integer, I created a string field for the ReportedFiscalYear and the ActualFiscalYear.

I also had to update the FiscalYear method in the DateExtensions class, to handle cases where the fiscalYear was being passed in as 0 in some test cases.

backend/core/Extensions/DateExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static int GetFiscalYear(this DateTime date)
         /// <returns></returns>
         public static string FiscalYear(this int fiscalYear)
         {
-            return $"{(fiscalYear - 1).ToString().Substring(2, 2)}/{(fiscalYear).ToString().Substring(2, 2)}";
+            return fiscalYear > 0 ? $"{(fiscalYear - 1).ToString().Substring(2, 2)}/{(fiscalYear).ToString().Substring(2, 2)}" : "";
```
LawrenceLau2020, line 36: During debugging, it seems that one of the tests which instantiates a new project, passes in the fiscalYear as 0 which caused this to fail.

backend/api/Areas/Reports/Mapping/Project/ProjectMap.cs
```diff
@@ -34,6 +35,8 @@ public void Register(TypeAdapterConfig config)
                 .Map(dest => dest.Description, src => src.Description)
                 .Map(dest => dest.ReportedFiscalYear, src => src.ReportedFiscalYear)
                 .Map(dest => dest.ActualFiscalYear, src => src.ActualFiscalYear)
+                .Map(dest => dest.ReportedFiscalYearString, src => src.ReportedFiscalYear.FiscalYear())
```
LawrenceLau2020, line 38: I needed to add 2 new fields to the mapping, so now the report will show both the fiscal year as well as the formatted fiscal year "string" for both the reported and actual fiscal year.

backend/core/Extensions/DateExtensions.cs
Fosol: You probably could have just updated the test to set the default value of "20/21".

Fosol (approved):

PR #1139, opened 2021-08-09
PIMS-39: Updating 3 templates to project owner

LawrenceLau2020:
I updated 3 templates for the 30, 60, & 90 day notifications from "owning agency" to "project owner".

After submitting a project locally, it seemed to work, but when I try to view the project I'm getting an error now....not sure if it's related to this yet or not, as my local PIMS seems to be acting a bit slow right now.

image


Fosol (approved):
I tested locally and didn't have any issues show up.  Might be your local setup for some reason.

LawrenceLau2020:
It must have been a temporary network issue, I'm not getting the error anymore.
So, if we don't need a roll back script, I'm going to push to dev....

LawrenceLau2020:
Also, the "Owning Agency" audience is not being used on any of the templates.

PR #1140, opened 2021-08-11
PIMS-42: The project initiator doesn't seem to be getting the "Project Owner" notifications

LawrenceLau2020:
It looks like there's a bug which occurs if someone who is not the "initiator" or creator of the project approves the project, the CreatedById gets overwritten by the updater of the project.

I think there was a copy and paste error in the code...

### DIFF
--- a/backend/dal/Helpers/Extensions/ProjectExtensions.cs
+++ b/backend/dal/Helpers/Extensions/ProjectExtensions.cs
@@ -400,7 +400,7 @@ namespace Pims.Dal.Helpers.Extensions
             var updatedById = originalProject.UpdatedById;
             context.Entry(originalProject).CurrentValues.SetValues(updatedProject);
             originalProject.Agency = agency; // Don't want to allow agency to change through this method.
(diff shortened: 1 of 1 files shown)


backend/dal/Helpers/Extensions/ProjectExtensions.cs
```diff
@@ -400,7 +400,7 @@ public static void Merge(this Entity.Project originalProject, Entity.Project upd
             var updatedById = originalProject.UpdatedById;
             context.Entry(originalProject).CurrentValues.SetValues(updatedProject);
             originalProject.Agency = agency; // Don't want to allow agency to change through this method.
-            originalProject.CreatedById = updatedById; // Don't want these updated externally.
```
LawrenceLau2020, line 403: the updatedById used to be createdById, I think it was a mistake?
This seems to only occur on a project approval, I couldn't get it to happen when just updating a project.

Fosol: You're probably correct.

Fosol (approved):

PR #1145, opened 2021-09-02
PIMS-34: Manage administrative areas

jdtoombs:
Manage administrative areas via the administration section.

Originally was setting it up to be able to manage abbreviation, boundary type etc. Backend is set up for so but right now frontend is just to change the name. Let me know if we want the extra fields. 

Kept row click as the way to open/edit administrative areas to stay consistent with the application, tooltip in the header advises users that clicking the row will allow you to edit corresponding admin area.

manageadminareas

Update - Duplication Check:
duplicate


LawrenceLau2020:
Thanks Johnny! Just wondering what happens if you try entering an existing administrative area?

jdtoombs:
> Thanks Johnny! Just wondering what happens if you try entering an existing administrative area, will it detect that the name is already being used and prevent a duplicate from being created?

@LawrenceLau2020 There is currently no duplicate check but I can add something in that checks the existing names!

jdtoombs:
@LawrenceLau2020 
Updated with a duplication check (see second GIF in description) - working on some unit tests too. Just wanted to get the code up for review in the meantime!

LawrenceLau2020 (approved):
Look great from my limited perspective.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        #region Endpoints
+
+        /// <summary>
+        /// GET - Returns a paged list of administrative areas from the datasource.
+        /// </summary>
+        /// <returns>Paged object with an area of administrative areas.</returns>
+        [HttpGet]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Api.Models.PageModel<Model.AdministrativeAreaModel>), 200)]
```
Fosol, line 60: Says you're returning a Page of AdministrativeArea, but you are returning an array.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        /// <summary>
+        /// GET - Returns a agency for the specified 'id' from the datasource.
+        /// </summary>
+        /// <param name="id">The unique 'id' for the admin area to return.</param>
+        /// <returns>The agency requested.</returns>
+        [HttpGet("{id}")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Api.Models.PageModel<Model.AdministrativeAreaModel>), 200)]
```
Fosol, line 76: Says you're returning a Page of AdministrativeArea but you are returning one.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// <summary>
+        /// Check if name of admin area is available
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("check/name-available")]
+        [HasPermission(Permissions.SystemAdmin)]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(Model.CheckAreaAvailabilityModel), 200)]
+        [SwaggerOperation(Tags = new[] { "admin-administrative-area" })]
+        public IActionResult IsAdministrativeAreaAvailable(string name, int id)
```
Fosol, line 168: Don't really understand why the `id` is also included in this endpoint?

backend/dal/Services/Admin/Concrete/AdministrativeAreaService.cs
```diff
@@ -45,6 +48,47 @@ public AdministrativeArea Get(string name)
...
+        /// </summary>
+        /// <param name="page"></param>
+        /// <param name="quantity"></param>
+        /// <returns></returns>
+        public Paged<AdministrativeArea> Get(int page, int quantity)
+        {
+            return Get(new AdministrativeAreaFilter(page, quantity));
+        }
+
+        public Paged<AdministrativeArea> Get(AdministrativeAreaFilter filter = null)
```
Fosol, line 69: Should include documentation

backend/dal/Services/Admin/Concrete/AdministrativeAreaService.cs
```diff
@@ -45,6 +48,47 @@ public AdministrativeArea Get(string name)
                 .FirstOrDefault(c => c.Name == name);
         }
 
+        public AdministrativeArea Get(int id)
```
Fosol, line 51: Should include documentation

backend/dal/Services/Admin/Concrete/AdministrativeAreaService.cs
```diff
@@ -77,13 +121,37 @@ public override void Update(AdministrativeArea entity)
...
 
             this.Context.Entry(city).CurrentValues.SetValues(entity);
             base.Remove(city);
         }
+
+        /// <summary>
+        /// Add the specified administrative area to the datasource.
+        /// </summary>
+        /// <param name="entity"></param>
+        public override void Add(AdministrativeArea entity)
```
Fosol, line 137: Test if concurrency works correctly for Add, Update, and Delete.

Fosol (changes requested):
A great start

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,180 @@
...
+using Model = Pims.Api.Areas.Admin.Models.AdministrativeArea;
+using EModel = Pims.Dal.Entities.Models;
+using Pims.Dal.Keycloak;
+
+namespace Pims.Api.Areas.Admin.Controllers
+{
+    /// <summary>
+    /// AdministrativeArea class, provides endpoints for managing administrative areas.
+    /// </summary>
+    [HasPermission(Permissions.AgencyAdmin)]
```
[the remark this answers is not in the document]
jdtoombs: Good catch

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
jdtoombs: I was wondering if this was necessary but for example if you were updating an already existing area and it ran the check for duplication, would it say this name isn't available since it would be comparing against itself at one point? Or am I overthinking this one? I guess it is not that big of a deal either since there is one field to edit at the moment

Fosol: hmmm, you have a point.  You're essentially asking if the name exists but isn't the current administrative area you are editing.  Hmmm, it's a bit of an ugly implementation.  I would actually recommend you don't do this validation at all.  Just make sure the table has a unique constraint on the name.  Then if they try it the DB will throw an error.  You can then inform the user of the unique requirement failure after they attempt to save it.

I also suspect that you will need to update the state with a new list of administrative areas any time the user add/delete/update them.

You could also check the state that contains all the current administrative areas to see if the name has already been used.  If you want a quick user experience.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
Fosol: Doesn't look like you fixed this.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
Fosol: Doesn't look like you resolved this.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,187 @@
...
+            try
+            {
+                _pimsAdminService.AdministrativeArea.Update(entity);
+
+                var adminArea = _mapper.Map<Model.AdministrativeAreaModel>(entity);
+                return new JsonResult(adminArea);
+            }
+            catch (DbUpdateException ex)
+            {
+                if (ex.InnerException?.Message.Contains("duplicate") ?? false)
```
Fosol, line 124: Probably a little broad only searching for a single word "duplicate", but for our use-case it'll be fine.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
```diff
@@ -0,0 +1,187 @@
...
+                _pimsAdminService.AdministrativeArea.Add(entity);
+
+                var administrativeArea = _mapper.Map<Model.AdministrativeAreaModel>(entity);
+
+                return CreatedAtAction(nameof(GetAdministrativeArea), new {id = administrativeArea.Id},
+                    administrativeArea);
+            }
+            catch (DbUpdateException ex)
+            {
+                if (ex.InnerException?.Message.Contains("duplicate") ?? false)
```
Fosol, line 158: Since it's the same code twice I'd make it an extension method.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
Fosol: The error says adding, but here you're updating.

backend/dal/Migrations/v01.12.02/Up/PreUp/01-AdministrativeAreas.sql
```diff
@@ -0,0 +1,4 @@
+PRINT 'Removing duplicate salmon arm'
+
+DELETE FROM dbo.[AdministrativeAreas]
+WHERE [Id] = 248 -- Salmon Arm Tsey Key duplicate
```
Fosol, line 4: You should probably ensure any possible duplicates are dealt with.  There is no guarantee the other environment DBs have the same data as you do locally.
Additionally, make sure nothing references these as a foreign key.  I don't think they do, I think we use it only as a list.

backend/dal/Services/Admin/Concrete/AdministrativeAreaService.cs
```diff
@@ -45,6 +48,57 @@ public AdministrativeArea Get(string name)
...
 
+        /// <summary>
+        /// Get the administrative area for the specified id.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        public AdministrativeArea Get(int id)
+        {
+            return this.Context.AdministrativeAreas
+                .AsNoTracking()
```
Fosol, line 59: You should generally default to `AsNoTracking()` when making a request for a lot of rows which will only be used as readonly.  When requesting one you probably should default to using tracking.

frontend/src/features/admin/admin-areas/constants.ts
```diff
@@ -0,0 +1,55 @@
+/** interface containing the structure of data found in administrative areas */
+export interface IAdministrativeArea {
```
Fosol, line 2: Make files for each interface.

frontend/src/features/admin/admin-areas/constants.ts
```diff
@@ -0,0 +1,55 @@
...
+  type: string;
+}
+
+/** required interface for adding a new administrative area */
+export interface IAddAdminArea {
+  name: string;
+}
+
+/** generate administrative area ready for use in api endpoints */
+export const toApiAdminArea = (adminArea: IAdministrativeArea, name?: string): IApiAdminArea => {
```
Fosol, line 42: Move this to a `utils.ts` file in this folder.

frontend/src/features/admin/constants/index.tsx
```diff
@@ -35,3 +35,12 @@ export const columnDefinitions: ColumnWithProps<IAgencyRecord>[] = [
     },
   },
 ];
+
+export const adminAreasColumnDefinistions: ColumnWithProps<any>[] = [
```
Fosol, line 39: Don't place functions in an `index.ts` file.  Move them to a `utils.ts` file.  Index files should only be used to enable easy referencing of other files in the folder.

frontend/src/hooks/useApi.ts
```diff
@@ -1,4 +1,4 @@
-import CustomAxios from 'customAxios';
+import CustomAxios, { LifecycleToasts } from 'customAxios';
```
Fosol, line 1: Can you extract all your changes to a separate file called `useApiAdminAreas.ts`?  I'm in the middle of a massive refactor to support updates to projects.  This file is going to disappear sooner than later.

Fosol (changes requested):

frontend/src/hooks/useApi.ts
jdtoombs: Yup will do that!

frontend/src/features/admin/constants/index.tsx
jdtoombs: Woops yes this was a mistake

backend/dal/Migrations/v01.12.02/Up/PreUp/01-AdministrativeAreas.sql
jdtoombs: Hey! Just need to address this one still but knocked off the other requests. To clarify - do you mean add to this PreUp script to check for duplicates in the system?

Fosol: It can be in this script.  It doesn't need to be in the preup.

For example if you don't ensure there are no other duplicates.  Then the migration will fail.  You need to make sure there are no other possible duplicates.  I'd lean towards not hardcoding the delete for an ID.  As a practice the IDs could be different in each environment.

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
Fosol: Still does not look like you've resolved this.
`[ProducesResponseType(typeof(Api.Models.PageModel<Model.AdministrativeAreaModel>), 200)]`
doesn't match your return
`return new JsonResult(_mapper.Map<Model.AdministrativeAreaModel[]>(adminAreas));`

backend/api/Areas/Admin/Controllers/AdministrativeAreaController.cs
Fosol: Still does not look like you've resolved this.
`[ProducesResponseType(typeof(Api.Models.PageModel<Model.AdministrativeAreaModel>), 200)]`
doesn't match your return
`return new JsonResult(_mapper.Map<Model.AdministrativeAreaModel[]>(adminAreas));`

backend/dal/Migrations/v01.12.02/Up/PreUp/01-AdministrativeAreas.sql
```diff
@@ -0,0 +1,9 @@
+PRINT 'Removing duplicates'
+
+DELETE FROM dbo.[AdministrativeAreas]
+WHERE [Id] NOT IN (SELECT MAX(Id)
```
Fosol, line 4: If read it correctly, you're grouping by name and returning the max ID.  So if there are two of the items it will delete the first one.

frontend/src/features/admin/admin-areas/interfaces.ts
```diff
@@ -0,0 +1,39 @@
+/** interface containing the structure of data found in administrative areas */
+export interface IAdministrativeArea {
```
Fosol, line 2: Make a separate file for each interface so that it's easy to find them.

Fosol: If you want to keep everything clean, throw them in a folder called `interfaces` and add an `index.ts` file that points to each one.  A little extra boilerplate, but it makes it nice and clean.

frontend/src/utils/utils.ts
```diff
@@ -295,3 +296,28 @@ export function emptyStringToNull(value: any, originalValue: any) {
...
+ * Determine if the specified error is an axios error, improves intellisense.
+ * @param error The error being checked.
+ */
+export const isAxiosError = (err: any): err is AxiosError => {
+  if (!err) return false;
+  return (err as AxiosError).isAxiosError !== undefined;
+};
+
+/** generate administrative area ready for use in api endpoints */
+export const toApiAdminArea = (adminArea: IAdministrativeArea, name?: string): IApiAdminArea => {
```
Fosol, line 310: Doesn't look like a global util, more like a specific use-case for the Admin components.  I've move to a util file in the relevant area.

Fosol (changes requested):
Nearly there.

jdtoombs:
@Fosol Updated again 😄

frontend/src/features/admin/admin-areas/EditAdminArea.tsx
```diff
@@ -95,7 +95,8 @@ const EditAdminArea = (props: IEditAdminAreaProps) => {
             }
           } else if (!!values.name) {
             try {
-              await addAdminArea({ name: values.name });
+              const data = await addAdminArea({ name: values.name });
+              history.push(`/admin/administrativeArea/${data.id}`);
```
jdtoombs, line 99: Should change to "edit" state after area is created

Fosol (approved):

PR #1147, opened 2021-09-03
PIMS-46 Add DB Migration v01.12.01

Fosol:
Update makes it easier to copy a database from one environment into another (i.e. PROD to TEST).
Added `User.KeycloakUserId` to keep a reference to the Keycloak user.  Originally it was the `User.Id` that provided the reference, but this makes it more difficult to copy data from other environments.

DB Migration **v01.12.01**


LawrenceLau2020 (approved):

.gitignore
```diff
@@ -9,4 +9,4 @@ bash.exe.stackdump
 auth/keycloak/config/hsperfdata_jboss/
 auth/keycloak/config/*.log
 .DS_Store
-
+/backups
```
Fosol, line 12: Just to make sure the backups don't get added to source.

backend/entities/User.cs
```diff
@@ -14,6 +14,11 @@ public class User : BaseEntity
         /// </summary>
         public Guid Id { get; set; }
 
+        /// <summary>
+        /// get/set - Unique identifier to link user with Keycloak user.
+        /// </summary>
+        public Guid? KeycloakUserId { get; set; }
```
Fosol, line 20: Instead of using the `User.Id` it will now use the `User.KeycloakUserId` property to link the user to Keycloak.  This will allow copying data from other environments and make it easier to correct the link.

tools/keycloak/sync/realm.local.json
```diff
@@ -17,7 +17,10 @@
         "Protocol": "openid-connect",
         "BearerOnly": true,
         "AuthorizationServicesEnabled": false,
-        "RedirectUris": ["http://localhost:5000/api/*"],
+        "RedirectUris": [
+          "http://localhost:5000/api/*",
+          "http://localhost:61081/api/*"
```
Fosol, line 22: These changes make it easier for my local environment to work with the Keycloak Sync Tool.

jdtoombs (approved):

PR #1149, opened 2021-09-07
PIMS-46 Use KeycloakUserId

Fosol:
With the change to the User entity that now links to Keycloak through a separate property `KeycloakUserId` instead of the `Id`, additional changes needed to be made to support this.

DEV continued to work correctly because the `Id` and `KeycloakUserId` are identical values, however in TEST it will have different values based on what has been imported.

A better solution than the one I implemented would be to add the PIMS `User.Id` to the claims found in Keycloak to reduce the need to make an additional query/connection to the database for the `User.Id`, however this will take more time and requires additional updates to Keycloak.

### DIFF
--- a/backend/core/Extensions/IdentityExtensions.cs
+++ b/backend/core/Extensions/IdentityExtensions.cs
@@ -16,7 +16,7 @@ namespace Pims.Core.Extensions
         /// </summary>
         /// <param name="user"></param>
         /// <returns></returns>
-        public static Guid GetUserId(this ClaimsPrincipal user)
+        public static Guid GetKeycloakUserId(this ClaimsPrincipal user)
         {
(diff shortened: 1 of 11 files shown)


backend/core/Extensions/IdentityExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class IdentityExtensions
         /// </summary>
         /// <param name="user"></param>
         /// <returns></returns>
-        public static Guid GetUserId(this ClaimsPrincipal user)
+        public static Guid GetKeycloakUserId(this ClaimsPrincipal user)
```
Fosol, line 19: This value from the claim is the Keycloak User Id, not the PIMS Id now.

LawrenceLau2020 (approved):

PR #1151, opened 2021-09-10
PIMS-46 Fix Keycloak user update

Fosol:
When updating users with the PIMS User Admin UI, it fails to be able to update Keycloak because it is apparently using the wrong Id.  The mapping should be correct, but for some reason it appears like it may not be.  I've manually ensured the correct GUID is used and will need to test once released.

### DIFF
--- a/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
+++ b/backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
@@ -173,6 +173,7 @@ namespace Pims.Dal.Keycloak
 
             // Now update keycloak
             var kmodel = _mapper.Map<KModel.UserModel>(user);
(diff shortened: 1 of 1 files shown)


backend/dal.keycloak/Partials/PimsKeycloakUserService.cs
```diff
@@ -173,6 +173,7 @@ public async Task<Entity.User> UpdateUserAsync(Entity.User user)
 
             // Now update keycloak
             var kmodel = _mapper.Map<KModel.UserModel>(user);
+            kmodel.Id = user.KeycloakUserId.Value;
```
Fosol, line 176: I shouldn't have to do this, but the mapping doesn't appear to be correctly handling this.

PR #1154, opened 2021-09-21
PIMS-44: Change to PIMS Access Request Email

LawrenceLau2020:
Adding RPD IMIT Help email address to the "CC" field of the access request email template in the database.


backend/dal/Migrations/v01.12.03/Up/PostUp/01-NotificationTemplates.sql
```diff
@@ -0,0 +1,6 @@
+PRINT 'Updating NotificationTemplates'
+
+-- Include the CITZ_RPD_IMIT_HELP@gov.bc.ca email to the Access Request notification
+UPDATE dbo.[NotificationTemplates]
+SET [Cc] = 'CITZ_RPD_IMIT_HELP@gov.bc.ca'
```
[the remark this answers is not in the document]
LawrenceLau2020: Good question, I was going by Susan's notes in the ticket, but I'm kind of basing this on an assumption. She's been in meetings and hasn't replied yet, I'll wait for her reply and let you know.

Fosol:
Migration number will need to be updated `v01.13.01`

LawrenceLau2020:
> Migration number will need to be updated `v01.13.01`

I noticed that your migration didn't contain a v01.13.00 folder I'm assuming because there were no scripts to run, should I just create a v01.13.01 folder for my migration?

Fosol:
That is correct.  It was a physical change to the DB, but no scripts required.

LawrenceLau2020:
> That is correct. It was a physical change to the DB, but no scripts required.

I've created the migration update, not sure if we need a rollback?

backend/dal/Migrations/v01.13.01/Up/PostUp/01-NotificationTemplates.sql
```diff
@@ -0,0 +1,6 @@
+PRINT 'Updating NotificationTemplates'
+
+-- Include the CITZ_RPD_IMIT_HELP@gov.bc.ca email to the Access Request notification
+UPDATE dbo.[NotificationTemplates]
+SET [To] = 'CITZ_RPD_IMIT_HELP@gov.bc.ca'
```
Fosol, line 5: Make sure this doesn't result in sending multiple copies of this email when it attempts to perform this action.  I think it will because the code looks for all system admins and sends them an email.

LawrenceLau2020: It looks like I'll need to remove the migration as setting the email in that template will cause the RPD email to get multiple copies of the access request.

I can make a change to the api in the UserController.cs file and I guess I'll need to hardcode the email address into the AddAccessRequestAsync function.

Fosol: You shouldn't need to hardcode it.  Once you have updated the template (as you have in this PR) all you should need to do is change the code to only send one email instead of one for each system admin.

Fosol (changes requested):
You'll need a rollback script that removes the new template, or it will fail if we have to rollback and and then reapply.

Fosol (approved):

PR #1156, opened 2021-10-04
PIMS-44: Change to PIMS Access Request Email

LawrenceLau2020:
Added an overloaded method for GenerateNotification.

### DIFF
--- a/backend/dal/Services/Concrete/NotificationQueueService.cs
+++ b/backend/dal/Services/Concrete/NotificationQueueService.cs
@@ -403,6 +403,7 @@ namespace Pims.Dal.Services
         {
             if (template == null) throw new ArgumentNullException(nameof(template));
             if (model == null) throw new ArgumentNullException(nameof(model));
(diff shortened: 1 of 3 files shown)


backend/dal/Services/Concrete/NotificationQueueService.cs
```diff
@@ -389,6 +389,44 @@ public NotificationQueue GenerateNotification<T>(string to, NotificationTemplate
...
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="template"></param>
+        /// <param name="model"></param>
+        /// <param name="sendOn"></param>
+        /// <returns></returns>
+        public NotificationQueue GenerateNotification<T>(NotificationTemplate template, T model, DateTime? sendOn = null)
+        {
+            if (template == null) throw new ArgumentNullException(nameof(template));
+            if (model == null) throw new ArgumentNullException(nameof(model));
```
Fosol, line 405: Add a check to ensure it has a "to".
`if (String.IsNullOrEmpty(template.To)) throw new ArgumentException("Template requires a 'to'.", nameof(template));

Fosol (approved):

PR #1163, opened 2021-11-01
Only populate notification fields if notifications enabled

jdtoombs:
Simple condition added to ensure notification fields are not populated when notifications are disabled.


Fosol (approved):
I wonder if the obsolete warning will show up in every PR going forward?  I was hoping to postpone resolving this.

LawrenceLau2020 (approved):

PR #1164, opened 2021-11-01
PIMS-28: How Project Numbers display

LawrenceLau2020:
Changes to the property map and adding format extension to AllFieldsPropertyMap.
For the export to CSV, if a project contains multiple project numbers you need to "wrap text" on the project number column.

### DIFF
--- a/backend/core/Extensions/StringExtensions.cs
+++ b/backend/core/Extensions/StringExtensions.cs
@@ -104,5 +104,16 @@ namespace Pims.Core.Extensions
             if (string.IsNullOrEmpty(value)) return value;
             return value.Length <= maxLength ? value : value.Substring(0, maxLength);
         }
+
+        /// <summary>
+        /// Clean up the project numbers array format and get rid of double quotes, brackets, and commas.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static string FormatProjectNumbers(this string value)
+        {
+            if (string.IsNullOrEmpty(value)) return value;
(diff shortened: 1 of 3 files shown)


backend/core/Extensions/StringExtensions.cs
```diff
@@ -104,5 +104,16 @@ public static string Truncate(this string value, int maxLength)
...
+
+        /// <summary>
+        /// Clean up the project numbers array format and get rid of double quotes, brackets, and commas.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static string FormatProjectNumbers(this string value)
+        {
+            if (string.IsNullOrEmpty(value)) return value;
+            return value.Replace("\"", "").Replace("[", "").Replace("]", "").Replace(",", System.Environment.NewLine);
```
Fosol, line 116: Replacing "," with a newline probably won't look right when display in HTML.  Not certain what the best solution would be there.  Are you certain you want to format this in the API?  I'd think it would make more sense to format how it looks on the frontend app.

LawrenceLau2020: I don't think it affects the HTML view, but the newline was for the export to Excel to get the SPP#'s to display on separate lines in Excel.
image
[pims-inventory (52).xlsx](https://github.com/bcgov/PIMS/files/7454696/pims-inventory.52.xlsx)

Fosol:
I think it’s just wrapping because of the column size.  You could do the same thing on the frontend iterating through the array of project numbers.

Fosol (approved):
I see it looks like the API is returning a string originally.

PR #1166, opened 2021-11-02
PIMS-54

jdtoombs:
Transitions already enabled.

_**Transferred within GRE**_ already part of SPL grouping (#896), added **_On Hold_** to part of SPL grouping.

### DIFF
--- a/backend/dal/Migrations/v01.14.01/Down/PostDown/02-WorkflowProjectStatus.sql
+++ b/backend/dal/Migrations/v01.14.01/Down/PostDown/02-WorkflowProjectStatus.sql
@@ -1,0 +1,5 @@
+PRINT 'Delete WorkflowProjectStatus - SPL --> On Hold'
+
+DELETE FROM dbo.[WorkflowProjectStatus]
+WHERE [WorkflowId] = 6 -- SPL
+    AND [StatusId] = 31 -- On Hold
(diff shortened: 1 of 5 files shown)


Fosol:
We definitely want them to be grouped in SPL.

jdtoombs:
> We definitely want them to be grouped in SPL.

@Fosol 

Done, I did notice the _**PimsContextModelSnapshot.cs**_ had some weird behaviour after generating the migration - does this look correct to you?

backend/dal/Migrations/v01.14.01/Down/PostDown/02-WorkflowProjectStatus.sql
```diff
@@ -0,0 +1,5 @@
+PRINT 'Delete WorkflowProjectStatus - SPL --> On Hold'
+
+DELETE FROM dbo.[WorkflowProjectStatus]
```
Fosol, line 3: We can't remove this unless there are no projects currently using it.  Additionally, why are you deleting it?  It looks like your script below just adds it back?  Is PIMS-54 enabling the ability for "Not in SPL" to be put "On Hold", or "Transferred within the GRE"?

jdtoombs: I did this with a rollback in mind (I am probably misunderstanding the down scripts). But yes PIMS-54 is to enable Not in SPL to be put On Hold or Transferred within the GRE. The transitions are already enabled and Transferred within the GRE is part of SPL already, I just had to add On Hold to the grouping

Fosol: I think I was looking at the scripts backwards.  Should have noticed this was a down script.

Fosol (approved):

PR #1168, opened 2021-11-04
PIMS-24: add the LTSA wms layer using the new url

LawrenceLau2020:
adding new LTSA map layer URL and readme doc for local keycloak configuration.

Please check that the README didn't include anything that's supposed to be confidential.

### DIFF
--- a/frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
+++ b/frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
@@ -15,6 +15,8 @@ export const PARCELS_LAYER_URL =
   'https://' +
   environment +
   'apps.gov.bc.ca/ext/sgw/geo.allgov?service=WFS&version=2.0&request=GetFeature&typeName=geo.allgov:WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_FA_SVW&outputFormat=application/json';
+export const PARCELS_LAYER_URL_PUBLIC =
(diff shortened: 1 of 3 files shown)


frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
```diff
@@ -1,12 +1,31 @@
...
+  environment = '';
+} else if (process.env.NODE_ENV === 'test') {
+  environment = 'test.';
+} else if (process.env.NODE_ENV === 'development') {
+  environment = 'delivery.';
+}
+
 export const MUNICIPALITY_LAYER_URL =
   'https://openmaps.gov.bc.ca/geo/pub/WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP/wfs?SERVICE=WFS&REQUEST=GetFeature&VERSION=1.3.0&outputFormat=application/json&typeNames=pub:WHSE_LEGAL_ADMIN_BOUNDARIES.ABMS_MUNICIPALITIES_SP';
 export const PARCELS_LAYER_URL =
```
Fosol, line 14: We should probably keep this as a separate layer instead of removing it.

Fosol (approved):
Awesome work!  Just remember to not merge until we have a new signed off `master` branch.

Fosol (approved):
My only comment was that it may make more sense to still keep the prior layer and add this new one as the default one.  But I'm not sure if the prior layer has any value.

LawrenceLau2020:
> Does it have anything that the new layer doesn't? Are we worried about users who don't have the correct permissions not being able to see the new layer?

I can bring this up with Lindsay and Amit, and get their feedback.

LawrenceLau2020:
> > Does it have anything that the new layer doesn't? Are we worried about users who don't have the correct permissions not being able to see the new layer?
> 
> I can bring this up with Lindsay and Amit, and get their feedback.

It looks like we'll need to keep both layers for now for people who can't view the secured layer, and we will need to figure out a way to give the user the "correct" layer somehow. I'm looking at the axios code and I'm not sure if there's an easy way to do this as the first axios requests makes a "promise" to get the data and returns an error if it doesn't.

PR #1171, opened 2021-11-12
PIMS-49: Save notification checked state on save

jdtoombs:
Fix that saves the state of the checkbox on save while approving
49bugfix


LawrenceLau2020 (approved):
Looks good to me!

PR #1173, opened 2021-11-15
PIMS-24: making the dev case as first option

LawrenceLau2020:
For some reason in the dev environment the current code is not using the development url prefix, so trying the dev case at the beginning.


Fosol (commented):
This won't work.  All environments in Openshift are compiled with a production build.

PR #1175, opened 2021-11-17
DevOps always build

Fosol:
Presently it's impossible to force a build if files don't change, which results in never deploying.

This change will always build and always deploy when running the pipeline.


Fosol:
@LawrenceLau2020 I hope this will resolve the DevOps issue in DEV.  Or at least let us see an error for deployment.

LawrenceLau2020 (approved):

PR #1179, opened 2021-12-16
Pims 24 v2

LawrenceLau2020:
changes to the map layer to display a modal for the error when viewing a parcel and removing municipal layer


frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -184,12 +181,14 @@ const Map: React.FC<MapProps> = ({
...
   const smallScreen = useMediaQuery({ maxWidth: 1800 });
   const [mapWidth, setMapWidth] = useState(0);
-  const municipalitiesService = useLayerQuery(MUNICIPALITY_LAYER_URL);
   const layerUrl = useBoundaryLayer();
   const parcelsService = useLayerQuery(layerUrl);
   const [bounds, setBounds] = useState<LatLngBounds>(defaultBounds);
   const { setChanged } = useFilterContext();
   const [layerPopup, setLayerPopup] = useState<LayerPopupInformation>();
+  const popUpContext = React.useContext(PropertyPopUpContext);
+  console.debug(popUpContext.showBCEIDWarning);
```
Fosol, line 190: Might want to remove this

frontend/src/components/maps/leaflet/Map.tsx
```diff
@@ -329,6 +320,14 @@ const Map: React.FC<MapProps> = ({
 
   const [infoOpen, setInfoOpen] = React.useState(false);
   const [layersOpen, setLayersOpen] = React.useState(false);
+  const displayMessage = (
```
Fosol, line 323: Nice simple fix.  I like it.  Handles both scenarios.  We can look into a better one later.

Fosol (approved):

PR #1180, opened 2021-12-16
Fix Sonarqube log4j security issue

Fosol:
I've update the image, but the deployment appears to be really slow.  Not sure if it's working atm.

### DIFF
--- a/openshift/4.0/templates/sonarqube/deploy.yaml
+++ b/openshift/4.0/templates/sonarqube/deploy.yaml
@@ -1,0 +1,161 @@
+apiVersion: v1
+kind: Template
+message: "Login to SonarQube with the default admin user: admin/admin"
+metadata:
+  annotations:
+    description: The SonarQube OpenShift template
+    tags: "instant-app,sonarqube"
+  name: sonarqube
+parameters:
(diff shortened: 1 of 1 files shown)


LawrenceLau2020 (approved):

LawrenceLau2020:
Hi Jeremy, should we also update the owasp-zap plugin to v2.11.1 for Jenkins in this PR as well?
https://www.zaproxy.org/blog/2021-12-10-zap-and-log4shell/

Fosol:
I don't see it in our image registry.

LawrenceLau2020:
> I don't see it in our image registry.

Is it not just changing the version number using an install command using python? I see this in the "Dockerfile" in jenkins-slave-zap folder:

`pip install python-owasp-zap-v2.4`

Fosol:
Ah yes I suppose that might do it.  I suspect this isn't accessible externally though.

LawrenceLau2020:
> Ah yes I suppose that might do it. I suspect this isn't accessible externally though.

It's actually on the github repo: https://github.com/zaproxy/zaproxy/releases/tag/v2.11.1

PR #1181, opened 2021-12-22
Fix for log4j to upgrade Sonarqube

LawrenceLau2020:
using the lts-enterprise image from sonarqube: https://hub.docker.com/_/sonarqube?tab=tags
I believe the version is v8.9.6, and I'm assuming enterprise is the one to use.
Also modified the deployment strategy from rolling to recreate and added a longer timeout:
recreateParams:
          timeoutSeconds: 9600

### DIFF
--- a/frontend/src/features/projects/list/__snapshots__/ProjectListView.test.tsx.snap
+++ b/frontend/src/features/projects/list/__snapshots__/ProjectListView.test.tsx.snap
@@ -136,6 +136,12 @@ exports[`Project list view tests Matches snapshot 1`] = `
               </option>
               <option
                 class="option"
+                value="2023"
+              >
+                2022 / 2023
+              </option>
+              <option
+                class="option"
                 value="2022"
               >
                 2021 / 2022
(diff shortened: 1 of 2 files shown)


Fosol (approved):
Do you know if this image will work in Openshift?  A lot of official images use the root user which is against Openshift security rules.

LawrenceLau2020:
> Do you know if this image will work in Openshift? A lot of official images use the root user which is against Openshift security rules.

That I don't know....I did try to find out if anyone else was using this image.
The replication controller seems to be using this image, though.

LawrenceLau2020:
> Do you know if this image will work in Openshift? A lot of official images use the root user which is against Openshift security rules.

I'm going to try Sonarqube 9.2-community version, as Wade recommended using that one.

LawrenceLau2020:
> > Do you know if this image will work in Openshift? A lot of official images use the root user which is against Openshift security rules.
> 
> I'm going to try Sonarqube 9.2-community version, as Wade recommended using that one.

using Sonarqube 9.2-community version caused the pod to crash and fail. The logs mentioned something about upgrading to LTS first. Currently using "lts-community" version of Sonarqube now, and the pod seems stable.

I'll probably try to merge this PR to dev tomorrow to test the new Sonarqube version.

PR #1182, opened 2022-01-06
Pims 24 - using both public and private layers for parcels

LawrenceLau2020:
Had to add back the red toast error for the error handling as that was breaking too many frontend tests. 
Maybe we can deal with it in a future PR?

### DIFF
--- a/frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
@@ -1,9 +1,29 @@
 import { useEnvironment } from 'hooks/useEnvironment';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+
+export const Username = () => {
+  const keycloak = useKeycloakWrapper();
+  return keycloak.username;
+};
+
+export const FindUserType = (username: string) => {
+  let isIDIRUser = username.includes('idir');
+  let isBCeIDUser = username.includes('bceid');
+
+  if (isIDIRUser) {
+    return 'idir';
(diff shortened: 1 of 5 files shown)


LawrenceLau2020:
I just realized I had forgotten to add back in the municipality popup layer info.

frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
```diff
@@ -1,9 +1,30 @@
 import { useEnvironment } from 'hooks/useEnvironment';
+import useKeycloakWrapper from 'hooks/useKeycloakWrapper';
+
+export const Username = () => {
+  const keycloak = useKeycloakWrapper();
+  return keycloak.username;
+};
+
+export const FindUserType = (username: string) => {
```
Fosol, line 9: This function looks like it can return void too.  Not sure if that's desirable.

Fosol (approved):

frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
LawrenceLau2020: What would you recommend to make this cleaner?

Fosol: Generally it's bad practice to return void, or undefined.  It's even worse practice to return inconsistent results.  Since you're returning a string value, you should return an empty string if no match is found "".

jdtoombs (approved):

Fosol (approved):

PR #1184, opened 2022-01-13
PIMS-24: Removing the env for test

LawrenceLau2020:
removing the test environment for the map url, so we only use the dev url for DEV & TEST, and prod url for PROD.

### DIFF
--- a/frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
+++ b/frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
@@ -28,9 +28,6 @@ export const useBoundaryLayer = () => {
   let domain = '';
   let parcelURL = '';
   switch (env.env) {
-    case 'test':
-      domain = `${env.env}.`;
-      break;
     case 'development':
       domain = 'delivery.';
       break;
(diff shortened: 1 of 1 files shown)


frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
```diff
@@ -28,9 +28,6 @@ export const useBoundaryLayer = () => {
   let domain = '';
   let parcelURL = '';
   switch (env.env) {
-    case 'test':
```
Fosol, line 31: Won't this make `test` use the same as `prod`?  If we want to use the `dev` you need to add the `case 'test':` before the `case 'development':`

LawrenceLau2020: You're right, good catch!

frontend/src/components/maps/leaflet/LayerPopup/hooks/useBoundaryLayer.tsx
```diff
@@ -28,6 +28,9 @@ export const useBoundaryLayer = () => {
   let domain = '';
   let parcelURL = '';
   switch (env.env) {
+    case 'test':
```
Fosol, line 31: You can rewrite to
```
case 'test':
case 'development':
```

Fosol (approved):

PR #1185, opened 2022-01-13
PIMS-79: Notification sent to ADM in error

LawrenceLau2020:
It seems that the notification which gets sent when updating project notes is not checking for the project initiator and sending an email to the ADM.

Added an extra condition in what seems like a "generic" function, as I'm not sure of a better way of doing this, also could have checked for the template id instead.

### DIFF
--- a/backend/dal/Services/Concrete/NotificationQueueService.cs
+++ b/backend/dal/Services/Concrete/NotificationQueueService.cs
@@ -315,7 +315,7 @@ namespace Pims.Dal.Services
 
             var env = new EnvironmentModel(_options.Environment.Uri, _options.Environment.Name, _options.Environment.Title);
             var model = new ProjectNotificationModel(Guid.NewGuid(), env, project, project.Agency);
-            var email = !String.IsNullOrWhiteSpace(project.Agency.Email) ? project.Agency.Email :
+            var email = !String.IsNullOrWhiteSpace(project.Agency.Email) && !(template.Audience == NotificationAudiences.ProjectOwner) ? project.Agency.Email :
                 this.Context.Users.FirstOrDefault(u => u.Id == project.CreatedById)?.Email;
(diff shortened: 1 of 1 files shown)


backend/dal/Services/Concrete/NotificationQueueService.cs
```diff
@@ -315,7 +315,7 @@ public NotificationQueue GenerateNotification(Project project, NotificationTempl
 
             var env = new EnvironmentModel(_options.Environment.Uri, _options.Environment.Name, _options.Environment.Title);
             var model = new ProjectNotificationModel(Guid.NewGuid(), env, project, project.Agency);
-            var email = !String.IsNullOrWhiteSpace(project.Agency.Email) ? project.Agency.Email :
+            var email = !String.IsNullOrWhiteSpace(project.Agency.Email) && !(template.Audience == NotificationAudiences.ProjectOwner) ? project.Agency.Email :
```
Fosol, line 318: If I am reading this correctly.  Use the user's email who created the project if the template does not have the `Audience == ProjectOwner`?
Can you not write that `template.Audience != NotificationAudiences.ProjectOwner)`

Fosol (approved):

backend/dal/Services/Concrete/NotificationQueueService.cs
LawrenceLau2020: That's correct. There are 7 templates including the notes change template whose audience is "Project Owner". The other templates are the 3 disposal templates for denied, cancelled, and approved for ERP, and the 3 ERP templates for owning agencies 30, 60, & 90 day notifications.

Actually, I can use this:
`(!String.IsNullOrWhiteSpace(project.Agency.Email) && template.Audience != NotificationAudiences.ProjectOwner)`

PR #1190, opened 2022-01-25
PIMS-63 Add Tekton pipeline

Fosol:
I've turned off the Jenkins dev-pipeline for now.  GitHub no longer is sending webhooks to start it.  Now the webhooks start the Tekton pipeline.

## Summary
- Add Tekton DevOps pipeline for DEV environment.  API and APP are built and deployed to DEV when `dev` branch is updated.
- Fix app bug that was blocking Sonarqube scan.

## Outstanding
I still need to come up with a way to run the database migration with the Tekton Pipeline.


frontend/src/store/slices/hooks/useAccessRequest.ts
```diff
@@ -164,12 +164,13 @@ export const useAccessRequest = () => {
...
 
           dispatch(hideLoading());
+          return response;
         } catch (axiosError) {
           const err = axiosError as AxiosError;
-          return dispatch(error(actionTypes.GET_REQUEST_ACCESS, err?.response?.status, axiosError));
+          dispatch(error(actionTypes.GET_REQUEST_ACCESS, err?.response?.status, axiosError));
         }
       } finally {
-        return dispatch(hideLoading());
```
Fosol, line 172: Sonarqube scan was failing because of this.

LawrenceLau2020 (approved):

nginx/html/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+  <head>
+  <title>TNO - Today's News Online</title>
+  <style>
```
LawrenceLau2020, line 5: Just noticed the TNO in the title, probably doesn't matter, but would this file be used at anytime in PIMS?

Fosol: I copied it from TNO.  Needs to be updated for PIMS.  This is the maintenance page.  It should at some point be updated to look better.  But I don't have time.

PR #1191, opened 2022-02-03
PIMS-80: Property list building view financials not saving

LawrenceLau2020:
I've disabled validation on the assessed land field now which allows the editing and saving of financial values in the building view mode.
I noticed that there are a couple other minor bugs which can be addressed at a later date, one is that after saving a financial value, the lot size value switches to display "NaN". 
image
The value is still there, as we can switch to the parcels view and then back and the value is there, it's just not getting displayed when you click save for some reason. Maybe something to do with the state?

The other issue is that after saving your edits, if you click the cancel button to get out of "edit" mode, the assessed building value for the record/s that was changed goes to 0:
image

Just click the parcels view and then come back to the buildings view and the values will reappear:
image


Fosol (approved):
Thanks Lawrence

PR #1202, opened 2022-02-18
Pims 81: ERP Notification building - ft2 s/b m2

Sharala-Perumal:
Removed sqft from notification templates and replaced with sqM


Fosol (approved):
Sorry I missed this PR!

PR #1211, opened 2022-03-04
PIMS-68: Issue with PID returning wrong parcel

LawrenceLau2020:
Using a new NPM package called Polylabel: https://github.com/mapbox/polylabel which locates the most internal point within a polygon. This is used when doing a PID search or an address search when submitting a new land and it will help when doing searches for lands which have irregular polygon shapes where the center would have been calculated based on the center of the "bounding box" of the actual shape falling outside of the polygon and returning a different property.

image

image

By default Polylabel uses a precision of 1.0, but when used in the layer query the precision was set to 0.00001

`polylabel(coordinates, 0.00001)`


frontend/src/components/maps/leaflet/LayerPopup/hooks/useLayerQuery.tsx
```diff
@@ -85,17 +93,22 @@ export const saveParcelDataLayerResponse = (
   latLng?: LatLng,
 ) => {
   if (resp?.features?.length > 0) {
```
Fosol, line 95: Happy to see this finally getting implemented.

Fosol (approved):

frontend/src/components/maps/leaflet/LayerPopup/hooks/useLayerQuery.tsx
LawrenceLau2020: Same here, it's been sitting dormant for a long time. :)

PR #1217, opened 2022-03-07
PIMS-105: Issue with adding buildings to existing land

LawrenceLau2020:
Creating a copy of the matchingParcel object to prevent error when adding/associating a building to an existing land.


Fosol (commented):
Happy this resolve the issue.

Fosol (approved):

PR #1220, opened 2022-03-14
PIMS-102: renaming folder name to PostUp and adding a space

LawrenceLau2020:
just changing the u to upper case U, and adding a space to the sql in order to create a PR to fix the previous migration which didn't run properly because of folder being named Postup instead of PostUp.


backend/dal/Migrations/v01.16.00/Up/Postup/01-ViewProperties.sql
```diff
@@ -183,7 +183,7 @@ LEFT JOIN dbo.[ParcelBuildings] pb ON b.[Id] = pb.[BuildingId]
 LEFT JOIN dbo.[Parcels] p ON pb.[ParcelId] = p.[Id]
 JOIN dbo.[PropertyClassifications] c ON b.[ClassificationId] = c.[Id]
 LEFT JOIN dbo.[Users] u ON b.UpdatedById = u.Id
-LEFT JOIN dbo.[Agencies] a ON b.[AgencyId] = a.[Id]
```
LawrenceLau2020, line 186: hmm.....even though I renamed my Postup to PostUp, the repo isn't detecting it for some reason?

PR #1221, opened 2022-03-16
PIMS-109 Fix associated land

Fosol:
Financial values should now be properly displayed on the Associated Land Form during adding.


LawrenceLau2020 (approved):

LawrenceLau2020:
> Financial values should now be properly displayed on the Associated Land Form during adding.

Will the valuations also show up properly in the Land details record itself?

Fosol:
I believe so, but I've only quickly reviewed the land form by itself.  It had values, but I didn't compare before and after.

Sharala-Perumal (approved):

PR #1228, opened 2022-03-31
PIMS-113: Classification Status "Demolished" should be available for buildings

LawrenceLau2020:
I'm not 100% sure if this is the right claim to use, but I changed the claim from `ADMIN_PROPERTIES` to `PROPERTY_EDIT`

### DIFF
--- a/frontend/src/hooks/useLookupCodes.tsx
+++ b/frontend/src/hooks/useLookupCodes.tsx
@@ -63,7 +63,7 @@ export function useCodeLookups() {
     const classifications = getByType(API.PROPERTY_CLASSIFICATION_CODE_SET_NAME);
     return filter
       ? (classifications ?? []).map(c => mapLookupCode(c)).filter(filter)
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/useLookupCodes.tsx
```diff
@@ -63,7 +63,7 @@ export function useCodeLookups() {
     const classifications = getByType(API.PROPERTY_CLASSIFICATION_CODE_SET_NAME);
     return filter
       ? (classifications ?? []).map(c => mapLookupCode(c)).filter(filter)
-      : !keycloak.hasClaim(Claims.ADMIN_PROPERTIES)
+      : !keycloak.hasClaim(Claims.PROPERTY_EDIT)
```
Fosol, line 66: This probably isn't ideal as all users essentially will be able to select Demolish, Subdivide, and Disposed.  Are other users allowed to do this?  I'm pretty sure they definitely can't Dispose outside of the Disposal Project workflow.

Fosol (approved):

PR #1232, opened 2022-04-05
PIMS-114: Issue with 2022 Assessment not showing

LawrenceLau2020:
This fix addresses an issue where it was creating an extra row in the assessment years for a property after April 1. 
This also includes a database migration to fix data issues where records were entered in the 2022 row, but is being stored in the db as 2021 date. The fix will increase the year for these records, mainly parcels and some buildings, by one year.


LawrenceLau2020:
No, I have to increase the evaluation because these evaluations which were entered in the "2022" row were actually saved in the database with a 2021 date. I guess because the code before had detected a new fiscal year, so it would increase the row on the left side of the table, but the evaluations on the right side didn't match.

image

Fosol (approved):

PR #1237, opened 2022-04-21
PIMS-119 Fix project status transition

Fosol:
To enable change a project status from Not-in-SPL to Disposed and Not-in-SPL to Transferred within GRE I have added a new DB migration `v01.18.01`.


backend/dal/Migrations/PimsContextModelSnapshot.cs
```diff
@@ -2734,6 +2734,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<string>("Description")
                         .HasColumnType("nvarchar(max)");
 
+                    b.Property<string>("DisplayName")
```
Fosol, line 2737: Not certain why this was added in the auto generated code.

backend/dal/Migrations/v01.18.01/Up/PostUp/01-WorkflowProjectStatus.sql
```diff
@@ -0,0 +1,105 @@
+PRINT 'Adding ProjectStatusTransitions - !SPL -> Disposed'
+
+IF NOT EXISTS (
```
Fosol, line 3: Only add the records if they don't already exist.

LawrenceLau2020 (approved):

PR #1267, opened 2022-06-25
PIMS-69 Add Agency CC

Fosol:
Added the ability to CC agencies when sending ERP notifications.  If the agency has a CC value it will add that email.

We now have Development Containers (devcontainers) for the backend and frontend.  If you open the `/backend` or the `/frontend` folders VS Code will ask if you want to open them in a container.  This provides a way to ensure every developer has the exact same environment.   It also resolves the challenges of Node/NPM/.NET versions and SDKs.

There was an annoying spamming bug related to the implementation of the Typeahead component.  This has now been resolved.

I am seeing some weird styling issues throughout the app.  I'm not sure when these were introduced, but I don't believe my edits were directly related.

> Please note that in DEV and TEST emails will only be sent to the tester.

> New DB Migration `v01.18.03` adds the `Agency.CCEmail` property to the database.  Run `make db-update` to update your local database.

image


LawrenceLau2020:
Hi @Fosol , it looks like there were over 500 files changed, were the majority of these related to the styling issues you had mentioned? Just curious what the styling issues are exactly?

Fosol:
The majority are styling issues with linting after upgrading various libraries.  

Presently I'm trying to resolve unit-test bugs with the frontend.  I'm not sure why but many errors appear related to react-router.  However, I never upgraded that package.  Which means something funky with the npm version was going on and these types of issues are only showing up now that I've fixed the npm version.

frontend/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+# [Optional] Uncomment this section to install additional OS packages.
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+  && apt-get -y install --no-install-recommends git libstdc++6
+
+# [Optional] Uncomment if you want to install an additional version of node using nvm
+ARG EXTRA_NODE_VERSION=v14.19.3
+RUN su node -c "npm config delete prefix"
+RUN su node -c "source /usr/local/share/nvm/nvm.sh && nvm install ${EXTRA_NODE_VERSION}"
+RUN su node -c "npm config set prefix $NVM_DIR/versions/node/${EXTRA_NODE_VERSION}"
+# RUN su node -c "nvm use --delete-prefix ${EXTRA_NODE_VERSION} --silent"
```
Fosol, line 16: Don't know how to run the `nvm` command to get rid of the warning.  For some reason `nvm` isn't available at this point...

frontend/Dockerfile
```diff
@@ -12,7 +12,7 @@ ENTRYPOINT ["dumb-init", "--"]
 WORKDIR /usr/app
 
 # install a specific NPM version
-RUN npm install -g npm@8.8.0
+RUN npm install -g npm@6.14.7
```
Fosol, line 15: This is the correct version for npm, for this version of node.

frontend/package.json
```diff
@@ -164,11 +165,18 @@
     ]
   },
   "eslintConfig": {
+    "parser": "@typescript-eslint/parser",
```
Fosol, line 168: Needed to updated linting because it was resulting in errors withotu this.

frontend/.devcontainer/Dockerfile
Fosol: Not sure what you mean?

frontend/package.json
```diff
@@ -1,13 +1,12 @@
...
   "name": "frontend",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "@babel/core": "7.18.5",
     "@bcgov/bc-sans": "1.0.1",
     "@react-keycloak/web": "3.4.0",
     "@reduxjs/toolkit": "1.6.0",
-    "@types/react-bootstrap-typeahead": "3.4.7",
```
LawrenceLau2020, line 9: Just curious why react-bootstrap-typeahead and react-responsive were both removed from the package.json file?

frontend/.devcontainer/Dockerfile
Fosol: We discussed this in our meeting.  This doesn't work.

frontend/package.json
LawrenceLau2020: It seems like before there were 2 separate packages, one for the @types: `"@types/react-bootstrap-typeahead": "3.4.7"` and one for the dependency: `"react-bootstrap-typeahead": "5.1.2"`

and now the dependency has been downgraded to 3.4.7, but now we are missing the @types for react-bootstrap-typeahead? Would this cause an issue?

LawrenceLau2020 (commented):
It seems like before there were 2 separate packages, one for the @types: "@types/react-bootstrap-typeahead": "3.4.7" and one for the dependency: "react-bootstrap-typeahead": "5.1.2"

and now the dependency has been downgraded to 3.4.7, but now we are missing the @types for react-bootstrap-typeahead? Would this cause an issue?

frontend/package.json
Fosol: Sorry wrong line number.  I've removed duplicates.

frontend/package.json
```diff
@@ -17,15 +16,15 @@
...
     "lodash": "4.17.21",
     "moment": "2.29.3",
     "moment-timezone": "0.5.31",
+    "polylabel": "1.1.0",
     "query-string": "6.12.1",
     "react": "16.12.0",
     "react-app-polyfill": "1.0.5",
     "react-bootstrap": "1.0.0",
-    "react-bootstrap-typeahead": "5.1.2",
+    "react-bootstrap-typeahead": "3.4.7",
```
Fosol, line 27: The downgrade may not be needed.  It was to match the `react-bootstrap-typeahead` library.  Presently they are mismatched.

backend/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.224.2/containers/dotnet/.devcontainer/base.Dockerfile
+
+# [Choice] .NET version: 6.0, 5.0, 3.1, 6.0-bullseye, 5.0-bullseye, 3.1-bullseye, 6.0-focal, 5.0-focal, 3.1-focal
+ARG VARIANT="6.0-bullseye-slim"
```
LawrenceLau2020, line 4: Should we be using .NET version 5.0 here?

backend/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,70 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.224.2/containers/dotnet
+{
+	"name": "C# (.NET)",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"args": {
+			// Update 'VARIANT' to pick a .NET Core version: 3.1, 5.0, 6.0
+			// Append -bullseye or -focal to pin to an OS version.
+			"VARIANT": "6.0",
```
LawrenceLau2020, line 10: Should the "VARIANT" be .NET Core version 5.0 here as well?

backend/.devcontainer/Dockerfile
Fosol: I tried this first of all.  Found out that it wouldn't work for some reason.  It's backward compatible though.

backend/.devcontainer/devcontainer.json
Fosol: I tried this first of all.  Found out that it wouldn't work for some reason.  It's backward compatible though.

backend/dal/Services/INotificationQueueService.cs
```diff
@@ -23,7 +23,7 @@ public interface INotificationQueueService : IService
         IEnumerable<NotificationQueue> GenerateNotifications(Project project, int projectStatusNotificationId);
         IEnumerable<NotificationQueue> GenerateNotifications(Project project, ProjectStatusNotification options, DateTime? sendOn = null);
         NotificationQueue GenerateNotification(Project project, ProjectStatusNotification options, Agency agency, DateTime? sendOn = null);
-        NotificationQueue GenerateNotification<T>(string to, NotificationTemplate template, T model, DateTime? sendOn = null);
+        NotificationQueue GenerateNotification<T>(string to, string cc, NotificationTemplate template, T model, DateTime? sendOn = null);
```
LawrenceLau2020, line 26: Just curious will the cc email get sent to the creator/generator of the notification in the DEV & TEST environments?

Fosol: It follows the same rules everything else does.  in DEV and TEST it will CC whomever is testing.   This way at least the tester can confirm the CC value was set.

LawrenceLau2020 (approved):

PR #1276, opened 2022-09-06
changes to how the map gets the parcel layer

LawrenceLau2020:
# Description

Replacing the url for the parcel layer with the FA layer that contains the legal description which is now public. Since we are only using "public" layers, we are also removing code which used the previous detection of the user type in order to determine which layer to display. Broke out the user type code into a separate hook for future use.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue) Note, it's not really a bug, but removed unnecessary code.

# How Has This Been Tested?

Selecting parcels on the map page, and the submit land and building pages and reviewing the parcel details on the popup.

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings


frontend/src/hooks/useGetAccountType.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if (isIDIRUser) {
+    return 'idir';
+  } else if (isBCeIDUser) {
+    return 'bceid';
+  } else {
+    return '';
+  }
+};
+
+export const useGetAccountType = () => {
```
LawrenceLau2020, line 21: created new hook for getting the user account type

Fosol (approved):
Thanks

PR #1281, opened 2022-09-27
removing the useMemo and console.logs

LawrenceLau2020:
# Description

This fixes the bug which wasn't loading any properties on the View Property Information page.
Fixes the invalid hook error.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. 

# Checklist:

- [ ] I have performed a self-review of my own code
- [x] My changes generate no new warnings

### DIFF
--- a/frontend/src/customAxios.ts
+++ b/frontend/src/customAxios.ts
@@ -1,6 +1,5 @@
 import axios from 'axios';
 import { isEmpty } from 'lodash';
-import React from 'react';
 import { toast } from 'react-toastify';
 import { store } from 'store/store';
 
@@ -34,7 +33,6 @@ export const CustomAxios = ({
   baseURL?: string;
 } = {}) => {
   let loadingToastId: React.ReactText | undefined = undefined;
-
   const instance = axios.create({
     baseURL,
     headers: {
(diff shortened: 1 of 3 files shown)


frontend/src/customAxios.ts
```diff
@@ -33,60 +33,55 @@ export const CustomAxios = ({
...
-      headers: {
-        'Access-Control-Allow-Origin': '*',
-      },
-    });
-    instance.interceptors.request.use((config) => {
-      config!.headers!.Authorization = `Bearer ${token}`;
-      if (selector !== undefined) {
-        const state = store.getState();
-        const storedValue = selector(state);
+  const instance = axios.create({
```
Fosol, line 38: I don't know if it matters, but anything that uses this object will get redrawn infinitely because this hook is returning an object that changes every time the hook gets called.

LawrenceLau2020: @Fosol I did some digging and found that my commit on September 1st: https://github.com/bcgov/PIMS/commit/324a5962742e30eb0f4a65e890059f926a65cc9e had introduced this new "useMemo" hook, I guess because I have ESLint and format on save turned on. Your original "customAxios.ts" didn't have it and neither does the existing version in the dev branch.

Fosol: If it works then it should be okay.  Hooks can have some unexpected behaviour because of this.

Fosol (approved):

PR #1296, opened 2022-10-07
Pims 302 vulnerabilities fixes

BradyMitch:
# Description

Resolved 11 vulnerabilities in pims-63 branch by updating 9 packages and added an override for packages `@svgr/webpack` and `scss-tokenizer`.  

An override package in `frontend/package.json` means that when other packages use one of these packages as a dependency, it will force it to use the override version instead of whichever version its trying to use. Overrides can be used to solve vulnerabilities when the parent packages that are using them haven't updated to use a patched version of the dependency. Overrides should be a last resort.

### Vulnerabilities fixed in this pr:  

npm-audit-pims-63

### Package versions between pims-63 and this branch:  

@pmmmwh/react-refresh-webpack-plugin 0.5.4 >> 0.5.7  
@svgr/plugin-svgo 5.5.0 >> 6.2.0
@svgr/webpack 5.5.0 >> 6.3.1
css-select 4.2.1 >> 4.3.0
moment 2.29.1 >> 2.29.4
moment-timezone 0.5.34 >> 0.5.37
nth-check 2.0.1 >> 2.1.1
react-scripts 5.0.0 >> 5.0.1

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

I tested loading up the containers and logging into the app. I saw properties on the map and in the Inventory View. I compared to the pims-63 branch and didn't see any changes to the app or console logs.

I am not very experienced with how the app works, so please test and let me know if you find any discrepancies between this branch and pims-63. 

# Checklist:

- [x] I have performed a self-review of my own code
- [ ] My changes generate no new warnings


BradyMitch:
I found this useful when figuring out the override for the nth-check related vulnerability: [https://github.com/facebook/create-react-app/issues/12132](https://github.com/facebook/create-react-app/issues/12132)

frontend/package.json
```diff
@@ -7,22 +7,27 @@
...
+    "css-select": "^4.3.0",
     "dequal": "2.0.2",
     "formik": "2.2.9",
     "http-proxy-middleware": "2.0.3",
     "keycloak-js": "17.0.0",
     "leaflet": "1.7.1",
     "lodash": "4.17.21",
-    "moment": "2.29.1",
-    "moment-timezone": "0.5.34",
+    "moment": "^2.29.4",
```
Fosol, line 28: Normally you don't want to use the `^` version, as it will most likely be different versions used by every single build.

frontend/package.json
```diff
@@ -118,7 +124,10 @@
     "redux-devtools-extension": "2.13.9",
     "redux-mock-store": "1.5.4"
   },
-  "resolutions": {},
+  "overrides": {
```
Fosol, line 127: I like this

frontend/package.json
BradyMitch: Thanks, I think there is an npm config option I can set to not set a prefix when updating versions. I've committed a change to remove the prefix.

BradyMitch: save the exact version - `npm config set save-exact true`  
set a prefix - `npm config set save-prefix='~'`

BradyMitch:
@LawrenceLau2020 I removed the unused packages that were only used as dependencies of other packages, and did not need to be installed as core packages. I have a better understanding of npm audit now and what packages to update for vulnerability fixes.

I did leave the one package that was needed for the override fix. I also tested running the application and everything seems to still work.

LawrenceLau2020 (approved):

PR #1644, opened 2023-08-10
PIMS-142 Update react-toastify

TaylorFries:
## 🎯 Summary

  
[PIMS-142: ](https://citz-imb.atlassian.net/browse/PIMS-142)

Update react-toastify to 9.1.3 

Had to adjust tests to keep them working. 
In manual testing, the toastify notifications that pop up when axios is awaiting a request are working. I think they're just disappearing so quick when the request is successful that it doesn't find the text after, and the request is always immediately successful when axios is mocked in tests.

I added a mock reply that should keep it waiting instead of immediately fulfilling the promise.
Also removed some await options that wait for the promise to complete, because it doesn't need to in this case. We're only looking for the toastify notification before it completes.
~Dylan

## 🔰 Checklist


TaylorFries:
changes approved

dbarkowsky (approved):

PR #1661, opened 2023-08-23
[PIMS-145] yup package update

dbarkowsky:
## 🎯 Summary

  
[PIMS-145: ](https://citz-imb.atlassian.net/browse/PIMS-145?atlOrigin=eyJpIjoiMTc2NmNiMTg4MmZmNDViOGI0MzlkNTE4MDIzZGQzYjMiLCJwIjoiaiJ9) Update yup package

v1.x.x includes many syntax changes.
1. .then and .otherwise properties changed like this:
```ts
// from this
then: Yup.object({
        id: Yup.string().required(),
        label: Yup.string().required(),
        value: Yup.string().required()
      }),

// to this
then: () => // change is here 👈
      Yup.object({
        id: Yup.string().required(),
        label: Yup.string().required(),
        value: Yup.string().required()
      }),
```
2. .shape() usually isn't needed
```ts
Yup.object().shape({});
// becomes
Yup.object({})
```
There are a couple of cases where shape is needed, such as when two properties depend on each other. Yup specifically recommends against doing this, but it is used in some of our schemas.
[Source](https://github.com/jquense/yup/issues/720)

## 🔰 Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -111,6 +111,7 @@
     "@types/styled-components": "5.1.26",
     "@types/supercluster": "7.1.0",
     "@types/text-mask-addons": "3.8.1",
+    "@types/yup": "0.32.0",
     "@typescript-eslint/eslint-plugin": "6.2.1",
     "@typescript-eslint/parser": "6.2.1",
     "axios-mock-adapter": "1.21.5",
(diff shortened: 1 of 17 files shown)


TaylorFries (approved):
Updates appear to work locally with no errors from yup

PR #1682, opened 2023-08-29
[PIMS-140] Update react-leaflet

dbarkowsky:
## 🎯 Summary

  
[PIMS-140: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-140)

Updated react-leaflet to version 4.2.1

The parameter `onClose` for popups were removed. Need to use a ref now to track when it closes.
The `whenCreated` parameter was also removed from the Leaflet map, so it also had to be handled with a ref.

Current status:
~One test failing because only sees `mapInstance` as null.~

Test working now. The way to access the inner map has changed, and there isn't a lot of information out there about how to do this now that `whenCreated` is removed. There is a `useMap()` function that seems to work to get the inner map component of `MapContainer`. 
Finally found this here: https://stackoverflow.com/questions/65394203/leaflet-react-get-map-instance-in-functional-component

## 🔰 Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -53,7 +53,7 @@
     "react-draggable": "4.4.5",
     "react-error-boundary": "4.0.10",
     "react-icons": "4.10.1",
-    "react-leaflet": "3.2.5",
+    "react-leaflet": "4.2.1",
     "react-paginate": "8.2.0",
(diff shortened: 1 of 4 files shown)


TaylorFries (approved):
builds with no errors locally and all tests pass

PR #1695, opened 2023-09-11
[PIMS-222] Add bulk upload route

dbarkowsky:
## 🎯 Summary

  
[PIMS-222: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-222)

Added a route that sends a list of properties to the import/properties endpoint.
Should return an object indicating the response code, accepted properties, and rejected properties.

## 🔰 Checklist


frontend/src/hooks/useApi.ts
```diff
@@ -196,5 +202,34 @@ export const useApi = (props?: IApiProps): PimsAPI => {
...
   }, []);
 
+  /**
+   * @description Attempts to send a list of properties to be inserted into the database.
+   *              Will not overwrite existing parcelIds. Also doesn't return an error if those parcelIds already exist.
+   * @prop {IPropertyModel[]} properties A list of properties to be added.
+   * @returns A promise with the response code and list of properties not added.
+   */
+  axios.importProperties = useCallback(async (properties: IPropertyModel[]) => {
+    const { status, data } = await axios.post<IPropertyModel[]>(
```
LawrenceLau2020, line 212: I'm not sure if we need a while loop in here to send the post requests in batches of 100 if there are more than 100 properties in the upload....

dbarkowsky: I thought about including this in a frontend component. It's going to need to keep track of the progress of how many are sent, and I don't know a way to do that from within this function and still communicate that back to the component.

LawrenceLau2020: Sounds good!

LawrenceLau2020 (approved):

PR #1698, opened 2023-09-14
[PIMS-227] Fix: Buildings disassociating from land

dbarkowsky:
## 🎯 Summary

  
[PIMS-227: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-227)

Issue:
When updating buildings from the View Inventory page, it was causing some fields, including associated land parcels, to be cleared or reset to default values. 
This was because that information is never loaded on the page and was therefore sent back to the API without it, which interpreted it as intentionally removed.

Changes:
- The route for updating financials was restored to only update the financials.
- Added an assignment to update the `ClassificationId` and removed the attempted update to `Classification`. `ClassificationId` is the foreign key and should be updated instead.
- Restored the edit button.

Testing instructions:
1. Identify a building with associated land. 
2. Go to the View Inventory page and edit that building, then save your edits.
3. Ensure the building still has associated land and has saved the new classification.

## 🔰 Checklist

### DIFF
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -381,7 +381,7 @@ namespace Pims.Dal.Services
                 .Include(b => b.Fiscals)
                 .FirstOrDefault(b => b.Id == building.Id) ?? throw new KeyNotFoundException();
             this.ThrowIfNotAllowedToUpdate(existingBuilding, _options.Project);
-            existingBuilding.Classification = building.Classification;
+
             var user = this.Context.Users
                .Include(u => u.Agencies)
                .ThenInclude(a => a.Agency)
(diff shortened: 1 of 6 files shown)


backend/dal/Services/Concrete/BuildingService.cs
```diff
@@ -399,6 +399,7 @@ public Building UpdateFinancials(Building building)
             {
                 this.Context.SetOriginalRowVersion(existingBuilding);
                 this.Context.UpdateBuildingFinancials(existingBuilding, building.Evaluations, building.Fiscals);
+                existingBuilding.ClassificationId = building.ClassificationId;
```
LawrenceLau2020, line 402: I like the simplicity of the fix! Nice work :)

LawrenceLau2020 (approved):
Tested the saving of parcel and building financials from the Property List View, and all seems good.

dbarkowsky:
Don't merge yet. Working on some tests.

PR #1707, opened 2023-09-18
PIMS-225 and 226: replace the wms PMBC layer and autopopulate the legal description

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-225: ](https://citz-imb.atlassian.net/browse/PIMS-225) - autopopulate legal description from the land titles API
[PIMS-226: ](https://citz-imb.atlassian.net/browse/PIMS-226) - replace the wms PMBC layer with the public one

Replaced the pmbc parcel layer with the "public one" which doesn't include the "legal description" field.
Needed to create new functions to get the legal description from the LTSA api.
We seem to be missing the "survey designation" information, doesn't seem to be part of the order information from LTSA.

## 🔰 Checklist


frontend/src/features/properties/hooks/useParcelLayerData.tsx
```diff
@@ -152,28 +155,55 @@ const useParcelLayerData = ({
   nameSpace,
   agencyId,
 }: IUseParcelLayerDataProps) => {
+  // Get LTSA information
+  const [ltsa, setLtsa] = useState<ILTSAOrderModel | undefined>(undefined);
```
LawrenceLau2020, line 159: I'm not sure we even need the state?

LawrenceLau2020:
I tried breaking out the "getLTSAData" into a new hook/function, but was running into issues.

frontend/src/components/maps/leaflet/LayerPopup/LayerPopupContent.tsx
```diff
@@ -75,11 +75,15 @@ export const LayerPopupContent: React.FC<IPopupContentProps> = ({ data, config,
 
   useEffect(() => {
     getLTSAData();
-  }, [data]);
+  }, [data.PID]);
 
   const getLTSAData = async () => {
```
LawrenceLau2020, line 80: maybe we can break this function out into a "reusable" component.....instead of having it in 2 files.

frontend/src/features/properties/hooks/useParcelLayerData.tsx
```diff
@@ -152,28 +155,55 @@ const useParcelLayerData = ({
...
+  // make the call to the LTSA api
+  const getLTSAData = async (pid: string) => {
+    try {
+      const ltsaData = await api.getLTSA(pid);
+      setLtsa(ltsaData);
+      return ltsaData;
+    } catch (e) {
+      setLtsa(undefined);
+    }
+  };
```
dbarkowsky, line 170: Agree with you that we don't really need the state. Could just do this here. Plus some other changes seen below.
```suggestion
const api = useApi();
```

frontend/src/features/properties/hooks/useParcelLayerData.tsx
```diff
@@ -152,28 +155,55 @@ const useParcelLayerData = ({
...
+    // Define an async function to fetch LTSA data
+    const fetchLTSAData = async () => {
+      if (
+        !!formikRef?.current &&
+        isMouseEventRecent(parcelLayerData?.e) &&
+        !!parcelLayerData?.data
+      ) {
+        if (!parcelId) {
+          const pid = parcelLayerData?.data?.PID || parcelLayerData?.data?.PID_NUMBER?.toString();
+          const ltsaResponseData = await getLTSAData(pid);
```
dbarkowsky, line 187: Then we would only need...
```suggestion
          const ltsaResponseData = await api.getLTSA(pid);
```

frontend/src/features/properties/hooks/useParcelLayerData.tsx
```diff
@@ -152,28 +155,55 @@ const useParcelLayerData = ({
...
+          setShowOverwriteDialog(true);
+        }
       }
-    }
-  }, [formikRef, getByType, parcelId, parcelLayerData]);
+    };
+
+    // Call the async function
+    fetchLTSAData();
+  }, [formikRef, getByType, parcelId, parcelLayerData, ltsa]);
```
dbarkowsky, line 206: And we shouldn't need this
```suggestion
  }, [formikRef, getByType, parcelId, parcelLayerData]);
```

frontend/src/features/properties/hooks/useParcelLayerData.tsx
dbarkowsky: Agreed. See below comment.

dbarkowsky (changes requested):

dbarkowsky (approved):
Noting that we haven't met the test threshold for this change.
Other than that, I don't see any issues. 
Tested adding a new parcel and using the parcel popup.

PR #1709, opened 2023-09-19
PIMS-223 Upload Properties Page

dbarkowsky:
## 🎯 Summary

  
[PIMS-223: ](https://citz-imb.atlassian.net/browse/PIMS-223?atlOrigin=eyJpIjoiMWE1MDc2ZmY3Y2E4NGU4NTk3NWQ0YTNjMzgyMmI2MjkiLCJwIjoiaiJ9)

Adds a UI to allow for CSV upload of properties.

- Initial page with instructions and file select.
- Secondary page that shows upload progress.
- Option to download CSV of PIDs and whether or not the upload for that PID was successful.

You will need System Administrator role to test. The page is found under Administration > Upload.

## 🔰 Checklist


dbarkowsky:
Ready to release for review.
After asking Boyd about some more specific requirements, it seems that the immediate use will only be for Land. It should also work with buildings, but there's a caveat that the API doesn't send back a way to identify those buildings, so as long as it doesn't throw an error, then it's assumed buildings were added.

When we rewrite the API in the coming months, I would like to completely rethink the import route so that it gives better feedback.

dbarkowsky:
Message me if you want a minimal file to test with.

LawrenceLau2020 (commented):
Just starting to test.
It would be nice to have a "retry" button, if it fails, as it seems you need to come back to the page from the nav menu.
image

Refreshing the page, takes you back to the home page for some reason.

LawrenceLau2020 (commented):
I'm testing the upload of an existing single property to test how it updates the parcel, but for some reason after uploading, I'm unable to search for the parcel by PID.

The parcel still exists in the DB, but for some reason it's not showing up on the map or on the inventory page.

Still investigating the issue...

dbarkowsky:
> Just starting to test. It would be nice to have a "retry" button, if it fails, as it seems you need to come back to the page from the nav menu. image
> 
> Refreshing the page, takes you back to the home page for some reason.

Seems like I should add something in case the property name is undefined like that as well.

backend/dal/Services/Admin/Concrete/ParcelService.cs
```diff
@@ -364,8 +364,20 @@ public override void Update(Parcel parcel)
...
+                .Include(p => p.Fiscals)
+                .Include(p => p.Buildings).ThenInclude(pb => pb.Building).ThenInclude(b => b.Evaluations)
+                .Include(p => p.Buildings).ThenInclude(pb => pb.Building).ThenInclude(b => b.Fiscals)
+                .Include(p => p.Buildings).ThenInclude(pb => pb.Building).ThenInclude(b => b.Address)
+                .Include(p => p.Parcels).ThenInclude(pp => pp.Parcel)
+                .Include(p => p.Subdivisions).ThenInclude(pp => pp.Subdivision)
+                .SingleOrDefault(p => p.Id == parcel.Id) ?? throw new KeyNotFoundException();
+
             parcel.PropertyTypeId = originalParcel.PropertyTypeId;
+            this.Context.Entry(originalParcel.Address).CurrentValues.SetValues(parcel.Address);
```
dbarkowsky, line 380: @LawrenceLau2020  Do we need to do something similar for the buildings as well?

LawrenceLau2020: I think that depends on the requirements of this feature, I'm not sure if SRES needs to be able to bulk upload or bulk update buildings? I recall Lindsay saying it wasn't needed, but maybe we need to double check this.

backend/api/Areas/Tools/Controllers/ImportController.cs
```diff
@@ -65,7 +66,7 @@ public ImportController(ILogger<ImportController> logger, IPimsService pimsServi
         /// <returns>The properties added.</returns>
         [HttpPost("properties")]
         [Produces("application/json")]
-        [ProducesResponseType(typeof(IEnumerable<Model.ParcelModel>), 200)]
+        [ProducesResponseType(typeof(IEnumerable<Model.ImportPropertyModel>), 200)]
```
LawrenceLau2020, line 69: Used the ImportPropertyModel instead of the ParcelModel, as we need to be able to detect whether buildings were updated or added.

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -247,19 +248,45 @@ public IEnumerable<Entity.Parcel> AddUpdateProperties(IEnumerable<Model.ImportPr
...
+                        property.Updated = true;
+                    }
+                    propertiesAddedOrEdited.Add(property);
                 }
                 else if (String.Compare(property.PropertyType, "Building") == 0)
                 {
-                    var parcel = AddUpdateBuilding(property, pid, agency);
-                    if (!entities.Any(p => p.PID == parcel.PID))
+                    var isBuildingExisting = _pimsAdminService.Building.GetByNameAddressWithoutTracking(property.Name, property.CivicAddress);
+                    // need to check the count before a building gets added, as the previous variable gets updated once the building has been added
```
LawrenceLau2020, line 272: Created a new function to check for an existing building before adding or updating the building.

dbarkowsky:
Newest changes now have all properties returned from API with their upload status and an error message if needed. 
Frontend progress feed will display those errors as seen here:
image

backend/api/Areas/Tools/Helpers/ImportPropertiesHelper.cs
```diff
@@ -297,6 +297,13 @@ public IEnumerable<Model.ImportPropertyModel> AddUpdateProperties(IEnumerable<Mo
                             property.Added = true;
                             property.Updated = false;
                         }
+                        else if (isBuildingExisting.Count() >= 2)
+                        {
+                            // there seems to be more than one building with the same name and address.....need to determine which one to update?
+                            property.Added = false;
+                            property.Updated = false;
+                            property.Error = isBuildingExisting.Count() + " buildings were found with the same name. Couldn't tell which one to update.";
+                        }
```
dbarkowsky, line 306: Good idea. I wonder if we should just throw an exception here though, as the catch already handles setting Added and Updated to false and setting the error message.

```cs
throw new Exception(isBuildingExisting.Count() + " buildings were found with the same name. Couldn't tell which one to update.");
```

dbarkowsky: Also because if we don't throw, I think it will still attempt the update at the bottom of this `try` block.

frontend/src/features/properties/upload/Instructions.tsx
```diff
@@ -0,0 +1,87 @@
...
+                ))}
+              </ul>
+            </Accordion.Body>
+          </Accordion.Item>
+        </Accordion>
+      </li>
+    </ul>
+
+    {/* CAVEATS */}
+    <h4>Caveats</h4>
```
dbarkowsky, line 74:
```suggestion
    <h4>Notes</h4>
```
Lindsay asks for a wording change here + word colour red.

dbarkowsky: Adding now

Sharala-Perumal:
Tested the PR with 3 different property subsets.
Things observed:
 - Both parcels and buildings were uploaded properly for the most part.
 - Failed uploading a subdivision which is correct as Bulk upload feature is not designed to upload subdivisions.
 - When a building/parcel with a blank name is uploaded, the upload says completed successfully, but upload didn't work as intended.
 -  For eg, in my subset I included 2 buildings (Langley Header House and Langley Campus) with same Address, same PIN with 'Name' column as blank. The upload completed successfully, but ended up updating only one building record instead of 2 different records as shown in the pic.
 
Building with blank name

There are few buildings and parcels in PIMS that have the Name field as blank(as its not a mandatory field). If the upload data has blank name, there is a potential risk that the correct property record is not updated/uploaded.

LawrenceLau2020:
> Tested the PR with 3 different property subsets. Things observed:
> 
> * Both parcels and buildings were uploaded properly for the most part.
> * Failed uploading a subdivision which is correct as Bulk upload feature is not designed to upload subdivisions.
> * When a building/parcel with a blank name is uploaded, the upload says completed successfully, but upload didn't work as intended.
> * For eg, in my subset I included 2 buildings (Langley Header House and Langley Campus) with same Address, same PIN with 'Name' column as blank. The upload completed successfully, but ended up updating only one building record instead of 2 different records as shown in the pic.
> 
> Building with blank name
> 
> There are few buildings and parcels in PIMS that have the Name field as blank(as its not a mandatory field). If the upload data has blank name, there is a potential risk that the correct property record is not updated/uploaded.

I can add the property description to the "GetByNameAddressWithoutTracking" method call.

Sharala-Perumal:
> > Tested the PR with 3 different property subsets. Things observed:
> > 
> > * Both parcels and buildings were uploaded properly for the most part.
> > * Failed uploading a subdivision which is correct as Bulk upload feature is not designed to upload subdivisions.
> > * When a building/parcel with a blank name is uploaded, the upload says completed successfully, but upload didn't work as intended.
> > * For eg, in my subset I included 2 buildings (Langley Header House and Langley Campus) with same Address, same PIN with 'Name' column as blank. The upload completed successfully, but ended up updating only one building record instead of 2 different records as shown in the pic.
> > 
> > Building with blank name
> > There are few buildings and parcels in PIMS that have the Name field as blank(as its not a mandatory field). If the upload data has blank name, there is a potential risk that the correct property record is not updated/uploaded.
> 
> I can add the property description to the "GetByNameAddressWithoutTracking" method call, so that we are searching for buildings in the db with the name, address, and description.

Lawrence, I tested the changes and the properties are uploaded properly. I'll approve the request as it looks good to me.

Sharala-Perumal (approved):

PR #1818, opened 2023-10-16
PIMS-303 reformatting gray bar styling

TaylorFries:
## 🎯 Summary

  
[PIMS-303: ](https://citz-imb.atlassian.net/browse/PIMS-303)

Gray Bar styling: 
- issue noted where gray bar was twice as tall as it should be taking up a lot of screen space. 

## 🔰 Checklist

### DIFF
--- a/frontend/src/features/properties/filter/PropertyFilterOptions.tsx
+++ b/frontend/src/features/properties/filter/PropertyFilterOptions.tsx
@@ -37,7 +37,6 @@ export const PropertyFilterOptions: React.FC<IPropertyFilterOptions> = ({ disabl
 
   return (
     <InputGroup
-      style={{ marginLeft: '-5px', width: '136px', marginRight: '10px' }}
       fast={
         searchBy === 'address' ? (
           <GeocoderAutoComplete
@@ -61,7 +60,6 @@ export const PropertyFilterOptions: React.FC<IPropertyFilterOptions> = ({ disabl
       prepend={
         <Select
           field="searchBy"
-          customStyles={{ marginLeft: '10px' }}
           options={state.options}
           onChange={reset}
           disabled={disabled}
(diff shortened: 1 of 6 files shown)


TaylorFries:
All items fit into gray bar with latest update. Spacing between elements is not consistent but with the way that elements are created and fit into the gray bar I don't think there is a simple way for consistent spacing.

frontend/src/features/properties/filter/PropertyFilterOptions.tsx
```diff
@@ -52,16 +52,15 @@ export const PropertyFilterOptions: React.FC<IPropertyFilterOptions> = ({ disabl
...
-          />
+          /></div>
         ) : (
           false
         )
       }
       formikProps={null as any}
       prepend={
-        <Select
+        <Select className='test2'
```
dbarkowsky, line 62: Was this intentionally left in?

dbarkowsky:
I think we need to consider some minimum sizings for fields. It's not clear what some fields are for if the placeholder text isn't visible. 
Maybe we could also dictate some controlled breakpoints to prevent just one element from going to the next line. 

What do you think about the Surplus Properties button? Would it be easier if we asked about taking the size of the button down?

frontend/src/features/properties/filter/PropertyFilterOptions.tsx
TaylorFries: thanks for catching that! removed.

TaylorFries:
> I think we need to consider some minimum sizings for fields. It's not clear what some fields are for if the placeholder text isn't visible. Maybe we could also dictate some controlled breakpoints to prevent just one element from going to the next line.
> 
> What do you think about the Surplus Properties button? Would it be easier if we asked about taking the size of the button down? image

I resized the button - it is inline now (I also fixed the spacing in the gray bar within the surplus properties popup)
I have tried adding in minimum sizing for as many fields as I could. There are some that take the minimum sizing and break into two lines - I have left those or kept the min width to fit-content.

Sharala-Perumal (approved):

PR #1832, opened 2023-10-19
[PIMS-96] Table Filter Not Clearing

dbarkowsky:
## 🎯 Summary

  
[PIMS-96: ](https://citz-imb.atlassian.net/browse/PIMS-96?atlOrigin=eyJpIjoiMzgyZjgxNjZhNWY0NDYwMzg5MDc3MmIwMzRiZTg4ODIiLCJwIjoiaiJ9)

### Issue
When on the View Inventory page, the filtering controls that exist in the table header were not behaving properly when cleared. Specifically, removing a filter from Classification or Location would not update the filter.

### Cause
There were a few issues. 
- The useEffect in Table.tsx only ever set the filter with the props.filter passed in from the parent component. It didn't seem like that filter had the updated information.
- The values `administrativeArea` and `classificationId` were completely removed from the filter object when clearing the values from the table filter, but the filter needs empty strings, otherwise it will not override the existing value for that filter key.
- The updated values were not being passed back to the `onFilter` function, so they were never being set again.

### Fix
Made sure there are defaults for the missing fields and that the new filter object is included when setting the filter values.

### Future Work
The way the overarching filter at the top of the page and this individual table filter interact is incredibly unpredictable.
They are not sharing once source, so changing the value in one filter doesn't always update the other filter. 
This table/filter could use a redesign. It is supposed to be extremely universal, but we might be asking too much of it here.
Additionally, the style of the table could use a rework. See how squished everything is and how the headers don't line up with their columns.

image

Maybe we could consolidate the filter into a cleaner option like our overarching filter, but with a modal for in-depth filtering
A good example of this would be like the filter on the site realtor.ca

## 🔰 Checklist

### DIFF
--- a/frontend/src/components/Table/ColumnFilter.scss
+++ b/frontend/src/components/Table/ColumnFilter.scss
@@ -1,0 +1,27 @@
+@import '../../colors.scss';
+
+.filter-icon {
+  font-size: 1em;
+  margin: 0 0.5em;
+}
+
+.input-container {
+  position: absolute;
+  left: 0;
+  top: 2em;
+  z-index: 1000;
+  min-width: 20em;
+
+  .form-control {
+    border-radius: 0;
+    border-color: $light-variant-color;
+  }
+}
+
+.filter-wrapper {
+  display: flex;
+  align-items: center;
+  justify-items: center;
+  position: relative;
+  padding: 0 auto;
+}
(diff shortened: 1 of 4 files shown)


dbarkowsky:
Code Climate reporting 0% coverage again. This is the result from running coverage locally:
image

TaylorFries (approved):
builds well with no apparent issues. All filters clear properly and reset search results from View Property Inventory page. Approved

PR #1902, opened 2023-11-16
PIMS-153 & PIMS-493

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-153: ](https://citz-imb.atlassian.net/browse/PIMS-153)
pull PID info from PMBC layer - for the search function
[PIMS-493: ](https://citz-imb.atlassian.net/browse/PIMS-493)
Map Reset button isn't working properly

PIMS-153: You should be able to search for a PID in the map using the search filter based on existing inventory as well as from the parcel layer.
PIMS-493: After using a filter for a property search on the map, clicking the reset button should zoom out and show all inventory properties in BC

***Hopefully this bug is now fixed, if you try searching by PID in the parcel layer when the map first loads, it will find the result but will not zoom to the location. If you click the "refresh/reset" icon and then copy and paste the PID and search again, it will zoom in, and all subsequent searches should also zoom to the location.

## 🔰 Checklist


dbarkowsky (commented):
A couple of things I notice:
- On first load, it always zooms in on Dease Creek. 
- If try a search twice in a row with the same PID, it doesn't look for the parcel. I think it must be a useCallback somewhere. 
- If I search for a PID that isn't found, it still zooms to max zoom as if it did find one.

dbarkowsky (approved):
Tested the following scenarios:
- [x] Searching only location, without PID.
- [x] Searching for PID that exists in PIMS inventory.
- [x] Searching for PID that only exists in parcel layer.
- [x] Searching for PID that doesn't exist or has invalid input.
- [x] Searching for Location and PID, both existing and not.

Everything appears to be handled as expected. 
There are two remaining quirks, but I think they are low priority and not easily implemented without bigger changes.
- Clicking search without changing the parameters of that search doesn't trigger the search again. I think the callback in `useDeepCompareEffect` is preventing that.
- If you do click search, then move the map, it triggers the adjustment of map bounds and zoom.

Approving these current changes with the thought that we might have to change the way the UI handles these events in the future.

LawrenceLau2020:
> Tested the following scenarios:
> 
> * [x]  Searching only location, without PID.
> * [x]  Searching for PID that exists in PIMS inventory.
> * [x]  Searching for PID that only exists in parcel layer.
> * [x]  Searching for PID that doesn't exist or has invalid input.
> * [x]  Searching for Location and PID, both existing and not.
> 
> Everything appears to be handled as expected. There are two remaining quirks, but I think they are low priority and not easily implemented without bigger changes.
> 
> * Clicking search without changing the parameters of that search doesn't trigger the search again. I think the callback in `useDeepCompareEffect` is preventing that.
> * If you do click search, then move the map, it triggers the adjustment of map bounds and zoom.
> 
> Approving these current changes with the thought that we might have to change the way the UI handles these events in the future.

Agreed.

PR #1907, opened 2023-11-17
PIMS-89 Investigate Removing react-click-away-listener

dbarkowsky:
## 🎯 Summary

  
[PIMS-89: ](https://citz-imb.atlassian.net/browse/PIMS-89?atlOrigin=eyJpIjoiMjM3NjU4OTMxMWIzNDQxMmE4OTUxMjFlMTJjYTg1NTEiLCJwIjoiaiJ9)

Looked at removing/replacing the react-click-away-listener with something custom.
It was working in all situations except one, which was just a strange implementation, I think. 

Tried to write a custom one, which worked everywhere except the table filters. Ended up just keeping the existing package, as there's no strong reason to remove it at this point and it is working.

I'll include the custom code below for future use.

Also had to add console.error suppression for a recently updated test.

## 🔰 Checklist

### DIFF
--- a/frontend/src/components/SearchBar/FindMorePropertiesForm.test.tsx
+++ b/frontend/src/components/SearchBar/FindMorePropertiesForm.test.tsx
@@ -70,6 +70,9 @@ const store = mockStore({
   lookupCode: lCodes,
 });
 
+// Suppress console.error messages
+jest.spyOn(console, 'error').mockImplementation(() => {});
+
 const RenderHelper = () => (
   <Provider store={store}>
     <MemoryRouter initialEntries={[history.location]}>
--- a/frontend/src/components/maps/leaflet/Legend/LegendControl.scss
+++ b/frontend/src/components/maps/leaflet/Legend/LegendControl.scss
@@ -1,4 +1,5 @@
 @import '../../../../styles.scss';
+@import '../../../../colors.scss';
 
 .legendTooltip {
   .arrow::before {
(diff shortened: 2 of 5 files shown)


dbarkowsky:
The custom clickaway listener that didn't end up getting included:

``` ts
import React, { MutableRefObject, useMemo, useRef } from 'react';

/**
 * Hook that alerts clicks outside of the passed ref
 */
const useOutsideAlerter = (ref: MutableRefObject<unknown>, onClickAway: () => void) => {
  useMemo(() => {
    /**
     * Alert if clicked on outside of element
     */
    const handleClickOutside: EventListener = (event) => {
      if (
        ref.current &&
        ref.current !== null &&
        !(ref.current as HTMLDivElement).contains(event.target as Node)
      ) {
        onClickAway();
      }
    };
    // Bind the event listener
    document.addEventListener('mouseup', handleClickOutside);
    return () => {
      // Unbind the event listener on clean up
      document.removeEventListener('mouseup', handleClickOutside);
    };
  }, [ref]);
};

interface IClickAwayListenerProps {
  children: React.ReactElement;
  onClickAway: () => void;
}

/**
 * Component that alerts if you click outside of it
 */
const ClickAwayListener = (props: IClickAwayListenerProps) => {
  const { children, onClickAway } = props;
  const wrapperRef: MutableRefObject<HTMLDivElement | null> = useRef(null);
  useOutsideAlerter(wrapperRef, onClickAway);

  return <div ref={wrapperRef}>{children}</div>;
};

export default ClickAwayListener;

```

LawrenceLau2020 (approved):

PR #1917, opened 2023-11-20
PIMS-288 Parcel Details Changing Unexpectedly

dbarkowsky:
## 🎯 Summary

  
[PIMS-288: ](https://citz-imb.atlassian.net/browse/PIMS-288?atlOrigin=eyJpIjoiZmViZmU3Yzc0MmY3NDJlZWIxMGU1Yzk1NmM4NWQ5ZDYiLCJwIjoiaiJ9)

## For Buildings
Added a state variable to track if the pin was picked up. The lat and long should only update if that's the case. 

## For Parcels
Shouldn't update unless edit mode is true. Might be a good idea to rework the land select form for parcels to act more like the one for buildings, where you have to pick up the pin in order to change fields with it.

## 🔰 Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -7,7 +7,7 @@
     "build": "vite build",
     "test": "react-scripts test --watchAll=false",
     "coverage": "react-scripts test --coverage --watchAll=false",
-    "coverage:changed": "react-scripts test --coverage --watchAll=false --changedSince=dev",
+    "coverage:changed": "react-scripts test --coverage --watchAll=false --changedSince=main",
     "snapshots": "react-scripts test --watchAll=false --updateSnapshot",
     "cypress:open": "npx cypress open",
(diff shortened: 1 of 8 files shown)


LawrenceLau2020 (commented):
I tried editing an existing property from the "Review" tab, testing a situation where a property maybe located on the wrong parcel/PID, was able to click on the "new" parcel which populated the parcel details form, but there was not "Submit" button to save the changes.

I also tested changing the parcel from the "Step 1 - Parcel ID" tab, clicked on a different parcel, but the PID didn't update:
image

dbarkowsky:
@LawrenceLau2020 I'm not having the same issues. Let's hop on a chat when you have time, so I can see what it's doing for you.

LawrenceLau2020 (approved):
It seems to be working now, I'm able to select a new parcel and change the PID only from the Parcel ID "step", and I'm not seeing any issues.

LawrenceLau2020:
I tested submitting a new parcel as well as updating an existing parcel, and after clicking on the pin and dropping it on the map, the PID isn't getting populated on the form for some reason:

image

PR #1920, opened 2023-11-22
PIMS-83 Remove text-mask-addons

dbarkowsky:
## 🎯 Summary

  
[PIMS-83](https://citz-imb.atlassian.net/browse/PIMS-83?atlOrigin=eyJpIjoiZjg2ODI1NDViOWUwNDZhMzkzYjk2NTY5YzIyYTVjOTUiLCJwIjoiaiJ9): Remove text-mask-addons

Original goal was to make our own text mask that would handle currency. I tried a few different attempts, but there's something about how it stores the current value that wasn't being persisted each run. 

Ended up finding the code from the original dependency. With a little conversion for TypeScript, it seems to work the same as it did when we used the whole dependency.

## Testing
Find a currency field, like on the Valuation tab of a property. Try entering different values and see if the results are what you would expect.

## 🔰 Checklist


dbarkowsky:
Working on getting a test file in.

TaylorFries (approved):
Tested changes and numbers seem to update as expected. 
- only note is that it doesn't show decimal points, but the dependency also missed this. Maybe we could add a ticket to try and fix the decimals in the future?

dbarkowsky:
> Tested changes and numbers seem to update as expected.
> 
> * only note is that it doesn't show decimal points, but the dependency also missed this. Maybe we could add a ticket to try and fix the decimals in the future?

I think this is a conflict with the default settings for the masks. I'll ask what they prefer and either disallow the decimal by default or increase decimalLimit to 2.

PR #1926, opened 2023-11-23
PIMS-288 Parcel Details Change Unexpectedly Pt.2

dbarkowsky:
## 🎯 Summary

  
[PIMS-288](https://citz-imb.atlassian.net/browse/PIMS-288?atlOrigin=eyJpIjoiYzVhMzYwMmIzMzAwNDM5ZGEzODE2NTY0MTAyNDU2YzEiLCJwIjoiaiJ9)

This is a follow up PR to the original PIMS-288 #1917

Feedback from PO was that parcel details should only update when using the dropped marker, not just when in edit mode.

## Current Status
~I have it back in the original state before PR #1460, where I think this issue started. 
This means that only the lat and long update with the marker, but fields aren't auto-populated.~

~The drop pin seems to be working with the parcels. Info only changes when using the pin, not otherwise.
With buildings, there's an issue where using the drop pin picks up the info from the previous click. Not sure where it's being set or how to track it down.~

Issue with buildings fixed. It was setting the namespace to undefined before the location was called. One more reason to move away from this giant formik state in the future.

## Changes
The pins in the sidemenu for buildings and for parcels now update the location and some details of the properties _only when picked up and dropped._

I ended up making a universal pin component. There actually used to be one, but it was taken out at some point.

## Testing
There are three places these pins exist:
- Building form in edit mode on the first Building Details tab.
- Land form in edit mode on the Parcel Details tab. 
- Land form when adding a new Parcel.  You have to choose the second option "Select Parcel from Map"

## Issues
Two things I would like to fix, but haven't figured out how:
- The zoom and centre when dropping the pin is off. It positions the view too far to the left. Not sure where this is happening though, and the issue exists in DEV already.
- I couldn't get the second test in MapDropPin.test.tsx to run. It would be nice to have, but technically it's already getting tested in some of the tests for parent components.

## TODO
- ~Make fields populate on click, but only when dropping the marker pin.~
- ~Marker pin should work the same on both LandForm and BuildingForm~
- ~Fix bug where building pin jumps to previous click, not current click.~
- ~Consolidate drop pins into a uniform component.~
- ~Removed `styled-components` elements.~ 
- ~Update text for user.~
- ~Add tests for coverage~

## 🔰 Checklist


TaylorFries (approved):
dropping pin onto map updates some fields in all three spots. I think that the zoom issue is something larger than this ticket and should be investigated separately.

PR #1961, opened 2023-12-04
PIMS-511: Opening a property from the View Inventory page doesn't take you to location

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-511: ](https://citz-imb.atlassian.net/browse/PIMS-511) - Opening a property from the View Inventory page doesn't take you to location

Because the "search" function is called when viewing a property from the view inventory page, I added a way to check if the user is on the "map" page or viewing the property details page. This should be tested thoroughly from all areas of the PIMS application.

There is still an issue with the marker not being centered.

## 🔰 Checklist


frontend/src/components/maps/leaflet/InventoryLayer.tsx
```diff
@@ -266,7 +274,7 @@ export const InventoryLayer: React.FC<InventoryLayerProps> = ({
           // Fit to municipality bounds
           map.fitBounds((GeoJSON.geometryToLayer(municipality) as any)._bounds, { maxZoom: 11 });
         }
-      } else if (results.length > 0) {
+      } else if (results.length > 0 && isMapView && isSidebarOpen === 'false') {
```
LawrenceLau2020, line 277: I'm not sure if there's a more elegant way of doing this, such as preventing the search from getting called...

LawrenceLau2020: I see, is this also the same issue in production?

dbarkowsky (approved):
Tested clicking on properties from the View Inventory page. Goes to the map and zooms appropriately.
Made sure search filters still zoom as expected. 
Fixes bug with map resizing as well.
All looks good.

PR #1968, opened 2023-12-08
PIMS-405 Add Logging to Express API

dbarkowsky:
## 🎯 Summary

  
[PIMS-405](https://citz-imb.atlassian.net/browse/PIMS-405?atlOrigin=eyJpIjoiMzYwNDQ0MmE4NDBmNDUzOGExNzU0NTMxNDQ4NWRhZTMiLCJwIjoiaiJ9)

Decided to use both `morgan` and `winston` because they excel at different areas. `morgan` works with Express natively to get the details of http requests, which I don't see `winston` doing nicely. `winston` gives better output with more configurations, can replace console.log statements, and will be easier to eventually send to a service like Kibana if desired.

## Changes
- Added `morgan` and `winston` logging packages.
- Created middleware that takes `morgan` http logs and uses `winston` to log as JSON to the console
- Switched the source of the TESTING env. It is now dependant on the NODE_ENV, which is automatically `test` when running Jest tests. This will prevent the rate limiter and the logging from happening during tests.
- Added a no console.log rule to ESLint. You can still use them for local debugging, but final code should use the logger if there's a need to print something to console or log it.

Logger can be used like this:
``` ts
logger.info("Here's some info!");
logger.warn("Here's a warning!");
logger.error("Here's an error!");
```
There are other logger functions as well, although these should cover most situations.

Log output looks like this:
image

There was one line I couldn't get coverage on, because the NODE_ENV is `test` while the tests run, but that line only runs when not testing. I had to lower the branches test threshold in order to pass. This is something we might have to review later unless anyone wants to try and get coverage on that line.

## Testing
1. Stop the .NET API if it's running.
2. In `express-api`, run `npm test` to see that tests pass.
3. Start the API with `npm run dev`.
4. Use a tool like Postman or Thunderclient to send a GET request to localhost:5000/api/v2/health, and observe the console for a log message.

## 🔰 Checklist

### DIFF
--- a/express-api/.eslintrc.cjs
+++ b/express-api/.eslintrc.cjs
@@ -20,6 +20,7 @@ module.exports = {
     'no-extra-boolean-cast': 'off',
     'no-unsafe-optional-chaining': 'off',
     'no-prototype-builtins': 'off',
+    'no-console': 'error',
     '@typescript-eslint/ban-ts-comment': 'off',
     '@typescript-eslint/no-inferrable-types': 'off', // ie. const val: number = 4;
     '@typescript-eslint/no-empty-function': 'off', // ie. {}
--- a/express-api/constants/switches.ts
+++ b/express-api/constants/switches.ts
@@ -1,5 +1,5 @@
-const { TESTING } = process.env;
+const { NODE_ENV } = process.env;
 
 export default {
-  TESTING: `${TESTING}`.toLowerCase() === 'true', // Can be used to disable certain features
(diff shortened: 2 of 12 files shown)


LawrenceLau2020 (approved):
Looks good!
all 5 tests passed, and sending a request to the health endpoint gives this response:
`{"level":"http","message":{"method":"GET","responseTimeMs":0.272,"route":"/api/v2/health","status":200},"timestamp":"2023-12-08 02:11:54.116 PM"}`

PR #1970, opened 2023-12-08
PIMS-530: Error when trying to approve access requests

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-530:](https://citz-imb.atlassian.net/browse/PIMS-530) Error when trying to approve access requests

This ticket is to help debug the issue when trying to approve users that don't have any "roles".

We are currently unable to approve a user access request, because according to the application it's unable to retrieve the roles from Keycloak.
image

It looks like this is not the only user it happens to, but if a user doesn't have any roles, then we can't approve their access request using the access request page and get an error (see the Jira ticket for specifics). We are also unable to view or change the user's roles from PIMS. Strange thing is that the user can see his role, when clicking on the drop down by his username.

## 🔰 Checklist

### DIFF
--- a/backend/dal/Services/Admin/Concrete/UserService.cs
+++ b/backend/dal/Services/Admin/Concrete/UserService.cs
@@ -400,12 +400,14 @@ namespace Pims.Dal.Services.Admin
             // Keycloak Gold only wants the clientID as a number, which is always at the end of the id, after a "-"
             string frontendId = this.configuration["Keycloak:FrontendClientId"].Split("-").Last();
             HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, $"https://api.loginproxy.gov.bc.ca/api/v1/{getEnv()}/{idp}/users?email={email}");
+            _logger.LogDebug($"Test Logging the request for username to Keycloak '{request}'");
             request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
             HttpResponseMessage response = await _httpClient.SendAsync(request);
             if (!response.IsSuccessStatusCode) throw new Exception("Unable to get user's username from Keycloak Gold");
             string payload = await response.Content.ReadAsStringAsync();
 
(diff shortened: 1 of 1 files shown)


backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -335,13 +341,17 @@ public async Task<IEnumerable<string>> GetGoldUsersRolesAsync(string preferred_u
...
             HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, this.configuration["Keycloak:BaseURL"] + $"integrations/{frontendId}/{getEnv()}/users/{preferred_username}/roles");
+            _logger.LogDebug($"Test Logging the request to Keycloak '{request}'");
+
             request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
             HttpResponseMessage response = await _httpClient.SendAsync(request);
             if (!response.IsSuccessStatusCode) throw new Exception("Unable to fetch roles from Keycloak Gold");
             string payload = await response.Content.ReadAsStringAsync();
             JsonDocument json = JsonDocument.Parse(payload);
 
+            _logger.LogDebug($"Test Logging the response from Keycloak '{json.ToString()}'");
```
LawrenceLau2020, line 352: Log the response from Keycloak

backend/dal/Services/Admin/Concrete/UserService.cs
```diff
@@ -335,13 +341,17 @@ public async Task<IEnumerable<string>> GetGoldUsersRolesAsync(string preferred_u
...
             IEnumerable<string> roles = json.RootElement.GetProperty("data").EnumerateArray().Select(r => r.GetProperty("name").GetString());
+            _logger.LogDebug($"Test Logging the Keycloak roles: '{roles}'");
```
LawrenceLau2020, line 355: see what roles we are getting back.

dbarkowsky (approved):

PR #1980, opened 2023-12-14
PIMS-403 Add Keycloak to Express API

dbarkowsky:
## 🎯 Summary

  
[PIMS-403](https://citz-imb.atlassian.net/browse/PIMS-403?atlOrigin=eyJpIjoiMzNkNzAyMTgwN2IwNDRmMmJmZjM3NjFiZThmYzVhYmUiLCJwIjoiaiJ9)

**~Work in Progress~**
~Need to address some of the component's functionality with Brady. Expecting additional changes.~

Talked with Brady this morning. There will be some changes, but they won't affect this current use.

I will create another ticket that addresses our Keycloak integrations. They need to be cleaned up, which means having a single integration (with dev, test, and prod settings) that contains all our necessary roles and role mappings. This might involve removing some unused roles. 

Here's the repo for the Keycloak React package. The main readme contains useful info on how to protect routes.
https://github.com/bcgov/citz-imb-kc-express

Note: ~There is currently a console.log in the Keycloak package. I've raised this issue in that repo.~ Resolved

## Changes
Adds the initialization for Keycloak. 

There are no routes to protect at this point, but you can test it by altering the `/health` route. To do so:
1. Import `protectedRoute` from the Keycloak package:
``` ts
import { protectedRoute } from '@bcgov/citz-imb-kc-express';
```
2. Change the route in `express.ts` to look like this:
``` ts
app.use(`/api/v2`, protectedRoute(), router.healthRouter);
```
3. Make sure the Keycloak ENVs are set. I used the ones for the backend, because it has a secret key.
4. Try sending a request to the `/health` endpoint without a token. It should say you're unauthenticated.
5. Copy a token from DEV PIMS in the browser, and try the `/health` endpoint again with that token. It should allow it.

## 🔰 Checklist


TaylorFries (approved):
Adding protected port to health route requires token to access and once proper token is provided the endpoint can be reached. 
Approved

PR #1984, opened 2023-12-18
PIMS-426 Routes LTSA

dbarkowsky:
## 🎯 Summary

  
[PIMS-426](https://citz-imb.atlassian.net/browse/PIMS-426?atlOrigin=eyJpIjoiYTU3M2JjMjU3ZTk5NGY3ZDliZTZkZmRhOGI0Y2E3MGIiLCJwIjoiaiJ9)

## Changes
Creating route for LTSA GET request. 
Currently uses a stub response. Recommend using this until the controller logic is completed at a later date.

## Testing
Using a token from PIMS, call the `api/v2/ltsa/land/title` endpoint. Should get a 501 response.

## Future Work
- Need to determine how Keycloak will be handled in future integration tests. Currently in conversations with the SSO team on this. **Update**: They recommend mocking any of the authentication. I added an example of how to add user info into a Request in the unit test. Not sure yet if we can mock the entire keycloak package. Might need to just do a bypass of it entirely.
- Need to be specific in which roles can access this route. Right now it is all of them.

## 🔰 Checklist


TaylorFries (approved):
got 501: Not yet implemented which is what we expect at this point. Approved.

dbarkowsky:
Noting that the swagger comments for the LTSA controller don't seem to be picked up by the swagger-autogen. Waiting to see what happens with future routes. Tried a few things, but for some reason it's only that route while the health check works fine.

PR #1992, opened 2023-12-21
PIMS-419: Parcel stubs

GrahamS-Quartech:
## 🎯 Summary

[PIMS-419](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-419)

Routes and controllers for the parcels endpoints were added. All responses are merely stubs for now.
Tests were also added for the controller, as well as some rudimentary unit tests.
I've taken the liberty of adding a couple additional packages in this PR. 
Faker is for automatically generating mock data for testing. 
Sinon is to assist in stubbing and faking functionality outside of individual tests, which is useful for factory type classes and functions that are going to be reused and reset over and over again. That's why I've incorporated it into the mockRequest and mock Response.

Please see the comments I've added for some additional thoughts.

## 🔰 Checklist

- [x ] I have read and agree with the following checklist and am following the guidelines in our [Code of Conduct](CODE_OF_CONDUCT.md) document.


dbarkowsky (approved):
Changes look good. Happy to see a standardized mocking for requests and responses.

Noted one thing not specific to this PR:
We might have an issue with the swagger-autogen. It worked fine for /health, but it hasn't picked up the comments in any additional routes. It also doesn't seem to interpret string literals.
image


Another thing for the future:
We should think about whether all the PUT requests should really be PUT or if they'd be more accurate as PATCH. Easy to change later depending on how they get implemented.

PR #1995, opened 2023-12-23
PIMS-417 Users stubs

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-417: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-417)

Added Users controllers, routes, and tests.
Also added the Zod library so we can have some object validation for API requests and responses. We may be able to replace some of these schemas with TypeORM functionality later, but stuff like the filter bodies and keycloak user info will most likely remain regardless.

## 🔰 Checklist


dbarkowsky (approved):
Tested out routes. Everything works as expected.
Happy to see preliminary zod schema.

Note: I removed the ^ from installed dependencies. Generally we have a process where these get updated through tickets.

express-api/src/controllers/users/usersSchema.ts
```diff
@@ -0,0 +1,124 @@
+import { z } from 'zod';
```
LawrenceLau2020, line 1: Does having the usersSchema.ts mean that an IUser.ts interface is unnecessary, as I'm assuming this kind of replaces that?

PR #2001, opened 2024-01-02
PIMS-421 Routes - Properties

dbarkowsky:
## 🎯 Summary

  
[PIMS-421](https://citz-imb.atlassian.net/browse/PIMS-421?atlOrigin=eyJpIjoiNDU3NzhmMWM5MDg2NDhjN2E3YmQyN2ExOTMzZDBkZmEiLCJwIjoiaiJ9)

These were all search routes. There was one, `properties/search/names` that I don't think is actually used currently (at least I can't find an example), so I omitted it. 

Additionally, I changed the routes with `wfs` to `geo`. These are the ones that retrieve the lists of properties for map marker population right now. `wfs` isn't a very descriptive name, so I'm hoping that `geo` will make it a little more obvious what it's used for.

Some other considerations when we start implementing controllers...
It feels like the filter routes could just be part of the GET routes, but with some query parameters. There doesn't appear to be anything confidential transmitted, so the use of a POST might not be necessary.

## Changes
- Added stub controllers for properties
- Added routes that expose these controllers
- Added stub integration and unit tests.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/index.ts
+++ b/express-api/src/controllers/index.ts
@@ -4,6 +4,7 @@ import * as ltsa from '@/controllers/ltsa/ltsaController';
 import * as parcels from '@/controllers/parcels/parcelsController';
 import * as lookup from '@/controllers/lookup/lookupController';
 import * as users from '@/controllers/users/usersController';
+import * as properties from '@/controllers/properties/propertiesController';
 
 export default {
   healthCheck,
@@ -12,4 +13,5 @@ export default {
   ...lookup,
   admin,
   ...users,
+  ...properties,
 };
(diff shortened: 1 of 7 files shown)


TaylorFries:
While testing the routes I get the following response:

`"error": "User must have all of the following roles: [Admin]"`

The following log is printed in the express-api:

` {"level":"http","message":{"method":"GET","responseTimeMs":194.385,"route":"/api/v2/properties/search","status":403},"timestamp":"2024-01-03 11:14:06.934 AM"} `

TaylorFries (approved):
All property routes return expected results. tested a few other routes too and they all act as expected.

PR #2012, opened 2024-01-03
PIMS-422 Projects stubs

GrahamS-Quartech:
## 🎯 Summary

[PIMS-422: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-422)

* Added stubs for routes, controllers, and tests.
* Added some fairly robust factories in the tests but these objects are kind of enormous so I simplified slightly in some spots.
* I skipped routes that had /report as their root cause I think that belongs in the reports ticket. There were also some routes already implemented by the notifications ticket.
* Left some notes inline for the routes again, but these ones mostly make sense. Will add to the issue discussion as well.

## 🔰 Checklist


express-api/src/controllers/projects/projectsController.ts
```diff
@@ -0,0 +1,325 @@
...
+   */
+  return stubResponse(res);
+};
+
+/**
+ * @description Update the report for the specified project.
+ * @param {Request}     req Incoming request.
+ * @param {Response}    res Outgoing response.
+ * @returns {Response}      A 200 status with the deleted report.
+ */
```
dbarkowsky, line 128:
```suggestion
 * @description Delete the report for the specified project.
 * @param {Request}     req Incoming request.
 * @param {Response}    res Outgoing response.
 * @returns {Response}      A 200 status with the deleted report.
 */
```
I wonder if we should also consider what kind of response we want to return for deletions. 200 makes sense I guess if we're returning something, but I feel like I've seen a lot of cases in the original that could have just sent a 204 response back.

express-api/src/routes/projectRouter.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const router = express.Router();
+
+export const PROJECT_DISPOSAL = '/projects/disposal';
+export const PROJECT_REPORTS = '/projects/reports';
+
+//These originally had a separate route for numeric id and projectNumber, but I don't think express supports this pattern.
+router.route(`${PROJECT_DISPOSAL}/:projectId`).get(controllers.getDisposalProject);
+router.route(`${PROJECT_DISPOSAL}/:projectId`).put(controllers.updateDisposalProject);
+router.route(`${PROJECT_DISPOSAL}/:projectId`).delete(controllers.deleteDisposalProject);
```
dbarkowsky, line 12:
```suggestion
router.route(`${PROJECT_DISPOSAL}/:projectId`)
  .get(controllers.getDisposalProject)
  .put(controllers.updateDisposalProject)
  .delete(controllers.deleteDisposalProject);
```
If the route is the same, you can always chain the methods together like this. Might be easier to read.

express-api/src/routes/projectRouter.ts
```diff
@@ -0,0 +1,40 @@
...
+router.route(`${PROJECT_DISPOSAL}`).post(controllers.addDisposalProject);
+
+//This originally had like 3 different routes that all seem to do the same thing so let's just consolidate that to one.
+router.route(`${PROJECT_DISPOSAL}/workflows`).put(controllers.requestProjectStatusChange);
+
+router.route(`${PROJECT_REPORTS}`).get(controllers.getAllProjectReports);
+router.route(`${PROJECT_REPORTS}/:reportId`).get(controllers.getProjectReport);
+router.route(`${PROJECT_REPORTS}/:reportId`).put(controllers.updateProjectReport);
+router.route(`${PROJECT_REPORTS}/:reportId`).delete(controllers.deleteProjectReport);
```
dbarkowsky, line 21: Same comment as above.

express-api/src/controllers/projects/projectsController.ts
GrahamS-Quartech: Either way would definitely be fine, I guess it just depends on whether the frontend might be able to do something useful with the entire deleted resource or not (ie. show the user that the snapshots were deleted in addition to the requested report)

dbarkowsky (approved):

PR #2027, opened 2024-01-10
PIMS-418 Keycloak Service

TaylorFries:
## 🎯 Summary

  
[PIMS-418: ](https://citz-imb.atlassian.net/browse/PIMS-418)

Started to work on this before the dev conversation on how we would set up these processes. Starting a draft pull request so the other devs can see the work that was completed before the discussion and so that we can work on this together. 

Goal changed to create a Keycloak service. 

Updated by @dbarkowsky 
## Changes
- Added keycloak service under the `/services` folder. Went with a functional design, not OOP. Seems to fit better with Express.
- Added tests for these files
- Added interfaces and zod schemas where needed.

## Future Work
There are two functions that sync data from Keycloak and update application tables. Until there are services for roles and users, they aren't completable.

There are also two issues where the package used to access the CSS API needs some adjustment. I've opened issues on the repo, but I might have to make a PR to fix them. One of them will require changes in the Keycloak service, which I noted in the comments.
- https://github.com/bcgov/citz-imb-kc-css-api/issues/2
- https://github.com/bcgov/citz-imb-kc-css-api/issues/1
> 

## 🔰 Checklist


dbarkowsky:
Looking at this further, it doesn't look like the old frontend calls any of the Keycloak routes directly. Instead, they are called from within the .NET API. 
I'm going to suggest that we move this to a Keycloak service instead that can be utilized as necessary within the API but can't be called externally.

express-api/src/services/keycloak/keycloakService.ts
```diff
@@ -0,0 +1,202 @@
...
+  // Does user exist in Keycloak?
+  // Get their existing roles.
+  // Does user exist in PIMS
+  // If user exists in PIMS...
+  // Update the roles in PIMS to match their Keycloak roles
+  // If they don't exist in PIMS...
+  // Add user and assign their roles
+};
+
+interface IKeycloakUsersFilter {
```
GrahamS-Quartech, line 108: Could just use the exported IDIRUserQuery from the keycloak package here instead

express-api/src/services/keycloak/keycloakService.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * @description Retrieves a Keycloak user that matches the provided guid.
+ * @param {string} guid The guid of the desired user.
+ * @returns {IKeycloakUser} A single Keycloak user.
+ * @throws If the user is not found.
+ */
+const getKeycloakUser = async (guid: string) => {
+  // Should be by guid. Only way to guarantee uniqueness.
+  const user: IKeycloakUser = (await getKeycloakUsers({ guid: guid, firstName: '' })).at(0); // TODO: Remove firstName after fix is applied to package.
```
GrahamS-Quartech, line 139: This would probably result in an exception if getKeycloakUsers failed for some reason, doing `?.at(0);` should be enough.

express-api/src/services/keycloak/keycloakService.ts
dbarkowsky: Good point. I think I had that originally because of the package typing I didn't like. Replaced now.

express-api/src/services/keycloak/keycloakSchemas.ts
```diff
@@ -0,0 +1,37 @@
...
+});
+
+// The generic error object that Keycloak returns.
+export const keycloakErrorSchema = z.object({
+  message: z.string(),
+});
+
+// A list of roles returned by Keycloak.
+export const keycloakUserRolesSchema = z.object({
+  data: array(keycloakRoleSchema),
```
GrahamS-Quartech, line 16:
```suggestion
  data: z.array(keycloakRoleSchema),
```
Then remove array import at top of file.

GrahamS-Quartech (approved):

PR #2031, opened 2024-01-15
PIMS-622 Frontend Testing Workflow

dbarkowsky:
## 🎯 Summary

  
[PIMS-622](https://citz-imb.atlassian.net/browse/PIMS-622?atlOrigin=eyJpIjoiYTAxMjJhYzFiN2JlNGVmZjlmMDRkMjY3ODJjNjY1ZmQiLCJwIjoiaiJ9)

The goal of this ticket has changed since it was opened.
Original goal was to have test coverage of React components checked with a workflow.
After discussion with the PIMS team/IMB Full Stack devs and research online, the new goal of this ticket is to offer the following:
- A workflow that runs existing tests for the purpose of regression tests
  - This could be snapshots on components or regular tests on other code.
- An opportunity for coverage on future files that are not components

## Changes
- Adds a workflow that runs Jest tests. Will fail only if tests fail or coverage is not met where specified.
- Current coverage targets a non-existent folder. When we have logic in files separate from React components, we can check coverage of those files. These coverage paths will need to be updated at that point.
- Added one sample snapshot test for the Footer component.

## 🔰 Checklist


TaylorFries (approved):
tests run as expected locally and on github prs.

PR #2168, opened 2024-02-03
PIMS-680/1294: Admin Users View UI & Access Request Submission

GrahamS-Quartech:
## 🎯 Summary

[PIMS-680](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-680)

* Made a reusable "DataCard" component that you can use to show some arbitrary data in a simple card view with a little edit button.
* Made some modular Dialog components as well, which should hopefully serve most use cases.
* Made a user details page according to the screenshots in the PR. It doesn't link back to anything yet though, needs the users table PR.

Not wired up to the API in this state, for that I need 1260 merged in.

Was also considering making the form bits into their own reusable component, but I'm not sure how generic one could make it. I feel like there's usually a lot of customization that needs to be done in terms of how the fields are placed, what types of field to use, what labels to use, so much so that a reusable component might be more cumbersome than just dropping in the field components into a grid as I've done here. Any thoughts on this?

[PIMS-1294](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1294)

Now you should be able to submit an access request! Except there is no access request table anymore, this just creates a user with status "OnHold."

## 🔰 Checklist


react-app/src/components/dialog/DeleteDialog.tsx
```diff
@@ -0,0 +1,40 @@
...
+const DeleteDialog = (props: IDeleteDialog) => {
+  const { open, title, message, deleteText, onDelete, onClose } = props;
+  const [textFieldValue, setTextFieldValue] = useState('');
+  return (
+    <ConfirmDialog
+      open={open}
+      title={title}
+      onConfirm={onDelete}
+      onCancel={onClose}
+      confirmButtonText={deleteText}
```
dbarkowsky, line 23: Should this have a default? Could be undefined.
```suggestion
      confirmButtonText={deleteText ?? 'Delete'}
```

react-app/src/utils/formatters.tsx
```diff
@@ -0,0 +1,27 @@
+import { Chip } from '@mui/material';
+import React from 'react';
+
+export const dateFormatter = (input: any) => {
+  if (!(input instanceof Date)) {
+    input = new Date(input);
+  }
+  return `${input.getDay()}/${input.getMonth() + 1}/${input.getFullYear()}`;
```
dbarkowsky, line 8: Based on the screenshots, I think we're looking for a MMM DD, YYYY format. 
image

dbarkowsky (changes requested):
Like the addition of the formatters. 
A few changes requested, plus linting to take care of some missing newlines.

dbarkowsky (approved):
One suggestion for how the Roles are presented, but if we're getting this from the API in the future, shouldn't be an issue. May have to adjust the enum string value though.

Believe the container issue is fixed.

PR #2176, opened 2024-02-08
PIMS-1294: Access Request Workflow

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1294](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1294)

Now you should be able to submit an access request! Except there is no access request table anymore, this just creates a user with status "OnHold."

If you want to test this properly, you'll need to make an agency with the key "BCH". Or modify the hard coded values yourself, making the agency services to enumerate the available agencies is a separate ticket.

The other side of this ticket, which is converting users into Active state, is gonna happen in PIMS-680.

You may note that the access requests services and entities are still there, wasn't sure if I should delete everything yet... Feel free to weigh in.

## 🔰 Checklist


react-app/src/hooks/useDataLoader.ts
```diff
@@ -54,7 +59,15 @@ const useDataLoader = <AFArgs extends any[], AFResponse = unknown, AFError = unk
...
 
-  return { refreshData, isLoading, data, error };
+  const loadOnce = async (...args: AFArgs) => {
+    if (oneTimeLoad) {
+      return;
+    }
+    setOneTimeLoad(true);
+    return refreshData(...args);
+  };
+
```
dbarkowsky, line 69: A little unsure of what this is doing in the grand scheme of things. Would appreciate more info.

dbarkowsky (commented):
Like the changes here. 
Would recommend waiting for the updated entities from PIMS-407, then making tweaks.
There are also some failing checks for tests/linting at the moment.

GrahamS-Quartech:
Note: I've fixed some broken tests, though it's still getting a "failure" because of not meeting full branch coverage. This is just because of ignored tests for access requests which I don't think are really worth spending time fixing since we are probably getting rid of them.

react-app/src/hooks/useDataLoader.ts
GrahamS-Quartech: Handy for something like the auth user context where we only really need to instance data the one time. This way you can just call loadDataOnce and re-render's wont spam additional requests. You can do something similar by wrapping refreshData in some useEffect/useState stuff where this behavior is desired, but this is just a handy shortcut so you don't have to.

PR #2190, opened 2024-02-14
PIMS-1329: Apply lower snake case standard for entity names

Sharala-Perumal:
PIMS-1331: Apply lower snake case standard for entity names                               

## 🎯 Summary

  
[PIMS-1331](https://citz-imb.atlassian.net/browse/PIMS-1331)

## 🔰 Checklist

- [x] I have read and agree with the following checklist and am following the guidelines in our [Code of Conduct](CODE_OF_CONDUCT.md) document.

> - I have performed a self-review of my code.
> - I have commented my code, particularly in hard-to-understand areas.
> - I have made corresponding changes to the documentation where required.
> - I have tested my changes to the best of my ability.
> - My changes generate no new warnings.

PR #2191, opened 2024-02-15
PIMS-1327: Data fixes after seeding Postgres 

LawrenceLau2020:
## 🎯 Summary
  
[PIMS-1327: ](https://citz-imb.atlassian.net/browse/PIMS-1327)

This PR adjusts some of the evaluation key and fiscal key seeds and disables the agency seeding which is replaced with ETL of the agencies based on SQL queries because of the differences in agencies across all environments.

First step you need to do is to "revert" all the data migrations that have been run by running this command in the express-api folder, you may need to do this 3 or 4 times depending on how many migrations you have in your migrations table:
`npm run migration revert`

The very last "revert" should be the one that DROPS all the tables so you're starting from scratch.

Then run the data migration to "re-seed" the tables again as well as run the other migration changes:
`npm run migration run`

Then the next step is running an ETL task in DBeaver, but before you run the ETL task in DBeaver, copy the attached tasks.json file to the following location: 
**Windows users will need to locate the existing tasks.json file in their AppData folder here**: 
`"C:\Users\YOUR_USER_NAME\AppData\Roaming\DBeaverData\workspace6\General\.dbeaver"`
**Mac users are slightly different:, see: https://dbeaver.com/docs/dbeaver/Workspace-Location/**
`~/Library/DBeaverData/`

If the above location doesn't exist you can create a "test" task in your DBeaver application to get this folder location going if it doesn't already exist after having installed DBeaver. Otherwise just replace the existing tasks.json file with the one in this PR.

You can test this by running the "PIMS ETL" task which should import agencies, users, parcels and buildings, Parcel and building evaluations, as well as parcel_buildings and parcel_fiscals, and projects. As more project related tables are included in the ETL there might be a few more data fixes.

Note: the ETL task that you need to run is the one highlighted in yellow below, located within the "PIMS ETL" folder:
image

This will populate all the tables in the Postgres database that were not "seeded". 
Once https://github.com/bcgov/PIMS/pull/2190 gets merged, there will be some changes needed in each of the ETL "jobs" to incorporate the new column name changes in that ticket.

## 🔰 Checklist


dbarkowsky (approved):

express-api/src/typeorm/Entities/ProjectAgencyResponse.ts
```diff
@@ -26,7 +26,7 @@ export class ProjectAgencyResponse extends BaseEntity {
   OfferAmount: number;
 
   // Notification Relation
-  @Column({ name: 'NotificationId', type: 'int' })
+  @Column({ name: 'NotificationId', type: 'int', nullable: true })
   NotificationId: number;
```
LawrenceLau2020, line 30: For some reason this column is never used, as I checked in production to confirm there are no references to the Notification Id. Possibly be cause an agency could have more than one notification associated to it.

dbarkowsky (approved):
Tested from scratch, running migrations then ETL. 
Everything worked as expected. My Postgres tables are populated with data from the MS SQL Server.

PR #2197, opened 2024-02-16
PIMS-1328 Express Error Handling

dbarkowsky:
## 🎯 Summary

  
[PIMS-1328](https://citz-imb.atlassian.net/browse/PIMS-1328?atlOrigin=eyJpIjoiMDk0ODNiZjJjNGQwNGU2MjgwNTIzOWIwYzBkN2ExNjIiLCJwIjoiaiJ9)

## Changes
- Added a fallback route handler that returns 404 if that route doesn't exist.
- Added some preset error responses we can use if wanted.
- Added error handler middleware to catch errors and send back a formatted response. 
  - The `next()` function accepts string, Error, and ErrorWithCode.
- Added some process listeners that will close the app in a controlled manner, closing existing connections first.

## Testing
- Easiest way to test is to add a temporary endpoint in express.ts (I used /error). This endpoint can then just throw different kinds of errors to see different responses.
``` ts
app.use('/error', (req: Request, res: Response) => {
  throw new Error('hi')
})
```
- Recommended to use next() instead, as it also allows this to work with async controllers.
``` ts
app.use('/error', async (req: Request, res: Response, next: NextFunction) => {
  next(new Error('hi'))
})
```
or even just
``` ts
app.use('/error', async (req: Request, res: Response, next: NextFunction) => {
  next('hi')
})
```
- Test the SIGTERM one by using this somewhere
``` ts
app.use('/error', async (req: Request, res: Response, next: NextFunction) => {
  process.kill(process.pid, 'SIGTERM')
})
```
- Test the non-existent route by calling a route that doesn't exist (e.g. /help)

## 🔰 Checklist


TaylorFries (approved):
Bad endpoints caught and SIGTERM handled well. App never crashed!

PR #2198, opened 2024-02-17
PIMS-1330: Sync Keycloak Roles

GrahamS-Quartech:
## 🎯 Summary
[PIMS-1330 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1330)

The self endpoint will now sync the available keycloak roles with the db and then sync the state of the calling user's roles before returning the user object.

The UI will now populate roles dropdowns with roles from the database. When you submit the user status section of the user edit view, it will sync the user's role in keycloak.

Includes updates to various keycloak, roles, and user services.

## 🔰 Checklist


dbarkowsky (approved):
Approved. Had to adjust a few things to stop getting errors when not permitted to see the users table.
I'm wondering if it's worth looking into an asynchronous load of the user's roles in the future. Waiting for the Keycloak call in /self seems to take a while.
Thinking we should also consider a ticket to load some things like Agencies and Roles lazily after login. Would avoid the delay when bringing up the dropdowns.

LawrenceLau2020 (approved):
Looks good, user role displays correctly.

PR #2208, opened 2024-02-23
PIMS-1343: Properties Data Table UI

GrahamS-Quartech:
## 🎯 Summary

[PIMS-1343](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1343)

This is just the frontend component, designed according to the PDF attached to this ticket.

As part of this PR, I actually lifted out the custom implementation of the users table into its own reusable component since the design of this table is pretty much the same. 

I wasn't sure what other filtering options should exist alongside All Properties so that's the only one for now.

## 🔰 Checklist


react-app/src/components/property/PropertyTable.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  const classificationColorMap = {
+    0: { textColor: theme.palette.blue.main, bgColor: theme.palette.blue.light },
+    1: { textColor: theme.palette.success.main, bgColor: theme.palette.success.light },
+    2: { textColor: theme.palette.info.main, bgColor: theme.palette.info.light },
+    3: { textColor: theme.palette.info.main, bgColor: theme.palette.info.light },
+    4: { textColor: theme.palette.warning.main, bgColor: theme.palette.warning.light },
+    5: { textColor: theme.palette.warning.main, bgColor: theme.palette.warning.light },
+    6: { textColor: theme.palette.warning.main, bgColor: theme.palette.warning.light },
+  };
```
dbarkowsky, line 56: I wonder if there's a way to do this with a default for >=4 in the future. Not concerned right now.
Edit: Nvm, I see how you handled this below.

dbarkowsky (approved):
Looks good, and I'm a big fan of separating the table like this. 
Thanks for also doing this as just frontend. Makes it easier to review.

react-app/src/components/property/PropertyTable.tsx
GrahamS-Quartech: It's slightly clunky, and kind of assumes stable keys for the classification id. But I think it's good enough, may revisit when we wire up the api component here.

PR #2210, opened 2024-02-23
PIMS-1349: Create Building Service

Sharala-Perumal:
PIMS-1339: Create Building Service      
 - Created Building Service
 - Created mockup data for unit test
 - Created unit test for buildingService                                 

## 🎯 Summary

  
[PIMS-1339](https://citz-imb.atlassian.net/browse/PIMS-1349)

## 🔰 Checklist


express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @param   building     Incoming building data to be added to the database.
+ * @returns {Response}   A 201 status and the building data added.
+ * @throws ErrorWithCode If the building already exists or is unable to be added.
+ */
+export const addBuilding = async (building: Building) => {
+  const existingBuilding = await getBuildingById(building.Id);
+  if (existingBuilding) {
+    throw new ErrorWithCode('Building already exists.', 409);
+  }
+  const newBuilding = buildingRepo.save(building);
```
dbarkowsky, line 20: I think .save returns a promise, so let's await it.
```suggestion
  const newBuilding = await buildingRepo.save(building);
```
Same issues exists with addParcel. Going to fix those and add to this PR as we go.

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+import { ErrorWithCode } from '@/utilities/customErrors/ErrorWithCode';
+import { DeepPartial, FindOptionsOrder } from 'typeorm';
+import { BuildingFilter } from '@/services/buildings/buildingSchema';
+
+const buildingRepo = AppDataSource.getRepository(Building);
+
+/**
+ * @description          Adds a new building to the datasource.
+ * @param   building     Incoming building data to be added to the database.
+ * @returns {Response}   A 201 status and the building data added.
```
dbarkowsky, line 12: This would return the building, not a response. The controller returns the response.

I'm seeing that this issue exists with the parcelServices.addParcel as well.

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+export const getBuildingById = async (buildingId: number) => {
+  try {
+    const findBuilding = await buildingRepo.findOne({
+      where: { Id: buildingId },
+    });
+    return findBuilding;
+  } catch (e) {
+    throw new ErrorWithCode(e.message, e.status);
+  }
```
dbarkowsky, line 37: It's ok to not catch anything here. The findOne doesn't throw an error, it just returns null if not found.
```suggestion
  const findBuilding = await buildingRepo.findOne({
    where: { Id: buildingId },
  });
  return findBuilding;
```

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const buildingRepo = AppDataSource.getRepository(Building);
+
+/**
+ * @description          Adds a new building to the datasource.
+ * @param   building     Incoming building data to be added to the database.
+ * @returns {Response}   A 201 status and the building data added.
+ * @throws ErrorWithCode If the building already exists or is unable to be added.
+ */
+export const addBuilding = async (building: Building) => {
```
dbarkowsky, line 15: This incoming building won't have all fields, like id, so we need a partial.
Either that or we define a new interface for what adding a building should take.
```suggestion
export const addBuilding = async (building: DeepPartial<Building>) => {
```

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+    throw new ErrorWithCode(e.message, e.status);
+  }
+};
+
+/**
+ * @description Update a building with matching Id
+ * @param       buildingId  - Number representing building we want to update.
+ * @returns     findBuilding - Building data matching Id passed in.
+ */
+export const updateBuildingById = async (building: Building) => {
```
dbarkowsky, line 45: The incoming Building might not have all fields. DeepPartial needed
```suggestion
export const updateBuildingById = async (building: DeepPartial<Building>) => {
```

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+  const existingBuilding = await getBuildingById(building.Id);
+  if (!existingBuilding) {
+    throw new ErrorWithCode('Building does not exists.', 404);
+  }
+  const retBuilding = await buildingRepo.update(building.Id, building);
+  return retBuilding;
```
dbarkowsky, line 51: Unfortunately, the update doesn't return the building it updated. 
To get around this, the parcelService did something like this at the end:
``` ts
 // update function doesn't return data on the row changed. Have to get the changed row again
  const newParcel = await getParcelById(incomingParcel.Id);
  return newParcel;
```

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -0,0 +1,121 @@
...
+  return buildings;
+};
+
+/**
+ * @description Finds and updates building based on the Building Id
+ * @param incomingBuilding incoming building information to be updated
+ * @returns updated building information and status
+ * @throws Error with code if building is not found or if an unexpected error is hit on update
+ */
+export const updateBuilding = async (incomingBuilding: DeepPartial<Building>) => {
```
dbarkowsky, line 108: This might be a duplicate of the earlier updateBuildingById?

express-api/src/services/buildings/buildingSchema.ts
```diff
@@ -0,0 +1,18 @@
+import z from 'zod';
+
+export const BuildingFilterSchema = z.object({
+  pid: z.number().nonnegative().optional(),
```
GrahamS-Quartech, line 4: These z.number() calls should all be z.coerce.number(). This is because when you go to use this on the query params in the controller, all the numeric values you are trying to parse will be strings (ie. "23" instead of 23).

express-api/src/services/buildings/buildingServices.ts
GrahamS-Quartech: Alternatively, directly return a Promise (ie. `return buildingRepo.save(building);`). The caller should be awaiting the Promise of this async function anyways.

express-api/src/services/buildings/buildingServices.ts
GrahamS-Quartech: Yeah, seconding this. I think retrieving a value by id and getting null back is not an exception. It's better to let the caller decide whether this is an error or not since in some cases it probably is expected.

express-api/tests/testUtils/factories.ts
```diff
@@ -231,3 +232,54 @@ export const produceEmail = (props: Partial<IEmail>): IEmail => {
   };
   return email;
 };
+
+export const produceBuilding = (): Building => {
```
GrahamS-Quartech, line 236: It would be better to have more of these values filled in for future use of this function in other tests, but probably sufficient for now.

dbarkowsky (approved):

GrahamS-Quartech (approved):

PR #2219, opened 2024-02-27
PIMS-1334 Create update parcel service

TaylorFries:
## 🎯 Summary

  
[PIMS-1334: ](https://citz-imb.atlassian.net/browse/PIMS-1334)

Changes:

- addParcel checks to ensure incoming pid is a number 9 digits long exactly 
- functionality of updateParcel 
- testing for all the changes noted above

## 🔰 Checklist


express-api/src/services/parcels/parcelServices.ts
```diff
@@ -73,6 +73,25 @@ export const getParcels = async (filter: ParcelFilter, includeRelations: boolean
...
+  if (findParcel == null) {
+    throw new ErrorWithCode('Parcel not found', 404);
+  }
+  try {
+    const updateParcel = await parcelRepo.update({ Id: findParcel.Id }, incomingParcel);
+    return updateParcel.raw[0];
+  } catch (e) {
+    throw new ErrorWithCode(e.message);
+  }
+};
```
dbarkowsky, line 93: - I think we should use DeepPartial<Parcel> as the incoming type. Otherwise, we have to supply a fully described Parcel object each time, even if we're just updating one field. Realizing now that this same problem exists in the Agency service as well.
- Assuming we use DeepPartial:
  - When I send an object without a PID, it still finds and changes the first parcel in the database it finds. We should check if a PID was included and throw an error if not. 
  - I'm not getting anything back in the return value. updateParcel.raw is just an empty array, even when a record is affected. I see userServices uses generatedMaps, but that seems to be blank as well. Maybe we should be retrieving the record after? That's the only way I see online people have done it for now.

dbarkowsky: There's apparently a way to do it in one request with query builder, but it's very long and I dislike using the query building if not absolutely necessary.
``` ts 
const result = await parcelRepo
      .createQueryBuilder()
      .update(Parcel, incomingParcel)
      .where('pid = :pid', { pid: incomingParcel.PID })
      .returning('*')
      .updateEntity(true)
      .execute();
```

TaylorFries: changed incoming type to DeepPartial, added checks for PID existence and correct format, and added a call to get the updated parcel to return.

GrahamS-Quartech (approved):

PR #2226, opened 2024-03-04
PIMS-1345/1435: Add Property Form

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1345](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1345)
[PIMS-1435](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1435)

Designed a form based off of Aiden's designs. Should be pretty close to the intent with a few intentional changes:
* No map component. We don't have a decision on our general map solution right now.
* I didn't include the section at the bottom with the autocomplete containing buildings. I don't know what this should do with our current data structure as it is. I did leave in a toggle on parcel for ownership though since there is a boolean field for that in the db.
* I included PIN in Building since I'm pretty sure we still want them to be able to input that, it's just optional. If you switch between property types you should see it toggle the required flag on that field in the form.
* I included net book value (which I've been assuming actually means fiscals) in the building version of the form since I'm pretty sure we also want to be able to input for that?

## 🔰 Checklist


react-app/package.json
```diff
@@ -23,6 +23,8 @@
     "@mui/icons-material": "5.15.6",
     "@mui/material": "5.15.6",
     "@mui/x-data-grid": "6.19.2",
+    "@mui/x-date-pickers": "^6.19.5",
+    "dayjs": "^1.11.10",
```
TaylorFries, line 27:
```suggestion
    "@mui/x-date-pickers": "6.19.5",
    "dayjs": "1.11.10",
```
Removing '^'s so that the ticket creation script wont flag false positives.

TaylorFries (approved):
The form looks great! I agree that a lot of the finer things can be tuned once we are more established.

LawrenceLau2020 (approved):
Looks good, and I agree that we can always refine details in the future.

PR #2234, opened 2024-03-07
PIMS-1367/1399/1400 Bug Round Up

dbarkowsky:
## 🎯 Summary

  
[PIMS-1367](https://citz-imb.atlassian.net/browse/PIMS-1367?atlOrigin=eyJpIjoiYzM4OGQzODNhNmQwNDllYjhjM2I1NTIwZmY4ZjFmZDQiLCJwIjoiaiJ9)
[PIMS-1399](https://citz-imb.atlassian.net/browse/PIMS-1399?atlOrigin=eyJpIjoiZmFhMzVlNDUyYmEzNGIzNThjNTQ1NmZkZjVkMWFhMzgiLCJwIjoiaiJ9)
[PIMS-1400](https://citz-imb.atlassian.net/browse/PIMS-1400?atlOrigin=eyJpIjoiYmJjZTMyNjAwMWVkNGM0ODgwNGQwZjYxYjIwMGU5YjUiLCJwIjoiaiJ9)

Includes a handful of bug fixes to recent problems.

## Changes
- The users agencies in the Building and Parcel services now look at the children of the parent agency as well. They were previously only looking at the exact agency the user belonged to. This was a problem because the user could see these properties, but they couldn't get additional information, etc. 
- Hide a toastify popup when Leaflet attempted to change the map position after the map was destroyed. This error message wasn't meant for Leaflet errors and there isn't any negative effect of the error occurring, so if the error caught is from leaflet, it no longer triggers the toastify popup.
- When filtering the map by agencies, only the agencies or child agencies available to the user are presented
- When filtering the map by agencies, selecting the parent agency will also show the properties that belong to the children agencies. 

## Testing
Try the following with you system admin role and when only having the Real Estate Manager role:
(You may need to assign yourself a parent agency depending on what you currently have. I recommend Health, as it has a large number of children agencies with properties.)
- Filter properties on map and inventory page.
- Click on properties that belong to your agency and some that belong to the children agencies.
- Open the map, then before it loads, go to the user page (shouldn't see toastify error).

## 🔰 Checklist

### DIFF
--- a/backend/dal/Services/Concrete/BuildingService.cs
+++ b/backend/dal/Services/Concrete/BuildingService.cs
@@ -60,7 +60,7 @@ namespace Pims.Dal.Services
                 .ThenInclude(a => a.Agency)
                 .ThenInclude(a => a.Children)
                 .SingleOrDefault(u => u.Username == this.User.GetUsername()) ?? throw new KeyNotFoundException();
-            var userAgencies = user.Agencies.Select(a => a.AgencyId).ToList();
+            IEnumerable<int> userAgencies = _service.User.GetAgencies(user.Id);
             var viewSensitive = this.User.HasPermission(Permissions.SensitiveView);
             var isAdmin = this.User.HasPermission(Permissions.AdminProperties);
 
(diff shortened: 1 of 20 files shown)


dbarkowsky:
Ready for review. I just need to add some tests to cover the stuff in the frontend I touched.

TaylorFries (approved):
no longer hit bugs in local with changes applied. looked through file and saw no issues.

PR #2236, opened 2024-03-09
BC Assessment Popup

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-###: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-###)

* Adds a button and dialog to the map parcel popup in similar style to LTSA info for the private BC Assessment API.
* Uses the somewhat naive approach of checking the current browser URL to determine which version of the API to call. Not possible to set this using Openshift.
* Note that calls to the API will bounce unless you are authorized on their end.

## 🔰 Checklist


frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
```diff
@@ -32,6 +35,32 @@ export const parcelLayerPopupConfig = {
...
+  // GEN_NET_IMPROVEMENT_VALUE: {
+  //   label: 'Net Improvement Value:',
+  //   display: (data: any) => (
+  //     <>{data.GEN_NET_IMPROVEMENT_VALUE ? `$${data.GEN_NET_IMPROVEMENT_VALUE}` : `N/A`}</>
+  //   ),
+  // },
+  // FOLIO_ID: {
+  //   label: 'Folio ID:',
+  //   display: (data: any) => <>{data.FOLIO_ID ? `${data.FOLIO_ID}` : `N/A`}</>,
+  // },
```
dbarkowsky, line 63: I'm good with leaving this here. Easy to uncomment if we want it back.

dbarkowsky (approved):
Testing in OpenShift. Working well. 
Ideally we find a workaround for local development at some point that isn't modifying the hosts file.

frontend/src/components/maps/leaflet/LayerPopup/constants/index.tsx
GrahamS-Quartech: Yeah I felt there was a non-zero chance we might move the information around so I figured I would leave it there for now.

PR #2249, opened 2024-03-13
PIMS-1434: Parcel Controller

GrahamS-Quartech:
## 🎯 Summary
[PIMS-1434 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1434)

Implements endpoints for managing parcels. I left some endpoints that I am not sure we need for moving forward with the current MVP goal unimplemented.

I also made some small adjustments to the services behavior:
* Services will try to obtain a unique row by primary key instead of by PID, which I feel is appropriate given PID has no unique constraint. There is still a function to get a single row by PID though.
* Removed some error throwing from the services where it did not seem appropriate to me and instead let the controllers handle it.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/parcels/parcelsController.ts
+++ b/express-api/src/controllers/parcels/parcelsController.ts
@@ -48,13 +48,14 @@ export const updateParcel = async (req: Request, res: Response) => {
    */
   try {
     const parcelId = Number(req.params.parcelId);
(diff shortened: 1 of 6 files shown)


dbarkowsky (approved):
Happy with these additions. 
There are some additional checks that I think we need to perform for authorization purposes, but I've created an additional ticket to add those in later.

PR #2253, opened 2024-03-15
PIMS-1430 User Table Tidy

dbarkowsky:
## 🎯 Summary

  
[PIMS-1430](https://citz-imb.atlassian.net/browse/PIMS-1430?atlOrigin=eyJpIjoiODhkNGRlNzZlMGFiNDhkMGIwYzVlYjBmNzk3NDc4NWQiLCJwIjoiaiJ9)

Provides some updates around the User Table and User Details experience.

## Changes
- Display Name field removed. Replaced with Provider column (IDIR or BCeID)
- Solved issue where rendering would fail if pimsUser.data was undefined
- Clicking Users in the header now returns you to the user table if you were on User Detail
- User Detail has its own URL, so the back button in the browser works and you can save the URL to go straight to that user
- Column sizes adjusted
- Updated quick select options to include additional roles, statuses
- Solved issue where trying to highlight text in rows would navigate you to User Detail upon mouseUp event. Now checks for selection length before deciding to navigate or not.
- Provided option to pass initial state of table through props. We might not want to default sort by CreatedOn for every table.

## Future Work
I wanted to implement something that would save the filter when navigating between the User Table and the User Details pages. I think there's a solution where we could save the filter and quick filter settings to a cookie and check that cookie upon load, but I'm going to start a new ticket for this solution to be generic for all tables. 

## 🔰 Checklist


TaylorFries (approved):
user table updates look good with test data. no issues detected

PR #2255, opened 2024-03-15
PIMS-1440 getSelf Optimization

dbarkowsky:
## 🎯 Summary

  
[PIMS-1440](https://citz-imb.atlassian.net/browse/PIMS-1440?atlOrigin=eyJpIjoiZmNiNjFhMjExZGYyNDg2M2EzZjdjZTA1Mjc2NzcyZDAiLCJwIjoiaiJ9)

The wait times for syncing Keycloak roles and user roles was longer than expected (2.5 seconds per call).

Trying to optimize these syncs and getSelf controller by moving them to post-login and cron jobs.
The result is a much faster getSelf call, but with these tradeoffs:
- If you change possible roles in Keycloak (not through PIMS), it could be up to 10 minutes before it's reflected in PIMS tables. This kind of change would also need database and code changes, so I question whether we need this sync at all, or if it should just be handled alongside a PR.
- If you change a user's role in Keycloak (not through PIMS), the user table won't be updated to reflect that until the post-login hook is called again. No issue with security here, as we shouldn't be checking the user table for role permissions. Check the Keycloak token instead. It just means when looking at a user on the User Table or their User Details, their role might be out of date for a short while.

## Changes
- Moved syncKeycloakUser to afterUserLogin hook in Keycloak middleware. The user's roles should be synced at that time.
- Moved syncKeycloakRoles to a cron job that runs every 10 minutes. 

## 🔰 Checklist


TaylorFries (approved):
username update seems to work as expected. unsure how to test cron other than merging and seeing how it goes.

PR #2262, opened 2024-03-19
PIMS-1469 401 Redirect

dbarkowsky:
## 🎯 Summary

  
[PIMS-1469](https://citz-imb.atlassian.net/browse/PIMS-1469?atlOrigin=eyJpIjoiMDA1YjM5NzYzOWYxNDFhNGEyNjkxNzQxMDM1YzRjNDIiLCJwIjoiaiJ9)

## From the ticket
Right now, when the token expires user can still attempt to make calls. These calls will 401, which causes the frontend components to error out when the data isn’t there. 

Possible solutions: 
- something that tells users their session expired
  - maybe with some advanced warning?
- can we auto refresh the token? (security concern)
- redirect to login page, but remember where you were before so site can return user there

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/properties/propertiesController.ts
+++ b/express-api/src/controllers/properties/propertiesController.ts
@@ -1,5 +1,6 @@
 import { Request, Response } from 'express';
 import { stubResponse } from '@/utilities/stubResponse';
+import propertyServices from '@/services/properties/propertiesServices';
 
 /**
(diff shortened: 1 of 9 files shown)


dbarkowsky:
Expecting a Keycloak update that will make this easier. Closing this for now.

PR #2264, opened 2024-03-19
PIMS-1467: parcel hooks in useParcelApi

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1467: ](https://citz-imb.atlassian.net/browse/PIMS-1467) - Implenenting parcel hooks in useParcelApi

## 🔰 Checklist


GrahamS-Quartech:
I think a function for adding a parcel should also be part of this ticket, because we will need to call that when hitting submit on the add property form.

react-app/src/hooks/api/useParcelsApi.ts
```diff
@@ -1,52 +1,45 @@
...
+  ProjectNumbers: string;
+  Corporation: string;
+  Ownership: 50;
+  IsSensitive: true;
+  UpdatedOn: Date;
+  Evaluations: Evaluation | null;
+  Fiscals: Fiscal | null;
+}
+
+const useParcelsApi = (absoluteFetch: IFetch) => {
```
GrahamS-Quartech, line 35: These functions should probably have the above interface in their return signature.

LawrenceLau2020:
I've made changes to the return signature, let me know if I've missed something, I wonder if we also need a patch for updating partial parcel fields, as well as needing a getParcelByPID?

react-app/src/hooks/api/useParcelsApi.ts
```diff
@@ -1,56 +1,68 @@
...
+  const addParcel = async () => {
+    const { parsedBody } = await absoluteFetch.post('/parcels');
+    return parsedBody as Parcel;
+  };
+  const updateParcel = async () => {
+    const { parsedBody } = await absoluteFetch.put('/parcels');
+    return parsedBody as Parcel;
+  };
```
dbarkowsky, line 44: Should this be ById? The controller is expecting that in the params.
We also need to pass a body.
```suggestion
  const updateParcel = async (id: number, parcel: Partial<Parcel>) => {
    const { parsedBody } = await absoluteFetch.put(`/parcels/${id}`, parcel);
    return parsedBody as Parcel;
  };
```

react-app/src/hooks/api/useParcelsApi.ts
```diff
@@ -1,56 +1,68 @@
...
+  UpdatedOn: Date;
+  Evaluations: Evaluation[] | null;
+  Fiscals: Fiscal[] | null;
+}
+
+const useParcelsApi = (absoluteFetch: IFetch) => {
+  const addParcel = async () => {
+    const { parsedBody } = await absoluteFetch.post('/parcels');
+    return parsedBody as Parcel;
+  };
```
dbarkowsky, line 40: Needs a body passed. This body is what we hope will be added in the API.
```suggestion
  const addParcel = async (parcel: Parcel) => {
    const { parsedBody } = await absoluteFetch.post('/parcels', parcel);
    return parsedBody as Parcel;
  };
```

react-app/src/hooks/api/useParcelsApi.ts
```diff
@@ -1,56 +1,68 @@
...
+  Agency: Agency | null;
+  Address1: string;
+  ProjectNumbers: string;
+  Corporation: string;
+  Ownership: string;
+  IsSensitive: true;
+  UpdatedOn: Date;
+  Evaluations: Evaluation[] | null;
+  Fiscals: Fiscal[] | null;
+}
```
dbarkowsky, line 34: Some of these might not be necessary, like PID and PIN. 
Maybe we should mark some as optional?
e.g.
``` 
PID?: number;
PIN?: number;
```

LawrenceLau2020: ah, good catch.

react-app/src/hooks/api/useParcelsApi.ts
```diff
@@ -29,16 +29,16 @@ export interface Parcel {
   Ownership: string;
   IsSensitive: true;
   UpdatedOn: Date;
-  Evaluations: Evaluation[] | null;
-  Fiscals: Fiscal[] | null;
+  Evaluations?: Evaluation[] | null;
+  Fiscals?: Fiscal[] | null;
 }
```
LawrenceLau2020, line 34: also made Evaluations and Fiscals nullable

GrahamS-Quartech (approved):

PR #2267, opened 2024-03-20
PIMS-1470: Create hooks for building services

Sharala-Perumal:
PIMS-1470: Create hooks for building services                                       

## 🎯 Summary

  
[PIMS-1470 ](https://citz-imb.atlassian.net/browse/PIMS-1470)

## 🔰 Checklist


express-api/src/routes/buildingsRouter.ts
```diff
@@ -22,11 +21,8 @@ router
   .get(catchErrors(getBuilding))
   .put(catchErrors(updateBuilding))
   .delete(catchErrors(deleteBuilding));
-router
-  .route(BUILDINGS_ROUTE)
-  .get(catchErrors(filterBuildingsQueryString))
-  .post(catchErrors(addBuilding));
-router.route(`${BUILDINGS_ROUTE}/filter`).post(catchErrors(filterBuildingsRequestBody));
+router.route(BUILDINGS_ROUTE).get(catchErrors(getBuildings)).post(catchErrors(addBuilding));
+router.route(`${BUILDINGS_ROUTE}/filter`).post(catchErrors(getBuildings));
```
dbarkowsky, line 25: We can get rid of the filter route. That kind of get request is covered in the above route now.

Also, check out the parcelsRouter for an example of this: Because the `/buildings` part is now assumed from express.ts, we can drop the BUILDINGS_ROUTE prefix on all these routes.

react-app/src/hooks/api/useBuildingsApi.ts
```diff
@@ -1,55 +1,104 @@
...
+export interface PropertyType {
+  Name: string;
+}
 
-const useBuildingsApi = () => {
+export type GeoPoint = {
+  x: number;
+  y: number;
+};
+export interface Building {
```
dbarkowsky, line 32: Since the parcels hooks got merged, you have access to many of these types/interfaces already. No reason to duplicate them.
There's also a Property interface that you could extend for the Building instead of having it stand on its own.

react-app/src/hooks/api/useBuildingsApi.ts
```diff
@@ -1,55 +1,104 @@
...
+  Agency?: Agency | null;
+  PID?: number;
+  PIN?: number;
+  IsSensitive: boolean;
+  Evaluations?: Evaluation[] | null;
+  Fiscals?: Fiscal[] | null;
+}
+
+const useBuildingsApi = (absoluteFetch: IFetch) => {
+  const addBuilding = async (building: Building) => {
```
dbarkowsky, line 73: After fixing up the types/interfaces, consider using a more specify type for building. When adding a building, there are a number of properties it won't have yet (like ID), but this suggests it could be provided.

GrahamS-Quartech (approved):

PR #2268, opened 2024-03-20
PIMS-1474: Save Evaluations and Fiscals alongside Parcels

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1474 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1474)

* Fixed various issues with entity definitions for parcels, parcel fiscals, and parcel evaluations.
* Parcel evaluation will use Year (int) instead of Date (timestamp)
* The service for getting a parcel by id now returns its child relations as well.

## 🔰 Checklist


express-api/src/typeorm/Entities/abstractEntities/Property.ts
```diff
@@ -57,7 +57,7 @@ export abstract class Property extends BaseEntity {
   @Column({
     type: 'point',
     transformer: {
-      to: (obj: GeoPoint) => `${obj.x},${obj.y}`, // Going to the database. Format into point.
+      to: (obj: GeoPoint) => `${obj?.x},${obj.y}`, // Going to the database. Format into point.
```
dbarkowsky, line 60: Why the ? and why only on the first one?

express-api/src/services/parcels/parcelServices.ts
```diff
@@ -90,10 +90,8 @@ const updateParcel = async (incomingParcel: DeepPartial<Parcel>) => {
   if (findParcel == null || findParcel.Id !== incomingParcel.Id) {
     throw new ErrorWithCode('Parcel not found', 404);
   }
-  await parcelRepo.update({ Id: findParcel.Id }, incomingParcel);
-  // update function doesn't return data on the row changed. Have to get the changed row again
-  const newParcel = await getParcelById(incomingParcel.Id);
-  return newParcel;
+
+  return parcelRepo.save(incomingParcel);
```
dbarkowsky, line 94:
```suggestion
  return await parcelRepo.save(incomingParcel);
```
This feels safer if we await the promise.

express-api/src/services/parcels/parcelServices.ts
```diff
@@ -113,7 +111,18 @@ const getParcelByPid = async (parcelPid: number) => {
  * @returns     findParcel Parcel data matching ID passed in.
  */
 const getParcelById = async (parcelId: number) => {
-  return parcelRepo.findOne({ where: { Id: parcelId } });
+  return parcelRepo.findOne({
```
dbarkowsky, line 114:
```suggestion
  return await parcelRepo.findOne({
```
Same await concern

dbarkowsky:
I assume we'll need this same setup for the Buildings + their evaluations and fiscals. Same ticket or separate?

express-api/src/services/parcels/parcelServices.ts
GrahamS-Quartech: The caller will need to await the returned Promise of this function either way.

express-api/src/typeorm/Entities/abstractEntities/Property.ts
GrahamS-Quartech: Crtl+Z mistake woops.

GrahamS-Quartech:
> I assume we'll need this same setup for the Buildings + their evaluations and fiscals. Same ticket or separate?
I was gonna say separate, but it is basically the exact same process so... Maybe I'll just work on adding that into this PR too.
> Also, we should probably add a migration for this.
Forgot about this, I will generate them.

GrahamS-Quartech:
I needed to modify the generated migration quite a bit, what TypeORM spat out was kinda weird. Please test running and reverting the migration if you can, though it seems fine through my testing.

dbarkowsky:
> I needed to modify the generated migration quite a bit, what TypeORM spat out was kinda weird. Please test running and reverting the migration if you can, though it seems fine through my testing.

Doesn't seem to run for me at the moment. The composite keys needed to be tweaked for the ParcelEvaluations, ParcelFiscals, BuildingEvaluations, and BuildingFiscals. 
Are you running the migration with data already in the database? I notice it fails for me when it tries to add a constraint because there's apparently more than one record with the same evaluation key id and building id.

dbarkowsky (approved):
Tested with cleaned up data from prod. Seems to be working well in both directions. 
I added some indexes for the 3-way composite primary key.

PR #2270, opened 2024-03-21
PIMS-1483: Lookup and Administraive Areas Hooks

GrahamS-Quartech:
## 🎯 Summary

Added hooks for obtaining various lookup values needed for the add property form. Also implemented them into the form itself.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/property/AddProperty.tsx
+++ b/react-app/src/components/property/AddProperty.tsx
@@ -10,6 +10,9 @@ import dayjs from 'dayjs';
 import BuildingIcon from '@/assets/icons/building.svg';
 import ParcelIcon from '@/assets/icons/parcel.svg';
 import BoxedIconRadio from '../form/BoxedIconRadio';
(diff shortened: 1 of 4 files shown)


dbarkowsky (approved):
Looks good. Hooks appear to be working.

Some thoughts for another ticket:
- Would be nice if date fields had a popup for the picking
- Should format currency fields to comma delimit and prevent more than 2 decimals
- Classifications should be mandatory
  - Same with Administrative Area, Lat/Lng, 
- Postal Code not mandatory
- PID not mandatory for Parcel, because could just have PIN. Needs at least one.
- I don't think we need the section on owning a parcel anymore. If they don't own it, we don't add it. This means it could come out of the entity as well. 

Will save these for next standup.

GrahamS-Quartech:
All valid thoughts, a lot of these things are just assumptions I made from the design or things other people seemed to think were required, but yeah let's just go over these in stand up later and see what people think 👍

PR #2283, opened 2024-03-27
PIMS-1404 Agency Details

dbarkowsky:
## 🎯 Summary

  
[PIMS-1404](https://citz-imb.atlassian.net/browse/PIMS-1404?atlOrigin=eyJpIjoiMzAxZmNmMzFlN2JiNGY0NmI1MDY5YmMxMWQ5ZTVlNjgiLCJwIjoiaiJ9)

## Changes
- Add Agencies Details page, which includes:
  - Editing agency details and notification settings
  - Deleting agency (only if there are no FK constraints. We need a more user-friendly outcome for these)
- useAgencies API hooks

## Testing
- Navigate to the Administration > Agencies page. `/admin/agencies`
- Select an agency and try editing, etc.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/agencies/AgencyDetails.tsx
+++ b/react-app/src/components/agencies/AgencyDetails.tsx
@@ -208,7 +208,14 @@ const AgencyDetail = ({ onClose }: IAgencyDetail) => {
                 name={'ParentId'}
                 label={'Parent Agency'}
                 options={agencyOptions}
-                disableOptionsFunction={(option) => option.value === +id}
+                disableOptionsFunction={
+                  (option) =>
+                    option.value === +id || // Can't assign to self
+                    agencyOptions
+                      .find((parent) => parent.value === +id)
+                      .children.map((child: Agency) => child.Id)
+                      .includes(option.value) // Can't assign to current children
+                }
               />
             </Grid>
           </Grid>
--- a/react-app/src/components/agencies/AgencyDetails.tsx
+++ b/react-app/src/components/agencies/AgencyDetails.tsx
@@ -1,0 +1,232 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import DataCard from '../display/DataCard';
+import { Box, Grid, Typography } from '@mui/material';
+import { dateFormatter, statusChipFormatter } from '@/utils/formatters';
+import DeleteDialog from '../dialog/DeleteDialog';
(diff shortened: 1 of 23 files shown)


react-app/src/components/agencies/AgencyDetails.tsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  const notificationsSettingsData = {
+    SendEmail: data?.SendEmail,
+    To: data?.Email ? data.Email.split(';') : [],
+    CC: data?.CCEmail ? data.CCEmail.split(';') : [],
+  };
+
+  const customFormatter = (key: keyof AgencyStatus, val: any) => {
+    console.log(val);
```
GrahamS-Quartech, line 58: Remove

GrahamS-Quartech:
Couple things I noticed:
* The email input works nicely with the enter key but I was expecting it to do something similar when I pressed comma or semi-colon... Instead, it will just eat the input but do nothing with it when you hit submit. I think forcing the use of enter key could be fine if it's indicated somehow, but I think better would be allowing enter, comma, or semi-colon.
* You can set the agency's parent to itself, which I imagine isn't valid? But doesn't seem to cause an error necessarily.

dbarkowsky:
> Couple things I noticed:
> 
> * The email input works nicely with the enter key but I was expecting it to do something similar when I pressed comma or semi-colon... Instead, it will just eat the input but do nothing with it when you hit submit. I think forcing the use of enter key could be fine if it's indicated somehow, but I think better would be allowing enter, comma, or semi-colon.
> * You can set the agency's parent to itself, which I imagine isn't valid? But doesn't seem to cause an error necessarily.

- Email chips should now delimit via comma or semicolon as well. The secret was not specifying a key for the TextField and then also having the value attached to inputValue on the Autocomplete, not the TextField.
- An agency's parent can no longer be set to itself or one of its current children. Marked those options as disabled in the autocomplete.
- Still need to validate email input, but I'm thinking that is part of the future ticket for zod validation.

GrahamS-Quartech:
>     * Still need to validate email input, but I'm thinking that is part of the future ticket for zod validation.

Yeah this was the only thing I was gonna comment on, otherwise all works nicely now. Approving.

GrahamS-Quartech (approved):

PR #2287, opened 2024-03-28
PIMS-1507: Geocoder Controller

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1507](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1507)

Implemented geocoder controller and geocoder controller tests.

## 🔰 Checklist


dbarkowsky (approved):
Tested. Working well. Only suggestion is to change `pids` to `pid` for the interface and response body. Seems like it only returns a single pid as a string.
Also wondering if it's worth converting that to a number.

GrahamS-Quartech:
> Tested. Working well. Only suggestion is to change `pids` to `pid` for the interface and response body. Seems like it only returns a single pid as a string. Also wondering if it's worth converting that to a number.

According to their official API specifications, that value could be a list as a comma separated string? Though I couldn't seem to get any of these occurrences to happen.

PR #2295, opened 2024-04-02
PIMS-1519: Resolving unused Building columns

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1519](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1519)

This sets BuildingOccupantTypeId to nullable and drops TransferLeaseOnSale as a column.

BuildingOccupantTypeId seems like it may have been set at some point in old PIMS, but as of today there is no way to set it. I don't think we want to be forced to set it in new PIMS but there may be some meaning to the old data, so we can keep that while allowing us to ignore it on new rows by setting it to nullable.

TransferLeaseOnSale is a boolean field that defaults to false and does not appear to ever be set otherwise by the frontend or API. I don't believe this stores any useful information and so I think it should be deleted. If it's too annoying to fix this in the ETL, making this field nullable as well or setting a default on the column in new PIMS (it currently does not even though it exists in MSSQL) are also options.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/administrativeAreas/administrativeAreasController.ts
+++ b/express-api/src/controllers/administrativeAreas/administrativeAreasController.ts
@@ -1,12 +1,7 @@
 import { NextFunction, Request, Response } from 'express';
 import { stubResponse } from '@/utilities/stubResponse';
-import { KeycloakUser } from '@bcgov/citz-imb-kc-express';
-import {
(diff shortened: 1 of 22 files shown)


dbarkowsky (approved):
Tested migrations. Working both directions.
I did remove two more instances of transferLeaseOnSale from other interfaces and schemas.

PR #2296, opened 2024-04-03
Agencies Cannot Be Own Parent Zero ID Bug

GrahamS-Quartech:
Fixed an issue where the controller would incorrectly skip the own parent check if the ID was set to zero. 

In Javascript, `0 == false` so the line `if (updateInfo.ParentId && updateInfo.ParentId === updateInfo.Id)` would evaluate false in this scenario.

## 🔰 Checklist


dbarkowsky (approved):
I was worried about undefined values, but this seems to work with the single =.
`undefined != null` = false
`undefined !== null` = true

PR #2297, opened 2024-04-03
PIMS-1437: Checking for User Agency

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1437: ](https://citz-imb.atlassian.net/browse/PIMS-1437)

Creating functions in the user controller to handle filtering by a user's agencies.

To test, change your user to a non-admin user, and you should only be able to see your own agencies. If your agency is a parent agency, you should see all users in both the "parent" and "children" agencies as well.

## 🔰 Checklist


dbarkowsky (approved):
Like these changes. Was able to only get my agencies' users when not admin.

We should make a ticket for next sprint that applies this same logic to the properties.

PR #2298, opened 2024-04-03
PIMS-1518: Add Administrative Area

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1518 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1518)

Pretty simple add form as there's not many fields for this data type. Should return you to the table view on submission.

Note that because the sequence for the administrative_area table's PK is not advanced during seeding, you may get failed POST requests after a couple adds. To fix this, simply advance the sequence manually. You can do this with `SELECT setval('administrative_area_id_seq', (SELECT MAX(id) FROM administrative_area));`. Another ticket will follow to address this with a migration.

## 🔰 Checklist


LawrenceLau2020 (approved):
Looks good, I was able to add a few new admin areas ("1000 Mile house" and "1001 Mile house", the first one was inserted with id =1, so had to use that setval to enable the correct sequencing.

The only thing I noticed, was that the sorting was a bit strange as I was expecting the new ones to show at the top of the list, but I had to manually sort by Name ascending to get it to display.

Perhaps the default sort order isn't by name ascending?

image

PR #2302, opened 2024-04-05
PIMS-1443 Table Filter Saving

dbarkowsky:
## 🎯 Summary

  
[PIMS-1443](https://citz-imb.atlassian.net/browse/PIMS-1443?atlOrigin=eyJpIjoiYTBlMmNlNjg4MGE5NGFmYmIwZTljMzA4MDEwYWVhYTIiLCJwIjoiaiJ9)

## Changes
- In DataTable, tables check for previous state based on assigned name.
- This state is saved when navigating away from the page and reloaded after render.
- Potentially sets filters, etc. based on query params or stored session cookie
- Query params get priority. If not set, then it will look for a cookie for table state. 

## Testing
Please play around with different filters. 
See if they are remembered and sharable.
Example query:
`http://localhost:3000/users?columnSortName=FirstName&columnSortValue=asc&columnFilterName=Username&quickSelectFilter=Active&columnFilterValue=idir&keywordFilter=transpor`
This should give you the users table sorted by first name, filtered by the quick select for Active users, filtered again by IDIR for Provider, and then filtered again with the keyword search for "transpor".

## 🔰 Checklist


GrahamS-Quartech (approved):
Tested and seems to work well with all combinations. The combination of url params and session storage seems like a pretty robust implementation 👍

PR #2305, opened 2024-04-05
PIMS-1503: Add Project Form

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1503 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1503)

* New form component for adding a project.
* Contains a custom component for searching properties and adding them to a table. Including a property removes it as an option in the autocomplete. Deleting the property re-adds it as an option.

## 🔰 Checklist


react-app/src/components/projects/AddProject.tsx
```diff
@@ -0,0 +1,375 @@
...
+          <Grid item xs={12}>
+            <AutocompleteFormField
+              required
+              name={'TierLevelId'}
+              label={'Assign tier'}
+              options={tierData?.map((tier) => ({ value: tier.Id, label: tier.Name })) ?? []}
+            />
+          </Grid>
+          <Grid item xs={12}>
+            <TextFormField required fullWidth name={'Notes'} label={'Notes'} />
```
dbarkowsky, line 245: I don't think Notes is a required field.
Also making it a bit bigger with multiline would be nice.
```suggestion
            <TextFormField multiline minRows={3} fullWidth name={'Notes'} label={'Notes'} />
```

react-app/src/components/projects/AddProject.tsx
```diff
@@ -0,0 +1,375 @@
...
+          >
+            <Delete />
+          </IconButton>
+        );
+      },
+    },
+  ];
+  return (
+    <Box display={'flex'} flexDirection={'column'} gap={'1rem'}>
+      <Autocomplete
```
dbarkowsky, line 131: Suggest adding a freesolo prop. Otherwise I get console warnings about field contents not being options.

react-app/src/components/projects/AddProject.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+    if (input.PID) {
+      return `${input.Type} - PID: ${input.PID}`;
+    } else if (input.PIN) {
+      return `${input.Type} - PIN: ${input.PIN}`;
+    } else if (input.Address1) {
+      return `${input.Type} - Address: ${input.Address1}`;
+    } else {
+      return `${input.Type} - No identifying info.`;
+    }
```
dbarkowsky, line 60: If we get a result with no PID/PIN/Address, should we really show an option?

dbarkowsky (changes requested):
A few small comments before approval. 

In the future, I would like to pretty-up the look of the autocomplete label, but that can be another ticket. Functionality works well here.

react-app/src/components/projects/AddProject.tsx
GrahamS-Quartech: This is just here to complete the logic block, realistically this should never actually happen.

dbarkowsky (approved):

PR #2306, opened 2024-04-05
PIMS-1531: Property Quicksearch Backend

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1531 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1531)

This adds an endpoint that you can use to search across multiple fields of parcels and buildings and get results that loosely match the provided keyword. You can optionally supply a take param to limit the results (it applies to each repo so the final count is effectively doubled).

I tried to make this work without the queryBuilder, but I couldn't seem to get it working with the integer columns without casting them using TypeORM's `Raw( )`, and the behavior of that function seems to be pretty bugged.

In the future, maybe we could also optionally limit/expand which fields are searched  across but I think this is fine for our current needs.

## 🔰 Checklist


express-api/src/services/properties/propertiesServices.ts
```diff
@@ -0,0 +1,42 @@
+import { AppDataSource } from '@/appDataSource';
+import { Building } from '@/typeorm/Entities/Building';
+import { Parcel } from '@/typeorm/Entities/Parcel';
+
+const propertiesFuzzySearch = async (keyword: string, limit?: number) => {
```
dbarkowsky, line 5: Maybe we should give this a default of undefined. In testing, it required me to fill the second argument with undefined manually.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -0,0 +1,42 @@
...
+    .leftJoinAndSelect('parcel.AdministrativeArea', 'adminArea')
+    .leftJoinAndSelect('parcel.Evaluations', 'evaluations')
+    .leftJoinAndSelect('parcel.Fiscals', 'fiscals')
+    .where(`parcel.pid::text like '%${keyword}%'`)
+    .orWhere(`parcel.pin::text like '%${keyword}%'`)
+    .orWhere(`agency.name like '%${keyword}%'`)
+    .orWhere(`adminArea.name like '%${keyword}%'`)
+    .orWhere(`parcel.address1 like '%${keyword}%'`)
+    .take(limit)
+    .getMany();
```
dbarkowsky, line 18: If you wanted to do this without the query builder, this seems to be working for me:
``` ts
 const parcels = await AppDataSource.getRepository(Parcel).find({
    relations: {
      Agency: true,
      AdministrativeArea: true,
      Evaluations: true,
      Fiscals: true,
    },
    where: [
      {
        Agency: {
          Name: Like(`%${keyword}%`),
        },
      },
      {
        AdministrativeArea: {
          Name: Like(`%${keyword}%`),
        },
      },
      {
        Address1: Like(`%${keyword}%`),
      },
      !isNaN(parseInt(keyword)) ? { PID: parseInt(keyword) } : undefined,
      !isNaN(parseInt(keyword)) ? { PIN: parseInt(keyword) } : undefined,
    ],
  });
```

dbarkowsky (commented):
Would this be easier to do if we made a view? I'm thinking we probably need to change the other materialized view to just a plain view anyway. It doesn't seem to be auto-updating.
A view that just has the fields we need for this fuzzy search could be useful

express-api/src/services/properties/propertiesServices.ts
GrahamS-Quartech: This should result in PID and PIN only working for exact rather than partial matches though right?

GrahamS-Quartech:
> Would this be easier to do if we made a view? I'm thinking we probably need to change the other materialized view to just a plain view anyway. It doesn't seem to be auto-updating. A view that just has the fields we need for this fuzzy search could be useful

Materialized views do not update until instructed to, so we would need a cronjob or some other mechanism to instruct it to. A regular view could be nice for convenience, though it probably wouldn't have any particular performance benefit.

express-api/src/services/properties/propertiesServices.ts
dbarkowsky: Hmm, seems to be. Should PIDs and PINs partially match? 
Also, I think this is case-specific. There's ILike provided by TypeORM as well, which is case-insensitive.

GrahamS-Quartech: I dunno, I assumed they should but I guess you could argue they should be exact. Maybe a Lindsay question.

dbarkowsky (approved):
After some more experimentation, I think what you had is the only way if we need partial matches on PID/PIN. Limitation of TypeORM.

PR #2315, opened 2024-04-09
PIMS-1544: Parcel Layer Integration for Property Management

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1544](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1544)

* This PR adds a small map component that appears in the general information section for buildings and parcels adding/editing.
* You should be able to enter a PID or a PIN and have the map automatically fix itself to the lat/lng at the centroid of the relevant parcel. 
* The map should automatically set itself to the relevant location when you go to edit an existing parcel/building. Note that I was making it default the location to (0, 0) previously so if you see that in some of the parcels you already made locally that is expected.
* When you click on the map, you should be able to see a super basic popup showing the PID/PID at that location. I figure we can flesh this out more later. It only triggers when you're zoomed in enough.

Please test this functionality out locally and let me know if there's anything weird happening.

## 🔰 Checklist


dbarkowsky (commented):
Two strange behaviours:
- If a popup is open, you cannot scroll away from it. As soon as it hits the outside of the map, scrolling is stopped.
- If you blur the PID field while a popup is open, it doesn't seem to centre the pin. It just brings it a little bit closer to that parcel. Maybe we should just close the popup on blur if possible before the map.setView is called.

dbarkowsky:
Pushed two changes:
- Moving @types dependencies to dev-dependencies and removing carets
- Closing the popup on blur events
Tried out closing the popup on drag to avoid the boundary issue, but it doesn't feel natural to use, so left that out.

dbarkowsky (approved):
Tested again after most recent change. Can now pan away from the popup.

PR #2319, opened 2024-04-11
PIMS-1520: Geocoder integration for street address field and map

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1520 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1520)

* The street address field will now call the geocoder API in the background to provide suggestions. 
* Should show Loading while the request is still pending.
* If you select an element from the dropdown, it should automatically set the map using the lat/lng contained in the api response from geocoder.
* If you just use freetext, it should not set the map.
* Note that if you enter a valid PIN or PIN afterwards, it will move the map once more. Though in theory it shouldn't make much difference assuming the address and PID line up.

## 🔰 Checklist


dbarkowsky (approved):
One console log that can be removed. Otherwise geocoder integration is working well.

PR #2320, opened 2024-04-11
PIMS-1545: AgenciesController auth checks

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1545: ](https://citz-imb.atlassian.net/browse/PIMS-1545)

Added a frontend authorization check to the agency management page and changed the route to a protected route.
You can test using a role of "General User" which shouldn't be able to access the Agency admin page: http://localhost:3000/admin/agencies and should be redirected to the homepage.
Only admin and auditor roles should be able to access it.

## 🔰 Checklist

### DIFF
--- a/express-api/src/routes/agenciesRouter.ts
+++ b/express-api/src/routes/agenciesRouter.ts
@@ -15,12 +15,12 @@ const router = express.Router();
 // Endpoints for Admin Agencies
 router
   .route(`/`)
-  .get(getAgencies)
+  .get(protectedRoute(), catchErrors(getAgencies))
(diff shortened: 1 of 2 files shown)


dbarkowsky (approved):
Works well for me. 
In the future, let's see if we could extract this check into a single function. Then it could be used in many places without duplicate code.

GrahamS-Quartech:
This is acceptable if we only need this for agencies pages right now but this is really something that should be implemented in the AuthRouteGuard or some similar wrapper component.

PR #2322, opened 2024-04-11
PIMS-1536: Projects Details Frontend

Sharala-Perumal:
PIMS-1536: Projects Details Frontend                                      

## 🎯 Summary

  
[PIMS-1536](https://citz-imb.atlassian.net/browse/PIMS-1536)

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/lookup/lookupController.ts
+++ b/express-api/src/controllers/lookup/lookupController.ts
@@ -7,10 +7,12 @@ import {
   ClassificationPublicResponseSchema,
   PredominateUsePublicResponseSchema,
   RegionalDistrictPublicResponseSchema,
+  TierLevelPublicResponseSchema,
 } from './lookupSchema';
 import { BuildingPredominateUse } from '@/typeorm/Entities/BuildingPredominateUse';
 import { BuildingConstructionType } from '@/typeorm/Entities/BuildingConstructionType';
 import { RegionalDistrict } from '@/typeorm/Entities/RegionalDistrict';
+import { TierLevel } from '@/typeorm/Entities/TierLevel';
 
 // TODO: What controllers here could just be replaced by existing GET requests?
 
@@ -139,9 +141,14 @@ export const lookupProjectTierLevels = async (req: Request, res: Response) => {
             "bearerAuth": []
       }]
    */
-
-  // TODO: Replace stub response with controller logic
-  return stubResponse(res);
+  const uses = await AppDataSource.getRepository(TierLevel).find();
+  const filtered = uses.filter((u) => !u.IsDisabled);
+  const parsed = TierLevelPublicResponseSchema.array().safeParse(filtered);
+  if (parsed.success) {
+    return res.status(200).send(parsed.data);
+  } else {
+    return res.status(400).send(parsed);
+  }
 };
 
 /**
--- a/express-api/src/controllers/lookup/lookupSchema.ts
+++ b/express-api/src/controllers/lookup/lookupSchema.ts
@@ -24,3 +24,9 @@ export const RegionalDistrictPublicResponseSchema = z.object({
   Id: z.number(),
   Abbreviation: z.string(),
 });
+
+export const TierLevelPublicResponseSchema = z.object({
+  Name: z.string(),
+  Id: z.number(),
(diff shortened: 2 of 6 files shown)


express-api/src/controllers/projects/projectsController.ts
```diff
@@ -22,7 +25,37 @@ export const getDisposalProject = async (req: Request, res: Response) => {
...
+    where: { Id: In(parcelIds ?? []) },
+  });
+  const buildings = await AppDataSource.getRepository(Building).find({
+    relations: {
+      Agency: true,
+      Evaluations: true,
+      Fiscals: true,
+    },
+    where: { Id: In(buildingIds ?? []) },
+  });
```
dbarkowsky, line 57: I think you could also do this (albeit with the parcels and buildings together) by querying ProjectProperties:
``` ts
  const result = await AppDataSource.getRepository(ProjectProperty).find({
    where: { ProjectId: project.Id },
    relations: {
      Parcel: {
        Evaluations: true,
        Fiscals: true,
        Agency: true,
      },
      Building: {
        Evaluations: true,
        Fiscals: true,
        Agency: true,
      },
    },
  });
```

dbarkowsky: Or like this from the project originally if we want to update the getProject service:
``` ts
const result = await AppDataSource.getRepository(Project).find({
    where: { Id: project.Id },
    relations: {
      ProjectProperties: {
        Parcel: {
          Agency: true,
          Evaluations: true,
          Fiscals: true
        },
        Building: {
          Agency: true,
          Evaluations: true,
          Fiscals: true
        }
      }
    },
  });
```

react-app/src/components/projects/ProjectDialog.tsx
```diff
@@ -0,0 +1,215 @@
...
+  });
+  useEffect(() => {
+    //console.log(`useEffect called! ${JSON.stringify(initialValues, null, 2)}`);
+    financialFormMethods.reset({
+      AssessedValue: initialValues?.Assessed,
+      NetBookValue: initialValues?.NetBook,
+      EstimatedMarketValue: initialValues?.Market,
+      AppraisedValue: initialValues?.Appraised,
+      Metadata: initialValues?.Metadata,
+    });
```
dbarkowsky, line 106: None of these values show for me. Just blank fields.

react-app/src/components/projects/ProjectForms.tsx
```diff
@@ -0,0 +1,176 @@
...
+          fullWidth
+          name={'Metadata.programCost'}
+          label={'Estimated program recovery fees'}
+        />
+      </Grid>
+    </Grid>
+  );
+};
+
+export const ProjectDocumentationForm = () => {
```
dbarkowsky, line 143: I'm not sure we should be allowing users to change any of these fields.
They weren't editable on old PIMS once the project was submitted. 
Just by being at this phase of the workflow, they must be complete.

GrahamS-Quartech: Yeah, this is sorta placeholder until we merge in the updated tasks api stuff. Maybe instead we present this form after they try changing the projectstatus, since we do need them to choose whether the tasks are completed at some point, but not necessarily at any time on this page.

express-api/src/controllers/projects/projectsController.ts
GrahamS-Quartech: The second example is similar to what Sharala had previously and it was not working. First example is good though, does simplify the logic a bit. I'll give that a try.

dbarkowsky: > The second example is similar to what Sharala had previously and it was not working. First example is good though, does simplify the logic a bit. I'll give that a try.

Not working how? I seem to be able to get the project object back with all the relations.

GrahamS-Quartech: Updated to that first method, you're right it works fine

dbarkowsky (approved):
Approving with the thought that there will be some bug fixes/optimizations coming at a later date.

Note: we probably need to convert these values so they aren't strings.
image

PR #2323, opened 2024-04-12
PIMS-1541 Project Details Services

dbarkowsky:
## 🎯 Summary

  
[PIMS-1541](https://citz-imb.atlassian.net/browse/PIMS-1541?atlOrigin=eyJpIjoiYTg0MjViNWQ0ZDdjNGUyZWFmYjcxZjRlM2Y0YTlmYjgiLCJwIjoiaiJ9)

## Changes
Added functions for the following:
- updating a project
- getting a single project by ID
- deleting a project
- adding and deleting project/property relations

There are a few sections that will need revision at a later date:
- Checks related to how statuses can move (so statuses aren't moved out of order). Will tackle after we confirm the projects workflow.
- Notifications get deleted from the queue upon project deletion, but when the notifications service is done, that responsibility should be done there so it can also cancel the notifications queued with CHES.
- If we make drafts a possibility one day, then project numbers maybe can't always start with SPP.

## 🔰 Checklist


express-api/src/services/projects/projectsServices.ts
```diff
@@ -123,8 +205,188 @@ const addProjectBuildingRelations = async (project: Project, buildingIds: number
...
+
+/**
+ * Updates a project with the given changes and property IDs.
+ *
+ * @param project - The project object containing the changes to be made.
+ * @param propertyIds - The IDs of the properties to be associated with the project.
+ * @returns The result of the project update.
+ * @throws {ErrorWithCode} If the project name is empty or null, if the project does not exist, if the project number or agency cannot be changed, or if there is an error updating the project.
+ */
+const updateProject = async (project: DeepPartial<Project>, propertyIds: ProjectPropertyIds) => {
```
TaylorFries, line 252: Hit the following error 
```
{"level":"info","message":"query: START TRANSACTION","timestamp":"2024-04-17 01:49:26.892 PM"}
{"level":"info","message":"query: ROLLBACK","timestamp":"2024-04-17 01:49:26.892 PM"}
{"level":"warn","message":"Cannot query across one-to-many for property ProjectProperties","timestamp":"2024-04-17 01:49:26.893 PM"}
{"level":"error","message":"Error updating project.","timestamp":"2024-04-17 01:49:26.894 PM"}
```

when trying to run this query:
```
const sample_project = await projectServices.getProjectById(9);
const result = await projectServices.updateProject(sample_project, {parcels:[7479], buildings:[]});
```

express-api/src/services/projects/projectsServices.ts
```diff
@@ -45,17 +110,26 @@ const addProject = async (project: DeepPartial<Project>, propertyIds: ProjectPro
     const newProject = await projectRepo.save(project);
     // After project is saved, add parcel/building relations
     const { parcels, buildings } = propertyIds;
-    if (propertyIds.parcels) await addProjectParcelRelations(newProject, parcels);
-    if (propertyIds.buildings) await addProjectBuildingRelations(newProject, buildings);
+    if (parcels) await addProjectParcelRelations(newProject, parcels);
+    if (buildings) await addProjectBuildingRelations(newProject, buildings);
```
TaylorFries, line 114: should we add checks to see if a project parcel/building already exists on the table with these values?

dbarkowsky: Added some better checks for the following:

- Does that property already have a relation with a project that isn't cancelled or denied?
- Is there already a relation for this property and project? If so, don't try to add again.

express-api/src/services/projects/projectsServices.ts
dbarkowsky: I think this one is now solved...
Had to change the .update to .save. It seems to expect something different in terms of what updated properties you  can pass.

express-api/src/services/projects/projectsServices.ts
TaylorFries: No longer adds a duplicate entry!
I am noticing that this process will add 2 rows to the project_property table, splitting the parcel and building relations, if this is what we want it to do i am happy with it, but it seems odd to me. 
for example if I run: 
```
const sample_project = await projectServices.getProjectById(6);
const result = await projectServices.addProject(sample_project, {parcels:[6475], buildings:[89]});
```

It results in two entries to the project_property table

express-api/src/services/projects/projectsServices.ts
TaylorFries: no longer hitting this error!

GrahamS-Quartech (approved):
Seems very solid. I don't really understand what the Tasks are meant to do but I see that's noted as TODOs so I figure it's fine to get this merged in and wired up as much as it can on the frontend and we can just sorta move forward from there.

PR #2333, opened 2024-04-17
PIMS-1584: Create reusable backend check for users

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1584: ](https://citz-imb.atlassian.net/browse/PIMS-1584)

Adding a reusable function that can be used in any controller to check to see if the user’s agencies match the agency of the object that the user is trying to access. See the Jira ticket for an example.

## 🔰 Checklist


express-api/src/utilities/authorizationChecks.ts
```diff
@@ -56,3 +56,16 @@ export const isUserActive = async (kcUser: KeycloakUser): Promise<boolean> => {
   const user = await getUser(kcUser.preferred_username);
   return user.Status === 'Active';
 };
+
+export const checkUserPermission = async (
```
dbarkowsky, line 60: Just thinking we should be more specific with this name maybe. It really only checks if the user has permissions to view something. 
If this was any other operation, we wouldn't want to give the auditor role the option to change something.
```suggestion
export const checkReadPermission = async (
```

dbarkowsky (changes requested):

PR #2338, opened 2024-04-22
PIMS-1589: Solve ID Collision Issue

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1589: ](https://citz-imb.atlassian.net/browse/PIMS-1589)

Have added a migration to address the sequencing for the following tables:  
  'agency',
  'administrative_area',
  'building',
  'parcel',
  'project',
  'project_note',
  'notification_queue',
  'notification_template',
  'project_property',
  'project_report'.

Before running the migration the following query: `SELECT nextval('agency_id_seq')` will most likely return a low id number that is not in the correct sequence.

After running the migration, you should be able to run the same command for the above tables with the correct id set.

## 🔰 Checklist

### DIFF
--- a/express-api/src/typeorm/Migrations/1713814960226-UpdateIDSequences.ts
+++ b/express-api/src/typeorm/Migrations/1713814960226-UpdateIDSequences.ts
@@ -1,0 +1,43 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const tables = [
+  'agency',
+  'administrative_area',
+  'building',
+  'parcel',
+  'project',
+  'project_note',
+  'notification_queue',
+  'notification_template',
+  'project_property',
+  'project_report',
+];
+
+export class UpdateIDSequences1713814960226 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    console.log('Executing up method of migration...');
+    for (const table of tables) {
+      console.log(`Updating sequence for table ${table}...`);
+      await queryRunner.query(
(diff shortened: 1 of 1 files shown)


GrahamS-Quartech:
The full list of sequences present in the database is as follows:

-     migrations_id_seq
-     administrative_area_id_seq
-     building_construction_type_id_seq
-     building_occupant_type_id_seq
-     building_predominate_use_id_seq
-     property_classification_id_seq
-     property_type_id_seq
-     building_id_seq
-     evaluation_key_id_seq
-     fiscal_key_id_seq
-     notification_template_id_seq
-     tier_level_id_seq
-     project_risk_id_seq
-     project_id_seq
-     notification_queue_id_seq
-     parcel_id_seq
-     project_note_id_seq
-     project_number_id_seq
-     project_property_id_seq
-     report_type_id_seq
-     project_report_id_seq
-     project_snapshot_id_seq
-     project_status_notification_id_seq
-     project_status_history_id_seq
-     task_id_seq
-     project_type_id_seq
-     agency_id_seq
-     project_num_seq

Something like migrations_id_seq does not need manual correction. Others like fiscal_key_id which we would probably never let the user modify anyways may be fine to leave as is as well. But just wanted to check that there aren't any other sequences from this list that should also be included in this migration. I do think you've covered all the crucial ones.

LawrenceLau2020:
Thanks for that list @GrahamS-Quartech, I've added the project snapshot table to the list of tables to be re-sequenced.

PR #2346, opened 2024-04-24
PIMS-1565: Parcel and Building controllers auth check

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1565: ](https://citz-imb.atlassian.net/browse/PIMS-1565) - Parcel and Building controllers auth check

Enabling filtering on the property inventory page based on the user's agency.
Testing as a general user, should only be able to see the properties that belong to the same agency or is a child agency of the user.

## 🔰 Checklist


express-api/src/services/buildings/buildingSchema.ts
```diff
@@ -3,7 +3,9 @@ import z from 'zod';
 export const BuildingFilterSchema = z.object({
   pid: z.coerce.number().nonnegative().optional(),
   classificationId: z.coerce.number().nonnegative().optional(),
-  agencyId: z.coerce.number().nonnegative().optional(),
+  agencyId:
+    z.coerce.number().nonnegative().optional() ||
+    z.array(z.number().int().nonnegative()).optional(),
```
GrahamS-Quartech, line 8: I had to check to make sure this works because it's not a documented way to do unions, but it does work... Pretty neat. I think that it may only work due to the first schema being `optional()`, which I think means that side of the logical or will evaluate to undefined when it fails? So may behave differently with non-optional schemas, just something to keep in mind.

express-api/src/controllers/projects/projectsController.ts
```diff
@@ -287,7 +259,24 @@ export const searchProjects = async (req: Request, res: Response) => {
...
+  }
+  const filterResult = filter.data;
+
+  let projects;
+  if (isAdmin(kcUser) || isAuditor(kcUser)) {
+    projects = await projectServices.getProjects(filterResult as ProjectFilter, includeRelations);
+  } else {
+    // get array of user's agencies
+    const usersAgencies = await userServices.getAgencies(kcUser.preferred_username);
+    filterResult.agencyId = usersAgencies;
```
GrahamS-Quartech, line 276: Doing this does mean that if the user did specify a certain agency they wanted to filter on in the query params, it would be overridden here instead. I imagine it could still be valid for a non-admin user to filter on a specific agency... Like if they wanted to get projects from just a specific child agency under their parent agency for example.

GrahamS-Quartech (approved):

express-api/src/controllers/projects/projectsController.ts
LawrenceLau2020: That's a good point, If a user tries to filter by an agency that is outside of their own, then yes it overrides it. By default it's always returning the records based on the user's agencies. Right now, a user can still filter the projects by child agency if they belong to a parent agency.

PR #2347, opened 2024-04-24
PIMS-1469 401 Redirect

dbarkowsky:
## 🎯 Summary

  
[PIMS-1469](https://citz-imb.atlassian.net/browse/PIMS-1469?atlOrigin=eyJpIjoiMTMxYWZiMmYyY2I5NDBmMmIzMGJiZTNlYTRiNjc2NGIiLCJwIjoiaiJ9)

## Changes
- Restore option to login with BCeID
- If useFetch function gets a 401, trigger a login call with current location as redirect URL
- Update some fields in the Access Request page to match new SSOUser parameters

## Other
There is one thing I couldn't get to work right:

There's a state where no calls are made because the AuthRouteGuard returns the circular waiting screen. No redirect is called in this case, because the keycloak .isAuthenticated property is always false for a split second, resulting in infinite re-logins if called. 
If a fix for that is released at some point, maybe we can have another login call for that scenario.

## 🔰 Checklist


TaylorFries (approved):
token expiry redirected as expected, no issues detected.

PR #2353, opened 2024-04-29
PIMS-1625: Parcel and Building view details agency authorization

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1625: ](https://citz-imb.atlassian.net/browse/PIMS-1625) - Parcel and Building view details agency authorization

Adding authorization checks to prevent a user from viewing a parcel or a building not belonging to their agency.
To test you can try as a "General user" by editing the url and changing the id of the parcel or building to one that has an agency which doesn't belong to your user's agency.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/buildings/buildingsController.ts
+++ b/express-api/src/controllers/buildings/buildingsController.ts
@@ -60,11 +60,10 @@ export const getBuilding = async (req: Request, res: Response) => {
 
   const kcUser = req.user as unknown as SSOUser;
   const building = await buildingService.getBuildingById(buildingId);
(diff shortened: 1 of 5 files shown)


dbarkowsky:
Right now, it seems like this isn't checking if I'm an admin or auditor. I should be able to see properties based on that.
Isn't this why we made the checkUserAgencyPermission function under /utilities?

LawrenceLau2020:
> checkUserAgencyPermission

Good catch! I'll refactor these 2 gets to use that function.

dbarkowsky (approved):
Seems to be working for me. 
Prevented me from retrieving details of a parcel that wasn't under my agency.

PR #2358, opened 2024-04-30
PIMS-1570 Excel Export Refine

dbarkowsky:
## 🎯 Summary

  
[PIMS-1570](https://citz-imb.atlassian.net/browse/PIMS-1570?atlOrigin=eyJpIjoiZWUzMWE5MTJlMDNjNDY1ZThiZDQwOTI3N2I5MmFhZjEiLCJwIjoiaiJ9)

In response to the request for customized columns, there's now an option to provide your own source of data for the Excel export.
In most cases, this can be taken directly from the table and just remapped. 
Potentially, additional calls to the API could be made if the data isn't already in the frontend.

## Changes
- Optional prop for the FilteredTableDataGrid component.
- Adding customized exports for existing tables that need them.
- Separating service calls for data if extra joins were needed for export (made things slower for initial load)
- Noticed that geocoder address1 function was giving back unordered addresses. Slight fix there. Used [this document](https://www2.gov.bc.ca/assets/gov/data/geographic/location-services/geocoder/batch_address_data_prep.pdf) for reference.
- After merge...
  - including tasks as part of the project object
  - tried to implement this update with cascade, but that just caused issues, reverted

## Known Issue
- Tasks aren't being picked up on the ProjectDetails page correctly. I see them in the response, and they seem to be getting saved, but there's a UI issue.

## 🔰 Checklist


GrahamS-Quartech (approved):
I feel like it could have been better to consolidate these new excel service functions into the existing getter functions, since they mostly perform the same task. Asking the caller to provide specific details about which relations they do/do not want could be cumbersome though, so I think this is ok. Seems like everything is working so approving.

PR #2363, opened 2024-05-02
PIMS: 1656 & 1658 - Create getTitleSummary and createOrder endpoints for LTSA

Sharala-Perumal:
PIMS-1656/1658:  Create getTitleSummary and createOrder endpoints for LTSA
                          

## 🎯 Summary

  
[PIMS-1656](https://citz-imb.atlassian.net/browse/PIMS-1656)
[PIMS-1658](https://citz-imb.atlassian.net/browse/PIMS-1658)

## 🔰 Checklist


dbarkowsky (commented):
Standardized some naming that was causing a build issue.
Added some strong typing to these functions. 
Tested, and it all seems to be working as expected.

Now we just need tests.

express-api/src/controllers/ltsa/ltsaController.ts
```diff
@@ -20,7 +15,10 @@ export const getLTSA = async (req: Request, res: Response) => {
             "bearerAuth": []
       }]
    */
+  const pid = req.query.pid as string;
+  const getLandTitleInfo = await ltsaService.processLTSARequest(pid);
 
-  // TODO: Replace stub response with controller logic
-  return stubResponse(res);
+  if (!getLandTitleInfo) {
```
dbarkowsky, line 21: I think we might never trigger this. The service throws errors if it fails. It doesn't just return a response with no data.

Sharala-Perumal: Implemented suggested change

express-api/src/services/ltsa/ltsaServices.ts
```diff
@@ -0,0 +1,120 @@
...
+  // make a request to get an access token from LTSA
+  const ltsatoken = await getTokenAsync();
+
+  // make a request to get title summary info from LTSA
+  const titleSummaryResponse: ILtsaTitleSummaryResponse = await getTitleSummary(
+    ltsatoken.accessToken,
+    pid,
+  );
+  // Access the first title summary in the list
+  const titleSummary: ILtsaTitleSummaryModel = titleSummaryResponse.titleSummaries[0];
```
GrahamS-Quartech, line 21: Why does getTitleSummary return an array? Is it sufficient to just take the first element whatever that may be?

express-api/src/services/ltsa/interfaces/ILtsaTokens.ts
```diff
@@ -0,0 +1,4 @@
+export interface ILtsaTokens {
```
GrahamS-Quartech, line 1: I feel like an interface this small doesn't really necessitate its own file. You could probably consolidate all these interfaces to one place.

express-api/src/services/ltsa/ltsaServices.ts
Sharala-Perumal: > Why does getTitleSummary return an array? Is it sufficient to just take the first element whatever that may be?

There are often cases where multiple owners are registered on a title . eg) joint tenancy or people who have fractional interest in properties.So the title summary brings in more than one ownership detail. (refer attached image)
multipleowners

I'm taking the first element to extract title number and land district code to be displayed in the parcel info page. Since, title number and land district code are same across all ownership, I'm just getting the values from the first element.

express-api/src/services/ltsa/interfaces/ILtsaTokens.ts
Sharala-Perumal: > I feel like an interface this small doesn't really necessitate its own file. You could probably consolidate all these interfaces to one place.

LTSA process is broken into 3 bits - get Tokens, get Title summary, get order Info. I created one interface per bit for readability purposes. If the standard is to consolidate all into one, I'm happy to do that :)

dbarkowsky (approved):
Works well in testing. Implementation matches that from old PIMS.

PR #2364, opened 2024-05-03
PIMS-1670: Notification Services

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1670](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1670)

* Adds services for building notifications for access requests and project status transitions
* Adds a service for sending a given notification in the queue table, which triggers the CHES service.

In the interest of moving the basic project functionality along, I think we start with just this functionality and add tooling for cancelling / getting notifications later (the CHES service already has a lot of this anyways).

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/notifications/notificationServices.ts
+++ b/express-api/src/services/notifications/notificationServices.ts
@@ -1,0 +1,68 @@
+import { AppDataSource } from '@/appDataSource';
+import { NotificationQueue } from '@/typeorm/Entities/NotificationQueue';
+import { NotificationTemplate } from '@/typeorm/Entities/NotificationTemplate';
+import { randomUUID } from 'crypto';
+import nunjucks from 'nunjucks';
+
+interface BasicNotificationData {
+  Title: string;
+  Uri: string;
+}
+
+interface AccessRequestData {
+  FirstName: string;
+  LastName: string;
+}
+
+enum NotificationStatus {
+  Accepted = 0,
+  Pending = 1,
+  Cancelled = 2,
+  Failed = 3,
+  Completed = 4,
+}
+
+const Title = 'PIMS';
+const Uri = '';
+
+const generateAccessRequestNotification = async (
+  accessRequest: AccessRequestData,
+  templateId: number,
+  to?: string,
+): Promise<NotificationQueue> => {
+  const template = await AppDataSource.getRepository(NotificationTemplate).findOne({
+    where: { Id: templateId },
+  });
+  const body = nunjucks.renderString(template.Body, {
+    Title: Title,
+    Uri: Uri,
+    AccessRequest: accessRequest,
+  });
+  const queueObject = {
+    Key: randomUUID(), // The .NET backend seems to just generate this in the API. Not sure what purpose this serves when there's also a serialized int id.
+    Status: NotificationStatus.Pending,
+    Priority: template.Priority,
+    Encoding: template.Encoding,
+    SendOn: new Date(),
+    Subject: template.Subject,
+    BodyType: template.BodyType,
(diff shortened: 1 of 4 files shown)


express-api/src/services/notifications/notificationServices.ts
```diff
@@ -0,0 +1,206 @@
...
+  });
+  return {
+    ...project,
+    Properties: flattenedProperties,
+  };
+};
+
+const generateAccessRequestNotification = async (
+  accessRequest: AccessRequestData,
+  templateId: number,
```
dbarkowsky, line 66: Does this need a template Id specified? Isn't this always the notification template for access requests?

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -0,0 +1,206 @@
...
+    } else if (template.Audience == NotificationAudience.OwningAgency) {
+      returnNotifications.push(
+        insertProjectNotificationQueue(template, projStatusNotif, project, project.Agency),
+      );
+    } else if (template.Audience == NotificationAudience.Agencies) {
+      //Dunno if we really have/plan to have the ProjectAgencyResponses implemented?
+    } else if (template.Audience == NotificationAudience.ParentAgencies) {
+      //Same here
+    } else if (template.Audience == NotificationAudience.WatchingAgencies) {
+      //Same here
```
dbarkowsky, line 164: We can probably fill these in later, but I think there might be a need to notify agencies on a broad basis like this when properties enter the ERP.

dbarkowsky (commented):
Currently getting a non-null constraint for CreatedById when saving to NotificationQueue. Who should be the user here? Maybe system?

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -0,0 +1,206 @@
...
+        ...projectProperty.Parcel,
+        Type: 'Parcel',
+      };
+    } else {
+      return {
+        ...projectProperty.Building,
+        Type: 'Building',
+      };
+    }
+  });
```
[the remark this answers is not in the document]
GrahamS-Quartech: 🤦 Fixed

express-api/src/services/notifications/notificationServices.ts
GrahamS-Quartech: I think yes because there are two separate access request templates in the db. This is similar to how the old API handles it too.

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -0,0 +1,206 @@
...
+          project.Agency,
+          overrideTo,
+        ),
+      );
+    } else if (template.Audience == NotificationAudience.OwningAgency) {
+      returnNotifications.push(
+        insertProjectNotificationQueue(template, projStatusNotif, project, project.Agency),
+      );
+    } else if (template.Audience == NotificationAudience.Agencies) {
+      //Dunno if we really have/plan to have the ProjectAgencyResponses implemented?
```
LawrenceLau2020, line 160: Currently, Josh, manually adds an agency's interest within an ERP project if an agency expresses interest, which will then trigger 3 emails for the 30/60/90 notification being sent to CHES for this agency which wanted to be "notified".

Since we don't have any UI designs for this, we'll probably need to bring Aiden and Josh into this conversation to see what's needed.

In the C# GenerateNotifciations service, there is a check for the template audience where if it's parent agencies, it will get all the emails for each of these to generate a notification to send.

I believe the majority of this content is still relevant: https://github.com/bcgov/PIMS/wiki/Notifications

LawrenceLau2020 (approved):

PR #2371, opened 2024-05-07
PIMS-1543: Unable to add a net book or assessed values when none exist

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1543: ](https://citz-imb.atlassian.net/browse/PIMS-1543)

Allowing users to add new net book and assessed values as well as editing existing values, sorting by year descending.

Please test adding and editing net book and assessed values to buildings and parcels when none exist, as well as when there are existing financial values.
In the case where a property's net book or assessed values history does not have a value for the current year (2024), there should be a new text box to add a new value.

## 🔰 Checklist


GrahamS-Quartech:
I fixed some issues and optimized the behavior of the submit function for evaluations since buildings were not able to be saved.

There's other things I'm noticing with updating evaluations though. You can modify the year of an evaluation, but there does not appear to be accounted for in the backend logic, so if you change a row from 2021 -> 2023, reloading the page will show you two evaluations with 2021 and 2023 respectively. You can also enter any number of past or future years if you continuously save using the free row that gets added unless you enter the current year.

We need to either decide on a more rigid workflow for how these get added or expand the functionality to have more fleshed out CRUD features on these dialogs.

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -262,44 +263,149 @@ export const PropertyAssessedValueEditDialog = (props: IPropertyAssessedValueEdi
...
 
+  const currentYear = new Date().getFullYear();
+  const years = [currentYear];
+  const defaultParcelValues = years.map((year) => ({
+    Year: year,
+    Value: 0,
+    EffectiveDate: Date(),
+    EvaluationKeyId: 0,
+    ParcelId: initialValues?.Id || null,
+  }));
```
dbarkowsky, line 280: Noticing that this is the only time that I think currentYear or years is used. So years will always be an array of 1. 
It seems like in that case we could just shorten this down to this:
```suggestion
  const defaultParcelValues = [{
    Year: new Date().getFullYear(),
    Value: 0,
    EffectiveDate: Date(),
    EvaluationKeyId: 0,
    ParcelId: initialValues?.Id || null,
  }];
```
Also, what if it's a building instead of a parcel?

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -262,44 +263,149 @@ export const PropertyAssessedValueEditDialog = (props: IPropertyAssessedValueEdi
...
+        RelatedBuildings: initialRelatedBuildings?.map((building) => ({
+          Id: building.Id,
+          Evaluations: building.Evaluations?.map((evalu) => ({
+            ...evalu,
+            Value: String(evalu.Value).replace(/[$,]/g, ''), // Obviously this double map is pretty evil so suggestions welcome.
+          })).sort((a, b) => b.Year - a.Year),
+        })),
+      });
+    }
+  }, [initialValues, initialRelatedBuildings]);
```
dbarkowsky, line 315: A near-identical useEffect exists on lines 256 to 270. Is it possible to consolidate them?

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -262,44 +263,149 @@ export const PropertyAssessedValueEditDialog = (props: IPropertyAssessedValueEdi
...
+            const buildingUpdatePromises = formValues.RelatedBuildings.map(async (building) => {
+              const updatedBuilding: Partial<Building> = {
+                ...building,
+                Evaluations: evaluationMapToRequest(building.Evaluations),
+              };
+              await submitBuilding(building.Id, updatedBuilding); // Update the building
+            });
+            await Promise.all(buildingUpdatePromises);
           }
+          await parcelUpdateResponse;
```
dbarkowsky, line 347: What is this line doing? It looks like the value would have been resolved on line 336.

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -262,44 +263,149 @@ export const PropertyAssessedValueEditDialog = (props: IPropertyAssessedValueEdi
...
+                years={mostRecentYears}
+                title={
+                  propertyType === 'Building' ? 'Assessed Building Value' : 'Assessed Land Value'
+                }
+              />
+              {/* Map through initialRelatedBuildings and render AssessedValue components */}
+              {initialRelatedBuildings?.map((building, idx) => {
+                const buildingEvals = building?.Evaluations?.map((evalu) => evalu.Year) ?? [];
+                if (!buildingEvals.includes(currentYear)) {
+                  // Add currentYear to yearsFromEvaluations array
```
dbarkowsky, line 392: `yearsFromEvaluations` array -> Did the name change at some point?

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -324,18 +430,75 @@ export const PropertyNetBookValueEditDialog = (props: IPropertyNetBookValueEditD
...
+  const currentYear = new Date().getFullYear();
+  const years = [currentYear];
+  const defaultValues = years.map((year) => ({
+    FiscalYear: year,
+    Value: 0,
+    EffectiveDate: undefined,
+    FiscalKeyId: 0,
+    PropertyType: initialValues?.PropertyTypeId || null,
+    Id: initialValues?.Id || null,
+  }));
```
dbarkowsky, line 444: Same sort of issue as above. `years` will only ever be an array of 1, so why not:
```suggestion
  const defaultValues = [{
    FiscalYear: currentYear,
    Value: 0,
    EffectiveDate: undefined,
    FiscalKeyId: 0,
    PropertyType: initialValues?.PropertyTypeId || null,
    Id: initialValues?.Id || null,
  }];
```

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -324,18 +430,75 @@ export const PropertyNetBookValueEditDialog = (props: IPropertyNetBookValueEditD
...
         ...fisc,
         Value: String(fisc.Value).replace(/[$,]/g, ''),
         EffectiveDate: dayjs(fisc.EffectiveDate),
-      })),
-    });
+      }));
+      // Check if currentYear is not in yearsFromEvaluations array
+      if (!fiscalYears.includes(currentYear)) {
+        // Add currentYear to yearsFromEvaluations array
+        fiscalYears.unshift(currentYear);
```
dbarkowsky, line 462: I don't think this is used again later, so is it necessary to add the current year?

react-app/src/components/property/PropertyDialog.tsx
```diff
@@ -324,18 +430,75 @@ export const PropertyNetBookValueEditDialog = (props: IPropertyNetBookValueEditD
...
+    }
   }, [initialValues]);
 
+  const fiscalMapToRequest = (fiscals: Partial<ParcelFiscal>[] | Partial<BuildingFiscal>[]) => {
+    return fiscals
+      .filter((fiscal) => fiscal.Value != null)
+      .map((fiscal) => ({
+        ...fiscal,
+        // BuildingId: (fiscal as BuildingFiscal).BuildingId,
+        // ParcelId: (fiscal as ParcelFiscal).ParcelId,
```
dbarkowsky, line 484: I'm not 100% sure how the building/parcel Id get set if not specified.

react-app/src/components/property/PropertyForms.tsx
```diff
@@ -454,35 +454,92 @@ interface INetBookValue {
...
+    const yearCounts = yearValues.reduce((acc, year) => {
+      acc[year] = (acc[year] || 0) + 1;
+      return acc;
+    }, {});
+
+    for (const year in yearCounts) {
+      if (yearCounts[year] > 1) {
+        return `There is already a net book value for the year ${year}`;
+      }
+    }
```
dbarkowsky, line 479: I can't seem to get this to iterate. I don't think objects are iterable in Javascript. You would need something like Object.entries(yearCounts) to get an iterable array.

react-app/src/components/property/PropertyForms.tsx
```diff
@@ -491,50 +548,105 @@ interface IAssessedValue {
...
+                return result;
+              },
+            }}
+          />
+          <TextFormField
+            InputProps={{
+              startAdornment: <InputAdornment position="start">$</InputAdornment>,
+            }}
+            sx={{ minWidth: 'calc(33.3% - 1rem)' }}
+            name={`${topLevelKey ?? ''}Evaluations.${hasCurrentYear ? 0 : currentYear}.Value`}
```
dbarkowsky, line 613: For these names, are we only ever showing the most recent two? If so, we could probably just use 0 and 1 as the indexes.

react-app/src/hooks/api/useBuildingsApi.ts
```diff
@@ -3,12 +3,13 @@ import { IFetch } from '../useFetch';
...
 import { EvaluationKey } from '@/interfaces/IEvaluationKey';
 import { FiscalKey } from '@/interfaces/IFiscalKey';
+import { DeepPartial } from 'react-hook-form';
 
 export interface BuildingEvaluation extends BaseEntityInterface {
   BuildingId: number;
   Building?: Building;
   Year: number;
-  Value: string;
+  Value: string | number;
```
dbarkowsky, line 12: Just to confirm, it's being received as a string but sent as a number? We should standardize this.

dbarkowsky (changes requested):
There are some breaking changes right now we need to figure out.

express-api/src/services/buildings/buildingServices.ts
```diff
@@ -48,7 +44,15 @@ export const getBuildingById = async (buildingId: number) => {
       Evaluations: true,
       Fiscals: true,
     },
-    where: { Id: buildingId },
+    where: { Id: buildingId, Evaluations: { EvaluationKeyId: 0 } },
```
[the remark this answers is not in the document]
LawrenceLau2020: good catch

dbarkowsky (approved):
Assessment fields appear to be working for both buildings and parcels.
One thing I think we should add for the future:
- If you're adding an assessment that overwrites an existing one (e.g. 2024 when there's already a 2024 entry), it should warn you that the old value will be overwritten.

PR #2372, opened 2024-05-07
PIMS-1661 Property Marker Slideout

dbarkowsky:
## 🎯 Summary

  
[PIMS-1661](https://citz-imb.atlassian.net/browse/PIMS-1661?atlOrigin=eyJpIjoiMjBmYzZjYjA2NDA2NDVlNmE4MzNhNTdiZTFkZTEzZmYiLCJwIjoiaiJ9)
## Changes
- Adds a basic slide out for details on a clicked property.
- Adds relations for Regional District and parent Agency when getting property by id

## Known Issues
- I find the loading time to open the popout too long. I think it's slow for the page to re-render when the selected property changes, but I'm not sure why.
- The clusters re-combine after clicking a marker. Not intended, but hoping this can be resolved with future supercluster experimentation.

## 🔰 Checklist


dbarkowsky:
Just realized that I need some differences for buildings. Back to draft.

LawrenceLau2020 (approved):
Just a slight lag after clicking on the marker to display the popup, but other than that it looks good to me.

react-app/src/components/map/MapPropertyDetails.tsx
```diff
@@ -0,0 +1,278 @@
...
+          <LeftGridColumn>PID</LeftGridColumn>
+          <RightGridColumn>{pidFormatter(propertyData?.PID)}</RightGridColumn>
+        </>
+      ) : (
+        <>
+          <LeftGridColumn>PIN</LeftGridColumn>
+          <RightGridColumn>{propertyData?.PIN}</RightGridColumn>
+        </>
+      )}
+      <LeftGridColumn>Name</LeftGridColumn>
```
GrahamS-Quartech, line 206: It could be worth combining the LeftGridColumn and RightGridColumn into one component since they pretty much always come in pairs as far as I can tell.

You could take it a step further and store the properties you want to display here in an object, then just enumerate the object and create a LeftRightGridColumn component for each entry. Manually specifying each component is fine too though, not like it really hurts performance or anything.

dbarkowsky: Joined them into a single component where possible. 
Also addressed the issue with opening the pages in new tabs instead.
Hoping to see how the supercluster ticket works out, then will try to address they cluster behaviour and lag when components rerender.

PR #2382, opened 2024-05-10
PIMS-1590 Expand Parcel Layer Popup

dbarkowsky:
## 🎯 Summary

  
[PIMS-1590](https://citz-imb.atlassian.net/browse/PIMS-1590?atlOrigin=eyJpIjoiNTMwZWMyNzUwMGQwNGE4ZDg1OTFhOGM2YzgyN2M2ODciLCJwIjoiaiJ9)

## Changes
- Fixed some styling issues with the MapPropertyDetails (MetresSquared fits in grids better)
- Added more info to view on the parcel layer popup
- Changed the state control for the popup so the map doesn't rerender upon click
- Added a PID list to the popup when a parcel has multiple parcels (aka a strata)
- Changed the hooks so they return more than one feature from the parcel layer

## Testing
- Navigate to /dev and click on parcels

I noticed that there isn't often differences in the parcel layer data when many parcels are stacked, like in a strata.
This popup might benefit from some changes in UI if it seems too repetitive. I want to see how the LTSA and BCA data fits in first though.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/map/MapPropertyDetails.tsx
+++ b/react-app/src/components/map/MapPropertyDetails.tsx
@@ -65,6 +65,16 @@ export interface GridColumnPairProps {
   rightValue: any;
   alignment?: string;
 }
+
+/**
+ * Renders a pair of grid columns with a left value and a right value.
+ *
+ * @param {GridColumnPairProps} props - The props for the GridColumnPair component.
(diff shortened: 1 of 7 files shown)


LawrenceLau2020 (approved):
Looks great! I like the way we can see multiple PIDs if the parcel is a strata.

PR #2383, opened 2024-05-13
PIMS-1685: Agency Response Workflow

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1685 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1685)

The goal of this ticket is to integrate the frontend components for agency interests with the backend for updating the project and sending notifications.

* Added submit and retrieve logic between frontend and backend for agency interest responses.
* Fixed several logic issues in the sequence of events for updating the project as well as issues where notifications would be created but not sent.
* Added editable fields to the agency interest table for Received On and Note.
* Fixed minor issues with the autocomplete form field that could raise an error and make it stop working as expected.

**Note:** Currently interests only get set to 1 for "Subscribed". Need Josh to confirm expected behavior from the different interest types.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/notifications/notificationServices.ts
+++ b/express-api/src/services/notifications/notificationServices.ts
@@ -7,7 +7,7 @@ import { ProjectStatusNotification } from '@/typeorm/Entities/ProjectStatusNotif
 import { User } from '@/typeorm/Entities/User';
 import { UUID, randomUUID } from 'crypto';
 import nunjucks from 'nunjucks';
-import { IsNull } from 'typeorm';
+import { In, IsNull } from 'typeorm';
 import chesServices, {
   EmailBody,
   EmailEncoding,
@@ -36,6 +36,12 @@ export enum NotificationAudience {
   ParentAgencies = 'ParentAgencies',
   Default = 'Default',
   WatchingAgencies = 'WatchingAgencies',
+}
+
+export enum AgencyResponseType {
+  Unsubscribe = 0,
+  Subscribe = 1,
+  Watch = 2,
 }
 
 const Title = 'PIMS';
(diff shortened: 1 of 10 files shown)


express-api/src/services/notifications/notificationServices.ts
```diff
@@ -163,11 +169,42 @@ const generateProjectNotifications = async (project: Project, previousStatusId:
...
+        },
+      });
+      agencies.forEach((agc) =>
+        returnNotifications.push(
+          insertProjectNotificationQueue(template, projStatusNotif, project, agc),
+        ),
+      );
     } else if (template.Audience == NotificationAudience.WatchingAgencies) {
-      //Same here
+      // Do we need this distinction?
```
dbarkowsky, line 207: What is a watching agency if not a Subscribed agency?

react-app/src/utilities/helperFunctions.ts
```diff
@@ -14,3 +14,19 @@ export const arrayUniqueBy = <T>(arr: Array<T>, key: (a: T) => string) => {
...
+ *
+ * Given:
+ *  enum Action { Subscribe = 0, Unsubscribe = 1 }
+ *  enumReverseLookup(Actions, 0) -> "Subscribe"
+ *
+ * @param enumObj Typescript enumerated type
+ * @param value number that enum resolves to
+ * @returns {string}
+ */
+export const enumReverseLookup = <T>(enumObj: T, value: number): keyof T | undefined => {
```
dbarkowsky, line 30: Something like this that also did a little bit of formatting would be good for display purposes.

react-app/src/components/projects/AgencyResponseSearchTable.tsx
```diff
@@ -121,6 +137,7 @@ const AgencySearchTable = (props: IAgencySearchTable) => {
             }}
             component="li"
             {...props}
+            key={option.value}
```
dbarkowsky, line 140: I get conflicting ids for keys when opening this autocomplete. Might have to be more specific for the key.

dbarkowsky (commented):
Noticed that if I insert interest without specifying a date, it allows it but defaults to Dec 31, 1969. Should this field be mandatory? Should it just default to the current date?

react-app/src/utilities/helperFunctions.ts
GrahamS-Quartech: What kind of formatting did you have in mind?

express-api/src/services/notifications/notificationServices.ts
GrahamS-Quartech: I don't know. I've been talking to Josh about it a bit and the distinction is pretty slim in the old PIMS implementation. The only real difference between what gets sent between Subscribed and Watch is that Subscribed gets an extra email right as the project enters Approved for ERP, otherwise they both just get sent emails after 30, 60, 90 days. The other weird thing is that with the way everything is set up, it looks like an agency on Subscribed needs to be a parent agency to get any notifications, but if they are on Interested it doesn't matter.

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -169,42 +170,80 @@ const generateProjectNotifications = async (project: Project, previousStatusId:
...
-        },
-      });
+      const agencies = await AppDataSource.getRepository(Agency)
+        .createQueryBuilder('a')
+        .leftJoin(
+          ProjectAgencyResponse,
+          'par',
+          'a.id = par.agency_id AND par.project_id = :projectId',
+          {
+            projectId: 5287,
```
dbarkowsky, line 180: Is this intentionally hard coded for some reason? There's a few in this file.

dbarkowsky (commented):
Seem to be getting this error now.
image

GrahamS-Quartech:
> Seem to be getting this error now. 

Should be fixed I think, lemme know. Changing the empty value to null suppresses this value but I've also had issues with it being null in the past.

dbarkowsky (approved):
Working well after that last change!

PR #2388, opened 2024-05-14
PIMS-1674: Snack Bar

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1674](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1674)

Can't take too much credit for this, as it's largely borrowed from previous work @dbarkowsky did in the link from the Jira ticket. I've modified it lightly to a bit more generic and compatible with the usage in our project. 

I've implemented some basic usage of this context in the user page only. I think that there should be a follow up task to this to improve on the general submission flow, giving us a way to consistently call PUT/POST requests that will expose a loading prop and set the snackbar context depending on the status returned by the response. We can get this snackbar popping up everywhere in that task.

## 🔰 Checklist


dbarkowsky (approved):
Working great. I made one change to centralize the styles in the context provider. Just want to keep it consistent when it gets used around the app.

PR #2400, opened 2024-05-22
PIMS-1703: Soft Delete Properties and Projects

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1703](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1703)

* Added migrations for soft deleting parcels, buildings, projects, and all their related entities.
* These migrations add deleted by id and deleted on columns. TypeORM automatically excludes rows with the delete on column as non-null, this is the soft deleted aspect.
* Soft deletion is incorporated into adding/removing properties/agencies during project updates. Note that because agency responses use composite keys, the previously "deleted" row will simply be restored instead of a new row being created.
* Frontend now correctly calls the delete endpoints using the useDataSubmitter hook. Delete dialog should show loading spinner and cause a snackbar popup on completion. 
* Deletion is done in transaction blocks so that a rollback will happen if a related entity does not delete correctly.
* If you try to delete a property that is linked to a project, it just returns a 403 right now (perhaps there's a better code for this?)

## 🔰 Checklist


dbarkowsky (approved):
Manual tests of deletions seems to be working as expected, removing them from user's view. 
Unfortunate we have to use update functions for this to save the user id, but I think that's the only way unless we write a custom function for the soft delete entity, which might not work because our IDs aren't consistent.

GrahamS-Quartech:
> Manual tests of deletions seems to be working as expected, removing them from user's view. Unfortunate we have to use update functions for this to save the user id, but I think that's the only way unless we write a custom function for the soft delete entity, which might not work because our IDs aren't consistent.

Yeah, softDelete function really seems to do nothing more than set the delete date field with no additional options like setting an additional field. I think using update make the intention less clear than calling softDelete but considering softDelete itself still results in an SQL UPDATE query, I think it's fine enough. Soft delete TypeORM seems like a feature that was added later in development, it's not documented very well. But so far it does work consistently enough so I think we can stick with it.

PR #2402, opened 2024-05-23
PIMS-1597 PID Formatting & Map Adjustments

dbarkowsky:
## 🎯 Summary

  
[PIMS-1567](https://citz-imb.atlassian.net/browse/PIMS-1597?atlOrigin=eyJpIjoiY2QwMWZkMjVkN2YxNDA4OGEzYjM1NzI5MmI4YjI3MzAiLCJwIjoiaiJ9)

## Changes
- Used pidFormatter utility where possible to make sure PID is always presented with hyphens
- Modified the PID text fields to format as values are typed in
  - I felt this was a little hacky, but it seems to work.
- Addressed concerns with map on property pages not zooming/dragging. Users had an issue with accidentally zooming as they scrolled, so I added a prop that focus on that directly. Map on property pages now can zoom and drag, but not from scrolling.
- Added ParcelMap prop to hide the controls. Because they were moved out of the InventoryLayer, they were visible on pages that didn't load properties. This prop hides those controls if wanted.

## Testing
- Find areas where PID is used. Make sure it is formatted correctly.
- Test PID form fields to make sure the function is as desired.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/form/TextFormField.tsx
+++ b/react-app/src/components/form/TextFormField.tsx
@@ -1,12 +1,14 @@
 import React from 'react';
 import { TextField, TextFieldProps } from '@mui/material';
 import { Controller, FieldValues, RegisterOptions, useFormContext } from 'react-hook-form';
+import { pidFormatter } from '@/utilities/formatters';
 
 type TextFormFieldProps = {
   defaultVal?: string;
   name: string;
   label: string;
   numeric?: boolean;
+  isPid?: boolean;
   rules?: Omit<
     RegisterOptions<FieldValues, string>,
     'disabled' | 'valueAsNumber' | 'valueAsDate' | 'setValueAs'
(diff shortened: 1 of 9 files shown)


GrahamS-Quartech (approved):
Only minor issue I've noticed with this is that when making a property, you can have it format the PID as `000-000` if you highlight the entire text and press letters. I'm still approving because this is pretty minor and it does still let you submit as if you entered `000-000-000`.

LawrenceLau2020 (approved):

dbarkowsky:
> Only minor issue I've noticed with this is that when making a property, you can have it format the PID as `000-000` if you highlight the entire text and press letters. I'm still approving because this is pretty minor and it does still let you submit as if you entered `000-000-000`.

Fixed this now. The zeroPadPID function didn't like it if it received a letter in the incoming PID, so I've just had it replace all non-digits before padding.

PR #2406, opened 2024-05-24
 PIMS-1593 NotOwned Column & PIMS-1601 Tenancy Field

dbarkowsky:
## 🎯 Summary

  
[PIMS-1593](https://citz-imb.atlassian.net/browse/PIMS-1593)
[PIMS-1601](https://citz-imb.atlassian.net/browse/PIMS-1601)

## Changes
- Removes the field on Parcels for "Not Owned"
  - This was a relic from when buildings required a parcel, whether gov-owned or not. Now that buildings can exist independently, we can remove this.
- Displayed a number of property values that were on the dialog but not the details page.
- Changed how Tenancy field is handled in the frontend. See notes below for explanation
## Tenancy
Talked with Josh about this field. At the moment, there is a lot of garbage data where users have mixed in text with what should have only been a number value representing a percentage. Unfortunately, most values are this text/number hybrid.

The tenancy field in new PIMS should prevent adding or saving records unless this field is a valid percentage. Users editing old properties will be forced to update this field if it does not conform. 

Hopefully over time we will have fewer entries that are text-based, then we can officially change that column to be a number, not a varchar. If they need functionality that stores multiple rows, each indicating a percentage for a given group (e.g. 50% Group A, 50% Group B) then this field should be replaced entirely with a related table.

## 🔰 Checklist


GrahamS-Quartech (approved):
I think the validation you've added is a good compromise between improving new entries and remaining compatible with existing data 👍

PR #2407, opened 2024-05-28
PIMS-1546 Disable Edit Button on User Detail Page for Non-Admins

TaylorFries:
## 🎯 Summary

  
[PIMS-1546: ](https://citz-imb.atlassian.net/browse/PIMS-1546)

Added line to get users keycloak role, if the user is an admin the edit button is enabled, if the user is not an admin the edit button is disabled. 

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/users/UserDetail.tsx
+++ b/react-app/src/components/users/UserDetail.tsx
@@ -98,11 +98,7 @@ const UserDetail = ({ onClose }: IUserDetail) => {
     mode: 'onBlur',
   });
 
-  let canEdit = false;
-  if (user.keycloak.hasRoles([Roles.ADMIN])) {
-    canEdit = true;
-  }
-
+  const canEdit = user.keycloak.hasRoles([Roles.ADMIN]);
   useEffect(() => {
     refreshData();
   }, [id]);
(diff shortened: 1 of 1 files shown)


react-app/src/components/users/UserDetail.tsx
```diff
@@ -96,6 +98,11 @@ const UserDetail = ({ onClose }: IUserDetail) => {
     mode: 'onBlur',
   });
 
+  let canEdit = false;
+  if (user.keycloak.hasRoles([Roles.ADMIN])) {
+    canEdit = true;
+  }
+
```
dbarkowsky, line 105: Could probably condense this.

```suggestion
  const canEdit = user.keycloak.hasRoles([Roles.ADMIN]);
```

react-app/src/components/users/UserDetail.tsx
```diff
@@ -27,8 +28,9 @@ interface UserProfile extends User {
 
 const UserDetail = ({ onClose }: IUserDetail) => {
   const { id } = useParams();
-  const { pimsUser } = useContext(AuthContext);
+  const user = useContext(AuthContext);
   const api = usePimsApi();
+  const pimsUser = { user };
```
dbarkowsky, line 33: This change is creating an extra layer to this object, which is why you needed the change on line 139. 
useContext will give back something like { pimsUser, keycloakUser }, so the original used destructuring to get pimsUser directly.
Would recommend keeping the original, but just getting keycloakUser through destructuring as well so you can use it for the hasRoles function below.

TaylorFries: added changes in newest commit

dbarkowsky (approved):
Edit buttons are successfully disabled when I'm not an administrator. 👍

PR #2411, opened 2024-05-28
PIMS-1708: Notes Backend

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1708](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1708)

* Notes are now included when getting a project by ID
* There's an additional handler for updating project notes while updating a project. Notes are only upserted and not deleted (I figure notes will just update to an empty string rather than have the row itself deleted)
* Added a migration that makes the index on the column pair (project_id, note_type) unique to make the conflict condition in the upsert possible. 

## 🔰 Checklist


dbarkowsky (approved):
Happy with these changes. Would be nice to fix that issue on the Details page where the Description field reads notes if possible.

PR #2412, opened 2024-05-28
PIMS-1596 Remove Sort Order Option

dbarkowsky:
## 🎯 Summary

  
[PIMS-1596](https://citz-imb.atlassian.net/browse/PIMS-1596?atlOrigin=eyJpIjoiNzQyZmU3NzY0NzY0NGFkZmI5MjQ3ZjAzYzlkYzk1ODMiLCJwIjoiaiJ9)

## Changes
- Removes option for the user to specify the sort order field on admin areas
- Adds default values for any SortOrder and IsDisabled columns so they don't have to be manually specified in the API if missing.

## Testing
- Try editing, adding, viewing admin areas

This change includes a migration. ~The date column was never dropped from when we changed the evaluation entity, so it got included in this migration as well.~ -Removed

## 🔰 Checklist


express-api/src/typeorm/Migrations/1716919102840-DefaultSortOrderIsDisabled.ts
```diff
@@ -0,0 +1,273 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DefaultSortOrderIsDisabled1716919102840 implements MigrationInterface {
+  name = 'DefaultSortOrderIsDisabled1716919102840';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "building_evaluation" DROP COLUMN "Date"`);
```
GrahamS-Quartech, line 7: This causes the migration to fail for me (does not exist). I imagine it's meant to be lower case, but I also think that this column should already be dropped by migration `EvaluationYearIntInsteadofDate1711054015973` (line 15).

express-api/src/typeorm/Migrations/1716919102840-DefaultSortOrderIsDisabled.ts
```diff
@@ -0,0 +1,273 @@
...
+      `CREATE INDEX "IDX_21a97bacb8d918364bfa10fb51" ON "workflow" ("is_disabled", "name", "sort_order") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_f85ac9b25814234deabdb943ea" ON "tier_level" ("is_disabled", "name", "sort_order") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_138e9747c308d68098b566c694" ON "project_risk" ("is_disabled", "code", "name", "sort_order") `,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_a1a5a88a8fb4465f95934e6963" ON "task" ("is_disabled", "is_optional", "name", "sort_order") `,
```
GrahamS-Quartech, line 135: I'm not really sure why TypeORM is creating these composite indexes. These don't seem like very optimal indexes to me.

express-api/src/typeorm/Migrations/1716919102840-DefaultSortOrderIsDisabled.ts
dbarkowsky: Hmm. Perhaps I was missing that migration run somehow. Will remove from this PR and just drop manually for my local.

express-api/src/typeorm/Migrations/1716919102840-DefaultSortOrderIsDisabled.ts
dbarkowsky: I think it's re-creating any indexes that were affected by the column change.

GrahamS-Quartech (approved):
Working well for me now that the migration is fixed

PR #2422, opened 2024-05-29
PIMS-1602 Error Refresh

dbarkowsky:
## 🎯 Summary

  
[PIMS-1602](https://citz-imb.atlassian.net/browse/PIMS-1602)

## Changes
- Moves ConfigContext up one component, now around the error boundary
- Adds some logic to ErrorFallback.tsx that records the count of errors previously and redirects to the root page if it has already happened.

Idea is that the root page will hopefully have no errors, so users won't be stuck in the repeating error loop if they are trying to reload a page that is causing the error in the first place. 

## Testing
I used this replacement for UsersManagement to trigger errors:
<details><summary>UsersManagement</summary>
<p>

```typescript
const UsersManagement = () => {
  const navigate = useNavigate();
  // Getting data from API
  const usersApi = usePimsApi();
  const { data, refreshData, isLoading, error } = useDataLoader(usersApi.users.getAllUsers);

  return (
    <>
      {error.map((e) => (
        <>hi</>
      ))}
      <UsersTable
        rowClickHandler={(params) => {
          // Checking length of selection so it only navigates if user isn't trying to select something
          const selection = window.getSelection().toString();
          if (!selection.length) {
            navigate(`/users/${params.id}`);
          }
        }}
        data={data}
        refreshData={refreshData}
        isLoading={isLoading}
        error={error}
      />
    </>
  );
};
``` 
</p>
</details> 

## 🔰 Checklist


dbarkowsky:
Converted back to draft because I noticed an issue with the reporting.

GrahamS-Quartech (approved):
Working well. I think using session storage is okay for this, can't really leverage any react hooks with the way this works. The only other way I think we could approach this is using the URL params (ie `?errorRedirect=1`) but not sure if there's any particular advantage over this.

PR #2423, opened 2024-05-29
PIMS-1713: Project Note Enumeration

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1713](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1713)

Created migration that adds note_type as its own table. 
Hardcoded inserts from C# API included and tried my best to line up status ids and project note relationships.

## 🔰 Checklist

### DIFF
--- a/express-api/src/typeorm/Entities/NoteType.ts
+++ b/express-api/src/typeorm/Entities/NoteType.ts
@@ -1,0 +1,32 @@
+import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from 'typeorm';
+import { BaseEntity } from './abstractEntities/BaseEntity';
+import { ProjectStatus } from './ProjectStatus';
+
+@Entity()
+export class NoteType extends BaseEntity {
+  @PrimaryGeneratedColumn()
+  Id: number;
+
+  @Column({ type: 'character varying', length: 150 })
+  Name: string;
+
+  @Column('boolean', { default: false })
+  IsDisabled: boolean;
+
+  @Column('int', { default: 0 })
+  SortOrder: number;
+
+  @Column('text', { nullable: true })
+  Description: string;
+
+  @Column('boolean')
+  IsOptional: boolean;
+
+  // Status Relation
+  @Column({ name: 'status_id', type: 'int', nullable: true })
+  StatusId: number;
+
+  @ManyToOne(() => ProjectStatus, (ProjectStatus) => ProjectStatus.Id, { nullable: true })
+  @JoinColumn({ name: 'status_id' })
+  Status: ProjectStatus;
+}
--- a/express-api/src/typeorm/Migrations/1717022772567-NoteTypeTable.ts
+++ b/express-api/src/typeorm/Migrations/1717022772567-NoteTypeTable.ts
@@ -1,0 +1,93 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { User } from '../Entities/User';
+
+export class NoteTypeTable1717022772567 implements MigrationInterface {
+  name = 'NoteTypeTable1717022772567';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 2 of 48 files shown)


express-api/src/typeorm/Entities/NoteType.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Column('boolean')
+  IsOptional: boolean;
+
+  // Status Relation
+  @Column({ name: 'status_id', type: 'int', nullable: true })
+  StatusId: number;
+
+  @ManyToOne(() => ProjectStatus, (ProjectStatus) => ProjectStatus.Id, { nullable: true })
+  @JoinColumn({ name: 'status_id' })
+  Status: ProjectStatus;
```
dbarkowsky, line 31: Should we be more specific here? e.g. StatusId => ProjectStatusId

express-api/src/typeorm/Migrations/1717022772567-NoteTypeTable.ts
```diff
@@ -0,0 +1,93 @@
...
+        (14, 'SalesHistory', 'Sales history notes.', true, '${created_by_id}', 21),
+        (15, 'CloseOut', 'Close out form notes.', true, '${created_by_id}', 21),
+        (16, 'Comments', 'General comments.', true, '${created_by_id}', 21),
+        (17, 'Appraisal', 'Appraisal note.', true, '${created_by_id}', 11),
+        (18, 'Offer', 'A purchaser has made an offer.', true, '${created_by_id}', 32),
+        (19, 'Remediation', 'A disposed project remediation note.', true, '${created_by_id}', 21),
+        (20, 'SplRemoval', 'Rational for being removed from SPL.', true, '${created_by_id}', 22),
+        (21, 'Documentation', 'Notes related to the provided documentation.', true, '${created_by_id}', NULL),
+        (22, 'ErpNotification', 'Notes displayed on ERP notifications.', true, '${created_by_id}', NULL) `,
+    );
```
dbarkowsky, line 44: Putting all thoughts on the status id mapping here:
- **Exemption**: Are these notes for SRES to enter or clients? If clients, I think it would be id 8 "Submitted Exemption". If SRES, maybe 13? "Exemption Review"
- **ContractInPlace**:  Hard to know if this should be applicable to both 42 and 43 if we're using both.
- **Reporting**: I have no idea what this field is for. Is this just presented in reports and nowhere else?
- **CloseOut**: I think this one might come up at multiple points. It's definitely not restricted to just SPL.
- **Offer**: 32 (Disposed) could work as long as they only enter this info at the end of the workflow. I wonder if they ever enter the offer info before it's completed though.

I'm sure there will be some changes once notes start getting put on the page. Might have to fine-tune these as we go.

dbarkowsky (approved):
Willing to approve as is. I don't think we can get the note type mappings right the first try, but the right mapping will reveal itself as we start working with them and as we get feedback from Josh.

express-api/src/typeorm/Entities/NoteType.ts
GrahamS-Quartech: Maybe... I wanted to keep it consistent with Task though.

express-api/src/typeorm/Migrations/1717022772567-NoteTypeTable.ts
GrahamS-Quartech: Dunno if it answers your questions but for more context, I'm pretty sure that Exemption is supposed to be this field:
image
Reporting is this field:
image

As for CloseOut, this field and pretty much all the other 21 status id fields are from that close out form. So you're right, the status isn't necessarily correct, it's kind of a special case. I wasn't sure where to put there.

Now Offer is a bit interesting. I thought it only appeared in the tab that reads Disposed, so I gave it Disposed status. But if you do a global search for "offersNote" in old PIMS you'll notice it actually appears in multiple places... So I guess it's valid under multiple statuses. Tying into your point about "ContractInPlace" too, maybe this means that we need another relational layer to these, a join table that can map note types to multiple statuses?

PR #2424, opened 2024-05-30
PIMS-1605 Custom Sort for User Status

TaylorFries:
## 🎯 Summary

  
[PIMS-1605: ](https://citz-imb.atlassian.net/browse/PIMS-1605)

Wrote custom sort for user status such that: 

- on hold users will be first
- disabled and denied users will be on the bottom. 

If there is a better sorting option please let me know. 

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/users/UsersTable.tsx
+++ b/react-app/src/components/users/UsersTable.tsx
@@ -3,9 +3,12 @@ import { FilterSearchDataGrid } from '@/components/table/DataTable';
 import { Box, SxProps, useTheme, ListSubheader, MenuItem } from '@mui/material';
 import {
   GridColDef,
+  GridComparatorFn,
   GridEventListener,
   gridFilteredSortedRowEntriesSelector,
+  GridRenderCellParams,
   GridRowId,
+  gridStringOrNumberComparator,
   GridValidRowModel,
 } from '@mui/x-data-grid';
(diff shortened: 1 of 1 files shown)


react-app/src/components/users/UsersTable.tsx
```diff
@@ -59,6 +60,14 @@ interface IUsersTable {
   error: unknown;
 }
 
+const statusComparitor: GridComparatorFn = (v1, v2) => {
```
GrahamS-Quartech, line 63: This works but it would be better if it indicated the intention of this sort more clearly at a glance. Adding a comment explaining as much could be sufficient. Though I will suggest one alternative approach:

```
const statusOrder = ['OnHold', 'Active', 'Disabled', 'Denied'];
const indx1 = statusOrder.indexOf(v1);
const indx2 = statusOrder.indexOf(v2);
if (indx1 < 0 || indx2 < 0) return 0;
else return indx2 - indx1;
```
This way you can see that the sort order is clearly demonstrated by the order in which elements appear within the array. The if statement at the bottom is just to guard against any unknown values appearing (indexOf returns -1 on failure).

TaylorFries: I like this a lot. thanks for the help!

GrahamS-Quartech (approved):

PR #2427, opened 2024-05-31
PIMS-1716: Prompt for Project Notes

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1716](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1716)

* Added hook for getting note types from API.
* Note fields now appear on edit dialog during status change much like tasks.
* Documentation history has been updated to show notes and tasks in accordion components. 

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/projects/ProjectDetail.tsx
+++ b/react-app/src/components/projects/ProjectDetail.tsx
@@ -103,6 +103,9 @@ const ProjectDetail = (props: IProjectDetail) => {
     //in the documentation history section.
     const reduceMap = data?.parsedBody.Tasks.reduce(
       (acc: Record<string, IStatusHistoryStruct>, curr) => {
+        if (!curr.IsCompleted) {
+          return acc; //Since this is just for display purposes, no point showing non-completed in results.
+        }
         const fullTask = tasks.find((a) => a.Id === curr.TaskId);
         const fullStatus = statuses.find((a) => a.Id === fullTask.StatusId) ?? {
           Name: 'Uncategorized',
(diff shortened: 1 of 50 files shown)


react-app/src/components/projects/ProjectDetail.tsx
```diff
@@ -77,17 +101,31 @@ const ProjectDetail = (props: IProjectDetail) => {
...
     //get data for the status and task names to be displayed when we enumarete the tasks associated to the project itself
     //in the documentation history section.
-    return data?.parsedBody.Tasks.reduce((acc: Record<string, Array<any>>, curr) => {
-      const fullTask = tasks.find((a) => a.Id === curr.TaskId);
-      const fullStatus = statuses.find((a) => a.Id === fullTask.StatusId);
+    const reduceMap = data?.parsedBody.Tasks.reduce(
+      (acc: Record<string, IStatusHistoryStruct>, curr) => {
+        const fullTask = tasks.find((a) => a.Id === curr.TaskId);
+        const fullStatus = statuses.find((a) => a.Id === fullTask.StatusId);
+        if (!acc[fullStatus.Name]) {
```
dbarkowsky, line 108: If fullStatus doesn't have a value (no status with that id found), then this causes a rendering error. I noticed it on Approved For Exemption status.

react-app/src/components/projects/ProjectDetail.tsx
```diff
@@ -77,17 +101,31 @@ const ProjectDetail = (props: IProjectDetail) => {
...
+          acc[fullStatus.Name].Tasks.push({ ...curr, Name: fullTask.Name });
+          return acc;
+        }
+      },
+      {},
+    );
+    return data?.parsedBody.Notes.reduce((acc: Record<string, IStatusHistoryStruct>, curr) => {
+      const fullNote = noteTypes.find((a) => a.Id === curr.NoteTypeId);
+      const fullStatus = statuses.find((a) => a.Id === fullNote.StatusId);
       if (!acc[fullStatus.Name]) {
```
dbarkowsky, line 121: Same issue here.

dbarkowsky (commented):
Two things I noticed:
- The note titles are visible even if there aren't any notes there. Seems like it could be cleaned up by just hiding them.

image
- The note fields in the dialog don't show up for me if the project is already in that status, even if there are notes populated already.

image
image

react-app/src/components/projects/ProjectDetail.tsx
GrahamS-Quartech: Tasks should always have a status, unlike notes where I've left some null, so this could indicate a different issue. But I gave them both a fallback value regardless.

GrahamS-Quartech:
> Two things I noticed:
> 
>     * The note titles are visible even if there aren't any notes there. Seems like it could be cleaned up by just hiding them.
Fixed this.
>     * The note fields in the dialog don't show up for me if the project is already in that status, even if there are notes populated already.
> 
This is intentional, the tasks works the same way in that they only show up if the status is actually being changed. But we could make it so you just always see the fields associated with the status currently in the dropdown if you think that makes more sense.

dbarkowsky:
> This is intentional, the tasks works the same way in that they only show up if the status is actually being changed. But we could make it so you just always see the fields associated with the status currently in the dropdown if you think that makes more sense.

I guess this depends on whether all note fields are supposed to be filled at the moment it changes statuses. Could be that they need to add notes at a later date.

LawrenceLau2020:
I'm not sure if we may want to hide the tasks related to "exemptions" if a project's status is not "Approved for Exemption" otherwise it seems like they need to check this off, below is a screenshot of http://localhost:3000/projects/5286 and the status is In ERP.

image

dbarkowsky (approved):
I like where this is at. Makes it more clear when I can just see the notes/tasks from the update dialog at all times.

PR #2430, opened 2024-06-03
PIMS-1719: Send notifications for access request

GrahamS-Quartech:
Fixed some bugs in the access request notif generation service. Added…function calls for submitting notifications when a user submits an access request. Fixed a minor frontend issue causing the access request page to crash.

  

## 🎯 Summary

  
[PIMS-1719](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1719)

* Fixed some bugs in the access request notif generation service. 
* Added function calls for submitting notifications when a user submits an access request. 
* Fixed a minor frontend issue causing the access request page to crash.

## Needs discussion
* How to handle administrators under new role format? Previously notifications would be sent to "system administrators" plus relevant users that were both administrators and part of the same agency. Now everyone is just an administrator without differentiating systems level.
* The to column of notification_queue has max length of 500. Most agencies with any admin users have enough of them to quickly overflow this value.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/users/usersController.ts
+++ b/express-api/src/controllers/users/usersController.ts
@@ -6,6 +6,9 @@ import { stubResponse } from '@/utilities/stubResponse';
 import { UserFiltering, UserFilteringSchema } from '@/controllers/users/usersSchema';
 import { z } from 'zod';
 import { isAdmin, isAuditor } from '@/utilities/authorizationChecks';
+import notificationServices from '@/services/notifications/notificationServices';
+import getConfig from '@/constants/config';
+import logger from '@/utilities/winstonLogger';
 /**
  * @description Function to filter users based on agencies
  * @param {Request}     req Incoming request.
@@ -110,6 +113,35 @@ export const submitUserAccessRequest = async (req: Request, res: Response) => {
     req.body.Position,
     req.body.Note,
   );
+  const config = getConfig();
+  const user = await userServices.getUser(req.user.preferred_username);
+  const adminUsers = await userServices.getUsers({
+    role: 'Administrator',
+    agencyId: user.AgencyId,
+  });
+
+  try {
+    const notif = await notificationServices.generateAccessRequestNotification(
+      {
+        FirstName: user.FirstName,
+        LastName: user.LastName,
+      },
+      config.accessRequest.notificationTemplate,
+      adminUsers.map((user) => user.Email).join(';'),
+    );
+    const notifRPD = await notificationServices.generateAccessRequestNotification(
+      {
+        FirstName: user.FirstName,
+        LastName: user.LastName,
+      },
+      config.accessRequest.notificationTemplateRPD,
+    );
+    await notificationServices.sendNotification(notif, req.user);
+    await notificationServices.sendNotification(notifRPD, req.user);
+  } catch (e) {
+    logger.error(`Failed to deliver access request notification: ${e.message}`);
+  }
+
   return res.status(200).send(result);
 };
 
--- a/express-api/src/controllers/users/usersController.ts
+++ b/express-api/src/controllers/users/usersController.ts
@@ -119,7 +119,10 @@ export const submitUserAccessRequest = async (req: Request, res: Response) => {
     role: 'Administrator',
     agencyId: user.AgencyId,
   });
-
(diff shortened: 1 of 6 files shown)


express-api/src/controllers/users/usersController.ts
```diff
@@ -110,6 +113,38 @@ export const submitUserAccessRequest = async (req: Request, res: Response) => {
...
+  const config = getConfig();
+  const user = await userServices.getUser(req.user.preferred_username);
+  const adminUsers = await userServices.getUsers({
+    role: 'Administrator',
+    agencyId: user.AgencyId,
+  });
+  const adminEmails = adminUsers.map((user) => user.Email);
+  while (adminEmails.join(';').length > 500) {
+    adminEmails.shift();
+  }
```
dbarkowsky, line 125: Is this how the old PIMS was deciding who to send these requests to? I thought it went to a shared inbox for Lindsay/Josh.

dbarkowsky (commented):
I think we need to check with Todd/Josh about who should be getting these access request emails. I definitely didn't get any from old PIMS, and I've been listed as an admin there.

LawrenceLau2020 (commented):
I'm just comparing this to the "current" PIMS, and after submitting a new project, it should send a "New Disposal Project Submitted" template notification (template id ==1).

LawrenceLau2020:
I'm now testing the access request page, and it looks like generateAccessRequestNotification isn't being used and no notification is being sent for access requests. Will look deeper into this.

LawrenceLau2020:
I've pushed some changes, right now it's sending 2 access requests notifications for some reason.

express-api/src/services/ches/chesServices.ts
```diff
@@ -136,19 +139,14 @@ const sendEmailAsync = async (email: IEmail, user: SSOUser): Promise<IEmailSentR
   email.from = email.from ?? cfg.ches.defaultFrom;
 
   if (cfg.ches.bccCurrentUser) {
-    email.bcc = [user.email, ...(email.bcc ?? [])];
+    email.bcc = [user?.email, ...(email.bcc ?? [])];
```
GrahamS-Quartech, line 142: By making user an optional argument to this function and accessing this field with the null conditional operator, we could run into cases where this introduces 'undefined' to the BCC list. Getting rid of this option altogether could be another option though, dunno if this actually gets used in live.

express-api/src/services/users/usersServices.ts
```diff
@@ -150,6 +151,19 @@ const addKeycloakUserOnHold = async (
     Note: note,
     CreatedById: systemUser.Id,
   });
+  const newAccessRequest: AccessRequestData = {
```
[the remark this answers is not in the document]
LawrenceLau2020: Good catch @GrahamS-Quartech , when I merged the branch, I guess I missed this because I had made the changes on a totally different branch and didn't even see your changes. I'll revert this back.

express-api/src/services/ches/chesServices.ts
LawrenceLau2020: Good point, I can remove it.

LawrenceLau2020: I've reverted some changes, and the access request notification seems to be sending successfully. Might need someone else to test as well.

express-api/src/services/ches/chesServices.ts
```diff
@@ -158,7 +153,12 @@ const sendEmailAsync = async (email: IEmail, user: SSOUser): Promise<IEmailSentR
       email.delayTS += Number(cfg.ches.secondsToDelay);
     }
   }
-  email.to = email.to.filter((a) => !!a);
+  email.to = cfg.ches.overrideTo
+    ? cfg.ches.overrideTo
+        .split(';')
+        .map((email) => email.trim())
+        .filter((email) => !!email)
+    : [user.email];
```
dbarkowsky, line 161: I think this last part is supposed to be email.to.filter((a) => !!a) still. Otherwise we never send things to the to address.

dbarkowsky (approved):
Tested with override on. Received email.
Some tasks for the future:
- How do we update the status of the notifications when looking at projects in the future?
- The email template could use some formatting. Hyperlink doesn't appear to work.

PR #2437, opened 2024-06-06
PIMS-1746: Entities for project timestamps and project monetary values

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1746](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1746)

* Added new entities that will allow us to separate out monetary and timestamp style fields from the metadata column into their own tables and further associate each field with a different status.
* I realize there's a bit of inconsistency between the format of names between tasks and the other tables. Not sure if we should strive to normalize this? I basically just tried to maintain consistency with the old field names as much as possible to hopefully make ETL easier.
* Once again tried to map statuses as best I could but not 100% sure on all of them.

## 🔰 Checklist

### DIFF
--- a/express-api/src/typeorm/Entities/ProjectTimestamp.ts
+++ b/express-api/src/typeorm/Entities/ProjectTimestamp.ts
@@ -23,6 +23,6 @@ export class ProjectTimestamp extends SoftDeleteEntity {
   @JoinColumn({ name: 'timestamp_type_id' })
   TimestampType: TimestampType;
 
-  @CreateDateColumn()
-  Value: Date;
+  @CreateDateColumn({ name: 'date' })
+  Date: Date;
 }
--- a/express-api/src/typeorm/Entities/ProjectTimestamp.ts
+++ b/express-api/src/typeorm/Entities/ProjectTimestamp.ts
@@ -1,0 +1,28 @@
+import { Entity, CreateDateColumn, Index, ManyToOne, PrimaryColumn, JoinColumn } from 'typeorm';
+import { Project } from './Project';
+import { SoftDeleteEntity } from './abstractEntities/SoftDeleteEntity';
+import { TimestampType } from './TimestampType';
+
+@Entity()
+@Index(['ProjectId', 'TimestampTypeId'])
+export class ProjectTimestamp extends SoftDeleteEntity {
+  // Project Relation
+  @PrimaryColumn({ name: 'project_id', type: 'int' })
+  ProjectId: number;
+
+  @ManyToOne(() => Project, (Project) => Project.Id)
+  @JoinColumn({ name: 'project_id' })
+  Project: Project;
+
+  // Task Relation
+  @PrimaryColumn({ name: 'timestamp_type_id', type: 'int' })
+  @Index()
+  TimestampTypeId: number;
+
+  @ManyToOne(() => TimestampType, (Timestamp) => Timestamp.Id)
+  @JoinColumn({ name: 'timestamp_type_id' })
+  TimestampType: TimestampType;
+
+  @CreateDateColumn()
+  Value: Date;
+}
--- a/express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
+++ b/express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
@@ -1,0 +1,165 @@
+import { In, MigrationInterface, QueryRunner } from 'typeorm';
+import { TimestampType } from '../Entities/TimestampType';
+import { ProjectStatus } from '../Entities/ProjectStatus';
+import { NoteType } from '../Entities/NoteType';
+import { MonetaryType } from '../Entities/MonetaryType';
+import { Task } from '../Entities/Task';
+import { User } from '../Entities/User';
+
+export class MoveMetadataToTables1717629146741 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const systemUser = await queryRunner.manager.findOne(User, { where: { Username: 'system' } });
+    const systemId = systemUser.Id;
+    const result = await queryRunner.manager.save(ProjectStatus, {
+      Id: 44, //For some reason there is no sequence on project_status
+      Name: 'Close Out',
+      Code: 'CLS-OUT',
+      Description: 'Status related to filling out the close out form.',
+      IsMilestone: true,
+      IsTerminal: true,
+      Route: '/place/holder',
+      CreatedById: systemId,
+    });
+    await queryRunner.manager.insert(TimestampType, [
+      { Name: 'AdjustedOn', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectCloseOut
+      { Name: 'ApprovedForSplOn', IsOptional: true, StatusId: 21, CreatedById: systemId }, //ProjectSPLApproval, ProjectERPComplete
+      {
+        Name: 'ClearanceNotificationSentOn',
+        IsOptional: true,
+        StatusId: 14,
+        CreatedById: systemId,
+      }, //ProjectERPComplete
+      { Name: 'DisposedOn', IsOptional: true, StatusId: 32, CreatedById: systemId }, //ProjectERPDisposed, ProjectCloseOut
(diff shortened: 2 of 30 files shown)


express-api/src/typeorm/Entities/ProjectTimestamp.ts
```diff
@@ -0,0 +1,28 @@
...
+  @PrimaryColumn({ name: 'timestamp_type_id', type: 'int' })
+  @Index()
+  TimestampTypeId: number;
+
+  @ManyToOne(() => TimestampType, (Timestamp) => Timestamp.Id)
+  @JoinColumn({ name: 'timestamp_type_id' })
+  TimestampType: TimestampType;
+
+  @CreateDateColumn()
+  Value: Date;
```
dbarkowsky, line 27: Do we want this to be Value or does it make more sense as Date?

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
```diff
@@ -0,0 +1,165 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const systemUser = await queryRunner.manager.findOne(User, { where: { Username: 'system' } });
+    const systemId = systemUser.Id;
+    const result = await queryRunner.manager.save(ProjectStatus, {
+      Id: 44, //For some reason there is no sequence on project_status
+      Name: 'Close Out',
+      Code: 'CLS-OUT',
+      Description: 'Status related to filling out the close out form.',
+      IsMilestone: true,
+      IsTerminal: true,
```
dbarkowsky, line 19: Is it terminal? I'm not sure we're even using this column's values at the moment. If close out has to be done before the Disposal status, this could be an issue.

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
```diff
@@ -0,0 +1,165 @@
...
+      Route: '/place/holder',
+      CreatedById: systemId,
+    });
+    await queryRunner.manager.insert(TimestampType, [
+      { Name: 'AdjustedOn', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectCloseOut
+      { Name: 'ApprovedForSplOn', IsOptional: true, StatusId: 21, CreatedById: systemId }, //ProjectSPLApproval, ProjectERPComplete
+      {
+        Name: 'ClearanceNotificationSentOn',
+        IsOptional: true,
+        StatusId: 14,
```
dbarkowsky, line 29: This happens when it leaves ERP? Not sure it belongs with Approved for ERP status then. Almost like we'd need to set it when entering the next status.

dbarkowsky: I guess this works if the frontend allows you to enter a value, change the status, and still save the value you previously entered. There's a few like this.

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
```diff
@@ -0,0 +1,165 @@
...
+      }, //ProjectERPComplete
+      { Name: 'DisposedOn', IsOptional: true, StatusId: 32, CreatedById: systemId }, //ProjectERPDisposed, ProjectCloseOut
+      { Name: 'ExemptionApprovedOn', IsOptional: true, StatusId: 15, CreatedById: systemId }, //ProjectERPExemption
+      { Name: 'FinalFormSignedOn', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectCloseOut
+      { Name: 'InterestedReceivedOn', IsOptional: true, StatusId: 14, CreatedById: systemId }, //ProjectERPComplete
+      { Name: 'OfferAcceptedOn', IsOptional: true, StatusId: 32, CreatedById: systemId }, //ProjectERPDisposed, ProjectNotSPL, ProjectSPLContractInPlace
+      { Name: 'RemovalFromSplApprovedOn', IsOptional: true, StatusId: 21, CreatedById: systemId }, //ProjectSPLApproval
+      { Name: 'RemovalFromSplRequestOn', IsOptional: true, StatusId: 21, CreatedById: systemId }, //ProjectSPLApproval
+      { Name: 'RequestForSplReceivedOn', IsOptional: true, StatusId: 14, CreatedById: systemId }, //ProjectERPComplete, ProjectSPLApproval
+      { Name: 'TransferredWithinGreOn', IsOptional: true, StatusId: 14, CreatedById: systemId }, //ProjectERPComplete, ProjectNotSPL
```
dbarkowsky, line 40: Why not status 20? "Transferred within the GRE"?

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
```diff
@@ -0,0 +1,165 @@
...
+        CreatedById: systemId,
+      },
+      { Name: 'OfferAmount', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectERPDisposed, ProjectNotSPL, ProjectSPLContractInPlace
+      { Name: 'ProgramCost', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectSummary, ProjectPropertyInformation, ProjectCloseOut
+      { Name: 'RealtorCommision', IsOptional: true, StatusId: result.Id, CreatedById: systemId },
+      { Name: 'SalesCost', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectSummary, ProjectPropertyInformation, ProjectCloseOut
+    ]);
+    await queryRunner.query(`SELECT setval('task_id_seq', (SELECT MAX(id) FROM task), true) `);
+    await queryRunner.manager.insert(Task, [
+      { Name: 'Exemption requested', IsOptional: true, StatusId: 4, CreatedById: systemId }, //Seems to be associated with approving the project, so probably place in the Add form in new design?
```
dbarkowsky, line 74: Yes, this used to be entered during the initial project submission in old PIMS.

dbarkowsky (commented):
Some things to clarify before we proceed.

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
```diff
@@ -0,0 +1,165 @@
...
+      Description: 'Status related to filling out the close out form.',
+      IsMilestone: true,
+      IsTerminal: true,
+      Route: '/place/holder',
+      CreatedById: systemId,
+    });
+    await queryRunner.manager.insert(TimestampType, [
+      { Name: 'AdjustedOn', IsOptional: true, StatusId: result.Id, CreatedById: systemId }, //ProjectCloseOut
+      { Name: 'ApprovedForSplOn', IsOptional: true, StatusId: 21, CreatedById: systemId }, //ProjectSPLApproval, ProjectERPComplete
+      {
```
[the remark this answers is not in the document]
dbarkowsky: That was what we were talking about today in standup. Those values can be retrieved from CHES, so we probably don't need to store them.

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
GrahamS-Quartech: Yeah this could make the most sense... Another one that is referenced more than once. But I think it's only showed conditionally in those other places.

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
GrahamS-Quartech: I'm not sure what the next closest status would be. Maybe "In ERP"?

express-api/src/typeorm/Migrations/1717629146741-MoveMetadataToTables.ts
GrahamS-Quartech: I'll change this to be false for now, but yeah we aren't doing anything with this value on the new frontend at the moment.

express-api/src/typeorm/Entities/ProjectTimestamp.ts
GrahamS-Quartech: I'll make it Date.

dbarkowsky (approved):
Changes make sense. We can fine tune the statuses as needed.

PR #2442, opened 2024-06-10
PIMS-1749: Frontend timestamps and monetary support

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1749](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1749)

* Added hooks for enumerating new field types.
* Edit dialog will now display fields dependent on status for monetary and timestamp types.
* Financial dialog will insert into a Monetary array instead of a Metadata object on submission
* Submission of new field types should work for both updating and adding projects

~~Still needs PIMS-1747 merged in for further integration.~~

## 🔰 Checklist


react-app/src/components/projects/ProjectDetail.tsx
```diff
@@ -116,31 +124,56 @@ const ProjectDetail = (props: IProjectDetail) => {
...
         if (!acc[fullStatus.Name]) {
-          acc[fullStatus.Name] = { Notes: [{ ...curr, Name: fullNote.Description }], Tasks: [] };
-          return acc;
-        } else {
-          acc[fullStatus.Name].Notes.push({ ...curr, Name: fullNote.Description });
-          return acc;
+          acc[fullStatus.Name] = { Tasks: [], Notes: [], Timestamps: [], Monetaries: [] };
         }
+        acc[fullStatus.Name].Notes.push({ ...curr, Name: fullNote.Description });
+        return acc;
```
dbarkowsky, line 144: Is the acc here actually affecting the original reduceMap object? Tried finding documentation to see whether acc is a reference or a copy, but it's not clear.

react-app/src/components/projects/ProjectDetail.tsx
```diff
@@ -246,7 +288,12 @@ const ProjectDetail = (props: IProjectDetail) => {
         <DataCard
           loading={isLoading}
           customFormatter={customFormatter}
-          values={FinancialInformationData}
+          values={Object.fromEntries(
+            Object.entries(FinancialInformationData).map(([k, v]) => [
+              k,
+              formatMoney(v != null ? Number(String(v).replace(/[$,]/g, '')) : 0), //This cast spaghetti sucks but hard to avoid when receiving money as a string from the API.
```
dbarkowsky, line 294: How difficult would it be to slap a transform on these columns?

react-app/src/components/projects/ProjectDialog.tsx
```diff
@@ -167,46 +263,76 @@ interface IProjectFinancialDialog {
...
+  );
   useEffect(() => {
     //console.log(`useEffect called! ${JSON.stringify(initialValues, null, 2)}`);
     financialFormMethods.reset({
-      Assessed: +initialValues?.Assessed?.toString().replace(/[$,]/g, ''),
-      NetBook: +initialValues?.NetBook?.toString().replace(/[$,]/g, ''),
-      Market: +initialValues?.Market?.toString().replace(/[$,]/g, ''),
-      Appraised: +initialValues?.Appraised?.toString().replace(/[$,]/g, ''),
-      Metadata: initialValues?.Metadata,
+      Assessed: +(initialValues?.Assessed ?? 0).toString().replace(/[$,]/g, ''),
```
dbarkowsky, line 294: There's a console.log right up here on line 292. ^^^

dbarkowsky (commented):
This Appraisal note type seems to have a missing status. It's falling under uncategorized:
image

Noticing that when I load a project and edit the main area, it sometimes shows all the fields.
image

When editing these fields, I tried checking one of the tasks and entering the Interest Component monetary field. It saved successfully in the database, but still showed as 0 in the Documentation section on the details page. Reloading the page threw this error:
image

react-app/src/components/projects/ProjectDetail.tsx
GrahamS-Quartech: Yeah, it's affecting the original. Basically the first reduce starts with an empty object, puts a bunch of stuff in it, then returns a reference to that new object which we assign to reduceMap. Then we pass this reference as the starting accumulator to each following the reduce, and each one modifies it further. In JS, primitive types are passed by value, while object types (including arrays) are passed by reference.

GrahamS-Quartech:
> This Appraisal note type seems to have a missing status. It's falling under uncategorized

This note type is actually "ErpNotification". I think in this case that field may have just been used weirdly. 

> Noticing that when I load a project and edit the main area, it sometimes shows all the fields. 
I still have a hard time getting this to trigger but I think the commit I just added might help solve this.

react-app/src/components/projects/ProjectDetail.tsx
GrahamS-Quartech: Probably wouldn't be too difficult and I think it would go a long way (should be another ticket though). There's not really any reason we can't just keep it as a number consistently, since the API accepts a number when you go to call save/insert/upsert.

dbarkowsky:
Noticing that when I edit the financial information, some of the values don't appear mapped correctly.
e.g.
image

Results in:
image

GrahamS-Quartech:
> Noticing that when I edit the financial information, some of the values don't appear mapped correctly.

Accidentally had the keys swapped around during submission. Should work fine now.

dbarkowsky (approved):
Recent changes seem to have fixed the noted issues. 👍

PR #2448, opened 2024-06-11
PIMS-1750: Modify ETL to extract metadata values into appropriate tables/columns

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1750: ](https://citz-imb.atlassian.net/browse/PIMS-1750)

**ETL Changes:**
No longer importing the "LeasedLandMetadata" field from the Buildings table or the "Metadata" column from the ProjectSnapshots table.

Added the OCG gain loss metadata field to the MonetaryType table in Postgres, and changed the CompletedOn date column in ProjectTasks to a regular column to because a date was being created even when the “isCompleted” boolean value was false.

Created a data migration to handle the alter project_task table, and modified the “1717629146741-MoveMetadataToTables.ts” migration to include the OCGGainLoss field as a monetary type.

**Steps for Testing:**
To test this locally, you'll need to run these commands to drop and recreate your local Postgres schema:

`DROP SCHEMA PUBLIC CASCADE`
`CREATE SCHEMA PUBLIC`

Then stop and restart the express api backend.

Then copy the tasks.json in this PR to replace your existing tasks.json in this directory: `"C:\Users\lawrlau\AppData\Roaming\DBeaverData\workspace6\General\.dbeaver"`
For Mac users, I believe it’s this: `“~/Library/DBeaverData/"`

Once you have copied the new tasks.json file, just start your local DBeaver application, navigate to your "Database Tasks" section, or go to Windows->Database Tasks, if it isn't visible in DBeaver.
Then, right click on the Prod to Local task and click on "Run Task":
image

This should be tested thoroughly, to make sure we've got all the metadata values associated to the correct data types, etc, as I did a lot of copying and pasting and might have missed something or added something incorrectly.

Note: after running the ETL, you may need to run the username updater script if you find yourself stuck at the access request page.

`cd ../tools/usernameUpdater/`

`npm run start`

## 🔰 Checklist


dbarkowsky (commented):
ETL worked without errors.
I spot checked some values between projects and the new tables. Seems good.
Noticed that we still have the old columns in the Projects table that correspond to the data that we're tracking in monetary and timestamp tables. Did we want to remove this?
image

dbarkowsky (approved):
Checked records in database and through frontend. Changes appear to be represented in full.

PR #2454, opened 2024-06-12
PIMS-1653: Adding all the layers to the map

Sharala-Perumal:
PIMS-1653:  Adding all the layers to the map                                  

## 🎯 Summary

  
[PIMS-1653](https://citz-imb.atlassian.net/browse/PIMS-1653)

## 🔰 Checklist


GrahamS-Quartech:
I toggled through all the different layers and they all appear to be working correctly as far as I can tell. The layer control should be positioned differently though, it's completely hidden when the property sidebar is open (which is the default). I also think that the parcel layer should probably be enabled by default (since I think this is the one that people actually use most of the time?)

Sharala-Perumal:
> I toggled through all the different layers and they all appear to be working correctly as far as I can tell. The layer control should be positioned differently though, it's completely hidden when the property sidebar is open (which is the default). I also think that the parcel layer should probably be enabled by default (since I think this is the one that people actually use most of the time?)

Good suggestions, Graham. Map should now render Parcel Boundaries as default layer and the layers control should display on the top left corner.

react-app/src/components/map/MapLayers.tsx
```diff
@@ -1,25 +1,181 @@
...
-        url={PARCEL_LAYER_URL}
-        format="image/png; mode=8bit"
-        transparent={true}
-        opacity={0.5}
-        layers="WHSE_CADASTRE.PMBC_PARCEL_FABRIC_POLY_SVW"
-      />
-    </>
-  );
-};
+    <LayersControl.Overlay name="Administrative Boundaries">
```
GrahamS-Quartech, line 99: I can see the idea behind setting up these parent components and having the enumerated children nested with a bit of left margin to make them look grouped together but I don't think it's working. It seems like the control displays identically when these parent LayersControl.Overlay's are removed. There are some leaflet plugins out there that provide this kind of functionality if it's really needed.

GrahamS-Quartech:
> Good suggestions, @GrahamS-Quartech . Map should now render Parcel Boundaries as default layer and the layers control should display on the top left corner.

These changes look good.

dbarkowsky (approved):

GrahamS-Quartech (approved):

PR #2457, opened 2024-06-13
PIMS-1743: Server Side Pagination

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1743](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1743)

* GET endpoint can now take an additional sorting key to specify a relation to sort on in addition to the actual property name.
* The options provided to the where section of TypeORM's find were changed to be an array (this constructs the query with OR instead of AND)
* Added helper methods to make it easier to add parameters to that array in a more generic way. Expects the operator strings provided by MUI's filtering model.
* FilterSearchDataGrid now expects an api call as a parameter (dataSource). It uses this to query the database when changing the pagination, sorting, or filtering mode of the data grid.

## 🔰 Checklist


react-app/src/components/table/DataTable.tsx
```diff
@@ -439,14 +577,27 @@ export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
...
         onStateChange={(e) => {
           // Keep track of row count separately
-          setRowCount(Object.values(e.filter.filteredRowsLookup).filter((value) => value).length);
+          if (!props.dataSource) {
+            setRowCount(Object.values(e.filter.filteredRowsLookup).filter((value) => value).length);
+          }
         }}
         onFilterModelChange={(e) => {
           // Can only filter by 1 at a time without DataGrid Pro
+          console.log(e);
```
dbarkowsky, line 586:
```suggestion
```

react-app/src/components/table/DataTable.tsx
```diff
@@ -182,12 +190,99 @@ type FilterSearchDataGridProps = {
 } & DataGridProps;
 
 export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
+  const [dataSourceRows, setDataSourceRows] = useState([]);
   const [rowCount, setRowCount] = useState<number>(0);
+  const [tableModel, setTableModel] = useState<ITableModelCollection>({
+    pagination: {
+      page: props.initialState?.pagination?.paginationModel?.page ?? 0,
+      pageSize: props.initialState?.pagination?.paginationModel?.pageSize ?? 10,
```
dbarkowsky, line 198: Same comment as before regarding default pagination size.

dbarkowsky (commented):
With the quickfilter, it sends a request on every change. I wonder if there's a way to wait to delay the search for a few milliseconds so users can type a string of letters before it queries.

The pre-set filter dropdowns don't appear to work anymore. The filter is set, but it's not affecting the results. This could be because the filter on IsDisabled doesn't seem to be working in general.

The counts in the table headers don't update. (Always 0)

react-app/src/components/table/DataTable.tsx
```diff
@@ -182,12 +190,99 @@ type FilterSearchDataGridProps = {
...
+    if (props.dataSource) {
+      dataSourceUpdate(tableModel);
+    }
+  }, [tableModel]);
+
+  useEffect(() => {
+    if (props.initialState?.sorting?.sortModel) {
+      tableApiRef.current.setSortModel(props.initialState.sorting.sortModel);
+    }
+  }, []); //Empty array makes this fire only on first render.
```
dbarkowsky, line 285: Is this not handled by the initialState prop on the DataGrid?

react-app/src/components/table/DataTable.tsx
```diff
@@ -343,7 +468,7 @@ export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
   // Sets quickfilter value of DataGrid. newValue is a string input.
   const updateSearchValue = useMemo(() => {
     return debounce((newValue) => {
```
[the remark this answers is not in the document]
GrahamS-Quartech: Added it to the debounce. Think it feels a bit better now?

react-app/src/components/table/DataTable.tsx
GrahamS-Quartech: Was running into issues with the sort model changing unexpectedly before but I must have corrected this somewhere since it seems like this isn't required anymore

GrahamS-Quartech:
> The pre-set filter dropdowns don't appear to work anymore. The filter is set, but it's not affecting the results. This could be because the filter on IsDisabled doesn't seem to be working in general.

This should work now, the isDisabled filter is present on agencies and admin areas at least.
 
> The counts in the table headers don't update. (Always 0)

I've made it so that this is just hidden when in server side pagination mode, at least for now.

react-app/src/components/table/DataTable.tsx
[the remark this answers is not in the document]
dbarkowsky: It's definitely subjective, but I think it feels better. Tried bumping the timing up to 300ms, and that sort of feels like a happy medium to me where it's still snappy but doesn't make a ton of calls before I'm done typing.

dbarkowsky (approved):
This is working really well so far, especially after the date column changes. Excited to see how it will apply to the property inventory in the future.
Noting that the date filter treats `is before` and `is on or before` as the same at the moment. Can be fixed with future tickets.

PR #2458, opened 2024-06-13
PIMS-1768: Updated DBeaver ETL task for Dev/Test/Prod

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1768: ](https://citz-imb.atlassian.net/browse/PIMS-1768)

Tried changing the port of the existing localhost postgres database connection to use the remote OC dev postgres database, but then all the table mapping gets mismatched for some reason.

Created an new ETL task, "PIMS ETL Prod to Dev" for running against Dev, Test, and Prod Openshift environments because it doesn't seem that DBeaver allows you to use the same database task against multiple databases. 

image

I also, included a "data-sources.json" file which should include the connection string that the new ETL task uses, you probably will need to enter the db name, username and password, port number should be 127.0.0.1:22224.
image

I also removed 2 older ETL tasks that were outdated, so you should now only see these 2 tasks, one for running locally and the other one for running against the Openshift environments.

## 🔰 Checklist


dbarkowsky (approved):
Tested this by doing the following:
- Dropped/renewed existing schema from TEST
- Pushed updated images so migrations would run
- Ran this new version of the ETL. No errors.
- Ran the usernameUpdater script.

Everything seems to have worked as expected.

PR #2459, opened 2024-06-13
 PIMS-1758 Map Parcel Search

dbarkowsky:
## 🎯 Summary

  
[PIMS-1758](https://citz-imb.atlassian.net/browse/PIMS-1758?atlOrigin=eyJpIjoiNGM3ZDkwMDcxNWExNDUwYTk1YWFiZTdmMzcxMjUzODAiLCJwIjoiaiJ9)

## Changes
- When searching for a PID or PIN in the map, if it's not found in the PIMS inventory, it searches the parcel layer
- If anything is found, it renders those parcels as polygons and zooms to show them.
- Removed a lot of old map code that we were previously use to display spider effects, old icons, and in-map property details.

Wasn't sure if we should show a popup when that happens, but there are cases where a result may contain multiple polygons, so it wasn't clear where to centre the popup.

## 🔰 Checklist


LawrenceLau2020 (approved):
Tested searching a PID using the inventory layer, parcel layer, as well as a parcel with an irregular shape and it was successful in all cases. If the PID wasn't found, it gives an appropriate message. Also, tested a PIN search. Working well.

PR #2460, opened 2024-06-14
PIMS-1594 Parent-Child Agency Relationship Limit

dbarkowsky:
## 🎯 Summary

  
[PIMS-1594](https://citz-imb.atlassian.net/browse/PIMS-1594?atlOrigin=eyJpIjoiYTBjMmI2NjE2MGU2NDA5ZjliZGIxZTFiNmI0M2E2OTIiLCJwIjoiaiJ9)

Intended to disallow multiple layers of parent/child relationships with agencies. It's either a parent, child, or neither.

## Changes
- Add backend checks to prohibit bad parent agency assignments
- Change how parents/children are identified in the agencyOptions. Lost parent field (boolean) and replaced with parentId (number). Can check if it's a possible parent if it doesn't have a parentId.
- Allow certain cases where users can clear the agency. Couldn't clear the parent agency field before. Also allowed clearing the user agency when editing. We should probably restrict editing this field to admins only in another ticket.
- Add prop to allow for disabling the AutocompleteFormField.

## 🔰 Checklist


LawrenceLau2020 (approved):
This is working well, prevents me from choosing child agencies as a parent agency and also prevents changing the parent agency on existing parent agencies.
I don't see any cases where a parent agency would become a child agency.

PR #2461, opened 2024-06-14
PIMS-1771: Parcel building union view 

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1771](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1771)

Just adding a new view so that we can more easily display the parcels and buildings in the same table on the front end. Mostly to facilitate the server side pagination ticket, since it's not really possible to sort and paginate effectively unless we are performing that action on a single table (view).

## 🔰 Checklist


GrahamS-Quartech:
> Should we add lot size to this for Parcels? It's used in the inventory table.

True, didn't notice this before, I will add it. Also, I welcome suggestions for better naming for this view/entity, couldn't think of anything more appropriate.

LawrenceLau2020:
> > Should we add lot size to this for Parcels? It's used in the inventory table.
> 
> True, didn't notice this before, I will add it. Also, I welcome suggestions for better naming for this view/entity, couldn't think of anything more appropriate.

Could this be called a property_list view?

GrahamS-Quartech:
> Could this be called a property_list view?

I think it sounds too "plural". Other tables are named as a singular. PropertyUnion feels like it has this problem a bit too but it's a bit more ambiguous at least.

dbarkowsky (approved):
Works well for me. Not picky about the name. PropertyUnion describes it well enough.

PR #2462, opened 2024-06-17
PIMS-1774 and PIMS-1757

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1774: ](https://citz-imb.atlassian.net/browse/PIMS-1774) - Make exemption requested field not required
[PIMS-1757: ](https://citz-imb.atlassian.net/browse/PIMS-1757) - New Disposal Project Submitted notification email not being sent

PIMS-1757 - Altered the project_status_notification seed from "6" to NULL for the from_status_id column on templates 1 & 2. The project status id of 6, which is "Review" isn't really used in the new PIMS.

PIMS-1774 - Altered the migration which seeded the "Exemption Requested" task from StatusId of 4 to 8 for "Submitted Exemption".

Creating a new project should trigger a "New Disposal Project Submitted" notification to be sent. This should also happen on ERP exempted projects.

## 🔰 Checklist


dbarkowsky (commented):
If I submit a project and request the exception, it doesn't actually complete the task that says exemption was requested. I think there's a disconnect somewhere with how the add project checkbox gets assigned to this task value.
image

GrahamS-Quartech:
> If I submit a project and request the exception, it doesn't actually complete the task that says exemption was requested. I think there's a disconnect somewhere with how the add project checkbox gets assigned to this task value.

This was because the exemption field left on the frontend was the old metadata version. I've pushed a fix to ensure it maps to the task instead.

dbarkowsky (approved):
This is working well for me after the most recent changes.
I think this is the first time we've passed user info down to a service, but I think this is okay because it's not being used for authorization purposes.
Alternative would be moving handleProjectNotifications up to the controller, but I like that those `handle...` functions are scoped within the service still.

PR #2469, opened 2024-06-19
PIMS-1741 Lookup Context

dbarkowsky:
## 🎯 Summary

  
[PIMS-1741](https://citz-imb.atlassian.net/browse/PIMS-1741?atlOrigin=eyJpIjoiZDhhYWUwZjJhZDVlNDY1MGIxZjdjNWJhZWY0Nzc1NzAiLCJwIjoiaiJ9)

Goal is to add a lookup context to the frontend that would reduce the number of joins needed in other database calls.
Lookup context matches this interface:
``` ts
export interface LookupAll {
  Risks: Partial<ProjectRisk>[];
  TimestampTypes: Partial<MetadataType>[];
  MonetaryTypes: Partial<MetadataType>[];
  NoteTypes: Partial<MetadataType>[];
  PropertyTypes: Partial<MetadataType>[];
  Tasks: Partial<Task>[];
  ProjectStatuses: Partial<ProjectStatus>[];
  ProjectTiers: Partial<TierLevel>[];
  ConstructionTypes: Partial<BuildingConstructionType>[];
  PredominateUses: Partial<BuildingPredominateUse>[];
  Classifications: Partial<PropertyClassification>[];
  Roles: Partial<Role>[];
  Agencies: Partial<Agency>[];
  AdministrativeAreas: Partial<AdministrativeArea>[];
  RegionalDistricts: Partial<RegionalDistrict>[];
}
```
**This PR only adds the context for use. Will continue this ticket with a second PR that starts to swap out the joins for context usage.**

## Changes
- Reinstated frontend /dev route, removing route that was used for LTSA testing
- Filled out /lookup/all controller logic.
- Added LookupContext to the frontend

## Testing
If you navigate to /dev, there's a log that should print the context to the console.

## 🔰 Checklist


GrahamS-Quartech:
The context seems to work well. I think one thing that might be desirable to include in this is some functions for retrieving a lookup item by it's numeric id. There would be some upfront cost in collecting objects into a feasible format for this (probably organized into objects of form [primarykey]: { ...restOfRow } ), but past that initial cost we would then be able to retrieve anything in O(1) time. This would keep things super speedy when we go to use these in the tables.

dbarkowsky:
@GrahamS-Quartech  I like the way that works. Good idea. 
Testing out the context in practice with the AdministrativeAreaDetail, I have some errors where I think the context isn't finished populating by the time we're trying to retrieve data from it. Will continue to investigate, but I don't think we should merge this until it's working properly.

GrahamS-Quartech (approved):

PR #2472, opened 2024-06-20
PIMS-1801 Replace Joins with Lookup

dbarkowsky:
## 🎯 Summary

  
[PIMS-1801](https://citz-imb.atlassian.net/browse/PIMS-1801)

Reducing the need for database joins by using the lookup context instead.
## Changes
- Removed a number of dataloader uses where context usage would work instead.
- Removed joins from services where it would no longer affect frontend behaviour
- Uncommented a line to restore quicksearch functionality on client-filtered tables
- MultiselectFormField now accepts nested indent formatting

## Issues
Didn't remove the joins on get______ services. This mostly just applies to tables. Replacing the joins on these tables with just the Id value affects the server-side pagination addition in a negative way. Can continue to look for a solution, but would like to see how the pagination is applied to the Disposal Inventory table first.

## Testing
Just use the app as normal. Note if any values appear missing or if any errors are thrown.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/lookup/lookupController.ts
+++ b/express-api/src/controllers/lookup/lookupController.ts
@@ -357,7 +357,7 @@ export const lookupAll = async (req: Request, res: Response) => {
   });
   const ProjectStatuses = await AppDataSource.getRepository(ProjectStatus).find({
     select: { Name: true, Id: true },
-    order: { SortOrder: 'asc' },
+    order: { SortOrder: 'asc', Name: 'asc' },
     where: { IsDisabled: false },
   });
   const ProjectTiers = await AppDataSource.getRepository(TierLevel).find({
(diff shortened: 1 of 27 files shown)


LawrenceLau2020 (approved):
Seems to work pretty well, tested a search filter and then sorting by each column and the pagination was keeping the filters.
I like how you can filter by almost any column by just typing what you want to filter by in the search box. For some reason typing "Land" won't filter by only land, but that's ok, as you can just apply a filter on the column itself.

I did run into a memory issue "not enough memory to open this page" while typing in the search box, but it's probably related to having too many tabs open.

GrahamS-Quartech (approved):
Good stuff. Even without being able to integrate these changes with the tables, I think it's still a big net positive. It's great not having to see a delay when dialogs try to load in the options for their drop downs especially. This context paired with the active inventory server side ticket should make the app feel snappier than ever.

PR #2476, opened 2024-06-22
PIMS-1804: Server side pagination for active inventory

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1804](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1804)

* Added the api hook for server side table pagination for the active inventory page.
* Improved filtering handling for PID/PIN case so that you can search with padded zeroes and hyphens and still have it work out.

## 🔰 Checklist


LawrenceLau2020 (approved):
Definitely much faster, and PID searches with hyphens works well.
The only change that I'm not sure how important it is to users is that we're missing the total row count for properties.

PR #2478, opened 2024-06-25
PIMS-1803: Bulk upload API

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1803 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1803)

* Adds a new endpoint that will accept a file using multer.
* Endpoint will return a 200 OK response ASAP, but not until after forking a node worker thread that will handle the file.
* The worker thread will attempt to bulk upsert the rows present in the spreadsheet using the SheetJS xlsx package. 
* Results are collected in a new import_result table, so make sure you run the migration. 

Please check general business logic in properties services and see if it makes sense with your own understanding and the previous implementation in the .NET API.

Note that a rollback transaction is wrapping the upload process right now. I will remove this before merging this branch.

## 🔰 Checklist


express-api/src/services/properties/propertiesServices.ts
```diff
@@ -110,6 +125,395 @@ const getPropertiesForMap = async (filter?: MapFilter) => {
...
+          const parcelToUpsert = await makeParcelUpsertObject(
+            row,
+            user,
+            roles,
+            lookups,
+            queryRunner,
+            existentParcel,
+          );
+          queuedParcels.push(parcelToUpsert);
+          results.push({ action: existentParcel ? 'updated' : 'inserted' });
```
dbarkowsky, line 459: If we push this success action here, could it still fail when the queuedParcels/Buildings goes to insert?
I'm thinking this might return false positives.

express-api/src/routes/propertiesRouter.ts
```diff
@@ -27,6 +29,10 @@ router.route('/search/geo').get(activeUserCheck, catchErrors(getPropertiesForMap
 router.route('/search/page').get(activeUserCheck, catchErrors(getPropertiesPaged));
 router.route('/search/page/filter').post(activeUserCheck, catchErrors(getPropertiesPagedFilter));
 
+const upload = multer({ dest: 'uploads/' });
```
dbarkowsky, line 32: Does this mean it will keep a local copy of the file in the container? Or it's just temporary?

dbarkowsky (commented):
I like where this is going. It seems like it's going to be really efficient compared to the old way of doing it.
Would like to have more information on the results though. I don't see how we could effectively identify which row caused the error. A row number would be good.
image

express-api/src/routes/propertiesRouter.ts
GrahamS-Quartech: It does temporarily create a copy of the file in the container but propertiesController.ts:210 will attempt to destroy it before exiting after processing the file.

express-api/src/services/properties/propertiesServices.ts
GrahamS-Quartech: Well in theory, anything that would cause the parcelToUpsert object to be invalid would have already thrown an error inside the makeParcelUpsertObject function. But that's a good point, the insert could still fail for some unexpected reason when it actually makes the save call. I suppose this may be another reason the old API does everything one at a time, but in the ideal case where all rows from the spreadsheet contain new PIDs it's much faster to do batched insert. I'll have to think about this a bit.

dbarkowsky: I definitely like the batch better. Just wondering if there's a way to see which items failed in that batch insert, then update the results maybe.

GrahamS-Quartech:
> I like where this is going. It seems like it's going to be really efficient compared to the old way of doing it. Would like to have more information on the results though. I don't see how we could effectively identify which row caused the error. A row number would be good.

Well, this results array does reflect the order in which rows are processed so the array index itself can serve as a row number when we go to display this on the frontend. But adding a row number key would be fine if we wanted to make this JSON blob more human readable.

dbarkowsky (approved):
Works well. Appreciate the enhanced feedback.

PR #2488, opened 2024-06-27
PIMS-1821: Fixing agency filtering on properties and projects

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1821: ](https://citz-imb.atlassian.net/browse/PIMS-1821)

## Changes
- Moved getProjects to use querybuilder instead of .find. Needed to get both AND and OR nested WHERE options.

## Issues
- Status isn't seen as a valid alias for filtering. Works for sorting though, but trying to filter on any field created by a join seems to crash it.
  - It's not clear what it wants for the Status alias.
  - Would it be easier to just make a view for this table? Then no joins in query.
- Preset filters (from the dropdown) not working as expected, especially when setting it to the `all` option.
- Using column filter with quicksearch filter overwrites the first filter. This makes it possible to search for records that should be omitted based on the other filter.

## 🔰 Checklist


express-api/src/services/properties/propertiesServices.ts
```diff
@@ -104,6 +104,11 @@ const collectFindOptions = (filter: PropertyUnionFilter) => {
     options.push(constructFindOptionFromQuery('AdministrativeArea', filter.administrativeArea));
   if (filter.propertyType)
     options.push(constructFindOptionFromQuery('PropertyType', filter.propertyType));
+  if (filter.agencyId && filter.agencyId.length > 0) {
+    filter.agencyId.forEach((id) => {
+      options.push(constructFindOptionFromQuery('AgencyId', `in,${String(id)}`));
+    });
+  }
```
dbarkowsky, line 111: I think doing it this way will add to the OR query. It really needs to be combined in a way that it is a requirement, not an option. Potentially like this, someone could search by agency name and get those properties even if they aren't in their own agencies, right?

TaylorFries:
I can add both filters and the results match. 
Issue: It seems like I can continue paging through non results? Like I can just keep clicking to the next page forever. 
Also the number of results always says "of more than x" where x is the upper limit of results on that page.
Screenshot 2024-07-05 at 8 27 51 AM

TaylorFries (approved):
double filtering works as expected on projects and inventory.

PR #2489, opened 2024-06-28
PIMS-1822/25 Property Errors

dbarkowsky:
## 🎯 Summary

  
[PIMS-1822](https://citz-imb.atlassian.net/browse/PIMS-1822)
[PIMS-1825](https://citz-imb.atlassian.net/browse/PIMS-1825)

## Changes
- Property Table only loads properties relevant to a user's agencies if not admin/auditor
- Trying to view Property Details of a property that doesn't exist or is outside of general user's agencies redirects to "/" page
- Suppress extra LTSA calls when pid is 0
- Add entry for classification lookup that handles undefined key.

I would like to find a more standard solution to viewing the request status from the data loader in the future. Maybe even the .ok property of the response as well. This should work until then.    

## Testing
- Try this with both admin and general user roles
- View property table, filter by agency -> should see different results depending on role
- View property details pages for properties in your agency and outside of it.

## 🔰 Checklist


TaylorFries (approved):
All changes listed work as expected. No issues detected.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -92,7 +92,14 @@ const sortKeyMapping = (
...
   const options = [];
-  if (filter.agency) options.push(constructFindOptionFromQuery('Agency', filter.agency));
+  // Has to handle if agency searched by name or set by users agencies
+  if (filter.agency || filter.agencyId)
+    options.push({
+      ...constructFindOptionFromQuery('Agency', filter.agency),
+      AgencyId: filter.agencyId
+        ? In(typeof filter.agencyId === 'number' ? [filter.agencyId] : filter.agencyId)
+        : undefined,
+    });
```
dbarkowsky, line 102: I'm having second thoughts about this solution.
This still only contributes to the OR query. If a user has agencyID # 1 only, but they also search by pid, I think they would be able to get a PID outside of their agency...

dbarkowsky: Fixed with recent commits.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -134,12 +149,35 @@ const collectFindOptions = (filter: PropertyUnionFilter) => {
...
+    .where(
+      new Brackets((qb) => {
+        options.forEach((option) => qb.orWhere(option));
+      }),
+    );
+
+  // Restricts based on user's agencies
+  if (filter.agencyId?.length) {
+    query.andWhere('agency_id IN(:list)', {
+      list: filter.agencyId.join(','),
```
GrahamS-Quartech, line 164: I think this actually needs to be `query.andWhere('agency_id IN (:...list)', { list: filter.agencyId } )`.
https://orkhan.gitbook.io/typeorm/docs/select-query-builder#using-parameters-to-escape-data

PR #2500, opened 2024-07-04
PIMS-1851 Submit Button Spacing

dbarkowsky:
## 🎯 Summary

  
[PIMS-1851](https://citz-imb.atlassian.net/browse/PIMS-1851)

## Changes
- Small amount of padding added under Add_____ pages. 

Tried to do this in the base layout at first, but there is some weird interaction with how the main component uses 100vh. This was simpler, but will have to be done for any future pages like this.

## 🔰 Checklist

### DIFF
--- a/.github/workflows/PROD-Api-V2-image-deploy.yml
+++ b/.github/workflows/PROD-Api-V2-image-deploy.yml
@@ -20,83 +20,42 @@ jobs:
     name: Deploy to OpenShift
     runs-on: ubuntu-latest
     steps:
-      # Checkout the repository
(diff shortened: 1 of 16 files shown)


TaylorFries (approved):
all submit buttons have padding

GrahamS-Quartech:
I made some changes to this so that you don't need to add padding to every page where this occurs. It should behave more consistently between browsers too. Also noticed that Chrome adds these extra unnecessary scrollbars to the DataTable, so I removed the overflow prop to make them go away.

PR #2504, opened 2024-07-05
PIMS-1828: PID Formatting on Map Filter

TaylorFries:
## 🎯 Summary

  
[PIMS-1828: ](https://citz-imb.atlassian.net/browse/PIMS-1828)

Added isPid bool to the textFormField so that when a user uses the filter by pid it gets formatted as a pid

## 🔰 Checklist


LawrenceLau2020 (approved):
The formatting looks good, and the filter works!
Just a minor issue, if you mistyped say the first 3 digits of the PID, eg. 123-456-789 and tried to change just the first 3 digits to 321 it doesn't seem to work, but highlighting the entire PID and then retyping the correct PID "321-456-789" works.

TaylorFries:
> The formatting looks good, and the filter works! Just a minor issue, if you mistyped say the first 3 digits of the PID, eg. 123-456-789 and tried to change just the first 3 digits to 321 it doesn't seem to work, but highlighting the entire PID and then retyping the correct PID "321-456-789" works.

I will create a ticket in the backlog to maybe look into how this pidformatting works and if we want to update it.

PR #2513, opened 2024-07-09
PIMS-1817 Tooltips for Tiers

dbarkowsky:
## 🎯 Summary

  
[PIMS-1817](https://citz-imb.atlassian.net/browse/PIMS-1817?atlOrigin=eyJpIjoiOWVmNWM5MWM5ZDY0NDU1OGFiZjdjM2FkNDFiYWIzNDciLCJwIjoiaiJ9)

## Changes
- /all lookup now returns descriptions where available
- SelectFormField and AutocompleteFormField now display tooltips for options if defined
- Added optional `tooltip` property to ISelectMenuItem

You should now see tooltips on the Tiers (Projects) dropdown in Add Project. I didn't add these anywhere else at the moment.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/lookup/lookupController.ts
+++ b/express-api/src/controllers/lookup/lookupController.ts
@@ -356,7 +356,7 @@ export const lookupAll = async (req: Request, res: Response) => {
     },
   });
   const ProjectStatuses = await AppDataSource.getRepository(ProjectStatus).find({
-    select: { Name: true, Id: true },
+    select: { Name: true, Id: true, Description: true },
     order: { SortOrder: 'asc', Name: 'asc' },
     where: { IsDisabled: false },
   });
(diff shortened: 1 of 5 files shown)


GrahamS-Quartech:
Works as advertised but not sure how I feel about putting tooltips over dropdown options, they kind of get in the way of hovering over subsequent items. Perhaps we could just have secondary descriptive text instead? Something like this:

image

dbarkowsky:
> Works as advertised but not sure how I feel about putting tooltips over dropdown options, they kind of get in the way of hovering over subsequent items. Perhaps we could just have secondary descriptive text instead? Something like this:

This looks a lot cleaner than the tooltips for sure. Did you already make these changes or is this just a mockup?

GrahamS-Quartech (approved):
Secondary text looking good. Would be nice to have it in more places but that's something we can pick away at. Verified that the nested indent style for access request page still works.

PR #2514, opened 2024-07-09
PIMS-1823: Fix count for server-side pagination

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1823: ](https://citz-imb.atlassian.net/browse/PIMS-1823) - Fix count for server-side pagination

## Changes
- Used a separate query for the "totalCount" in the project and property services because the TypeORM "findAndCount" function still returns the "filtered" number of rows and not the actual count.
- Modified the FilterSearchDataGridProps to take a rowCountProp number as a property that can be passed in.
- You should be able to see the total count reflected in the properties and projects list page as well as a snackbar notification at the bottom of the screen. This number should change when applying a filter on the list being displayed.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/projects/projectsServices.ts
+++ b/express-api/src/services/projects/projectsServices.ts
@@ -816,8 +816,6 @@ const getProjects = async (filter: ProjectFilter) => {
     });
   }
 
-  const totalCount = await query.getCount();
-
   if (filter.quantity) query.take(filter.quantity);
   if (filter.page && filter.quantity) query.skip((filter.page ?? 0) * (filter.quantity ?? 0));
   if (filter.sortKey && filter.sortOrder) {
@@ -830,7 +828,7 @@ const getProjects = async (filter: ProjectFilter) => {
       logger.error('PropertyUnion Service - Invalid Sort Key');
     }
   }
-  const projects = await query.getMany();
+  const [projects, totalCount] = await query.getManyAndCount();
   return { projects, totalCount };
 };
 
--- a/express-api/src/services/projects/projectsServices.ts
+++ b/express-api/src/services/projects/projectsServices.ts
@@ -816,6 +816,8 @@ const getProjects = async (filter: ProjectFilter) => {
     });
   }
 
+  const totalCount = await query.getCount();
+
   if (filter.quantity) query.take(filter.quantity);
   if (filter.page && filter.quantity) query.skip((filter.page ?? 0) * (filter.quantity ?? 0));
   if (filter.sortKey && filter.sortOrder) {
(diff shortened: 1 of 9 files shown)


express-api/src/services/projects/projectsServices.ts
```diff
@@ -816,6 +816,8 @@ const getProjects = async (filter: ProjectFilter) => {
     });
   }
 
+  const totalCount = await query.getCount();
+
```
LawrenceLau2020, line 820: had to use a separate query here after the agencyId filters and before the page filtering.

express-api/src/services/projects/projectsServices.ts
```diff
@@ -828,7 +830,8 @@ const getProjects = async (filter: ProjectFilter) => {
       logger.error('PropertyUnion Service - Invalid Sort Key');
     }
   }
-  return await query.getMany();
+  const projects = await query.getMany();
+  return { projects, totalCount };
```
dbarkowsky, line 834: There's an existing call that seems to work here and would do this all in one. 
```suggestion
  const [properties, totalCount] = await query.getManyAndCount();
  return { properties, totalCount };
```

express-api/src/services/projects/projectsServices.ts
dbarkowsky: See the comment on the lines below. I think we could remove this if that works for you.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -614,7 +616,8 @@ const getPropertiesUnion = async (filter: PropertyUnionFilter) => {
       logger.error('PropertyUnion Service - Invalid Sort Key');
     }
   }
-  return await query.getMany();
+  const properties = await query.getMany();
+  return { properties, totalCount };
```
dbarkowsky, line 620: Same idea as with projects.
```suggestion
  const [properties, totalCount] = await query.getManyAndCount();
  return { properties, totalCount };
```

dbarkowsky (commented):
I wonder if we need to snackbar popup on successful retrievals. The number of records is already shown above, so maybe that's duplicated info on screen. The error one could be useful though.

express-api/src/services/properties/propertiesServices.ts
LawrenceLau2020: Tested using getManyAndCount and it seems to work.

dbarkowsky (approved):
These changes are working well for me.
They don't apply to Agencies or Admin Areas yet, but those services may change a bit still in other tickets, so I'm not concerned. They currently show 0s.

PR #2515, opened 2024-07-09
PIMS-1802: Bulk Upload Page

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1802](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1802)

* Adds a bulk upload page with similar elements to the PIMS v1 version.
* Can be navigated to from the admin menu in the header. As such, it is a protected route.
* You should be able to click the dashed line area or drag and drop into it to attach a file.
* Once a file is attached, you can hit upload and the file will begin processing.
* The loading bar will progress based on the contents of the API call made every X seconds using the interval inside the useEffect.
* You can download the results of your most recent upload as an XLSX file or just view in the MUI data grid component.
* There are also some small changes to the backend. Mostly just adding some missed fields into the import service and modifying the way that the import_result row is first created so that it can be returned in the initial call the to the import endpoint.

## 🔰 Checklist


dbarkowsky (commented):
Like the UI here. 

- Land Area doesn't seem to be populated, even when the row has a value. Makes the details page look like this:
image
- PID/PIN may have to be joint optional fields. Right now PID is mandatory, but there are parcels that will only have a PIN.

GrahamS-Quartech:
> Like the UI here.
> 
>     * Land Area doesn't seem to be populated, even when the row has a value. Makes the details page look like this:

I think this might just be the test CSV having the wrong header (should be LandArea and not Land (space) Area).

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -447,7 +448,7 @@ const importPropertiesAsJSON = async (
   // let queuedParcels = [];
   // let queuedBuildings = [];
   const queryRunner = AppDataSource.createQueryRunner();
-  await queryRunner.startTransaction();
+  // await queryRunner.startTransaction();
```
dbarkowsky, line 451:
```suggestion
```

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -508,7 +509,7 @@ const importPropertiesAsJSON = async (
     logger.warn(e.message);
     logger.warn(e.stack);
   } finally {
-    await queryRunner.rollbackTransaction(); //NOTE: This rollback provided for testing convenience. Will be removed for final merge.
+    // await queryRunner.rollbackTransaction(); //NOTE: This rollback provided for testing convenience. Will be removed for final merge.
```
dbarkowsky, line 512:
```suggestion
```

dbarkowsky (commented):
Working well for me. Knocking it out of the park. ⚾ 
A few comments to discuss then let's get it in.

GrahamS-Quartech:
> PID/PIN may have to be joint optional fields. Right now PID is mandatory, but there are parcels that will only have a PIN.

Just noting here that Steven helped us decide that PID will remain mandatory for now.

dbarkowsky (approved):

PR #2519, opened 2024-07-10
PIMS-1816: Exemption Field On Add Project

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1816 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1816)

* Adds an exemption note field towards the bottom of the add project page.
* Will only be enabled if the exemption task has also been ticked.

Note: The note that I believe should be the correct one here is mapped to a different status from the exemption task. Submitted exemption vs Approved for exemption. Anyone have any thoughts on whether this is acceptable or not?

## 🔰 Checklist


react-app/src/components/projects/AddProject.tsx
```diff
@@ -40,32 +41,40 @@ const AddProject = () => {
...
+    [lookupData],
+  );
+  const salesCostType = useMemo(
+    () => lookupData?.MonetaryTypes?.find((a) => a.Name === 'SalesCost'),
+    [lookupData],
+  );
+  const programCostType = useMemo(
+    () => lookupData?.MonetaryTypes?.find((a) => a.Name === 'ProgramCost'),
+    [lookupData],
+  );
```
dbarkowsky, line 61: Similar comment as in other PR regarding finding items using name. Changing name could cause issues.

react-app/src/components/projects/AddProject.tsx
```diff
@@ -221,6 +230,14 @@ const AddProject = () => {
...
             </Grid>
           ))}
+          <Grid item xs={12}>
+            <TextFormField
+              disabled={watch.find((a) => a.TaskId === exemptionTask.Id)?.IsCompleted != true}
+              fullWidth
+              name={'ExemptionNote'}
+              label={'Exemption rationale note'}
+            />
+          </Grid>
```
dbarkowsky, line 240: In old PIMS, this was a required field when there was a request for exemption. 
Also suggest making it a multiline input and hiding when not required.

```suggestion
          {watch.find((a) => a.TaskId === exemptionTask.Id)?.IsCompleted === true ? (
            <Grid item xs={12}>
              <TextFormField
                fullWidth
                name={'ExemptionNote'}
                label={'Exemption rationale note'}
                required
                multiline
                minRows={2}
                rules={{
                  required: 'Rationale required when requesting exemption.',
                }}
              />
            </Grid>
          ) : (
            <></>
          )}
```

dbarkowsky (changes requested):
Some suggestions about how the field is presented. Works well otherwise.

react-app/src/components/projects/AddProject.tsx
GrahamS-Quartech: I think I'll just handle this pattern in all the spots its used in another ticket. Will make an enum for it as I believe the id's should remain consistent enough between migrations as long as we don't screw up the sequence somehow.

react-app/src/components/projects/AddProject.tsx
GrahamS-Quartech: I think some web design purists may argue that hiding fields this way is bad practice but I personally think it's alright here. Implemented these changes.

react-app/src/components/projects/AddProject.tsx
GrahamS-Quartech: Replaced these with the relevant enums.

dbarkowsky (approved):

PR #2520, opened 2024-07-10
PIMS-1868 More Quickfilter Fixes

dbarkowsky:
## 🎯 Summary

  
[PIMS-1868](https://citz-imb.atlassian.net/browse/PIMS-1868?atlOrigin=eyJpIjoiYzA0N2E4ZTAxMTA4NDAyNGFlYzU2NTBlOGI4YjAzYzAiLCJwIjoiaiJ9)

## Changes
- Modified backend services for Agencies and Admin Areas to use queryBuilder
- Modified projects service so that keywordFilter in frontend will be applied
- Changed columns to boolean type in MUI tables where possible
- Fix keywordFilter interaction so that the props are saved/read from the query string correctly

## Testing
- Please test these tables for sorting/filtering using the in-column filter, keyword filter, and pre-set select filter:
  - Agencies
  - Admin Areas
  - Properties
  - Projects

## Known Issue
- Some columns in the database have both NULL and empty strings. This causes weird sorting behaviour. Would be good to standardize these to default to NULL if the value is just an empty string.

## 🔰 Checklist


GrahamS-Quartech (approved):
Noticed no issues. Tables all appear to sort, filter, etc correctly. Did not notice any issues related to query strings in the URL either. Nicely done.

PR #2522, opened 2024-07-10
PIMS-1818: Acknowledge notifications before saving project in approval state

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1818](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1818)

Wound up going for a slightly different method than what was suggested in feedback. I don't really want to layer multiple modals on top of each other, so I think it's better to just show a checkbox right at the end of the form instead. You should only see it when you are transitioning to Approved for ERP from any other state, and you should be required to tick off the acknowledgement checkbox before the form will save.

Also, we may wind up changing what status gets used for this depending on whether we get rid of In ERP or Approved for ERP. But I'm pretty sure Approved for ERP is going to be the one we keep.

## 🔰 Checklist


react-app/src/components/projects/ProjectDialog.tsx
```diff
@@ -148,11 +150,20 @@ export const ProjectGeneralInfoDialog = (props: IProjectGeneralInfoDialog) => {
     );
   }, [statusTypes, initialValues]);
 
+  useEffect(() => {
+    setApprovedStatus(lookupData?.ProjectStatuses?.find((a) => a.Name === 'Approved for ERP')?.Id);
```
dbarkowsky, line 154: Is the name the best way to find this? I think we do this in a few places, but I wonder if that's opening ourselves to issues if the names of the statuses change. I almost feel like an enum would be safer.
And if it's always going to be a single value, why not just set it explicitly?

dbarkowsky (approved):
Works well. Good, un-intrusive solution.
One comment about how we find the value for the proper status, but I think we can proceed as is for now.

react-app/src/components/projects/ProjectDialog.tsx
GrahamS-Quartech: Not sure what you mean by set it explicitly, but yeah I have thought before that finding by name may not really be ideal. Putting the names in an enum would be an improvement at least since we could at least change the names anywhere they're used with just one line if we ever need to. Ultimately we kinda have to hardcode something though for these sorts of cases, whether it's a numeric id or a name.

PR #2523, opened 2024-07-10
Pims-1858 Truncate Address From Geocoder

TaylorFries:
## 🎯 Summary

  
[PIMS-1858: ](https://citz-imb.atlassian.net/browse/PIMS-1858)

Made the following changes:
- truncated address on first comma from geocoder
- updated address model for geocoder
  - is fields were being returned as a string not bool
  - also updated logic for creating address1
  - more work will have to go into this logic block before we use it
- added a 10m setback to geocoder call
  - theoretically this should move the point away from the road and toward the center of the parcel
  - I have found a few instances that it still misses the parcel 

## 🔰 Checklist


express-api/src/services/geocoder/geocoderService.ts
```diff
@@ -34,6 +34,8 @@ export const getSiteAddresses = async (
   url.searchParams.append('addressString', address);
   url.searchParams.append('minScore', minScore);
   url.searchParams.append('maxResults', maxResults);
+  url.searchParams.append('locationDescriptor', 'accessPoint');
+  url.searchParams.append('setBack', '10');
```
dbarkowsky, line 38: Was curious about what this locationDescriptor did, so I did some tests with it on/off:
All tests done with 467 Lampson St.
With lines 37 and 38 commented out, I get this:
image

With line 38 commented out only, I get this:
image

And with both lines present, I get this:
image

It seems like adding these lines might have reduced the accuracy of the location somehow.

Yet, with 1407 Haultain St (a weird split parcel), I get this with them both active:
image

And this with neither active:
image

With them, accuracy was 2/2 checked. Without, it was 1/2. so I think it's still worth it.

dbarkowsky (approved):

express-api/src/services/geocoder/geocoderService.ts
TaylorFries: It really could go either way - it just depends on each parcel. But I found the same, on parcels that geocoder sends a good point it makes the point worse but still acceptable, on parcels that geocoder sends a bad point it makes it better.

PR #2532, opened 2024-07-12
PIMS-1856: Cleanup Project Statuses

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1856](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1856)

Added a migration which:
 * disables deprecated statuses
 * consolidates several other statuses
 * adds duplicated fields under task and note_type for fields that may be entered on either Approval status.
 * ~~duplicates existing project_task and project_note entries for now duplicated field types, since they may have been entered under either approval state previously~~ Per discussion over teams, I actually removed this in favor of doing it in the ETL. 
 * fixes sort order for the consolidated field types so that they appear in a somewhat coherent grouping on the frontend

Some things I have not done but we could consider adding to this migration:
* changing the sort order of the project statuses, I think it looks quite a bit better with all the cruft disabled but we could still tidy this up further if we can decide on what that order would be
* some of the names for the consolidated tasks under the Approval status look a bit non-descript in the frontend, could consider changing their names or perhaps using the description for display instead

## 🔰 Checklist


GrahamS-Quartech:
~~duplicates existing project_task and project_note entries for now duplicated field types, since they may have been entered under either approval state previously~~ 
Per discussion over teams, I actually removed this in favor of doing it in the ETL. It's removed from the down case as well so you may need to clean up those rows manually in case of a revert.

dbarkowsky (approved):
Migration completes without errors.

PR #2534, opened 2024-07-15
PIMS-1752: Allow editing of previous netbook and asssessment values

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1752: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1752)

* The disabled prop has been removed on the non-year fields of the netbook and assessment forms for properties.
* An additional restriction has been added on adding a new row to these sections: you cannot re-use an existing year. Instead, just edit the relevant row and submit.
* You still must enter a year of current year or previous year for a new value.
* I removed some logic in the transform function for handling the submission, as it should no longer be relevant to do that filtering with the above changes.

## 🔰 Checklist


dbarkowsky (approved):
I like the way this works. It's more obvious to update the current year as well.

PR #2544, opened 2024-07-17
PIMS-1754: Display Project Notifications

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1754: ](https://citz-imb.atlassian.net/browse/PIMS-1754)

## Changes
- New data card for notifications with a paginated table displaying notifications for a project and their status based on the NotificationQueue database table.
- When the project page first loads, it should make a call to the ches api to update all notifications that are in a pending or active state.
- When opening the dialog, the notifications will display in a wider table.

image

## Testing
- To test this you will need to use the CHES prod credentials in your .env
- The projects with the most notifications will be the ones in "ERP" status.
- This query will give you all the project ids with "pending" notifications:
`SELECT project_id, count(send_on) FROM notification_queue WHERE status = 1 and send_on < '2024-07-29' GROUP BY project_id`
- Going to the project details page using the project id in the url should trigger the update the status to 4 or "completed" on the pending notifications which were supposed to be sent.
- May also want to compare the notifications against PIMS prod live.

## Considerations
- If we want to  to capture the instances where a notification doesn't get sent because of network related issues, etc., then we need to create new code to handle a possible 404 error that we get from CHES.
- We may want to disable the "Confirm" button as well as it doesn't currently do anything right now.

## 🔰 Checklist


express-api/src/services/notifications/notificationServices.ts
```diff
@@ -353,10 +372,104 @@ const sendNotification = async (
...
+      notification.UpdatedOn = new Date();
+      notification.UpdatedById = user.Id;
+      const updatedNotification = await query.manager.save(NotificationQueue, notification);
+
+      return updatedNotification;
+    } else {
+      throw new Error(`Failed to retrieve status for notification with id ${notificationId}.`);
+    }
+  } catch (error) {
+    await query.rollbackTransaction(); // Rollback transaction on error
```
LawrenceLau2020, line 422: We may want to catch 404 errors here?

react-app/src/components/projects/ProjectNotificationsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const ProjectNotificationsTable = (props: ProjectNotificationsTableProps) => {
+  const [paginationModel, setPaginationModel] = useState({
+    pageSize: 10,
+    page: 0,
+  });
+
+  useEffect(() => {
+    fetchData();
+  }, [paginationModel.page, paginationModel.pageSize]);
```
GrahamS-Quartech, line 28: Is server-side pagination being used here? It looks like it's configured to be client side.

express-api/src/controllers/notifications/notificationsController.ts
```diff
@@ -104,6 +107,40 @@ export const getNotificationInQueueById = async (req: Request, res: Response) =>
...
+    if (!projectId) {
+      return res.status(400).send({ message: 'Project ID is required.' });
+    }
+
+    // Fetch notifications in the queue for this project
+    const notificationFilter = {
+      projectId: Number(projectId), // Ensure projectId is a number
+      pageNumber: Number(req.query.pageNumber) || 0, // Default to page 0 if not provided
+      pageSize: Number(req.query.pageSize) || 0, // Default to 10 if not provided
+    };
```
GrahamS-Quartech, line 130: I think this could be handled more similarly to other controllers just for consistency sake, especially in regards to the key names for pageNumber and pageSize (elsewhere it is page and quantity).

express-api/src/controllers/notifications/notificationsController.ts
```diff
@@ -104,6 +107,40 @@ export const getNotificationInQueueById = async (req: Request, res: Response) =>
   return stubResponse(res);
 };
 
+/**
+ * @description Get all notifications for a specific project.
+ * @param {Request}     req Incoming request.
+ * @param {Response}    res Outgoing response.
+ * @returns {Response}      A 200 status with an object containing the specified notifications in the queue.
+ */
+export const getNotificationsByProjectId = async (req: Request, res: Response) => {
```
GrahamS-Quartech, line 116: Should this function check user agency permissions?

react-app/src/components/projects/ProjectDetail.tsx
```diff
@@ -398,6 +404,39 @@ const ProjectDetail = (props: IProjectDetail) => {
...
+        >
+          {!data?.parsedBody.Notifications?.length ? ( //TODO: Logic will depend on precense of agency responses
+            <Box display={'flex'} justifyContent={'center'}>
+              <Typography>No notifications were sent for this project.</Typography>
+            </Box>
+          ) : (
+            <ProjectNotificationsTable
+              rows={
+                data?.parsedBody.Notifications
+                  ? data?.parsedBody.Notifications.sort(
```
GrahamS-Quartech, line 424: Would probably make more sense to provide a default sort order to the GridColDef of this table component rather than sort here.

react-app/src/components/projects/ProjectDialog.tsx
```diff
@@ -489,3 +491,66 @@ export const ProjectAgencyResponseDialog = (props: IProjectAgencyResponseDialog)
...
+
+  const fetchUpdatedNotifications = async (projectId: number) => {
+    setSubmitting(true);
+    try {
+      const notifications = await api.notifications.getNotificationsByProjectId(projectId);
+      setRows(notifications);
+    } catch (error) {
+      console.error('Error fetching updated notifications:', error);
+    } finally {
+      setSubmitting(false);
```
GrahamS-Quartech, line 532: This is fine but alternatively you could leverage the useDataSubmitter hook which handles the boilerplate of exposing some loading state for the request.

express-api/src/controllers/notifications/notificationsController.ts
LawrenceLau2020: made some changes to this

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -355,10 +374,106 @@ const sendNotification = async (
...
+    if (
+      notification.Status === NotificationStatus.Pending ||
+      notification.Status === NotificationStatus.Accepted
+    ) {
+      const updatedNotification = await updateNotificationStatus(notification.Id, user);
+      updatedNotifications.push(updatedNotification);
+    } else {
+      updatedNotifications.push(notification);
+    }
+  }
```
GrahamS-Quartech, line 466: There's a few things I'm noticing here.
* updatedNotifications array's final value seems to go unused, I think that you need to include the results of updating these notifications in the return value of the function.
* updateNotificationStatus should not use await as there are no dependencies on the result of this call until the function returns, you should instead collect each promise in to an array and then await them with either Promise.all() or Promise.allSettled(), you can expect a pretty significant performance increase from this

react-app/src/components/projects/ProjectDialog.tsx
```diff
@@ -489,3 +491,46 @@ export const ProjectAgencyResponseDialog = (props: IProjectAgencyResponseDialog)
...
+
+export const ProjectNotificationDialog = (props: INotificationDialog) => {
+  const { initialValues, open, onCancel } = props;
+  const [rows, setRows] = useState<INotificationModel[]>([]);
+  useEffect(() => {
+    if (initialValues) {
+      setRows(
+        initialValues.Notifications?.map((resp) => ({
+          id: resp.Id,
+          agency: props.ungroupedAgencies?.find((agc) => agc.Id === resp.ToAgencyId)?.Name,
```
GrahamS-Quartech, line 511: This can also make use of the lookupcontext.

react-app/src/components/projects/ProjectNotificationsTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+  status: string;
+  sendOn: Date;
+  to: string;
+  subject: string;
+}
+export interface INotificationResponse {
+  items: INotificationModel[];
+  pageNumber: number;
+  pageSize: number;
+}
```
GrahamS-Quartech, line 18: It would be better if this typing was more similar to other hooks that return paged results.

LawrenceLau2020:
I made changes to all the above comments, except for removing the rollbackTransaction, due to time.

dbarkowsky (approved):
Added one condition that won't update the status if it's something unexpected, even if it's a string.
Happy with the way this works.
Some thoughts for future improvements:
- Option to cancel, resend individual notifications
- An internal status for when the message has already been removed from the CHES server. Right now, it would stay in the Pending status, which means it would be checked every time.

PR #2545, opened 2024-07-18
PIMS-1888 PID Search Formatting

TaylorFries:
## 🎯 Summary

  
[PIMS-1888: ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1888)

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/table/DataTable.tsx
+++ b/react-app/src/components/table/DataTable.tsx
@@ -476,7 +476,13 @@ export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
   // Sets quickfilter value of DataGrid. newValue is a string input.
   const updateSearchValue = useMemo(() => {
     return debounce((newValue) => {
-      tableApiRef.current.setQuickFilterValues(newValue.split(' ').filter((word) => word !== ''));
+      tableApiRef.current.setQuickFilterValues(
+        newValue
+          .replace(/^0+/, '')
+          .replaceAll('-', '')
+          .split(' ')
+          .filter((word) => word !== ''),
+      );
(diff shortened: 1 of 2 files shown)


react-app/src/components/table/DataTable.tsx
```diff
@@ -476,7 +476,13 @@ export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
...
   const updateSearchValue = useMemo(() => {
     return debounce((newValue) => {
-      tableApiRef.current.setQuickFilterValues(newValue.split(' ').filter((word) => word !== ''));
+      tableApiRef.current.setQuickFilterValues(
+        newValue
+          .replace(/^0+/, '')
+          .replaceAll('-', '')
+          .split(' ')
+          .filter((word) => word !== ''),
+      );
```
dbarkowsky, line 485: This solution would affect every search on every page being done with the keyword search bar. 

I would suggest instead looking at the `propertiesServices.ts` file in the backend. Currently on line 586, it adds the quickfilter options. If the foreach loop on `quickFilterFields` pushed `constructFindOptionFromQueryPid` instead for the PID and PIN fields, I think this would solve the issue more specifically.

dbarkowsky (changes requested):

react-app/src/components/table/DataTable.tsx
TaylorFries: okay i tried to implement this as you suggested. Still open to any and all feedback. Also would we want this for PID and PIN? are pins formatted the same as pids?

dbarkowsky (approved):
Working great. 👍

GrahamS-Quartech (approved):
👍 Reusing that function is definitely the way to go about it

PR #2560, opened 2024-07-22
PIMS-1990: Email Authorized env variable for testing emails in live

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1990](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1990)

Introduced ~~CHES_EMAIL_AUTHORIZED~~ CHES_SEND_TO_LIVE env and relevant entry in the config function. When set to false + no OVERRIDE_TO present, the SSO user's email field is used instead.

We can use this to allow PO's to verify email behavior in live. When they go to change a project status, they should see the emails that would usually be sent out to agencies and the like appear in their own inbox.

## 🔰 Checklist


dbarkowsky (commented):
I think there's something backward as to what the ENV behaviour should be if `true`.  See comments.
Also wondering if there's something more obvious we could name the ENV. `EMAIL_AUTHORIZED` makes it sound like we're checking for permissions, but it's not clear that it is reflecting the email back to the user.

GrahamS-Quartech:
> I think there's something backward as to what the ENV behaviour should be if `true`. See comments. Also wondering if there's something more obvious we could name the ENV. `EMAIL_AUTHORIZED` makes it sound like we're checking for permissions, but it's not clear that it is reflecting the email back to the user.

I did write the comment for this field backwards to how it is actually behaving, whoops. I'm open to suggestions for a better naming scheme, this is just taken from PIMS v1.

dbarkowsky (approved):

PR #2563, opened 2024-07-23
PIMS-1901 PID or PIN Required on Parcel Add

TaylorFries:
## 🎯 Summary

  
[PIMS-1901: ](https://citz-imb.atlassian.net/browse/PIMS-1901)

Added line to note the requirement of pid or pin when adding a parcel 

## 🔰 Checklist


GrahamS-Quartech:
I wonder if there should be an additional layer of boolean logic here... Since one is not required anymore if the other has been set. I imagine this is why we originally did not have the * indicator on these.

dbarkowsky (approved):
image

Appears only for parcel as expected.

Issue #2498, opened 2024-07-02
[Issue]: Cannot set location with map for "Create New Property"


BradyMitch:
### Details

Under the map it says "Fill fields or drag map to set location."

I can drag the map, zoom in on a parcel and even click on the parcel to see its PID but it does not set location.

Needs added functionality or changed language to suggest that the map can be used to find a PID.

### Steps to Reproduce (Optional)

1. Open "Active Inventory" tab.
2. Click on "Create New Property" ("+") button.
3. Scroll down to the map.
4. Move and zoom the map to a property, and click.

### Issue Type

Visual

### What browsers are you seeing the problem on?

Chrome

### What Operating System are you using?

Windows

### Environment (Optional)

Dev

GrahamS-Quartech:
Fixed by: https://github.com/bcgov/PIMS/pull/2566

PR #2566, opened 2024-07-23
PIMS-1874: Fix for onMove being registered inconsistently to the map

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1874](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1874)

Addresses this issue: https://github.com/bcgov/PIMS/issues/2498

* Moved onMove into the hook of a null component so that it will be consistently attached. Relying on the map ref in a useEffect is not consistent.
* Modified the way that the map is set to its default location for existing properties by allowing the map to take default location and zoom props.
* Eliminated some redundant state as the watch feature of react hook form can be used for the same purpose.

## 🔰 Checklist


react-app/src/components/property/PropertyForms.tsx
```diff
@@ -89,37 +92,11 @@ export const GeneralInformationForm = (props: IGeneralInformationForm) => {
   const postalRegex = /^[ABCEGHJ-NPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][ ]?\d[ABCEGHJ-NPRSTV-Z]\d$/i;
 
   const map = useRef<Map>();
-  const [position, setPosition] = useState<LatLng>(null);
-
+  const position = formContext.watch('Location');
```
dbarkowsky, line 95: Nice touch.

dbarkowsky (approved):
Works well in practice. Lat,Lng info was always updated as expected and map as centred on property when loaded.

PR #2569, opened 2024-07-24
PIMS-1919: Bulk upload resolve path for production

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1919](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1919)

* I added a ternary that reads the NODE_ENV and changes the path to point to the js version of the worker script instead of the ts version if the value is 'production'. This is maybe not super robust but from my research it doesn't seem like there's an obvious cleaner solution.
* The error event on the worker thread will now fire an update to the import result table and set the CompletionPercentage to -1 to indicate an unexpected failure.
* The frontend has been updated to handle the case where the -1 appears and display some different messaging accordingly. Note that the bulk upload will ideally never show the user this, but in a case like this we still don't want to leave them hanging on the processing screen forever.

## Testing

Test both with local instance of your API and a containerized version to make sure bulk upload works correctly regardless of whether ts has been compiled. 

## 🔰 Checklist


dbarkowsky (approved):
Changed some console.logs to logger.___ calls. 
Tested through frontend in dev build and then built container and tested like that.
All seems to be working fine.

PR #2570, opened 2024-07-24
Wrap getLookupValueById in useCallback

GrahamS-Quartech:
Wrapped the function for looking up values by id in a useCallback so that dependent components know to re-render when data is loaded. Most evident on the property page as the object used to display property data was memoized and did not have anything lookupContext related as a dependency. 

## 🔰 Checklist


TaylorFries (approved):
everything loads well and no issues detected.

PR #2578, opened 2024-07-26
PIMS-1925 Project Editing Admins Only

dbarkowsky:
## 🎯 Summary

  
[PIMS-1925](https://citz-imb.atlassian.net/browse/PIMS-1925)

## Changes
- Disabled edit buttons on ProjectDetails if lacking the Admin role
- Added check in update projects controller to look for Admin role first
## Testing
- Make sure you can still edit projects as an admin
- Make sure as a general user you can't see the edit options

## 🔰 Checklist


GrahamS-Quartech:
I wonder if the delete button should also be disabled if you are not an admin?

dbarkowsky:
> I wonder if the delete button should also be disabled if you are not an admin?

Did not see this until now. Have changed the delete disable condition to be admin only.

GrahamS-Quartech (approved):

PR #2580, opened 2024-07-26
PIMS-1927 Keycloak Roles Update 

dbarkowsky:
## 🎯 Summary

  
[PIMS-1927](https://citz-imb.atlassian.net/browse/PIMS-1927)

## Changes
Altered the syncKeycloakUserRoles service so it does the following:
- If the user has keycloak roles, apply them to the database
- If they don't have keycloak roles, but have database roles, apply them to Keycloak
- Otherwise, don't do anything.

## Testing
Try changing your role in both keycloak and the database manually. 
Log in again to see if the sync took affect. This only triggers on login, not a refresh.

## Concern
Because this only triggers on login, there's a situation where a user with no keycloak role may log in and receive the no-access page, even if they have a database role. If they refresh the page, it should resolve itself, but it would be nice to handle this somehow. Open to ideas.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/keycloak/keycloakService.ts
+++ b/express-api/src/services/keycloak/keycloakService.ts
@@ -168,35 +168,36 @@ const updateKeycloakRole = async (roleName: string, newRoleName: string) => {
 };
 
 const syncKeycloakUser = async (username: string) => {
-  // Does user exist in Keycloak?
-  // Get their existing roles.
-  // Does user exist in PIMS
-  // If user exists in PIMS...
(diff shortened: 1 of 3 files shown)


dbarkowsky:
Pulling back to draft for now. May have received intervention from the SSO team that will make this unnecessary.

dbarkowsky:
Closing this. SSO team transferred BCeID users. Used the role-mapping script to assign roles in new integration. Spot checking seems to suggest it was successful. Keycloak will remain the sole source of authorization truth.

PR #2589, opened 2024-07-30
PIMS-1945: Property views exclude deleted rows

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1945 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1945)

Added deleted_on clause to view definitions and created relevant migration.

Note that if you follow the workflow of search property on map -> open property from map in new tab -> delete property, if you look back at the original tab the property won't appear as gone until you take some action that queries the view again. Not sure if there's much we can do about this though.

## 🔰 Checklist


dbarkowsky (approved):
Tested by deleting a property and then seeing if it could be accessed. 
Request no longer returns the soft-deleted property. 👍

react-app/src/components/property/PropertyDetail.tsx
```diff
@@ -262,6 +263,12 @@ const PropertyDetail = (props: IPropertyDetail) => {
   const [openNetBookDialog, setOpenNetBookDialog] = useState(false);
   const [openAssessedValueDialog, setOpenAssessedValueDialog] = useState(false);
 
+  const deletionAction = async () => {
+    deletionBroadcastChannel.postMessage('refresh');
+    await deleteProperty();
+    navigate('/properties');
```
dbarkowsky, line 269: Should this be dependant on a successful delete?

react-app/src/components/map/ParcelMap.tsx
```diff
@@ -76,6 +77,15 @@ const ParcelMap = (props: ParcelMapProps) => {
     api.properties.propertiesGeoSearch(filter),
   );
 
+  const deletionBroadcastChannel = useMemo(() => new BroadcastChannel('property'), []);
+  useEffect(() =>
+    deletionBroadcastChannel.addEventListener('message', (event) => {
+      if (typeof event.data === 'string' && event.data === 'refresh') {
+        refreshData();
+      }
+    }),
```
dbarkowsky, line 86: If we were to store the map view before the refresh, then reapply it after, maybe that would avoid the zooming out issue?

GrahamS-Quartech: Good suggestion but changes to accommodate this would be potentially breaking, so we will come back to this later.

dbarkowsky (approved):

PR #2590, opened 2024-07-30
PIMS-1871: Fuzzy search is returning properties already in other projects

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1871: ](https://citz-imb.atlassian.net/browse/PIMS-1871) 

## Changes
- Added a subquery to the fuzzy search to exclude properties that are already in projects.

## 🔰 Checklist


dbarkowsky (commented):
The subquery is checking for cases where the property belongs to _any_ project, but are  users allowed to add properties to projects if they were previously part of a cancelled project? Would be nice if there was a way to do that.

LawrenceLau2020:
I've added a couple sub queries for both the parcelsQuery and the buildingsQuery to try to include properties that don't belong to any projects as well as belonging to "Cancelled" projects. Could use a second pair of eyes right now, try searching for a building that doesn't belong to any project, I think the query includes it, but the ordering isn't correct.

The parcel query needs testing as well and possible refactoring, as I was just trying to get the buildingsQuery to work properly first.

dbarkowsky (approved):

PR #2591, opened 2024-07-30
PIMS-1933: Cancel notifications on project deletion

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1933 ](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1933)
* Project controller will attempt to cancel all remaining notifications on project delete. 
* Modified controller to only accept admins for project deletion.

Notes:
* I didn't think that the ches service having issues is reason enough to not allow the project to be cancelled. So the transaction to delete the project may still be commited even if the ches calls all fail. 
* I'm not sure if we want to show to the user the count of which emails didn't get cancelled, since this could just be due to desync between our notification_queue table and ches. For example, if you move the project to approved for erp and immediately try to delete it, it may fail to cancel some notifications that have already completed because they are marked pending in our table. 

## 🔰 Checklist


dbarkowsky (approved):
RE your notes:
- Agreed that failed notification cancellation shouldn't stop the deletion of a project.
- I don't think we need to show them that X number of cancelations couldn't be confirmed. They can't really do anything about it if the project is deleted at that point anyway.

PR #2595, opened 2024-07-31
PIMS-1949: Keep DataTable state between pages more consistently 

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1949](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1949)

Made DataTable much more dependent on the query param state by using the react-router-dom useSearchParams hook as the main source of truth. 

The data source now only updates when you modify these query strings using the hook provided functions.

The main thing that seemed to be causing issues before was navigating to a detail page and then returning to the table. This seems to keep your previous settings pretty consistently now. Note that the keyword filtering seems a bit bugged but I believe its due to backend behavior unrelated to these changes. 

## 🔰 Checklist


dbarkowsky (commented):
Issues noted:
- For the quickSelectFilter, the ones that use booleans don't repopulate after refresh. The table is correct, just not the select menu.
- When a quickSelectFilter is set, editing a column filter doesn't set it back to default, even though they can't exist at the same time.

react-app/src/components/table/DataTable.tsx
```diff
@@ -516,11 +458,14 @@ export const FilterSearchDataGrid = (props: FilterSearchDataGridProps) => {
...
-                    pagination: { page: 0, pageSize: tableModel.pagination.pageSize },
-                  });
+                  //Clear search params related to filtering
+                  deleteSearchParamsKey([
+                    'keywordFilter',
+                    'quickSelectFilter',
+                    'columnFilterMode',
+                    'columnFilterName',
+                    'columnFilterValue',
+                  ]);
```
dbarkowsky, line 468: Should we clear the sort ones too?

react-app/src/components/property/PropertyTable.tsx
```diff
@@ -307,11 +307,11 @@ const PropertyTable = (props: IPropertyTable) => {
...
-          initialState={{
-            sorting: {
-              sortModel: [{ sort: 'desc', field: 'UpdatedOn' }],
-            },
-          }}
+          // initialState={{
+          //   sorting: {
+          //     sortModel: [{ sort: 'desc', field: 'UpdatedOn' }],
+          //   },
+          // }}
```
dbarkowsky, line 314: It's too bad this can't still be set within the table itself. I see it included in the routes, but if you don't come from the routes but also have no query params, will there be an initial state?

dbarkowsky (commented):
I still have an issue where the filters are saved and query called correctly, but the table model isn't updated after refreshing.
image
Apart from that, everything is working well.

LawrenceLau2020 (approved):
Works great, I can apply a filter to a table, view one of the records and go back to the main page while retaining the applied filter. Copying and pasting the url which includes a filter in a new tab also works.

dbarkowsky:
> I still have an issue where the filters are saved and query called correctly, but the table model isn't updated after refreshing.

This issue happens for me in Edge, but not in Firefox or Chrome.

GrahamS-Quartech:
Closing this, I don't think we need to pursue this further now that the other version of 1949 has been merged.

PR #2602, opened 2024-08-02
PIMS-1868 Follow Up

dbarkowsky:
## 🎯 Summary

  
[PIMS-1868](https://citz-imb.atlassian.net/browse/PIMS-1868?atlOrigin=eyJpIjoiNTc2Zjk0ODgxMDZhNDdkZmI4ZjgwNDFhMTg1YjA1M2QiLCJwIjoiaiJ9)

Following up on some comments about the ticket from Josh
## Changes
- Fixed the PID handling in the properties table when using the keyword filter search
- Added some more select field filter options
  - Josh wants the option to filter by projects in SPL, but I think that would require utilizing the workflows. Many statuses could fall into SPL territory, but we have no way of showing that in the frontend at the moment.

## 🔰 Checklist


LawrenceLau2020 (approved):
Looks good, I can edit a PID search in the map filter, and the quick filters on the tables work well. No issues found.

PR #2613, opened 2024-08-08
PIMS-1954 Workflow Update & Table Drops

dbarkowsky:
## 🎯 Summary

  
[PIMS-1954](https://citz-imb.atlassian.net/browse/PIMS-1954?atlOrigin=eyJpIjoiMDhmNWMzMWU3ODkzNDEyMGFkZTM0ZGU0Mjc1ZGIwODkiLCJwIjoiaiJ9)
This started as part of the movement to make project statuses groupable by workflow.
Recent changed made that unnecessary.

Now this PR removes old tables from the database and takes the workflow reference out of the projects.

## Changes
- Drop the tables ProjectNumber, ProjectReport, WorkflowProjectStatus, ProjectStatusTransition, Workflow.
- Remove WorkflowId FK from related tables, including Project.
- Remove any reference to Workflows in frontend and backend
- ProjectStatusHistory now only uses the integer as the primary key. It should insert repeated instances of a project leaving a status now, not just update the last one for that status.
- Fixes an issue on the ProjectDetails page that was giving me a crash while testing.

## Testing
The migrations will restore the old tables, but not the data within them. There is a risk of data loss if we decide we want these tables at some point. Locally, you can restore from a postgres dump file with a command like:
`docker exec -i postgres  psql -U postgres --single-transaction --set ON_ERROR_STOP=1 postgres < pims_dump`
The pims_dump file has to be in the `/data/db` folder in the attached volume in this example.
- Run the migrations
- Test creating, updating projects
- Inspect ProjectStatusHistory table to make sure it's saving as expected.

### Original

~PR aims to repurpose the Workflow table and WorkflowProjectStatus table. They were used in old PIMS but don't have an existing purpose in new PIMS.~

~Workflows are now seen as groupings of Project Statuses. It's a Many-to-Many relationship still.
Going to touch base with Josh on possible groupings that he would like to be able to select by. 
This should make queries for pre-set groups of statuses easier as well.~

## 🔰 Checklist


dbarkowsky:
See ticket for link to confluence board. Hoping Josh will decide on what groupings of statuses in workflows that he would like to filter by.

GrahamS-Quartech:
I seem to be getting a not-null constraint violation for workflow_id on saving to ProjectStatusHistory when updating project, probably need to either drop that column or just drop the constraint.

dbarkowsky:
Made some additional changes, which involved removing the Workflow entity entirely. It seems we may not need it for grouping or other business logic.

LawrenceLau2020 (approved):
Seems to be no conflicts with the migration even though this migration is shown as being before the legal land migrations that have already run. It still runs and completes with no errors.
image

Tested reverting migrations so that the migration history is the same in dev, this migration gets added as the last row:
image

Tested updating a project from various statuses, and it saves the status history nicely. No issues found.

PR #2616, opened 2024-08-09
PIMS-1928: Cancel and resend notifications

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1928](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1928)

* Adds new controllers and services for resending and cancelling notifications
* Adds new frontend hooks for resending and cancelling notifications
* Added buttons to the notification table on the project detail for resending or cancelling projects.

Note:
* The cancel button is greyed out for notifications that are already cancelled or completed, obviously cancelling them doesn't do anything.
* Resending a cancelled or completed notification does work, but the function to refresh notifications in getNotificationByProjectId is hardcoded to not update notifications in these statuses. Should we loosen that restriction? EDIT: This is still true, but I added a call to update just the single notification in the resend endpoint, so this change will be reflected in the frontend table.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/notifications/notificationsController.ts
+++ b/express-api/src/controllers/notifications/notificationsController.ts
@@ -67,7 +67,12 @@ export const resendNotificationById = async (req: Request, res: Response) => {
     }
   }
   const resultantNotification = await notificationServices.sendNotification(notification, kcUser);
-  return res.status(200).send(resultantNotification);
+  const user = await userServices.getUser(kcUser.preferred_username);
+  const updatedNotification = await notificationServices.updateNotificationStatus(
+    resultantNotification.Id,
+    user,
+  );
+  return res.status(200).send(updatedNotification);
 };
 
 export const cancelNotificationById = async (req: Request, res: Response) => {
--- a/express-api/src/controllers/notifications/notificationsController.ts
+++ b/express-api/src/controllers/notifications/notificationsController.ts
@@ -58,13 +58,8 @@ export const resendNotificationById = async (req: Request, res: Response) => {
     return res.status(404).send('Notification not found.');
   }
   const kcUser = req.user;
-  if (!(isAdmin(kcUser) || isAuditor(kcUser))) {
-    // get array of user's agencies
-    const usersAgencies = await userServices.getAgencies(kcUser.preferred_username);
-    const project = await projectServices.getProjectById(notification.ProjectId);
-    if (!usersAgencies.includes(project.AgencyId)) {
-      return res.status(403).send({ message: 'User is not authorized to access this endpoint.' });
-    }
+  if (!isAdmin(kcUser)) {
+    return res.status(403).send({ message: 'User is not authorized to access this endpoint.' });
   }
   const resultantNotification = await notificationServices.sendNotification(notification, kcUser);
   const user = await userServices.getUser(kcUser.preferred_username);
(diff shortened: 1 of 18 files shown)


express-api/src/controllers/notifications/notificationsController.ts
```diff
@@ -48,3 +50,49 @@ export const getNotificationsByProjectId = async (req: Request, res: Response) =
...
 };
+
+export const resendNotificationById = async (req: Request, res: Response) => {
+  const id = Number(req.params.id);
+  const notification = await notificationServices.getNotificationById(id);
+  if (!notification) {
+    return res.status(404).send('Notification not found.');
+  }
+  const kcUser = req.user;
+  if (!(isAdmin(kcUser) || isAuditor(kcUser))) {
```
dbarkowsky, line 61: If they are not an admin, should we even allow them to trigger this?

express-api/src/controllers/notifications/notificationsController.ts
```diff
@@ -48,3 +50,49 @@ export const getNotificationsByProjectId = async (req: Request, res: Response) =
...
+};
+
+export const cancelNotificationById = async (req: Request, res: Response) => {
+  const id = Number(req.params.id);
+  const notification = await notificationServices.getNotificationById(id);
+  if (!notification) {
+    return res.status(404).send('Notification not found.');
+  }
+  const kcUser = req.user;
+  if (!(isAdmin(kcUser) || isAuditor(kcUser))) {
```
dbarkowsky, line 85: Same question about admin-only logic.

express-api/src/routes/notificationsRouter.ts
```diff
@@ -9,12 +9,20 @@ export const DISPOSAL_API_ROUTE = '/projects/disposal';
...
+  .route(NOTIFICATION_QUEUE_ROUTE)
   .get(activeUserCheck, catchErrors(getNotificationsByProjectId));
 
+router
+  .route(`${NOTIFICATION_QUEUE_ROUTE}/:id`)
+  .put(activeUserCheck, catchErrors(resendNotificationById));
+
+router
+  .route(`${NOTIFICATION_QUEUE_ROUTE}/:id`)
+  .delete(activeUserCheck, catchErrors(cancelNotificationById));
```
dbarkowsky, line 26: This feels sort of misleading, as it doesn't delete the record, but I think this is probably the right way to go.

dbarkowsky (commented):
Getting this error regarding the ProjectNotificationsTable:
image

Trying to cancel a notification, it returns a status 400 with a NotificationQueue object.
image

The resend option is working as expected.

I wonder if we can't condense the buttons down to an icon. We're so tight for space to begin with.

express-api/src/routes/notificationsRouter.ts
GrahamS-Quartech: I feel similarly about using PUT for resend. The old implementation instead does a PUT to either /id/resend or /id/cancel but I think it's generally considered poor practice to make your endpoints verbs rather than just using the HTTP request type verbage.

GrahamS-Quartech:
> Getting this error regarding the ProjectNotificationsTable:

I don't believe this is introduced with this ticket but I will see about fixing it regardless.

> I wonder if we can't condense the buttons down to an icon. We're so tight for space to begin with.

I think this is a good idea, I'll choose some appropriate icons.

express-api/src/controllers/notifications/notificationsController.ts
GrahamS-Quartech: No idea, I wasn't sure how strict this should be so I just applied similar logic as elsewhere.

dbarkowsky (approved):
Buttons are working well for me now. Like the look of the icons chosen.

PR #2617, opened 2024-08-12
PIMS-1967: Create link to project from property page

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1967: ](https://citz-imb.atlassian.net/browse/PIMS-1967)

## Changes
- Added new controller, route, and service to find associated projects for a property
- New data card on the property details page

## Notes
- I can't seem to make the third column for the project description wrap around when this field contains text which goes beyond the width of the table.
image

- Waiting for some feedback from Josh as well, so some things may change.
## 🔰 Checklist

### DIFF
--- a/express-api/src/services/properties/propertiesServices.ts
+++ b/express-api/src/services/properties/propertiesServices.ts
@@ -162,7 +162,7 @@ const findLinkedProjectsForProperty = async (buildingId?: number, parcelId?: num
     .leftJoinAndSelect('pp.Project', 'p')
     .leftJoinAndSelect('p.Status', 'ps')
     .where(whereCondition)
-    .select(['p.*', 'ps.Name AS StatusName']);
+    .select(['p.*', 'ps.Name AS status_name']);
 
   const associatedProjects = buildingId || parcelId ? await query.getRawMany() : []; // Return an empty array if no ID is provided
 
--- a/express-api/src/services/properties/propertiesServices.ts
+++ b/express-api/src/services/properties/propertiesServices.ts
@@ -137,6 +137,33 @@ const propertiesFuzzySearch = async (keyword: string, limit?: number, agencyIds?
     Buildings: buildings,
   };
 };
+/**
+ * Finds associated projects based on the provided building ID or parcel ID.
+ *
+ * This function queries the `ProjectProperty` repository to find projects linked
+ * to either a building or a parcel. It returns an empty array if neither ID is provided.
+ *
+ * @param buildingId - Optional ID of the building to find associated projects for.
+ * @param parcelId - Optional ID of the parcel to find associated projects for.
+ * @returns A promise that resolves to an array of `ProjectProperty` objects.
+ *          If neither `buildingId` nor `parcelId` is provided, an empty array is returned.
+ */
+const findLinkedProjects = async (buildingId?: number, parcelId?: number) => {
+  const whereCondition = buildingId
+    ? { BuildingId: buildingId }
+    : parcelId
+      ? { ParcelId: parcelId }
+      : {}; // Return an empty condition if neither ID is provided
+
+  const associatedProjects =
+    buildingId || parcelId
+      ? await AppDataSource.getRepository(ProjectProperty).find({
+          where: whereCondition,
+        })
+      : []; // Return an empty array if no ID is provided
+
+  return associatedProjects;
+};
 
 /**
  * Retrieves properties based on the provided filter criteria to render map markers.
@@ -835,6 +862,7 @@ const propertyServices = {
   getImportResults,
   getPropertiesForExport,
   processFile,
+  findLinkedProjects,
 };
 
 export default propertyServices;
--- a/express-api/src/services/properties/propertiesServices.ts
+++ b/express-api/src/services/properties/propertiesServices.ts
@@ -148,21 +148,30 @@ const propertiesFuzzySearch = async (keyword: string, limit?: number, agencyIds?
  * @returns A promise that resolves to an array of `ProjectProperty` objects.
  *          If neither `buildingId` nor `parcelId` is provided, an empty array is returned.
  */
-const findLinkedProjects = async (buildingId?: number, parcelId?: number) => {
+const findLinkedProjectsForProperty = async (buildingId?: number, parcelId?: number) => {
   const whereCondition = buildingId
     ? { BuildingId: buildingId }
     : parcelId
       ? { ParcelId: parcelId }
       : {}; // Return an empty condition if neither ID is provided
 
+  const query = AppDataSource.getRepository(ProjectProperty)
+    .createQueryBuilder('pp')
+    .leftJoin('pp.Project', 'p') // Join the Project entity
+    .where(whereCondition) // Apply the condition
+    .select(['p.ProjectNumber', 'p.Id']) // Specify the fields to select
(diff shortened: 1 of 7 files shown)


react-app/src/hooks/api/usePropertiesApi.ts
```diff
@@ -179,6 +184,25 @@ const usePropertiesApi = (absoluteFetch: IFetch) => {
...
+    const queryParams = new URLSearchParams();
+    console.log('queryParams.toString():', queryParams.toString());
+    if (propertyId.buildingId !== undefined)
+      queryParams.append('buildingId', propertyId.buildingId.toString());
+    if (propertyId.parcelId !== undefined)
+      queryParams.append('parcelId', propertyId.parcelId.toString());
+    console.log('buildingId:', propertyId.buildingId);
+    console.log('parcelId:', propertyId.parcelId);
+
+    console.log('queryParams.toString():', queryParams.toString());
```
GrahamS-Quartech, line 198: Should remove these console.logs.

react-app/src/hooks/api/usePropertiesApi.ts
```diff
@@ -179,6 +184,25 @@ const usePropertiesApi = (absoluteFetch: IFetch) => {
     return parsedBody as ImportResult[];
   };
 
+  const getLinkedProjectsToProperty = async (propertyId: PropertyId) => {
+    console.log('propertyId', propertyId);
+    const queryParams = new URLSearchParams();
```
GrahamS-Quartech, line 189: You don't need to construct this here, the absoluteFetch.get takes params as its second argument.

GrahamS-Quartech:
The implementation for getting associated projects through the backend is functional so I'm not necessarily opposed, but I wonder if it could be better practice to implement this as part of the project filtering endpoint instead of a separate endpoint. Because that is what this operation entails really, it's enumerating projects by a specific property. It would just need to be implemented a bit differently from how the other filters are handled, since these properties are not directly available from the view. You could do something like this for example (untested):

```
const getPropertiesUnion = async (filter: PropertyUnionFilter) => {
  const options = collectFindOptions(filter);
  const query = AppDataSource.getRepository(PropertyUnion)
    .createQueryBuilder()
    .where(
      new Brackets((qb) => {
        options.forEach((option) => qb.orWhere(option));
      }),
    );

  /* ...omitted... */
   if (filter.hasParcels?.length) {
    const projectsHavingTheseParcels = await AppDataSource.getRepository(ProjectProperty).find({ where: {
      ParcelId: In(filter.hasParcels)
    }});
    query.andWhere('id IN (:...projectsParcels)', { projectsParcels: In(projectsHavingTheseParcels.map((p) => p.Id))})
  }
/* ...omitted... */
}

File: express-api/src/services/properties/propertiesServices.ts [hunk 3]
```diff
@@ -139,6 +139,40 @@ const propertiesFuzzySearch = async (keyword: string, limit?: number, agencyIds?
...
+    : parcelId
+      ? { ParcelId: parcelId }
+      : {}; // Return an empty condition if neither ID is provided
+
+  const query = AppDataSource.getRepository(ProjectProperty)
+    .createQueryBuilder('pp')
+    .leftJoinAndSelect('pp.Project', 'p')
+    .leftJoinAndSelect('p.Status', 'ps')
+    .where(whereCondition)
+    .select(['p.*', 'ps.Name AS StatusName']);
```
dbarkowsky, line 165: Noticed that this gets forced to lowercase in the return statement below. Might be worth just anticipating that here instead.
```suggestion
    .select(['p.*', 'ps.Name AS status_name']);
```

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -139,6 +139,40 @@ const propertiesFuzzySearch = async (keyword: string, limit?: number, agencyIds?
...
+
+  const associatedProjects = buildingId || parcelId ? await query.getRawMany() : []; // Return an empty array if no ID is provided
+
+  return associatedProjects.map((result) => ({
+    ProjectNumber: result.project_number,
+    Id: result.id,
+    StatusName: result.statusname,
```
dbarkowsky, line 172: Then this would change to
```suggestion
    StatusName: result.status_name,
```

react-app/src/components/property/PropertyDetail.tsx
```diff
@@ -82,6 +84,21 @@ const PropertyDetail = (props: IPropertyDetail) => {
...
+      if (parcelId) {
+        const projects = await api.properties.getLinkedProjectsToProperty({ parcelId: parcelId });
+        setLinkedProjects(projects);
+      } else if (buildingId) {
+        const projects = await api.properties.getLinkedProjectsToProperty({
+          buildingId: buildingId,
+        });
+        setLinkedProjects(projects);
+      }
+    };
```
dbarkowsky, line 98: If the parcelId or buildingId is undefined, is it possible to combine these statements?
Would suggest testing this locally before committing.
```suggestion
      const projects = await api.properties.getLinkedProjectsToProperty({ 
        parcelId,
        buildingId,
      });
      setLinkedProjects(projects);    
    };
```

dbarkowsky (commented):
Would be nice to shrink this down to the smallest height needed. A lot of extra white space right now.
image

GrahamS-Quartech:
> Would be nice to shrink this down to the smallest height needed. A lot of extra white space right now. 

In addition, I was gonna comment that this table could probably be styled more similarly to the other card tables like the financial section.

react-app/src/components/property/PropertyDetail.tsx
LawrenceLau2020: It is possible, but I had wrap api call in a try catch block, as without it won't return anything

LawrenceLau2020:
> if (filter.hasParcels?.length) {
>     const projectsHavingTheseParcels = await AppDataSource.getRepository(ProjectProperty).find({ where: {
>       ParcelId: In(filter.hasParcels)
>     }});
>     query.andWhere('id IN (:...projectsParcels)', { projectsParcels: In(projectsHavingTheseParcels.map((p) => p.Id))})
>   }

I'm not sure how this function would get called from the PropertyDetail form? Should this perhaps be a separate ticket?

GrahamS-Quartech:
> > if (filter.hasParcels?.length) {
> > const projectsHavingTheseParcels = await AppDataSource.getRepository(ProjectProperty).find({ where: {
> > ParcelId: In(filter.hasParcels)
> > }});
> > query.andWhere('id IN (:...projectsParcels)', { projectsParcels: In(projectsHavingTheseParcels.map((p) => p.Id))})
> > }
> 
> I'm not sure how this function would get called from the PropertyDetail form? Should this perhaps be a separate ticket?

This is meant to be an excerpt from the project service on the api.

react-app/src/components/property/AssociatedProjectsTable.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const rows: GridRowsProp = linkedProjects.map((project) => ({
+    id: project.Id,
+    ProjectNumber: project.ProjectNumber,
+    StatusName: project.StatusName,
+    Description: project.Description,
+  }));
+
+  return (
+    <div style={{ height: 400, width: '100%' }}>
```
dbarkowsky, line 56: I think this height is keeping it from shrinking.

dbarkowsky (approved):
Looks good and works well for me.

PR #2622, opened 2024-08-13
PIMS-1956: Single select filters

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1956](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1956)

Added single select style filters where applicable.
This applies to
Properties:
* Agency
* Administrative Areas
* Classification

Projects:
* Agency 
* Status

Administrative Areas:
* Regional District

Agencies:
* Parent Agency

Users:
* Status
* Agency
* Role

For users, I thought it could also make sense to apply it to provider, but I'm not sure we can enumerate the options reliably. 

Please try filtering each of the affected columns to test. Also try filtering using the preset options in the top right and make sure those still work with these changes.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/agencies/AgencyTable.tsx
+++ b/react-app/src/components/agencies/AgencyTable.tsx
@@ -1,4 +1,4 @@
-import React, { MutableRefObject, useContext, useState } from 'react';
+import React, { MutableRefObject, useContext, useMemo, useState } from 'react';
 import { CustomListSubheader, CustomMenuItem, FilterSearchDataGrid } from '../table/DataTable';
 import { Box, Chip, SxProps } from '@mui/material';
 import { GridApiCommunity } from '@mui/x-data-grid/internals';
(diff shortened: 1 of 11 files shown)


react-app/src/components/agencies/AgencyTable.tsx
```diff
@@ -60,6 +70,8 @@ const AgencyTable = (props: IAgencyTable) => {
       field: 'ParentName',
       headerName: 'Parent Agency',
       flex: 1,
+      type: 'singleSelect',
+      valueOptions: agenciesForLookup,
```
dbarkowsky, line 74: I think these should probably only be the parent agencies, not all agencies.

dbarkowsky (commented):
All working really well. The only other column this could maybe be applied to is the Property Type.

I think there is one situation where the select filters don't populate values as expected.
In properties, changing the filter to a broad search like "Core" works as expected for the results but doesn't populate the filter model as expected.
image

react-app/src/components/agencies/AgencyTable.tsx
GrahamS-Quartech: Good catch, fixed this.

GrahamS-Quartech:
> There is one situation where the select filters don't populate values as expected. In properties, changing the filter to a broad search like "Core" works as expected for the results but doesn't populate the filter model.

Fixed this, was just setting the value property slightly wrong.

GrahamS-Quartech:
> All working really well. The only other column this could maybe be applied to is the Property Type.

Also added this.

dbarkowsky (approved):

PR #2625, opened 2024-08-14
BulkUpload Fixes

TaylorFries:
## 🎯 Summary

  

list of changes: 

- Updated getClassificationOrThrow to match flow of other bulk upload functions.
- parcel: added:
  -  description -> '', 
  - isSensitive -> false, 
  - isVisibleToOtherAgencies -> true

  as variables that will use the value given in the bulk upload sheet if present, if not it will check if we are updating a parcel, if we are use the old value if we are adding a new parcel set the value to defaults (listed above)
- building: added:
  - description -> '', 
  - rentableArea -> 0, 
  - isSensitive -> false, 
  - isVisibleToOtherAgencies -> true, 
  - buildingFloorCount -> 0, 
  - tenancy -> '', 
  - totalArea -> 0
 
  as variables that will use the value given in the bulk upload sheet if present, if not it will check if we are updating a building, if we are use the old value if we are adding a new building set the value to defaults (listed above) 
- updated list of required/ optional headers 
- included check for all required headers to be present
- included check for multiple buildinfs found with name PID combo
- removed name field from parcel update/insert
- updated how IsSensitive and IsVisableToOtherAgencies is determined
- Removed test for active vs. non active statuses and all references to status in testing function for getClassificationOrThrow. 

Some things to note: 

- Will the people curating these sheets to be uploaded know the agencyCode?
- Assessed, AssessedYear, NetBook, FiscalYear can add new values but not update old ones. 
  - They also can only add one new value.

## Testing

try uploading a spreadsheet with bulk upload. Please check if all possible fields update correctly and new properties get added correctly. Bulk upload should be able to add new parcels or buildings as well as update any of the fields (excluding PID and name)
To test how the bulk upload handles matching building name/PID combo I used this query:

```
SELECT PID, name, count from 
  (SELECT PID, name, COUNT(*) as count
  from building
  group by PID, name) as T1
where count > 2;
``` 

On my local database to get a list of buildings that have name PID duplicates.  Then i created a test upload file for a building with that name and pid. 

## 🔰 Checklist


express-api/src/services/properties/propertiesServices.ts
```diff
@@ -411,33 +420,37 @@ const makeParcelUpsertObject = async (
...
     UpdatedById: existentParcel ? user.Id : undefined,
     UpdatedOn: existentParcel ? new Date() : undefined,
-    CreatedOn: existentParcel ? undefined : new Date(),
+    CreatedOn: existentParcel ? existentParcel.CreatedOn : new Date(),
     Location: {
       x: row.Longitude,
       y: row.Latitude,
     },
-    Address1: row.Address,
+    Address1: row.Address ?? existentParcel.Address1,
```
dbarkowsky, line 448: if address was not included, and the existentParcel is null, I think this will throw an error.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -499,31 +512,47 @@ const makeBuildingUpsertObject = async (
   const predominateUseId = getBuildingPredominateUseOrThrow(row, lookups.predominateUses);
   const adminAreaId = getAdministrativeAreaOrThrow(row, lookups.adminAreas);
 
+  const description = row.Description ?? (existentBuilding ? existentBuilding.Description : '');
+  const rentableArea = row.RentableArea ?? (existentBuilding ? existentBuilding.RentableArea : 0);
+  const isSensitive = setNewBool(row.IsSensitive, existentBuilding.IsSensitive, false);
+  const isVisibleToOtherAgencies = setNewBool(
+    row.IsVisibleToOtherAgencies,
+    existentBuilding.IsVisibleToOtherAgencies,
+    true,
```
dbarkowsky, line 521: Same comment at parcels. Only 7 with this true currently.

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -499,31 +512,47 @@ const makeBuildingUpsertObject = async (
...
-    IsSensitive: false,
-    IsVisibleToOtherAgencies: true,
-    PropertyTypeId: 0,
-    RentableArea: numberOrNull(row.RentableArea) ?? 0,
-    BuildingTenancy: row.Tenancy,
-    BuildingFloorCount: 0,
-    TotalArea: 0,
+    IsSensitive: isSensitive,
+    Description: description,
+    Address1: row.Address ?? existentBuilding.Address1,
```
dbarkowsky, line 548: Same comment as parcels. Possibly null.

react-app/src/pages/BulkUpload.tsx
```diff
@@ -214,16 +215,21 @@ const BulkUpload = () => {
...
                   'Description',
                   'Address',
                   'PIN',
                   'Assessed',
                   'AssessedYear',
                   'Netbook',
                   'FiscalYear',
-                  'RentableArea (for Buildings)',
+                  'IsSensitive',
+                  'isVisibleToOtherAgencies',
```
dbarkowsky, line 226: Should we capitalize this to match the other headers?
Looks like it's expecting capitalization in the backend as well.

dbarkowsky (commented):
A few comments. Theoretically looks good. Need to manually test still.

dbarkowsky:
A one thing noticed in testing:
- I think if the existant parcel is null, then the calls for `existentParcel.IsSensitive` will fail. Might need `existentParcel?.IsSensitive` instead.
image

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -571,6 +572,30 @@ export type BulkUploadRowResult = {
...
+    const row = sheetObj[rowNum];
+    if (row.PropertyType == 'Building') {
+      requiredHeaders.push('Name', 'PredominateUse', 'ConstructionType');
+      break;
+    }
+  }
+  for (let rowNum = 0; rowNum < requiredHeaders.length; rowNum++) {
+    if (!Object.keys(sheetObj[0]).includes(requiredHeaders[rowNum])) {
+      throw new ErrorWithCode('Please ensure all required headers are present', 400);
+    }
```
LawrenceLau2020, line 595: Should we display which headers are missing if some are not present in the sheet?

dbarkowsky: I added a specific message depending on which header is missing. Think that's good or do we show all at once?

dbarkowsky (approved):
I added the name field back in, but it should use the existing building's name first, otherwise use the name from the row. This way, it's not going to update the record's name field, but it should allow a building to be inserted with a name, which is what we'd expect given it's a mandatory field.

PR #2628, opened 2024-08-14
PIMS-1929: Unknown status for project notifications

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-1929: ](https://citz-imb.atlassian.net/browse/PIMS-1929)

## Changes
- Added 2 new notification statuses to handle 404, and 422 responses from the CHES api: https://ches.api.gov.bc.ca/api/v1/docs#tag/Message/operation/GetStatusQuery

- I can't seem to catch the 401 unauthorized response from CHES, this may need handling in the controller.

## Testing
- Since most of the notifications you will have in your localhost are from production, you can use the CHES dev credentials in your .env, then create a new project. To trigger a 404 not found, you can change one of the characters in the ChesMessageId column for the record in the NotificationQueue table where the notifications are stored.

- With existing projects from production, you should still see all the "completed" notifications, but the ones that were in a pending state will most likely display as "Not Found" if you are using the CHES dev credentials.

- Triggering a 422 is kind of tricky and I haven't found a way of doing it other than hard coding the request using an incorrectly formatted CHES Message id which is missing one character for example.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/notifications/notificationServices.ts
+++ b/express-api/src/services/notifications/notificationServices.ts
@@ -34,8 +34,6 @@ export enum NotificationStatus {
   Failed = 3,
   Completed = 4,
   NotFound = 5,
-  Invalid = 6,
-  Unauthorized = 7,
 }
 
 export enum NotificationAudience {
@@ -447,10 +445,6 @@ const convertChesStatusToNotificationStatus = (chesStatus: string): Notification
       return NotificationStatus.Completed;
     case '404':
       return NotificationStatus.NotFound;
-    case '422':
-      return NotificationStatus.Invalid;
-    case '401':
-      return NotificationStatus.Unauthorized;
     default:
       return null;
   }
--- a/express-api/src/services/notifications/notificationServices.ts
+++ b/express-api/src/services/notifications/notificationServices.ts
@@ -33,6 +33,9 @@ export enum NotificationStatus {
   Cancelled = 2,
   Failed = 3,
   Completed = 4,
+  NotFound = 5,
+  Invalid = 6,
+  Unauthorized = 7,
 }
 
 export enum NotificationAudience {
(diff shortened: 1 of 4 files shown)


express-api/src/services/notifications/notificationServices.ts
```diff
@@ -442,6 +445,12 @@ const convertChesStatusToNotificationStatus = (chesStatus: string): Notification
       return NotificationStatus.Failed;
     case 'completed':
       return NotificationStatus.Completed;
+    case '404':
+      return NotificationStatus.NotFound;
+    case '422':
+      return NotificationStatus.Invalid;
+    case '401':
+      return NotificationStatus.Unauthorized;
```
dbarkowsky, line 453: Do we need statuses for these? It seems like they describe the request, not the state of the message.

express-api/src/services/notifications/notificationServices.ts
```diff
@@ -481,7 +489,48 @@ const updateNotificationStatus = async (notificationId: number, user: User) => {
...
+        notification.Status = NotificationStatus.Unauthorized;
+        notification.UpdatedOn = new Date();
+        notification.UpdatedById = user.Id;
+        const updatedNotification = await query.manager.save(NotificationQueue, notification);
+        logger.error(
+          `Cannot authorize the request to the CHES server, check your CHES credentials.`,
+        );
+        query.release();
+        return updatedNotification;
+      }
```
dbarkowsky, line 524: Similar to the comment on the statuses, I don't think we should be changing the status of the notification if we get back a 422 or a 401. Logging it makes sense, but perhaps in that case we just shouldn't update the notification.

react-app/src/constants/chesNotificationStatus.ts
```diff
@@ -18,6 +20,10 @@ export const getStatusString = (status: number): string => {
       return 'Failed';
     case NotificationStatus.Completed:
       return 'Completed';
+    case NotificationStatus.NotFound:
+      return 'Not Found';
+    case NotificationStatus.Invalid:
+      return 'Invalid';
```
dbarkowsky, line 26: Same thought. Invalid doesn't seem like a description of the notification status.

dbarkowsky (approved):
Tested following the instructions in the PR.
Everything seems to work as expected. Messages set to Not Found if CHES gives 404 back.
Please merge after test coverage is added.

PR #2632, opened 2024-08-15
PIMS-1984: Fiscal year only takes year values.

GrahamS-Quartech:
## 🎯 Summary

  
[PIMS-1984](https://apps.itsm.gov.bc.ca/jira/browse/PIMS-1984)

Just adding a numeric prop to the text field could have solved this bug but it seemed like it may be better to just constrain this input a lot more in general. After all, there's really only two options that may be valid, max. So with that in mind I've tried to make this a single select style input instead. I also made it so that you can't add more rows once you've added current year and previous year, since past that point nothing else is valid.

## 🔰 Checklist


dbarkowsky (approved):
This is working well for me.
The maxRows is sometimes 1. I think that will be ok, but noting that we could up it if needed.

PR #2640, opened 2024-08-22
PIMS-2000: Make non-optional checkboxes mandatory

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-2000: ](https://citz-imb.atlassian.net/browse/PIMS-2000)

- Added the required option for the project tasks, now all the tasks in the "Confirm Tasks" section are required to be checked off before a user can submit the project.
- After the migration is run, the "Review Completed" task should be one of the required checkboxes that appear when changing a project status to either "Approved for ERP" or "Approved for Exemption"
- Also moved timestamp type, id#11 "RequestForSplReceivedOn" to the Approved for SPL Status and updated the task 'Preparation and Due Diligence complete' to a required field.

## 🔰 Checklist


dbarkowsky (commented):
Functionally, this is a good fix. 
Before we approve and merge, I wouldn't mind getting Josh's opinion on whether he wants to be restricted on all these fields.
For example, if we look at the list to get into Approved for ERP:
image
It feels like you wouldn't likely check off both Appraisal fields or both Consultation fields at the same time, but you can't save the project if they aren't.

dbarkowsky (commented):
- Contract is conditional is currently required, but that should be an optional task
- ~Noticing that the required field error message will stay between statuses, even if the new field is not required. I wonder if there's a way to force it to reset the error state.~ Solved this one. Change pushed.
image

PR #2641, opened 2024-08-26
PIMS-2002 Properties Export

dbarkowsky:
## 🎯 Summary

  
[PIMS-2002](https://citz-imb.atlassian.net/browse/PIMS-2002?atlOrigin=eyJpIjoiMmIxMTMxNDFlZWU0NGU5ZjhiZTM1MDRlMDE0YTAyOWEiLCJwIjoiaiJ9)
This is in response to the timeout issue sometimes happening when exporting the properties in a live environment.
Pretty sure the clause that slows things down is this:
`where: { Id: In(buildingIds) }`
When it is called getting parcels and buildings.
Suspect this is an issue with how TypeORM handles its queries. Almost looks like it's going a circular query at one point.

## Changes
- Broke up the relations and where properties in the TypeORM call into operations in code instead. For some reason this runs faster than letting TypeORM do the joins and filtering. These are the timings I got in local testing:
```
  // Getting evals, fiscals, and filtering separately: 850-1050ms
  // Getting as as part of joins, where clause: 1587-1672ms
```
- If the request for export data fails, it shows a snackbar message and resets the download icon.
- PID in export is now formatted with padded zeroes and hyphens
- IsSensitive column added to export

## Testing
It's hard to know if this will save time in OpenShift, where the issue seems to be.
Locally, you can test that the download works by exporting the properties table.
Test it doesn't work by stopping the API and hitting the download button again.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/properties/propertiesServices.ts
+++ b/express-api/src/services/properties/propertiesServices.ts
@@ -802,8 +802,6 @@ const getPropertiesForExport = async (filter: PropertyUnionFilter) => {
   // Use IDs from filtered properties to get those properites with joins
   const parcelQueryOptions: FindManyOptions<Parcel> = {
     relations: {
-      CreatedBy: true,
-      UpdatedBy: true,
       Evaluations: true,
       Fiscals: true,
     },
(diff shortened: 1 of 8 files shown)


LawrenceLau2020 (approved):
Seems to work well, downloaded all 11736 properties in about 1.5 seconds compared to 7.75 seconds using the TypeORM joins.

PR #2645, opened 2024-08-27
PIMS-1559: Update eslint from 8.57.0 to version 9.9.0 in express-api

Sharala-Perumal:
PIMS-1559: Update eslint from 8.57.0 to version 9.9.0 in express-api                          

This PR will create a new eslint.config.mjs file.
The old file eslintrc.js can be deleted after this PR is inherited.
## 🎯 Summary

  
[PIMS-1559](https://citz-imb.atlassian.net/browse/PIMS-1559)

## 🔰 Checklist


express-api/eslint.config.mjs
```diff
@@ -0,0 +1,32 @@
...
+    'no-console': 'error',
+    '@typescript-eslint/ban-ts-comment': 'off',
+    '@typescript-eslint/no-inferrable-types': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
+    '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/no-explicit-any': 'warn',
+    '@typescript-eslint/no-duplicate-enum-values': 'error',
+  },
+  // Ignore specific files and directories
+  ignores: ['node_modules/', 'package-lock.json', 'dist/', 'coverage/'],
```
dbarkowsky, line 29: I seem to get this warning from the coverage folder still. Not sure why it's not picking it up.
image

express-api/package.json
```diff
@@ -45,10 +45,12 @@
     "zod": "3.23.3"
   },
   "devDependencies": {
+    "@eslint/js": "9.9.1",
```
dbarkowsky, line 48: Are we still using this dependency? I don't see it imported.
Same with `@types/eslint__js` and the parent `typescript-eslint`

express-api/src/services/properties/propertiesServices.ts
```diff
@@ -818,7 +818,6 @@ const getPropertiesUnion = async (filter: PropertyUnionFilter) => {
 
   // Add quickfilter part
   if (filter.quickFilter) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dbarkowsky, line 821: Any idea why these are no longer being picked up as errors?
Perhaps this is not being defined in the configuration anymore.

express-api/eslint.config.mjs
Sharala-Perumal: > I seem to get this warning from the coverage folder still. Not sure why it's not picking it up. image

I removed the eslint-disable from sorter and block-navigation files in local environment, but it is not picked up as changes when I commit.

express-api/package.json
Sharala-Perumal: > Are we still using this dependency? I don't see it imported. Same with `@types/eslint__js` and the parent `typescript-eslint`

Good catch. We don't use these dependencies. I'll remove it.

express-api/eslint.config.mjs
dbarkowsky: Those files are autogenerated when you run tests and they shouldn't be committed. The config should be telling eslint to ignore files in those paths though.

express-api/src/services/properties/propertiesServices.ts
Sharala-Perumal: > Any idea why these are no longer being picked up as errors? Perhaps this is not being defined in the configuration anymore.

I have set the rule @typescript-eslint/no-explicit-any to 'warn' in eslint.config.js. This means ESLint will warn you whenever it detects the use of any, but it will not fail your build or force you to disable the rule with comments like eslint-disable-next-line.

PR #2656, opened 2024-09-04
PIMS-1859 PID Formatter Fix

dbarkowsky:
## 🎯 Summary

  
[PIMS-1859](https://citz-imb.atlassian.net/browse/PIMS-1859?atlOrigin=eyJpIjoiMDUwMzgxMjJmYWRjNGQ3ZWE0MGQwZTNjYjg2OWNhZmUiLCJwIjoiaiJ9)

## Changes
- Move some formatting to the onBlur prop
- Make sure any passed onBlur is also called
- Adjust onChange prop to accept full numeric PID or hyphenated PID

## Testing
Please try entering PID in many ways, some expected and unexpected.
Does it behave like you would expect?

## 🔰 Checklist


TaylorFries:
The changes work well, only thing to note is that when using PID filter on the map if you enter more than 9 numbers it errors out. maybe add a catch for that scenario?

dbarkowsky:
> The changes work well, only thing to note is that when using PID filter on the map if you enter more than 9 numbers it errors out. maybe add a catch for that scenario?

Good catch. I think there's some other behaviour that was unexpected as well...

dbarkowsky:
@TaylorFries Made a change that should prioritize the first 9 digits. Does this seem more intuitive?

PR #2675, opened 2024-09-13
PIMS-2027 BCSC Provider Options

dbarkowsky:
## 🎯 Summary

  
[PIMS-2027](https://citz-imb.atlassian.net/browse/PIMS-2027)

This PR adds the BCSC option whenever a provider is presented in the frontend.
The BCSC provider in the token is just the Keycloak integration client ID unfortunately, so it looks for that.

## Changes
- Added helper function to return Provider string
- Replaced individual checks with this helper function for consistency
- Added sso client id to config so we have something to check BCSC for
- Changed KeycloakUserId column from uuid type to varchar type. The id from Keycloak for BCSC is not a true uuid because it contains values outside [0-9a-f]. This would cause the database to reject the BCSC value when it was entered. Note, the KeycloakUserId field will not populate correctly still, because the sso-express package does not recognize BCSC as an option yet.

## Testing
- Log in with user who doesn't have access, check that provider on access form is correct
- Check provider column on user table
- Check provider area on user details

## 🔰 Checklist

### DIFF
--- a/express-api/src/constants/config.ts
+++ b/express-api/src/constants/config.ts
@@ -34,7 +34,7 @@ const config = {
   },
   keycloak: {
     client_id: process.env.SSO_CLIENT_ID,
-  }
+  },
 };
 
 const getConfig = () => {
--- a/express-api/src/constants/config.ts
+++ b/express-api/src/constants/config.ts
@@ -32,6 +32,9 @@ const config = {
     title: 'PIMS',
     uri: process.env.FRONTEND_URL,
   },
+  keycloak: {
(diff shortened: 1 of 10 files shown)


LawrenceLau2020 (approved):
Works great! I'm approving it, but I am seeing that the user's display_name, first_name, and last_name in the user table was stored in all caps.

dbarkowsky:
Note: Test user for BC Services Card has name info in all uppercase. Normal users shouldn't have this issue.

PR #2676, opened 2024-09-13
PIMS-2083 Remove Keycloak Roles Code

dbarkowsky:
## 🎯 Summary

  
[PIMS-2083](https://citz-imb.atlassian.net/browse/PIMS-2083)

This ticket removes code that handled anything related to updating Keycloak roles. 

This includes when a user logs in, is updated, when the cron job ran, etc. 

This will break the following until the auth checks are also updated to use the database instead:
- changing a user's role in PIMS will not have an effect on their permissions (it's still coming from Keycloak)
- changing a user's role in Keycloak will affect permissions, but the database value will not update

We should not push this past DEV until the rest of the work is done.

## 🔰 Checklist


LawrenceLau2020 (approved):
Works well, looks like it's no longer using the SSO Keycloak roles at all now. Changed user to a general user and it updated the role_id correctly in the database, also tested changing the role in Keycloak which has no effect as expected.

PR #2679, opened 2024-09-13
PIMS-2082 Backend Authorization Updates

dbarkowsky:
## 🎯 Summary

  
[PIMS-2082](https://citz-imb.atlassian.net/browse/PIMS-2082)

## Goal
- Ensure there is a consistent way to check the database for user role-based permissions
- Remove role checks that used the Keycloak roles (found in the req.user object)

## Changes
- Expanded upon the old `activeUserCheck` middleware. Renamed it to `userAuthCheck`. It now performs role checking functions as well. It also contains a function called `hasOneOfRoles` that you can use for role checking in other API locations. It then adds the user from the database plus this function to the request. It can be accessed liked `req.pimsUser`. 
- This `userAuthCheck` is applied in the paths to the routes. It must be proceeded by the `protectedRoute` middleware or it will not have the user.preferred_username field to look up the database user. In cases where all the routes in an area (like agencies) were Admin protected, I added it to the `express.ts` file, but if there were different permissions for each route in an area, I added it to the route files. Open to preference on this if we wanted it to be a little more repetitive in some cases but keep all these checks in the route files. 
- Replaced any locations where we were using the Keycloak roles for authorization checking. 
- Replaced unneeded request user lookups. Now that the user's info is passed down with the request, there were a lot of getUser calls that could be omitted.
- Updated tests to match this new system

## Use Example
```ts
router
  .route('/myRoute')
  .get(userAuthCheck(), catchErrors(myController))  // Active user required, any role
  .post(userAuthCheck({ requiredRoles: [Roles.ADMIN] }), catchErrors(myOtherController); // Same, but Admin role required.
``` 

## Testing
With your user already in the database:
- Try sending requests to a few different endpoints with different roles. You have to change the roles in the database now. Changing them in Keycloak will not have an effect on the backend checks. 
- Can use the react-app to view pages and see if you are locked out from certain requests. This can be a little odd, as the react-app is still using checks for Keycloak roles.  

## 🔰 Checklist


LawrenceLau2020 (approved):
Seems to work, after changing my role to general user, I'm only able to see properties under my agency and below, tested updating my user and I get a 403 error, and tested viewing another building and get a 403 and redirected to the map page.
Tested as an auditor as well, and the api is preventing me from making changes as expected.

PR #2700, opened 2024-09-24
PIMS-2117 Auth Header Bug

dbarkowsky:
## 🎯 Summary

  
[PIMS-2117](https://citz-imb.atlassian.net/browse/PIMS-2117)

## Changes
- Make the config context into a static import. Didn't need to be a context.
- Fix issue where access request provider was potentially uncontrolled.
- Update sso-express package to version with BCSC support.
- Update sso-react package to 1.0.0-alpha7.
- Separate sso information from AuthContext and rename context to UserContext. It only provides the PIMS user now. Use the `useSSO()` hook from the sso-react package for keycloak information.

## Issues
- Refreshing the page with sso-react 1.0.1 and beyond causes the auth header to be `Bearer undefined`. Not sure why this is happening. I've commented out a bit of code that retries requests on a 401 in order to make this easier to test. Restore before merge. It seems like the state in the sso-react provider is not set when `getAuthorizationHeaderValue` is called. Maybe we could delay our own calls until it is, but this should probably be fixed in the sso-react package itself.

## 🔰 Checklist


dbarkowsky:
I ended up trying a lot of things to make the sso-react@1.1.0-beta to work. I cannot get around the issue where the auth header is undefined after a refresh. It does not seem to come back unless you log out and log in again. Things I tried to make this work:

- Setting a delay/promise in our useFetch hook to try and wait for the header to be populated
- Altering the get self and lookup all calls to refresh if the sso state changes. It never does.
- Making getAuthorizationHeaderValue from the sso-react package asynchronous, so it would wait until there's a value to get rather than returning `Bearer undefined`. I still think there's a race condition between it refreshing your token on a refresh of the page and constructing the access token state.
- A bunch of other things I can't remember. 

At this point, I'm recommending that we stay on version 1.0.0-alpha7. It's the last-known working version, and as long as the backend is updated, it seems like we still save all the user information properly.

TaylorFries (approved):
changes work as expected. Rollback to alpha7 sso-react package resolves the refresh issue. N?o issues detected.
Should we open another branch once this is closed to link to https://github.com/bcgov/citz-imb-sso-react/issues/169?

LawrenceLau2020 (approved):
Page refreshing seems to be working with these updated packages. Tested with IDIR as well as a BCSC user, refreshed page and no issues detected.

dbarkowsky:
> changes work as expected. Rollback to alpha7 sso-react package resolves the refresh issue. N?o issues detected. Should we open another branch once this is closed to link to [bcgov/citz-imb-sso-react#169](https://github.com/bcgov/citz-imb-sso-react/issues/169)?

I'll make another ticket for the backlog as a "lookout for this"-kind of task. When the framework agnostic version eventually comes out, we may want to move to that.

PR #2716, opened 2024-10-03
PIMS-305 Spatial Analysis

dbarkowsky:
## 🎯 Summary

  
[PIMS-305](https://citz-imb.atlassian.net/browse/PIMS-305)
 WIP

## 🔰 Checklist


dbarkowsky:
Noting an issue with the ray-casting algorithm used to determine if points are inside polygons. It sometimes returns points to the left of polygons as well, and I think it's an issue with floating point math and lost precision.

dbarkowsky:
Winding-number algorithm seems to work better. Process roughly works like this:
image
The resulting angle for this picture would be 1.249 radians.

Repeating that for each pair of points in the polygon will result in a number. I believe the number will be roughly 2𝜋 for most shapes. If it's near zero, it's not in the polygon.

It will work for complex shapes, but not for shapes that cross over themselves.
Good example: 
image

Bad example:
image

This doesn't seem like a huge limitation if users are aware.

LawrenceLau2020 (approved):
Works great! I can draw multiple shapes to filter properties as well as combining the map filter to further filter the properties. Deleting shapes to remove the properties that were selected also works.

PR #2721, opened 2024-10-09
PIMS-632 Playwright

dbarkowsky:
## 🎯 Summary

  
[PIMS-632](https://citz-imb.atlassian.net/browse/PIMS-632)

This ticket aims to add Playwright as an option for end-to-end (e2e) testing.
It is not a complete testing suite for all the app's functions, but it provides a baseline we can add to over time.

If we want to test adding/editing/deleting in the future, it will be necessary to restore or stage the database in some way so that tests are idempotent. 

## Changes
- Added e2e folder that houses Playwright code
- Added README with usage instructions
- Added some tests for the following: 
  - logging in
  - redirecting to no-role/inactive user pages
  - using map and filtering
  - navigating to agencies table and filtering

## Testing
- Follow instructions in README in `/e2e` folder for installation.
- Test the various playwright commands listed there.

I've tried to combat flakey tests, but some of them (like the map filter) are really sensitive to timings.

## 🔰 Checklist


e2e/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
```
TaylorFries, line 5: add .env to gitignore?
```suggestion
node_modules/
/test-results/
/playwright-report/
/blob-report/
/playwright/.cache/
*.env
```

TaylorFries:
tests run well and pass when running in ui (`npx playwright test --ui`) but when running in headless mode (`npx playwright test`) some tests are failing at random.

e2e/.gitignore
dbarkowsky: Sure. Should be covered by the parent gitignore, but always good to be safe.

TaylorFries:
tests in headless mode now passing consistently (5 consecutive runs.) Had to restart the API every 2 tests to avoid hitting rate limit.

TaylorFries (approved):
tests run as expected. Aside from occasional timeouts and tests failing due to checking elements before they have fully rendered/ finished animations. See above comments for more information

dbarkowsky:
Going to merge this for now so I can work on a workflow that uses it. Need it in DEV first.

PR #2722, opened 2024-10-11
PIMS-2134 Export Tables with Dates

dbarkowsky:
## 🎯 Summary

  
[PIMS-2134](https://citz-imb.atlassian.net/browse/PIMS-2134)

Request was to have the dates in exported spreadsheet be just the date format, not the ISO date strings we store.

## Changes
- Moved from 'node-xlsx' to the `xlsx` (really sheetjs) package we use in the backend too. This allowed for date formatting.
- Created a function to handle and convert the date strings as needed for export
- Reworked the export from tables to use the `xlsx` process for making and downloading spreadsheet. This way we can specify that fields are actually configured as dates that excel will just recognize.
- Updated the export from bulk upload results to use the new package as well.

## Testing
If you need to seed the bulk upload results but can't legitimately do so because of the upload bug, try uploading a file, let it fail, then update that row with this SQL:
```sql
UPDATE import_result set results = '[{"rowNumber": 1, "action": "uploaded", "reason":"good job"}]', message = 'wow', completion_percentage = 1 where id = 1;
```

- Download spreadsheet from bulk upload results
- Download spreadsheets from other tables and ensure that dates are formatted correctly

## 🔰 Checklist

### DIFF
--- a/express-api/.gitignore
+++ b/express-api/.gitignore
@@ -1,1 +1,2 @@
 /dist
+/uploads
--- a/react-app/package.json
+++ b/react-app/package.json
@@ -27,7 +27,6 @@
     "@mui/x-date-pickers": "7.18.0",
     "@turf/turf": "7.1.0",
     "dayjs": "1.11.10",
-    "node-xlsx": "0.24.0",
     "react": "18.3.1",
     "react-dom": "18.3.1",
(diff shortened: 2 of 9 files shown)


TaylorFries (approved):
table downloads now format date columns as columns. no other issues detected.

PR #2723, opened 2024-10-11
PIMS-2135 Bulk Upload Worker Bug

dbarkowsky:
## 🎯 Summary

  
[PIMS-2135](https://citz-imb.atlassian.net/browse/PIMS-2135)

There was an issue with the bulk upload that was caused when moving the authentication checks out of Keycloak.
Two problems:
- The user passed into the worker now had the `hasOneOfRoles` function attached. Workers apparently cannot pass functions, promises, steams, etc.
- The role was initially sending the name, but later checks expected the Id.

## Changes
- Formatted a date on the Bulk Upload page to make it more human readable.
- Made sure that the Role and Agency properties of the User records are marked as potentially undefined (when you get the record without doing the JOIN)
- Updated the user object to remove the `hasOneOfRoles` function before it is passed into the worker. It's not used further in this process, and the type is marked only as User to reflect that.
- Corrected how the role is passed to the worker.

## Testing
- Upload a file to bulk upload an make sure it works.

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/properties/propertiesController.ts
+++ b/express-api/src/controllers/properties/propertiesController.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { Request, Response } from 'express';
 import propertyServices from '@/services/properties/propertiesServices';
 import {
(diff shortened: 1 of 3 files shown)


TaylorFries (approved):
Updates to process fix the error that was occurring. 
One side note to devs (made in Teams chat) to ensure their import_result table is cleaned before attempting to use the bulk upload feature. 
No new issues detected.

PR #2730, opened 2024-10-21
CHES Notifications & Logger Adjustment

dbarkowsky:
## 🎯 Summary

  
This is in response to the issue I had this morning with sending/resending notifications.
It turned out that my email in the database was not a real email from testing another ticket. This caused the CHES request to fail, but it wasn't very clear why it was doing so.

## Changes
- The logger now reports the stack if you log something that is of type Error. 
  - e.g. Will show stack: `logger.error(new Error('something went wrong'));`
  - e.g. Won't show stack: `logger.error('this is not type error');`
- If the API is not in its containerized form, it will also pretty print the logs for easier readability
  - e.g. image

- If the response is bad in some way or empty, sendAsync should now return null, which just marks attempted notifications as Failed.

## Testing
- Manually set up the cause for an error. Easiest way is to set your email to a string like `blank`.
- Trigger sending notifications through the resend option, submitting projects, changing statuses. 
- Check that log prints and notification is saved as Failed.

## 🔰 Checklist


TaylorFries (approved):
Failed email log printing as expected, and notifications set to failed in project page.

LawrenceLau2020 (approved):

PR #2733, opened 2024-10-22
PIMS-2036: Map Filter - ERP Projects

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-2036](https://citz-imb.atlassian.net/browse/PIMS-2036)

## Changes
- Added a new ToggleSwitch on the map page to display properties that are in ERP projects.

## Testing
- To test the filter, when selected it should only display properties that are in ERP.
- You can also test adding additional filters when this is selected as well, eg. Land or Building, Agency, etc.
- You will probably need to double check the number of properties in each project that is in "Approved for ERP" status just to make sure the results match.

## 🔰 Checklist


TaylorFries (approved):
toggle works well and filters map as expected (both as admin and general user) as well as working well with other filter options (even the polygon drawing!)

dbarkowsky (approved):
Looks good and works well too.

PR #2734, opened 2024-10-22
PIMS-1869 BCA XREF

dbarkowsky:
## 🎯 Summary

  
[PIMS-1869](https://citz-imb.atlassian.net/browse/PIMS-1869)

## Changes
- Created table to host XREF data
- Added controller and route to API to access this table
- Updated the ParcelPopup component to select the BCA data depending on the return value from this table.
- Tests and Documentation

## Testing
This is tough to test without the BC Assessment data in local, but it's possible to test individual parts with some substitutions.

First you have to run the migration then load the XREF data.
Get the data from here:
https://catalogue.data.gov.bc.ca/dataset/parcelmap-bc-jurisdiction-roll-number-pid-cross-reference-table
You will have to log in to see it.
I loaded the data with dbeaver. There's an option by right-clicking a table to import data > from CSV
Make sure you only import the three columns that are needed. You'll need to map them to the correct columns or additional columns with the CSV names will be created.

API
- Can test with Thunderclient or Postman. Use a GET request like this: `localhost:5000/v2/tools/jur-roll-xref?pid=206814`

Frontend
- You can test this by substituting the BC Assessment data with some static dummy data. Just make sure that the location you click has a PID that will correspond to an existing jurisdiction and roll number in the dummy data. 
- I just copied the returned object from PROD for a strata, then made sure that the useBCAssessment hook in the frontend returned that data instead of making the BCA call. 
- After that, click on the same strata as you took the data from in PROD

## 🔰 Checklist

### DIFF
--- a/express-api/src/controllers/tools/toolsController.ts
+++ b/express-api/src/controllers/tools/toolsController.ts
@@ -94,7 +94,7 @@ export const searchGeocoderAddresses = async (req: Request, res: Response) => {
 };
 
 export const getJurisdictionRollNumberByPid = async (req: Request, res: Response) => {
-  const pidQuery = String(req.query.pid);
+  const pidQuery = req.query.pid as string;
   if (parseInt(pidQuery)) {
     const result = await AppDataSource.getRepository(JurRollPidXref).findOne({
       where: { PID: parseInt(pidQuery) },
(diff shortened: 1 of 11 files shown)


LawrenceLau2020 (approved):
Tested using the strata PID: 000-899-275
checked a few properties and they seem to match with the online BCA data.
Works really well.

PR #2781, opened 2024-10-31
PIMS-2106: Address WAVA Scan Findings

LawrenceLau2020:
## 🎯 Summary

  
[PIMS-2106](https://citz-imb.atlassian.net/browse/PIMS-2106)
It seems that some of the findings of the WAVA report such as "Cookie with Insecure or Improper or Missing SameSite attribute" is out of our control, as these cookies are being sent by the service sending back the response (keycloak, siteminder, etc.)

## Changes
- Added security headers based on an existing repo mentioned by Garry in the Jira ticket.
- Modified the content security policy somewhat to allow for requests to unpkg.com, cdnjs.cloudflare so that the map displays correctly.
- Also removed "'Access-Control-Allow-Origin', '*'" as this can potentially be a security issue, please test to see nothing broke.
- This is a preliminary change. Feedback and additional suggestions are welcome!

## Testing
- To test you will need to run the react-app and the express-api as docker containers
`docker-compose build pims-app-v2 ` and `docker-compose build pims-api-v2 `
and then to run the containers: `docker-compose up pims-app-v2 -d`

- Once the application is running, you'll need to check the console for any errors, and also make sure everything is working as expected. I did do some testing with the map filter and creating of a project, which seemed to work. You may need to edit the content security policy if some url was missed (eg. an image not loading when it should be)
- This is kind of hard to test, other than making sure everything still works and then merging to dev to let the OWASP scan run in Openshift, but if you have time and want to try you can always try installing the opensource OWASP ZAP scanning tool, running it as a docker container, and follow the instructions mentioned from the documentation here: https://www.zaproxy.org/docs/docker/about/
   - The command I used to scan on a Windows machine was this which did a "baseline" scan: `docker run -v C:\zap:/zap/wrk:rw --network="host" ghcr.io/zaproxy/zaproxy:stable bash -c "zap-baseline.py -t http://host.docker.internal:3000 -r /zap/wrk/zap-scan-report.html && sleep 10"`
   - I believe the scan in openshift does either a "full" scan: https://www.zaproxy.org/docs/docker/full-scan/ or an "api" scan: https://www.zaproxy.org/docs/docker/api-scan/
- Also, note that currently we are not doing an "in depth" application scan by logging in and testing each of the pages which would involve adding some authentication to the existing OWASP scan and more configuration.

## 🔰 Checklist

### DIFF
--- a/react-app/nginx.conf
+++ b/react-app/nginx.conf
@@ -19,7 +19,7 @@ server {
   add_header Pragma "no-cache";
 
   # Content Security Policy
-  add_header Content-Security-Policy "default-src 'none'; frame-src 'self' *.gov.bc.ca; script-src 'nonce-windowsp' 'self' *.gov.bc.ca https://maps.googleapis.com https://unpkg.com; style-src 'self' 'unsafe-inline' https://unpkg.com https://cdnjs.cloudflare.com; font-src 'self' *.gov.bc.ca; img-src 'self' *.gov.bc.ca data: tile.openstreetmap.org https://maps.googleapis.com https://unpkg.com https://cdnjs.cloudflare.com https://openmaps.gov.bc.ca; connect-src 'self' *.gov.bc.ca https://openmaps.gov.bc.ca; manifest-src 'self';";
+  add_header Content-Security-Policy "default-src 'none'; frame-src 'self' *.gov.bc.ca; script-src 'nonce-windowsp' 'self' *.gov.bc.ca https://unpkg.com; style-src 'self' 'unsafe-inline' https://unpkg.com https://cdnjs.cloudflare.com; font-src 'self' *.gov.bc.ca; img-src 'self' *.gov.bc.ca data: tile.openstreetmap.org https://maps.googleapis.com https://unpkg.com https://cdnjs.cloudflare.com https://openmaps.gov.bc.ca; connect-src 'self' *.gov.bc.ca https://openmaps.gov.bc.ca; manifest-src 'self';";
 
   listen 3000;
   location / {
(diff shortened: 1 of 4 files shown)


react-app/nginx.conf
```diff
@@ -1,4 +1,26 @@
...
+  add_header X-Frame-Options "DENY";
+
+  add_header Permissions-Policy "camera=(), microphone=(), geolocation=(), fullscreen=()";
+
+  # Turn off all caddy caching
+  add_header Cache-Control "no-cache,no-store,must-revalidate";
+  add_header Pragma "no-cache";
+
+  # Content Security Policy
+  add_header Content-Security-Policy "default-src 'none'; frame-src 'self' *.gov.bc.ca; script-src 'nonce-windowsp' 'self' *.gov.bc.ca https://maps.googleapis.com https://unpkg.com; style-src 'self' 'unsafe-inline' https://unpkg.com https://cdnjs.cloudflare.com; font-src 'self' *.gov.bc.ca; img-src 'self' *.gov.bc.ca data: tile.openstreetmap.org https://maps.googleapis.com https://unpkg.com https://cdnjs.cloudflare.com https://openmaps.gov.bc.ca; connect-src 'self' *.gov.bc.ca https://openmaps.gov.bc.ca; manifest-src 'self';";
```
dbarkowsky, line 22: Where is the maps.googleapis.com used?

react-app/nginx.conf
```diff
@@ -1,4 +1,26 @@
 server {
+  # Enable HTTP Strict Transport Security (HSTS) to force clients to always
+  # connect via HTTPS (do not use if only testing)
+  add_header Strict-Transport-Security "max-age=31536000;";
```
dbarkowsky, line 4: Not sure how this still works with localhost. Doesn't seem to impede it, but it's not https there.

react-app/nginx.conf
LawrenceLau2020: Good catch, that was probably copied and pasted from the other repo.

dbarkowsky (approved):
Everything seems to work in local. Let's see how it affects DEV.

PR #2884, opened 2024-12-18
PIMS-2254 Late Notification Warning

dbarkowsky:
## 🎯 Summary

  
[PIMS-2254](https://citz-imb.atlassian.net/browse/PIMS-2254)

## Changes
- Re-introduced [`node-cron`](https://www.npmjs.com/package/node-cron) package to run this job on a schedule. We had previously removed it when we moved away from Keycloak-roles.
- Created function to check for late emails, update their status, and send a warning email if there are still Pending notifications
- Fixed a bug with the fuzzy search that only occurs when you have no parcels/buildings to search from (probably only a developer issue)
- Standardized the sending email for system messages to `PIMS System <system.pims@gov.bc.ca>` between the error report and notification warning emails.

## Testing
Make sure your .env file has these set:
```env
CHES_EMAIL_ENABLED=true
CHES_OVERRIDE_TO="yourEmailHere;"
```

- Set some existing notifications that were sent from the dev CHES server with past dates to Pending. For example:
```sql
UPDATE notification_queue set status = 1 where id in(4, 5, 9);
```
- To trigger this function, change the cron schedule in the `server.ts` file to `* * * * *` . This should make it fire every minute.
- The altered notifications should be updated back to their correct status.

To see a copy of the email, set some existing notifications to Pending like noted above, but comment out lines 43-47 of `failedEmailCheck.ts`. This will keep those notifications as Pending, so the email will be triggered.

## 🔰 Checklist


Sharala-Perumal (approved):
The new function does check for the past pending notifications, sends them to queue and update the status for completion. Tested to receive email notification on projects with pending statuses past their send date. 
This ticket can be merged to main when the promise call in Failedemailcheck to log error messages is altered.

dbarkowsky:
Update call has been edited so that any failed updates will not interfere with others.
They will still log errors from the CHES service and from this function.

PR #2886, opened 2024-12-19
PIMS-2258 Adjust when property agency changes are restricted

dbarkowsky:
## 🎯 Summary

  
[PIMS-2258](https://citz-imb.atlassian.net/browse/PIMS-2258)

## Changes
- Added some comments
- Added a condition to the if statement that controls if an agency change is denied for properties
  - It will now check that an agency was included in the request before checking if a user has permissions for that agency

## Testing
- As a General User, try to add/update assessment values for a property in your agencies.

## 🔰 Checklist


Sharala-Perumal (approved):
As a general user, I can now update or add new assessment values to the properties within my agency.

PR #2895, opened 2025-01-09
PIMS-2257 Export Properties from Map Page

dbarkowsky:
## 🎯 Summary

  
[PIMS-2257](https://citz-imb.atlassian.net/browse/PIMS-2257)

## Changes
- Added an export button on the map sidebar. It should download a file that contains all properties currently selected by an active filter (or all of them if no filter).
- Consolidated the Excel property map into its own file because it is now used in two different places
- Added a new frontend API hook and API route. Even though it uses the same controller as the map properties search, it is easier to represent this in documentation.
- Modified the propertiesController to return export-ready data if ExcelExport query param is detected
- Modified the propertiesService so that the export function can take an object with lists of parcel and buildings ids. These lists are the final filter to the property export if provided.

## Testing
- Test the map to make sure normal map functionality still works
- Test the export button on the map. Do you only get properties filtered by the map?

## 🔰 Checklist


Sharala-Perumal (approved):
The map functionality and filters work as expected. The export to Excel feature retrieves all properties when no filter is applied, and only the required properties when a filter is applied.

PR #2906, opened 2025-01-27
PIMS-2276 Access Request Name Case

dbarkowsky:
## 🎯 Summary

  
[PIMS-2276](https://citz-imb.atlassian.net/browse/PIMS-2276)

This is to fix the instance with BC Services Card where all letters in a user's name are capitalized. 

## Changes
- Added a function that capitalized the first letter of each word in a name
- Applied that function to the access request page

## Testing
- Log in with an account that doesn't already have a user
- The PIMS BCeID account works for this well, because it's mostly capitals.
- See that the capitalization has been updated.

## 🔰 Checklist


Sharala-Perumal (approved):
Tried logging in using BCeid and BC Services card test accounts and I see the Fully capitalized names converted into Init Caps.

PR #2913, opened 2025-02-06
PIMS-2282 Issue with Notification Cancellation Status

dbarkowsky:
## 🎯 Summary

  
[PIMS-2282](https://citz-imb.atlassian.net/browse/PIMS-2282)

## Changes
- Added a check to make sure a response was received from CHES before marking a notification as cancelled on our side. I suspect that was the issue that led to PIMS-2282 being reported.
- After talking with Christy, we noted that the Agency Response table doesn't populate, but it does send emails.
  - She was having to manually add and unsubscribe agencies at their request.
  - The table should now be populated with agencies that have emails and are set to receive email notifications. She can then opt out agencies when they request them.
- Noticed that there's a possibility for emails to try and send if the email field is blank but the agency is set to receive emails. Added some additional WHERE clauses to prevent that.

## Testing
- Open a new project. Note that there should be agencies in the response table now.
- Change project to Approved for ERP. Emails should be sent out to the owning agency, parent ministries, and other subscribed ministries.

## 🔰 Checklist


Sharala-Perumal (approved):
The agency interest table  seems to be populated with the new change. The notification is et to Cancelled when the agency is unsubscribed or deleted from the agency Interest table after receiving response from CHES.

PR #2915, opened 2025-02-10
PIMS-2275 Notifications Update on Agency Email Change

dbarkowsky:
## 🎯 Summary

  
[PIMS-2275](https://citz-imb.atlassian.net/browse/PIMS-2275)

## Changes
- Added a check when updating email settings of an agency
- If the email addresses are changed, it should cancel pending notifications for that agency and requeue them with the new emails.
- If the Send Notifications toggle is changed, it either queues new notifications for ERP projects for that agency or cancels them depending on the new Send Notifications value.
- Added functions used for queuing outstanding notifications for a project/agency combo.

### Changes Since Original Review
- Separated how the agency responses are updated. It now has its own route and some service to facilitate this. Watch notifications are modified (queued or cancelled) when these responses change, but the updateProject route/service no longer tries to do everything. It will only send relevant notifications according to what is there in the agency response table.
- Prevented users from having Send Email checked while having nothing in the To field.
- Made it so you can only add agencies to the watch list for a project if that agency has an email and send notifications checked. This prevents subscribed responses when there's no one to send the notifications to.
- Fixed a bug where some agencies weren't being queued, despite being in the watching list. The issue was with agencies that were not also parent agencies. 
- Updated a lot of test coverage.

## Testing
- Create a test agency or use an existing one. Better if it's not part of the agency your user is in.
- Create a new project and move it in to Approved for ERP status. Depending on the test agency's settings, emails may be triggered.
- Experiment with changing the agency's email addresses and toggling off/on the Send Notifications option.
- You should see notifications for this agency-project combo to be queued/cancelled depending on your selections.

## 🔰 Checklist


Sharala-Perumal (approved):
Created a new agency and enabled subscription to verify that notifications are received successfully. Also tested the unsubscribe functionality to ensure that pending notifications are cancelled appropriately . Switching the email address worked as expected, with queued notifications being transferred to the new email address. Additionally, toggled the "Send email" setting between "Yes" and "No" to confirm that new notifications are created and cancelled accordingly.

PR #2944, opened 2025-04-23
Update compose file and git-ignore

dbarkowsky:
## 🎯 Summary

  

Not related to a ticket. These are just the changes that I had to make to have the compose file work with Podman instead of Docker. 
I believe it should still be compatible with the docker-engine. 

## Changes
- Add database folder to gitignore. I don't see how this wasn't an issue before. Maybe Docker handled it automatically.
- Few auto-formatting changes to the compose, such as quotation marks.
- Move postgres environment ENVs from inline to env file.
- Specify where postgres saves its data with the PGDATA environment variable. For some reason, this version of postgres was not saving it in the volume.

## Testing
1. Make sure your .env is still populated. No change needed if PIMS is already running.
2. Try building the containers with your choice of software (Docker, Podman, etc.)

`docker-compose up --build -d` for example.

3. If you bring down all the containers and bring them up again, your data should persist. This wasn't an issue before, but for some reason Podman would take down my volumes when I ran `podman-compose down`

## 🔰 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ package-lock.json
 
 # .csv and txt files created by script
 tools/propertyCheck/data/*
+
+# docker volumes
+database/
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,4 @@ package-lock.json
 tools/propertyCheck/data/*
 
 # docker volumes
-database/
+database/postgres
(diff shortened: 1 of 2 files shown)


Sharala-Perumal (approved):
Refined compose file works fine. The named volume and local mount for dump file works as expected

PR #2946, opened 2025-04-24
PIMS-2307 Remove Watch Agency Response Type

dbarkowsky:
## 🎯 Summary

  
[PIMS-2307](https://citz-imb.atlassian.net/browse/PIMS-2307)
Recommend this is reviewed after PIMS-2275 is merged. Will keep in draft until then.

## Changes
- Remove Watch value from enums in frontend and backend
- Adjust notification generation area to look for Subscribe instead of NOT (watch or unsubscribe)
- Create migration that moves all Watch responses to Subscribe responses. After looking at PROD to see how Watch is being used, it seems like admins occasionally attempt to use it when an agency is interested in a project. Unfortunately, Watch status acts more like Unsubscribe status at the moment. This should bring the attempted use closer to the intended behaviour.

## Testing
- Make sure you have parent agencies and child agencies with emails and Send Notifications checked in your local database.
- Start a new project and move it to ERP. There should be agencies in the Agency Response table when the project is created. 
- See that Watch is no longer an option when trying to manually change the response.
- See that emails are queued for expected agencies.

## 🔰 Checklist

### DIFF
--- a/express-api/src/services/notifications/notificationServices.ts
+++ b/express-api/src/services/notifications/notificationServices.ts
@@ -48,7 +48,6 @@ export enum NotificationAudience {
 export enum AgencyResponseType {
   Unsubscribe = 0,
   Subscribe = 1,
-  Watch = 2,
 }
 export interface ProjectNotificationFilter {
   projectId: number;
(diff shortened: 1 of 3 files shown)


Sharala-Perumal (approved):
The option 'Watch' is removed from the SUBSCRIBE dropdown list and the agencies who opted for Watch are now subscribed and receive notifications. Works as expected.

PR #3011, opened 2025-06-25
logging additions and small fixes

dbarkowsky:
## Changes
- Added some logging points. Some of these are temporary and can be removed later.
- Fixed one issue where parent agencies were getting the watching agency notifications as well
- Added a check when resending notifications. If the notification is already pending, it will cancel that one now before resending.  

## 🔰 Checklist


Sharala-Perumal (approved):
Changed the agency email id from one to another and the witnessed the CHES message id getting updated which confirms the new email id is update in CHES.

PR #3026, opened 2025-07-28
PIMS-2340 Site Banner Option

dbarkowsky:
## 🎯 Summary

  
[PIMS-2340](https://citz-imb.atlassian.net/browse/PIMS-2340)

This change allows for an optional site banner at the top of the page. Now we can set messages for users to see by changing an environment variable in the backend.

## Changes
- BANNER_MESSAGE env. If blank, the banner won't be visible.
- Updates to the lookup endpoint and context to send/hold this banner message
- Changes to how the layout dimensions are calculated in some cases. This is because the banner message needed to push down the window's content, not overlap it.

Should be used for short messages only. 

## Testing
- Set an environment variable in your `.env` for `BANNER_MESSAGE`. 
- Start or restart your API.
- Navigate around some pages to see that the message fits appropriately.

## 🔰 Checklist

### DIFF
--- a/react-app/src/components/projects/AddProject.tsx
+++ b/react-app/src/components/projects/AddProject.tsx
@@ -153,8 +153,8 @@ const AddProject = () => {
               label={'Net Book Value'}
               rules={{
                 min: {
-                  value: 0.01,
-                  message: 'Must be greater than 0.',
+                  value: 0,
+                  message: 'Must be 0 or greater.',
                 },
               }}
               required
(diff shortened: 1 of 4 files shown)


Sharala-Perumal:
Tested this PR yesterday. The banner shows up on all pages once the user logs in. But it doesn't show up on the login page. If this is the intended behaviour, the PR can be approved. Please confirm.

Banner display with the app:
image

Banner not displaying in the login page:
image

Sharala-Perumal:
> Looked into this today. The intent was to have it show on the splash page as well. The issue was that the lookup route required authentication, but users aren't authenticated at that point.
> 
> I separated the banner message into its own separate API endpoint, then created a separate hook and context in the frontend to handle it. It should now appear anywhere on the application.

I tested after the above changes were implemented. Now the banner message is displayed on the login page before the users signs in.
image

